PR #28, opened 2024-06-10
Ports Druid from BlackStone

Stagrawl:
Boots Dendor & Noc acolytes out of the church for so many reasons.

Adds the Druid role for Dendor Enjoyers.

I can't remember what else this did so I'll edit it in.

Let's see if this errors!


Stagrawl:
I can't believe you made me compile this myself.

image

Stagrawl:
Oh, is that the plan?

PR #33, opened 2024-06-11
Removes adventurer's limited choice of classes

Jaybirdyy:
## About The Pull Request

This simply removes the PQ requirement to access the full list of adventurer classes. It bypasses the PQ check entirely, and sets the amount to a maximum value.

## Why It's Good For The Game

While this PQ check, and PQ as a whole makes sense in the context of BLACKSTONE itself, this type of gamification is not typically something seen in RP servers.

Leaving what may be a conscious decision for a character's story - such as its very own class, entirely up to RNG doesn't seem entirely ideal. This will allow full player freedom in their own character.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/adventurer.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/adventurer.dm
@@ -42,19 +42,8 @@ GLOBAL_VAR_INIT(adventurer_hugbox_duration_still, 3 MINUTES)
 			addtimer(CALLBACK(H, TYPE_PROC_REF(/mob/living/carbon/human, adv_hugboxing_start)), 1)
 		var/list/classes = GLOB.adv_classes.Copy()
 		var/list/special_classes = list()
-		var/classamt = 5
-		if(M.client)
-			// For every 5 positive PQ points, grant an extra choice for Adventurer classes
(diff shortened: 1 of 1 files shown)


K4rlox:
good PR

DiscordWizard:
I like this PR, I'd really appreciate the consistency with adventurer characters, it's hard to roleplay one out when the profession and class they will be is left random

ChungusGamer666:
Abysmal dogshit, adventurertide was genuinely killing Blackstone. You can't just come up and wave the RP wand away at something that is inherently a massive balance issue.

BlackjackIsTired:
Infinite class rolls would give you every rare class option in the codebase. I'd rather not have a Lord of Heartfelt every round alongside every rare(which would no longer be with this PR) adventurer class in the codebase.

Ultimately, "this type of gamification is not typically something seen in RP servers." is foundationally fragile and a poor reason to obliterate any semblance of balance every round.

As for player choice, I'm all for it. But it'll be hard for the players making other choices when the metagamers whom exist in every Community whether blatant or otherwise, group up and shove ten fireballs down your throat.

LordDoge:
If adventurertide was a problem in Blackstone, that's fair enough. At the end of the day, I don't really care whether or not it's implemented: it was just requested that the PR was made, so I did.

If this change will seriously fuck up the balance of the server, I see no reason to implement it.

PR #48, opened 2024-06-12
Spanifies the codebase (I WILL NOT KEEP FIXING MERGE CONFLICTS IF THIS IS LEFT FOR AGES)

Gandalf2k15:
## About The Pull Request

Spanifies the codebase, each file. There are about 20 edge cases that cannot be changed due to the fact they use changing spans, I mean, I could change it, but that code is old as fuck so I won't.

GOD I HATE IT.

I included the regex .py in the files :)

## Why It's Good For The Game

Naturally because this is outdated af the code doesn't use span defs, which is a lil annoying. Just one part of modernizing the codebase to a standard that is tolerable... I shouldn't have to look up what spans are available.

Also, yes, it works fine:
image


ChungusGamer666:
i will do this upstream enjoy the merge conflicts

Gandalf2k15:
> Shooting ourselves in the foot right here if we want to sync

will wait for bob to do it upstream then do our stuff here then

K4rlox:
i wish we can get fancy chat on blackstone and rotwood soon

Gandalf2k15:
> i wish we can get fancy chat on blackstone and rotwood soon

ill port it here tbh

PR #68, opened 2024-06-14
Allows cleaning the coom-decal from floors by biting it.

CalamaBanana:
## About The Pull Request

This PR allows to clean the evidence after the act in a slightly degrading way.
If one is lying down and can use their mouth,  using the "bite" action on a floor with a coom decal on it will start cleaning it.

Small note, cleanables (blood and others as well) do not seem to be targetable directly ingame in this codebase, so I had to sneak it into the floor code instead.

## Why It's Good For The Game

Highly improves the maid gameplay loop.

### DIFF
--- a/code/game/turfs/open/floor.dm
+++ b/code/game/turfs/open/floor.dm
@@ -288,26 +288,3 @@
 			return TRUE
 
 	return FALSE
-
-//Dirt decals were not targetable directly as of writing this. If they ever are, best move this to /obj/effect/decal/cleanable/coom/
-/turf/open/floor/onbite(mob/user)	
-	if(isliving(user))		
-		var/mob/living/L = user		
-		if(L.stat != CONSCIOUS)
-			return		
-		if(iscarbon(user))
-			var/mob/living/carbon/C = user
-			if(!C.lying)
-				return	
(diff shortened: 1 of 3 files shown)


CalamaBanana:
That is the same volume you have when drinking from the bog directly. I stole the code there.

hivemuscle:
perish

githubuser4141:
Why only coom and not other cleanables like blood or dirt?

CalamaBanana:
Because other cleanables should have deeper interaction. Slurping up dirt/blood should make one vomit and stuff like that. Would be a bit weird to be able to lick everything clean without consequence.

Right now it's just kink stuff.

Gyrandola:
This is not holy.

DeerJesus:
> Because other cleanables should have deeper interaction. Slurping up dirt/blood should make one vomit and stuff like that. Would be a bit weird to be able to lick everything clean without consequence.
> 
> And mostly, doing it with blood/dirt would just be weird, half the rounds you end up with servants crawling all over the floor to clean instead of grabbing a soap/rug. Right now it's just inconsequential kink stuff,

Week old floor cum would probably make you vomit too

PR #73, opened 2024-06-14
Racial Stat Rebalancing

KKLZ4:
## About The Pull Request

Rebalances all the currently in-use races Statlines in a 2, 1, 1, 0, 0, -1, -2 system, with female variants being given -1 STR and +1 INT, apart from FemDrow, who get +1 STR and -1 INT. This also gives Akula/Axians a racial stat line, since they apparently didn't have one to begin with.

## Why It's Good For The Game

Minimizes racial imbalance, which should hopefully prevent the horde of Drow Women flooding the server due to the simple fact that they are objectively the best race/gender for nearly every task. Meanwhile races like Half-Elf were honestly pretty dogshit and woefully underutilized

RacialBalance


KittyNoodle:
webedit spotted

pismorizzo:
Minor gripe but drow losing their +1 int for instead getting a constitution doesn’t really fit favor for me, makes them more bland and generic rather than being more frail, smart, and fast

Deek-Za:
Neat. Though it misses Kobolds. Currently the base stats are using a 2,1,1 0,-1,-1,-2 array instead of your 2, 1, 1, 0, 0, -1, -2. So they'd need to get a bump in either Str or Con and adjust the females accordingly.

PR #77, opened 2024-06-15
Bathwench Expansion 1/30 🧵

Stagrawl:
_Powercreeps Bath Wenches, adds Bath Boys._

Adds Male Bath Wenches by hugely popular demand, slightly standardises the skills becuase there's not really any point in rolling for such meagre pickings anyway. 

Adds a harp to their loadout so they can actually play soothing music for customers, ideally this would probably be good mapped in, but this stops late-spawns from not getting one or the instruments being stolen, so...

IDK.

Part 1 of 30.


Stagrawl:
Also they're better at riding.

Stagrawl:
The male title is bath boy to be fair. I should probably tweak the tutorial now that you mention it.

Swain is a beautiful word. I'll make it so.

Multigamer99:
based changes

PR #91, opened 2024-06-18
Gives an alternative to prisoner masks in the form of collars.

Multigamer99:
## About The Pull Request

Makes prisoner have subclasses to choose what type of cursed item you want to wear, with current options of a mask or a collar. Also adds a cursed collar.

## Why It's Good For The Game

I, for one, am not a fan of the prisoner's forced cursed masks as it hides their face, something you wouldn't necessarily want to do to an inmate. For example, you would not know who your inmate is in the event that they escape and got the mask removed by the clergy or a rogue astratan cleric. However, I could see some arguments for it, so instead of replacing it outright, it gives an option to be able to have your face shown. A collar was chosen so the player would still have something cursed worn to identify them as a prisoner yet still be minimally invasive/covering.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/peasants/bogprisoner.dm
+++ b/code/modules/jobs/job_types/roguetown/peasants/bogprisoner.dm
@@ -37,7 +37,6 @@
 		var/datum/antagonist/new_antag = new /datum/antagonist/prisoner()
 		H.mind.add_antag_datum(new_antag)
 	ADD_TRAIT(H, TRAIT_BANDITCAMP, TRAIT_GENERIC)
-	ADD_TRAIT(H, TRAIT_NOBLE, TRAIT_GENERIC)
 	if(H.gender == FEMALE)
 		H.change_stat("strength", -2)
 		armor = /obj/item/clothing/suit/roguetown/shirt/dress/gen/random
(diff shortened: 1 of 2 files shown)


Useroth:
Good PR but needs some different way to go about it.

image

Multigamer99:
oh pfft yeah i can fix that

that's a side effect of my beefing up how i implemented the advanced classes

lemme fix that real quick

Multigamer99:
Alright, that shoulda fixed it. Either you can test it or I can go test it once I drag myself to my computer.

Multigamer99:
oh til that labeling a git thing as "fix #(something)" does weird git shit

Multigamer99:
Alright, so after fixing my ineptitude with git and doing a rudimentary (and probably not quite sufficient but weh) test, it seems that the fix works.
image
image
image
image
image
image

Multigamer99:
Reworked into the two roles idea, with one being a political prisoner in the keep and the other a normal prisoner outside.

Multigamer99:
Dear god I hate git.

Either way, fixed now.

Multigamer99:
It'll probably need some form of testing but checks have passed and I gotta get ready for work so either someone else can do that or I can do it later.

Multigamer99:
The asked-for changes have been made. Lemme know if any other tweaks need to be made.

Useroth:
Seems to work good.

PR #92, opened 2024-06-18
Increases playerslots for towner roles, also reenables two previously locked towner roles

zakanater19:
## About The Pull Request

Like title, increases some towner role slots both roundstart and midjoin, and enables two jobs previously set at 0 (so unplayable but still coded) the two jobs are tailor & scribe

## Why It's Good For The Game
More towner roles, also more RP focused roles reintroduced


Hmeister-real:
Tailor is pointless, just a copy of Seamstress

PR #93, opened 2024-06-18
Lower adventurer slots.... but femknights

Useroth:


Zebraman:
Let half orc female to be knight, and then we're set

PR #96, opened 2024-06-18
Allows any age to be Archivist - Requested PR

BlackjackIsTired:
## About The Pull Request

Allows any age to be Archivist

## Why It's Good For The Game

More freedom for Archivist age.

Requested by @Useroth


BlackjackIsTired:
> A young adult should not be an archivist.

I don't disagree...

Useroth:
That's the same kind of adult as for example steward and counsillor.
Mind that the ages we have are: Adult, Middle-aged and Old

PR #106, opened 2024-06-19
Torso and Nudist back to pilgrims

Azarak:
## About The Pull Request
Sets torso and nudist back to just pilgirms
I mistakenly set them to back to adventurers after the sync made them available for towners/adv/pilgrims for free


Hmeister-real:
isnt there a button to "show challenge classes" which adds both when choosing a role?

PR #112, opened 2024-06-19
Displays pleasures on round end and re-instates Good Lover trait

Azarak:
## About The Pull Request
Pleasures were lost in the sync, this brings them back, also colors the text slightly pinkish
So was Good Lover, but it wasn't used so I actually put it to use and gave it to Bath Swains, Bards, Minstrels and Werewolves
The trait grants 1 Triumph to your partner upon climax in intimate areas, only once per round


Useroth:
Kek at werewolves

usernameasterisk:
> Kek at werewolves

image

PR #121, opened 2024-06-20
Adds title to skin.dmf

Alanii:
### **Adds title to skin.dmf to permit main-window name**

Adds "TITLE" to skin.dmf and allows people to change the Main Window title card.

Currently it has been set to "Ratwood Keep" (the name of discord server).

This will have people who are pedantic about things (like me) not get confused upon seeing Blackstone.

Please note: If you want the main-window to display the current map, then remove TITLE in future after renaming the maps.

Right now, Main Window was pulling whatever map was loaded, in this case, Blackstone. Perhaps someone in the future will submit a system similar to TGSTATION that permits Castle names, but that would be some work.

Maintainer may edit this PR to change the name of the main-window as please.

### DIFF
--- a/interface/skin.dmf
+++ b/interface/skin.dmf
@@ -16,6 +16,7 @@ window "mainwindow"
 		icon = 'icons\\tg_32.png'
 		macro = "default"
 		menu = "menu"
+		title = "Ratwood Keep"
 	elem "split"
 		type = CHILD
 		pos = 3,0


Alanii:
Additionally, if anyone has a flavicon they wish to use, I can change that as well.

Alanii:
> As far as I can tell it currently pulls the `world` name which is `Rotwood Vale` The current deployment on the server may have not reflected that Edit: Err actually it's just "RATWOOD" from this #108

Okay, if this PR is not needed, I can close it.

PR #124, opened 2024-06-20
Adds rune chat emotes

Azarak:
## About The Pull Request
Now your emotes will also show up on rune chat, with a gray color to distinguish from speech

Emotes can also define `runechat_short_msg` to have it display instead of `message` to shorten down any emotes if they're too long

Edit: Currently only custom emotes make the runechat show it up.

### DIFF
--- a/code/modules/mob/living/emote.dm
+++ b/code/modules/mob/living/emote.dm
@@ -1094,6 +1094,8 @@
 	message = "whimpers."
 	message_mime = "appears hurt."
 	message_muffled = "makes a muffled whimper."
+	emote_type = EMOTE_AUDIBLE
+	show_runechat = FALSE
 
 /mob/living/carbon/human/verb/emote_whimper()
 	set name = "Whimper"


Hmeister-real:
dont like this, theres a certain charm to not having rune chat emotes.

Useroth:
let's testmerge it like it is for now

Hmeister-real:
only custome emotes pls, at least

Jakksergal:
Only custom emotes or at the very least a reduced font size for normal emotes. With the rate that runechat displays messages I don't wanna wait for speech to show up.

zeratino:
I think that another person made a comment that strongly reflected my feelings, here.

I think this is a little immersion breaking for text to constantly pop up over someone, when nothing else (that isn't TM'd) does that as of now. 

It already plays a sound when an emote is made. I think that is enough.

zeratino:
As a side note, I think that the way this is done is botched at best and immersion breaking at worst. The only way to signify the difference between an emote and speech is through their text color, which most people don't recognize.

image

Hmeister-real:
please un-TM and close. Very emmersion breaking, very unhelpful, and useless for actually seeing emotes, since they are too long and show annoyingly, and stop speaking text from showing up, which is especially awful since people audibly emote when in pain, and when they groan. Maybe rune emotes work on other servers, but not here. Thanks for making this anyway

KittyNoodle:
all emotes are muffled now

Useroth:
> > please un-TM and close. Very emmersion breaking, very unhelpful, and useless for actually seeing emotes, since they are too long and show annoyingly, and stop speaking text from showing up, which is especially awful since people audibly emote when in pain, and when they groan. Maybe rune emotes work on other servers, but not here. Thanks for making this anyway
> 
> Yeah, making only custom emotes use it

My suggestion would be to: let it stay for non-custom emotes, strip punctuation from the edges, wrap them in * so that they differ from spoken text.

PR #130, opened 2024-06-20
Revert "Devotion and Medical revision"

Azarak:
Reverts Rotwood-Vale/Rotwood-Vale#119

Inconsistent list syntax (is it even lists or statements? new byond feature?)
Spells being free is bad
No self surgery is bad (would rather implement it being harder)


Useroth:
(merging just on the basis of inconsistent list syntax)

DeerJesus:
Need to readd cure rot and limb miracle and such

PR #139, opened 2024-06-21
Skill Tweaks + Class Updates

Stagrawl:
Increases Noble Reading, Drunkard > Gambler, Seamstress > Gender Neutral Seamster


Useroth:
I feel we could actually combine tailor and seamster here at this point

PR #144, opened 2024-06-21
Stylish Wizards [SUGGESTION] 

AniciaBelle:
## About The Pull Request

Seamstress, and any other with the weaving skill, can make COLORED ROBES (Blue, green, yellow, orange, white) which, otherwise, would be obtainable only by class spawn (RNG).

image

## Why It's Good For The Game

Because, if you spawn with an ugly robe, you could try to find a Seamstress to make another and look more STYLISH.

Looking good is important too.

### DIFF
--- a/code/modules/clothing/rogueclothes/robes.dm
+++ b/code/modules/clothing/rogueclothes/robes.dm
@@ -116,6 +116,18 @@
 	color = pick("#4756d8", "#759259", "#bf6f39", "#c1b144")
 	. = ..()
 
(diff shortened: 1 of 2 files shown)


Stagrawl:
At some point we're most likely going to have dyeing.

I like this conceptually but it'll bloat up the Seamstress list more than it already is.

Alejodbort:
yeah it will, but thats how the system currently works.

At least it'll have more options to craft.

PR #163, opened 2024-06-22
Reworks Church and Cemetery. Swaps their positions on the map.

Acantharctia:
## About The Pull Request

Swaps Church's with Graveyard's place on the town. Reworks both. Basically, entirely converts the church from its small and kicked to the corner state and puts it in front of the brand new town center. Swaps place with the cemetery which has also been reworked.

**The House of the Ten**

image
1st floor [Z level 4]
The main entrance. Contains the nave (the big corridor main hall), the altar, the kitchen and infirmary, two storage rooms, and accommodations for church staff. Also has a side entrance to the left of the Altar.
image
2nd floor [Z level 5]
Holds the priestess and Inquisitor's room, an extra room, the upper part of the nave and the bell room. Also two balconies for rogues to sneak in or from heretics to escape from.
image
Underground [Z level 3]
Holds the interrogations chambers along with a new 'bath' room who has an extra exit outside. Also where Morticians settle directly in front of the cemetery gates.

**The Cemetery**

image
Mostly set on Z level 3, with more space for graves for particularly bloodier weeks along with more coffins stored in a sealed room. 

image
Above on Z level 4 is where the public garden has been moved to, with both more space for farming and relaxing after a long dae of work.

## Why It's Good For The Game

The church is definitely **not** small in any way. This is not some local small institution that appeared one day with an idea in mind. These people represent Psydon's and his children's will, which in no way in small feat. The church is powerful, and so has been seem as battle with is, even if lost, always prove costly to whoever attacks it.

While I'd rather not mettle with balance about power itself (spells, classes, roles), I would like it to be portraited as it should, This is not your local Sunday school. This is something that spreads as the known word of Blackstone (or Ratkeep) itself.

Also, the old church felt like an after though. Quite literally kicked into the corner of the map, with while decent design on the main floor, which still had faults, lacked heavily on the underground and specially the second floor, which only lead to a balcony overseeing the altar and the bell room. Everything else was just a solid wall.

I hope that with this PR, I can enhance RP by giving quite literally more room for the church, enhance combat inside the church, which was an absolute pain because of how crowded it often got and also bring it closer to the level of mapping standards we have seem recently.


Alanii:
### Suggestions
- Change the extra "Priest Key" to a "graveyard key"
  - Priest key does not have access to graveyard.
image

- Replace these spawns with tables, and also place the bell ringer on one of them.
  - The tables would allow a better place to place donations and such, so it isn't too crowded later in the round.
image

- Maybe change this Pantheon cross with the statue of Astrata.
  - Astrata Acolytes are who may offer better miracles and healing abilities, so it would make sense
image


I would also suggest to place the confessional at one of the crosses that are in the church foyer, but it isn't needed, just a want.

Altangy:
The platforms under the chandeliers will prevent them from falling to the lower floor in the rare instance someone remembers they can drop on people

Acantharctia:
Ignore that push.
I accidentally removed the entire new church lol

MefAnneFeminene:
The small coffin-room for the graveyard is completely innaccessible because of a wall that exists on the same tile as the door. In general, this cemetary needs a massive rework. The ground is primarily stone, which doesn't really fit the look of the cemetary, there is grass where graves are gonna be, which need to be removed with a hoe (something that isn't provided to the Gravekeepers), and in general the layout of the cemetary just looks really weird and unfitting for a proper graveyard.

Also the gravekeepers lose their private quarters, meaning they have nowhere to sleep now.

PR #164, opened 2024-06-23
All ages for archivist

Useroth:


Hmeister-real:
I don't like this change at all and think it's unneccessary, but if it has to go through at least restrict them to middle+old. Adults shouldnt be archivists they arent wise enough

Useroth:
As in previous PR, I sort of disagree. The archivist is a glorified librarian at the moment.

PR #167, opened 2024-06-23
Goblin Update

PTFOperations:
Enables Goblins,
Enables Armor for Goblins,
Update Rockhill for Goblin Fort spawning.


PTFOperations:
Additional Dark Goblin Skin added

PTFOperations:
Sounds like the same should be done for the regular king

PTFOperations:
Mainly just trying to get the content out there if it's a problem that'll just be after care triage.

PR #172, opened 2024-06-23
Allows us to say 'Swell' and 'Pussy' without the words being changed

Dedwolf:
## About The Pull Request

Title

## Why It's Good For The Game

People in the middle ages would absolutely say 'swell' and 'pussy' and there's no reason why we as players shouldn't be able to. Pussy is just slang for vagina which already has limited slang terms that we can use, and swell is obviously a normal word.


zoktiik:
"Yeah that guy's constantly getting `cunt-bitten coward`whipped"
"I'll have you know that there's no `CUNT-BITTEN COWARD`EEEEEE—"

good riddance

Hmeister-real:
cunt-bitten cowards great and il miss it, but i can live with this

PTFOperations:
I rather enjoyed cunt bitten coward more

PR #177, opened 2024-06-23
stops you hearing smithy noises from anywhere

Slorbin:
## About The Pull Request

if someone wants to implement it better sure they can but the code is so abysmally awful because it plays the sound locally in each person's head using the same thing used for ss13 explosions or in game music so goodbye won't miss you

## why good

https://github.com/Rotwood-Vale/Rotwood-Vale/issues/175 fix


kepesk1:
Implement this immediately, I have to lower my game sound just because of this horribly loud audio piece. Hurts my ears.

DeerJesus:
> There's a better way to do this. Simply make it so only players within X dist get the sound.

if someone wishes to do that they can but you can already hear shit through walls so... why bother?

Hmeister-real:
At this point the noise is literally engrained into my head when i think of Roguetown. Good riddence

PR #181, opened 2024-06-24
Heartfelt & Witch Hunter subclasses: allow tolerated+ races

zoktiik:
## About The Pull Request
Changes Lord of Heartfelt, Hand of Heartfelt, and Witch Hunter to no longer be humen exclusive. Tested in-game before and after the change.

## Why It's Good For The Game
If King and other nobles can be whatever race aside from horc, kobold or goblin, why can't the Heartfelters? Same goes for the Inquisitor and the Witch Hunter, where both are practically identical.


Coolgat3:
merge now

PR #184, opened 2024-06-24
[TM] Revert "Adjusts Popup Window Sizes"

Useroth:
Reverts Rotwood-Vale/Ratwood-Keep#140

Until this is fixed

image


Alanii:
> Reverts #140
> 
> Until this is fixed
> 
> image

Revert the skin.dmf change if that is the case. 

That window change is from the skin.dmf edit.

The other windows are set through specific width and height settings and do not affect that window.

PR #187, opened 2024-06-24
House staff reorganization

Gyrandola:
## About The Pull Request

Reduce the butler/maid job slots to 1.
Increase the servant job slots to 6.
Rewrite both job descriptions to clarify the hierarchy between the two.

## Why It's Good For The Game

Maids (3 slots) and Servants (4 slots) constituted a redundancy, with the only difference between them being the youngfolk age lock, which is meaningless. I've made it so the Maid is considered the head of the house staff and essentially a majordomo who commands the servants.

### DIFF
--- a/code/game/objects/effects/landmarks.dm
+++ b/code/game/objects/effects/landmarks.dm
@@ -356,7 +356,7 @@ INITIALIZE_IMMEDIATE(/obj/effect/landmark)
 	icon_state = "arrow"
 
 /obj/effect/landmark/start/butler
-	name = "Butler"
+	name = "Head Butler"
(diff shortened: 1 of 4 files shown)


CalamaBanana:
If you do this, do give aprons to the servants, and push their skills to where maids where at before, please.

Personally, as a maid player, I'd rather you remove servants outright and increase the maid slots and call it a day if you feels this needs changing. Adding even more hierarchy just over-complicates things.

DorfGuy1:
As a maid it's fun to have people to order around. But I don't think 1 maid is enough. I'd rather there be a few.

PR #189, opened 2024-06-24
gives buffs to female fishers

Elli-Skala:
## About The Pull Request

gives female fishers a flint, pan, a poor coin purse from empty, and a shift start fishing rod to match male fishers

## Why It's Good For The Game

while I understand the theme we're going with here for a difference between the genders, i find it hard to believe just because a fisher was female they would not have a working fishing rod _as a fisher_ , while also lacking basic cooking gear for said fish and litterly not a single piece of copper to their name


Stagrawl:
Jank. Just put the base gear before the gendered kit.

PR #190, opened 2024-06-24
"allowed_patrons" code now tries to avoid switching faiths first

zoktiik:
## About The Pull Request
The logic used is a bit different than the title implies, but the effect is the same.
Jobs and subclasses that switch your patron when you choose them will now attempt to avoid patrons that are from a different faith, instead only picking a patron randomly from within the same faith if possible, then falling back to random selection.

Old God ("One-Psydonite") characters becoming an unsupported (sub)class, that _does_ support both Divine and Inhumen patrons, will randomly pick between them as normal; there is no special logic or cases here. 

With this PR, an Astratite character will only ever become a Pestrite barber-surgeon or a Ravoxian barbarian, never a Graggite (if possible). This does not prevent him from becoming a Psydonite witch hunter or a Zizoist necromancer.

In other words, this PR really only means that you'll be seeing a lot more of this:

image
image

## Why It's Good For The Game
Having to roll the dice on whether my character is switched to a different faith is bad.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/roguetown.dm
+++ b/code/modules/jobs/job_types/roguetown/roguetown.dm
@@ -50,12 +50,19 @@
 	var/datum/patron/old_patron = H.patron
 	if(length(allowed_patrons) && (!old_patron || !(old_patron.type in allowed_patrons)))
 		var/list/datum/patron/possiblegods = list()
+		var/list/datum/patron/preferredgods = list()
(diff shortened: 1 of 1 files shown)


zoktiik:
I am considering another PR after this to add `old_god` as an allowed patron to more, (semi-)secular (sub)classes, wherever it would make sense.

PR #191, opened 2024-06-25
Lets females spawn as manor guards/man at arms

ghost:
## About The Pull Request

This lets female characters spawn as man at arms. Role was gender locked to male only for some reason.

## Why It's Good For The Game

It doesn't really make much sense that manor guard/man at arms is locked to males only; especially considering knight (which spawns with significantly better gear and is a much more prestigious position in-roleplay) isn't. 

As is, this has only really led to people who want to play female manor guards just going watchman and then hanging around the manor for most of the round anyways, which is lame.


H0gberg:
if that's the case, might as well remove the locks for templars as well, seeing as some female roles can't become templars for whatever reason

PR #194, opened 2024-06-26
Reworks the acid pit into something non-lethal AND more interesting

Gyrandola:
## About The Pull Request

The keep's gate pit no longer drops you into round removing acid, instead giving you a one-way ticket into the dungeon.
It's a 2 z-levels fall, which often results in dislocated limbs and broken legs, and you're now at the mercy of the dungeoneer.

image

image

image


pismorizzo:
Main issue for me is this would now require the dungoneer to stand down here basically all round so the pit doesn’t become a way 1 ticket into the keep. 

Should probably drop them 2 floors into bedrock on the south side of unbreakable walls, preventing access to this keep.

PR #199, opened 2024-06-26
Merchant.Wardrobe.PriceChanges

Bapldap:
Merchant change, adds Wardrobe and adjusts some prices.

## About The Pull Request

-Adds the Wardrobe License. Several players expressed a desire for an ability to sell fancy clothing from the merchant, and this provides such a thing without making seamstress obsolete. It has various colored clothes, hats, shoes, and a few things that overlap with what seamstresses can craft. However, the things that overlap are designed to be an expensive alternative to a seamstress in case there are none in the round. The license is 50 mammons.
-Returns Potion prices to their original price. 18 Mammons/24 Mammons for Health and Manna potions is fair: No player will buy a health or manna potion for more than 20 mammons, and the price increase on the previous PR(raised to 45 for 3 potions) wasn't really reasoned. With the addition of alchemy, this returns potions to being one of the unique things merchants can (actually) sell.
-Adds Spice to the luxury selling list. Should enable mages with lots of money to reasonably make some spell scrolls, I think. There's no other way to get it in game except for static spawns.
-Adds Sacks to the tools selling list. Lots of people ask for these.
-Raises stunmaces price to 300 (why would you triple the amount of stunmaces you get per purchase AND lower the price by half).
-Pluralizes the purchase names because Blackstone did not spec into the 'buy in bulk' route

## Why It's Good For The Game

-Adds variety to merchant. Giving the merchant more things to sell people means more people will seek out the merchant. Adds a good way to acquire specific outfits without making crafting roles obsolete.
-Adjusts some prices to be more reasonable. Potion prices were fine as is before the last PR, raises stunmace prices (should probably still be raised more), slightly lowers bow prices (they are just regular bows)
-More doesn't always equal better but the Merchant /does/ need some more options for unique things they can provide other players.

Let me know if any changes need to be made. I tested the functionality of all of the changes myself.

### DIFF
--- a/code/game/objects/items/rogueitems/gems.dm
+++ b/code/game/objects/items/rogueitems/gems.dm
@@ -83,7 +83,7 @@
 	slot_flags = ITEM_SLOT_MOUTH
 	dropshrink = 0.4
 	drop_sound = 'sound/items/gem.ogg'
-	sellprice = 554
+	sellprice = 400
 
 /obj/item/riddleofsteel/Initialize()
 	..()


Inanichts:
Selling drip? YES

Bapldap:
Added another commit that fixes riddle of steel sellprice allowing for infinite money.

PR #210, opened 2024-06-28
More Slots, Slopfixes

Stagrawl:
Slightly reduces Adventurer Slots.
Adds an acolyte.
Adds two* bathswains. @Teragenic 
Adjusts prisoner female strength.
Adds a churchling slot.
Re-opens up the mage apprentice slots.
Slight unjankening of rand skills.

Adjusts the Prisoner antag to be a bit less... Weird? As it is, optimal play is to basically just round-end suicide if you care about the triumphs but not your character.

Escaping is more rewarding than just 1 triumph, though not as good as a full week of rest from a normal citizen who's actually being productive. Death is still unpunished since it's technically a triumph? Just surviving gives you a pity triumph and a bit of fluff to try again next week.

### DIFF
--- a/code/modules/antagonists/roguetown/roleobjs/prisoner.dm
+++ b/code/modules/antagonists/roguetown/roleobjs/prisoner.dm
@@ -35,14 +35,14 @@
 		count = 1
 
 	if(traitorwin)
-		owner.adjust_triumphs(count)
+		owner.adjust_triumphs(3)
 		to_chat(owner.current, span_greentext("I've ESCAPED THAT AWFUL CELL! THE WORLD IS MINE!"))
 		if(owner.current)
 			owner.current.playsound_local(get_turf(owner.current), 'sound/misc/triumph.ogg', 100, FALSE, pressure_affected = FALSE)
 	else
 		if(considered_alive(owner))
-			owner.adjust_triumphs(-2)
-			to_chat(owner.current, span_redtext("I'LL NEVER ESCAPE!!"))
+			owner.adjust_triumphs(1)
+			to_chat(owner.current, span_redtext("I didn't get away this week, but I live to try again!"))
 		else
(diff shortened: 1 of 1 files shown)


Teragenic:
Adds more whores, makes stuff less random with skills, fixes the prisoner oversight not accounting for the already applied racial debuff. Good PR.

However, I do not like the random reduction of adventurer slots. 

Also, changelog is false. "Adds a bathswain." It adds two.

That's my review.

Stagrawl:
Appreciate the feedback. 

I think the adventurer slot reduction is something we're going to be working towards and eventually outright replacing with another system. 

The huge increase to adventurers to the point that a 6 slot role became 20% of our player population is something that wounds me deeply in my heart and sovl. Opening these other town-based slots hopefully opens more avenues and opportunities for some displaced players who claim to want RP roles. 

Well, if you're a bard main? You can keep your good lover by becoming a Bathswain! 
Mage main? Try being an apprentice from time to time, or use your powers in the keep! (Likely to be expanded on later)
Paladin/Cleric main? IDK, churchling. I'm coping, I actually don't mind where they go.

pismorizzo:
Adventurer slots do need some slight lowering, but I want to remind that one of the main reasons Blackstones pop nuked so hard was when they forced the drifter queue and removed Adventurers. 

Perhaps 5 coders shouldn’t tell 30% of the player base they aren’t allowed to have fun anymore. 

Good changes though

DorfGuy1:
On one hand, I was there for Blackstone when they had like, way too many Adventurer slots, with GREAT gear and great skills. It was cancer. So less adventurers the better. 
On the other. Its like...frustrating when All the slots are filled an instant after roundstart, even if I'm constantly clicking the button to late join. Maybe cut the slots in half. Ad some more after an hour?

PR #216, opened 2024-06-28
Pink chat log

ghost:
## About The Pull Request

I just removed pink shadow. Maybe somebody knows how to fix it by another way - u are welcome. I dont know.

## After: 

# So, less eyes damage.


DeerJesus:
I have never seen this in my life how does this occur

RaShCat:
> I have never seen this in my life how does this occur
It's occurs not in the game. You need to right click and download the game chat.

RaShCat:
> Making this change and testing on W10 (without WINE) doesn't _seem_ to do anything, after some basic testing. With the in-game chat, this CSS declaration seems either completely covered up by other declarations or otherwise entirely ignored. The chat log is the same as it has been with this change. Perhaps the deleted declaration was part of an earlier attempt to mimic a certain other server's chat?

Did you even tried to download chatlog?

PR #233, opened 2024-06-29
[bugfix] port blackstone #1376 -  Disables autobite while zombies are unplayable #1376 

Slorbin:
## About The Pull Request

Because the autobite functions separately from the zombie's AI attacks, it effectively attacks twice per round. This is super rude since NPCs don't interact with stamina, which is intended to balance combat.

## Why It's Good For The Game

Zombies will still attack normally with AI, but until they are playable again, autobite is not necessary.


DeerJesus:
> Nuke the whole commented out section - Let's not keep dead code in the codebase as we have the git commit history

It's commented out in case we decided to bring back player zombies, however, yeah, i'll nuke it

PR #236, opened 2024-06-29
Removes Tiefling spanish accent

Azarak:
## About The Pull Request
Extremely on the nose accent
We probably need to go through most of the accents and re-think what they're supposed to be


ChungusGamer666:
dogshit soul removal

PR #238, opened 2024-06-29
Master's Rod can now silence people for 20 seconds

Gyrandola:
## About The Pull Request
Title, ~~with the minor addition of being able to use the rod on people who do not have a bank account (?).~~

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/special.dm
+++ b/code/game/objects/items/rogueweapons/melee/special.dm
@@ -73,6 +73,9 @@
 			if(H.anti_magic_check())
 				return
 		
+			if(!(H in SStreasury.bank_accounts))
+				return
+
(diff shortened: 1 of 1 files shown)


CVSthePharmacy:
People who don't have a bank account? that's INSANE. keep it as those who DO have an account. This goes against the reason it was even made, it's not supposed to be the meta weapon, it's supposed to be used on people who disobey you. Such as the maids erping instead of doing their job of making food, not on the vampire lord when you go out hunting.

CVSthePharmacy:
> > People who don't have a bank account? that's INSANE. keep it as those who DO have an account. This goes against the reason it was even made, it's not supposed to be the meta weapon, it's supposed to be used on people who disobey you. Such as the maids erping instead of doing their job of making food, not on the vampire lord when you go out hunting.
> 
> Okay.. you have a point. I'll fix it

ty I appreciate you for listening

PR #241, opened 2024-06-29
Hair update

Azarak:
## About The Pull Request
Imports Roguetown hairstyles and facial hairstyles
Removes Bedhead floorlength hairstyle (was waaay too long)


wow50000:
very long with fringe is not plain long from roguetown
image

PR #261, opened 2024-07-01
Restricts some Wild-kin wing choices

Azarak:
## About The Pull Request
Angel, Megamoth, Mothra, Skeleton, Robotic are no longer available for Wildkins


Elli-Skala:
Megamoth, Mothra aren't avaliable for moths for some reason

PR #271, opened 2024-07-02
[RESPRITE] Reskins "Love Potion"

Alanii:
## Replaces "Love Potion" sprite

All I did was take the bottles we currenty use, changed it pink and submitted it into the cooking.dmi

The "Love Potion" is just an item and not a reagent. Perhaps in future it will be turned into a reagent so wizard may practice alchemy and spice up the life of the court.

Original Sprite
image

New Sprite
image


YennyWR:
Could you also rewrite the flavortext when you try to pour it into a beaker, while you're at it?
It still mentions feeding to slimes.

Alanii:
> Could you also rewrite the flavortext when you try to pour it into a beaker, while you're at it? It still mentions feeding to slimes.

I did not know that is a thing, I'll look into it.

PR #272, opened 2024-07-02
Proper "last class" and "current class" role locks

Azarak:
## About The Pull Request
Reworks the "last class" restriction in a way where it keeps tracks of all roles a player has played in the current and previous round.
This makes it so people can't abuse the system, and also won't get punished if they take a break and skip a round

~~All jobs with the exception of Pilgrim, Towner, Lunatic and Beggar have this restriction~~
~~Adventurers dont have this restricton, but they cannot be played if you played them in the current round (unable to respawn as an adventurer)~~

This was modified to affect only a handful of the nobility roles and a couple others, like :
King, 
Consort, 
Hand, 
Knight, 
Priest, 
Court Magician, 
Court Physician, 
Dungeoneer, 
Prince,
~~Adventurer~~
And back to Adventurers not being able to be picked again if you played them *current* round

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/adventurer.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/adventurer.dm
@@ -21,7 +21,7 @@ GLOBAL_VAR_INIT(adventurer_hugbox_duration_still, 3 MINUTES)
 	show_in_credits = FALSE
 	min_pq = 0
 	max_pq = null
-	lastclass_forbidden = TRUE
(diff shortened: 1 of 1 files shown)


Wildwestmars:
Well, with this, we might as well raise the adventurer cap again for reasons that should be obvious enough.

Murderfoxen:
it says adventurer isnt locked from the last round, but it seems like it is? a bit confusing. Did you.. test this merge? o_o

PR #274, opened 2024-07-02
Skips the "I'm preparing to walk again." check

zoktiik:
## About The Pull Request
Comments out the check that prevents you from ghosting (as in, becoming a spectator), before the zombie's rise timer is up.

## Why It's Good For The Game
Players can't actually control zombies for the time being, so there's no point to making them wait to ghost before zombifying.


DeerJesus:
GOOD. SPEEDMERGE

PR #279, opened 2024-07-03
Rewards players for contributing to the grimdark atmosphere

Dedwolf:
## About The Pull Request

Title. 

## Why It's Good For The Game

No longer punishes players for giving in to dark desires. Rewards players for being brave enough to contribute to what truthfully should be a grim and uncomfortable setting.

### DIFF
--- a/code/datums/sexcon.dm
+++ b/code/datums/sexcon.dm
@@ -1160,7 +1160,7 @@
 								if(H.mind.antag_datums.len)
 									wuzantag = TRUE
 						if(!wuzantag)
-							adjust_playerquality(+1, H.ckey, reason="Raped as a non villain.")
+							adjust_playerquality(1, H.ckey, reason="Raped as a non villain.")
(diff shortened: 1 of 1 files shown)


Stagrawl:
So brave, merge PR.

Stagrawl:
It’s fixed! Merge PR!

Honestly someone needs to spice up the sexcon it’s so dull. I’ll do it today maybe.

Useroth (changes requested):
Add a check for your ckey @Dedwolf . If victim's ckey == your ckey it will be +1, else, still -2.

Edit:
Let me make it clear this is a NO.
It was just said in a joking way.

pismorizzo:
GIVING PQ for rape is pretty ehhhhhhh, can be farmed absurdly easily with this method especially since you specifically allowed for it to work on AFK players.

Probably should just not dock PQ but not give any either. Giving a delayed triumph also makes more sense since they would then need survive the round after raping someone versus instantly gaining something nobody can interact with or take away.

PR #289, opened 2024-07-03
Reverts fishing minigame

Azarak:
## About The Pull Request
Reverts https://github.com/Rotwood-Vale/Ratwood-Keep/pull/155
It's really menial and annoying.
There is a good bunch of changes bundles up with it but most of them are really unnessecary
The fix for icon state of a bait bundle is left
And fish rod having the cast line as first intent is left too


DeerJesus:
THANK FUCK

PR #296, opened 2024-07-04
Fixes bandit spawning

Azarak:
## About The Pull Request
This fixes a terrible bug where the spawning of bandits will wait for one bandit after another to press OK in the popup, making them stuck in the lobby screen for possibly infinite time


MickDick1:
Yeah. My bad. I'll make it work correctly a bit later. Was about to make this.

PR #313, opened 2024-07-05
Die, archetypes

Stagrawl:
Does what it says on the tin. These are bloatslop, pointless and just don't add anything.

Commented for reference but these should really be deleted later.


Stagrawl:
No, not today.

PR #330, opened 2024-07-06
Plants no longer grow underground

Azarak:
## About The Pull Request
Fixes plants being able to grow underground, or where they cant see the sky
However, this adds an option for some plants in future to be able to grow like that (such as mushrooms)


Elli-Skala:
Why is it good for the game?

Useroth:
What if we added plants that don't require it along with this? ie some dwarf fortress mushrooms

PR #332, opened 2024-07-06
Werewolf nerf

Gyrandola:
## About The Pull Request

Moderate nerf to the werewolf claws and chop intent, which still remain Top Tier.
Lowers unarmed skill to Master.

## Why It's Good For The Game

Now that the werewolf infection has finally been fixed, we have dozens of them roaming the map decapping every person they see. The nature of this antagonist is undeniably combat and frag-heavy, but having so many of these one-man armies can turn every round into a giant deathmatch. So let's make them a bit weaker.

### DIFF
--- a/code/modules/antagonists/roguetown/villain/werewolf/werewolf.dm
+++ b/code/modules/antagonists/roguetown/villain/werewolf/werewolf.dm
@@ -148,7 +148,7 @@
 	attack_verb = list("claws", "mauls", "eviscerates")
 	animname = "chop"
 	hitsound = "genslash"
-	penfactor = 50
+	penfactor = 35
 	candodge = TRUE
 	canparry = TRUE
 	miss_text = "slashes the air!"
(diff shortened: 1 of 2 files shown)


sirelgate:
Disagree with the sneaking point. Vampires can knock you out from a distance instantly. You're telling me a wolf can't sneak? Or a bear? 
Largely disagree with this. Don't say werewolves are frag heavy, when I normally see non-werewolves running headlong at the werewolves in groups. Even when there is other antags about. Their main goal in round is to fight the vampires who are undeniably cancer.

TheRealRatMan:
HELL NO, as is your making the GIANT target out of all antags even weaker whilst vampires still as they are currently remain stronger than werewolfs and can easily blend in among humans.

The only thing nerf wise, which should be done to werewolfs AND vampires is capping the total amount of thralls/lesser werewolfs that can be made.

PR #333, opened 2024-07-07
Howling logs, doesnt message admins

Azarak:


BlackjackIsTired:
This is so blessed. Noticed immediately.

PR #336, opened 2024-07-07
Lower XP gain from farming

Azarak:
## About The Pull Request
Halves the EXP gain from Farming

## Why It's Good For The Game
Non farmers skilling farming have it slightly too easy from what I noticed


Elli-Skala:
Why It's Good For The Game?

PR #340, opened 2024-07-07
Changes aghost perm requirement

Azarak:


Useroth:
No longer needed after config adjustment.

PR #358, opened 2024-07-08
Attempt to fix the Draconic language

### LINKED ISSUE #200: [BUG] Draconic language just does not work
## Describe the bug
Draconic language has a prefix that doesn't work and its text also has no special font/style. Please fix.

## Steps to Reproduce:
- Spawn as a race that gets Draconic language: Drakon, Sissean, or Kobold.
- Right click your SKILLS button to see that your language prefix is `,10`
- Try to use it, have your message come out in Imperial as `Unknown hisses, "0 [text]"` because prefixes need to be comma and 1 character
- Go into the code and change it to use a free prefix like `,r`
- Watch as your speech remains visually indistinguishable from normal Imperial speech

Hmeister-real:
## About The Pull Request

fixes #200


Hmeister-real:
It probably works, but i cant figure out how to make the language use a font like the other accents and it currently looks like normal speak to other draconic speakers

Hmeister-real:
> Edit: Managed to get it fully working after adding it to the list of languages allowed by the tongue object. Will decide on font before making a PR.
Sweet, thanks for doing this for me. <3. Il keep this drafted incase you need to copy anything over, when you make your PR let me know to close this one

PR #349, opened 2024-07-08
Builds Out Old Classes, Removes RNG

Stagrawl:
Bunch of small changes, main focus is to build out some of the older less developed classes and remove horrible skill RNG from others. While I was in there, also cleans up some loadout jank.

This isn't really a balance passover, mostly a style and good practice run through. It does slightly touch balance, but generally for the better.

Expands:

- Bath Maid - Whip Skill + Few Bump Ups
- Barkeep - Sword Skill + Better Hand-to-Hand Utility
- Mage Apprentice - Basically turns this into a mage. (Until Rituals are implemented)

Adjusts (Removal of RNG or Cleanups, tiny number adjustments.)

- Blacksmith
- Mason
- Merchant
-  Mortician
- Jester
- Magician
- Acolyte
- Carpenter
- Miner
- Woodcutter

Does nothing to slots.

PR #360, opened 2024-07-08
Gives masonry to Miners and Blacksmiths

Hmeister-real:
## About The Pull Request

Miners mine rocks, and then cant make shit with those rocks. Blacksmits simlarly somehow don't know basic masonry.

## Why It's Good For The Game

Allows for more building without having to have high Int


Teragenic:
Miners I agree with because it's rocks and they use rock pillars and rock power.

I do believe that Blacksmiths should not get it. They are specialized in a trade. As much as I personally don't want to grind masonry as a Blacksmith(God I love spamming rock floors) I do feel like it would subtract people from hiring masons/carpenters more than it already has

Elli-Skala:
I am unsure about giving miners carpentry. Though also some of theses skills are RNG and I thought we were moving away from skill rolls?

Hmeister-real:
> also some of theses skills are RNG and I thought we were moving away from skill rolls?

Carpentry was already RNG, and i made masonry RNG too because it makes for more diversity, i didn't know we were shifting away from it?

Hmeister-real:
> I do believe that Blacksmiths should not get it. They are specialized in a trade. As much as I personally don't want to grind masonry as a Blacksmith(God I love spamming rock floors) I do feel like it would subtract people from hiring masons/carpenters more than it already has

They only get level 1, which doesnt let them build walls, and if someone is willing to grind Masonry to wall building level from 1, i would assume they would do the same from 0, yes? Level 1 just allows for construction of very basic masonry things, and would let lone bogsmiths to build little bog smitherys by themselves, where you cant usually hire a carpenter/mason

Elli-Skala:
Atleast with https://github.com/Rotwood-Vale/Ratwood-Keep/pull/349 is

PR #366, opened 2024-07-09
Lets Mages be Zizoid

Slorbin:
Mages should be able to pick Zizo as a God because the power doesn't come FROM Noc like an acolyte or anything - Noc just kinda gave them the knowledge of magic. Non-Noccites would still be able to cast magic, just look at Zizo and his worshippers... Kinda like how Non-Matthiosians can use fire despite the fact that Matthios stole fire from the underworld and gave it to humans, yeah?

Magic is more science than 'divine gift'. Now, Zizoid mages can exist and be EVIL. The Shadow Wizard Money Gang shall rise again!


pismorizzo:
Kinda silly as hell you are forced to be part of the Divine Pantheon if you are a magic user. Noc is their Patron not their powers originator.

PR #370, opened 2024-07-09
[DRAFT] Species Balancing

Dedwolf:
## About The Pull Request

Another attempt as species balance, this PR aims to bring all species in line with one another while balancing out stats in a way that keeps or makes species more interesting to play while also taking into account lore and roleplay.

I am aware that some stats are 'worth' more to players in combat than other stats, that is outside the scope of this PR and my abilities. Perception for example, affects very little, I thought it affected critical strike chance, fov with helmets and accuracy with weapons, it doesn't do either. This is something that should be fixed eventually but right now is not a huge concern and can be implemented in the future.

Think about a typical roleplaying game, when you're creating a character and assigning stat allocations, in no world is +1 strength over baseline worth +2 perception and +1 agility. 

Anthro species have poorer stats than humans and humanoids as they are intended to have traits that allow them to do things a typical humanoid would not be expected to do, such as drinking bogwater, leaping great distances, or landing falls that'd otherwise injure someone. 

## Why It's Good For The Game

Balances species with respect to lore and roleplaying considerations. This PR is open to constructive feedback that can be explained reasonably and I'm happy to change things and implement your ideas for traits.


MefAnneFeminene:
Changes to Fortune are extremely vast, and giving extra fortune, or taking away fortune can break races. Fortune directly changes how often crits happen, and giving a race a higher or lower chance to crit would mean that they'd need to be compensated heavily elsewhere.

Remove the Fortune Changes to both Tabaxi and Half-kin. The only race to justify having a greater Fortune would be Aasimar, in which they get worse stats elsewhere overall. Not even Humens should have any higher fortune.

MefAnneFeminene:
Ideas for race traits

Elf - Walking on leaves
Lizard - Breathing underwater
Tabaxi - Punches can cause bleed

For wild-kins (probably half-kins too), I'd actually go the route of making them similar to humens, as there's gonna be so much variety in Wild-kin subraces (given the fact they encompass all anthros not already covered) that it'd be nigh-impossible to assign a trait that would realistically apply to all of them

Stagrawl:
I've already quietly raised some of my concerns with you about this. Not going to drag you through the dirt in public.

code/modules/mob/living/carbon/human/species_types/roguetown/human/humen.dm
Alanii, line 1: "weeping gods" is grammatically correct.

god's implies it is a singular god and not multiple. If you are going to imply humen were created by "god" then imply they're children of psydon instead so it fits within the current lore.

Stagrawl: > I thought the weeping god related to a specific god

It's Psydon.

PR #385, opened 2024-07-10
Torso is an actual challenge

Gyrandola:
## About The Pull Request

Gives the torso class randomized and, at best, average skills and stats.
Why? Because it's supposed to be a challenge, and people are reportedly coordinating with their metabuddies to have their torso character picked up and given limbs.


DeerJesus:
Yeah some guy coordinated with a meta buddy to get full limbs and just started murdering people and then that meta buddy grabbed their body after we FINALLY killed the billion hp torso to be revived THEYRE utterly INSANE to fight

PR #388, opened 2024-07-10
Removes Scrying Orb from Merchant

Jaaackx:
## About The Pull Request

As titled, removes Scrying Orb from Merchant entirely.

## Why It's Good For The Game

Because Scrying Orbs are a special magic item that are incredibly powerful / annoying to be on the receiving end of, so the Magician's should be the only real source of one.

No more getting scryed and hunted by a loser because you stole their ERP partner. Win for humenity. Win for my balls.

Making it more expensive is pointless because currency is ridiculously easy to make. Magician gets the monopoly.

### DIFF
--- a/code/modules/cargo/packsrogue/luxury.dm
+++ b/code/modules/cargo/packsrogue/luxury.dm
@@ -61,11 +61,6 @@
 	cost = 300
 	contains = list(/obj/item/clothing/ring/active/nomag)
 
-/datum/supply_pack/rogue/luxury/scrying
-	name = "Scrying Orb"
-	cost = 120
-	contains = list(/obj/item/scrying)
-
 /datum/supply_pack/rogue/luxury/listenst
 	name = "Emerald Choker"
 	cost = 250


Jaaackx:
> A salt PR that I actually think might not be a bad idea.

not even a salt PR honestly, I haven't been bothered one in weeks
I just think it stinks

Useroth:
I added orb logging for the purposes of spotting overuse.
I'd consider raising its cost to not be laughable tbh, rather than complete removal.

Up for discussion.

Altangy:
As a reminder, to scry someone, the game has to acknowledge you know them. While a lot of people start known at round start, you don't know everyone. Additionally, Merchants themselves choose if the scrying orb is for sale or not and at what price. If scrying orbs are falling into the hands of peasants too often for you, try playing merchant and charging people out the rear for one or just say you aren't selling them. Apparently most people have too much money anyway.

DeerJesus:
> I added orb logging for the purposes of spotting overuse.
> 
> I'd consider raising its cost to not be laughable tbh, rather than complete removal.
> 
> Also imo _some_ way of even approximately finding people should be reliably available (magician that may fuck off and ERP ignoring everything around the round or die somewhere + stealable item is not something I'd call reliable). Lifeweb does that with cameras that may be broken by people and locating people's rings (rings in Lifeweb are basically equivalent of ID card, holding people's access to the economy terminals and a lot of doors, so you may ditch yours to prevent being located but it's at a sacrifice.
> 
> Up for discussion.

I believe a solution to a magician that fucks off/it being stealable whilst letting us make the item a one of a king could be to make it a structure! I've been considering it for a while, but usually a crystal orb doesn't.. move location in a tower. It's on a pedestal. MAYBE, make the pedestal on wheels so it can be brought to the court, but not carryable, yes?

Jaaackx:
> > I added orb logging for the purposes of spotting overuse.
> > I'd consider raising its cost to not be laughable tbh, rather than complete removal.
> > Also imo _some_ way of even approximately finding people should be reliably available (magician that may fuck off and ERP ignoring everything around the round or die somewhere + stealable item is not something I'd call reliable). Lifeweb does that with cameras that may be broken by people and locating people's rings (rings in Lifeweb are basically equivalent of ID card, holding people's access to the economy terminals and a lot of doors, so you may ditch yours to prevent being located but it's at a sacrifice.
> > Up for discussion.
> 
> I believe a solution to a magician that fucks off/it being stealable whilst letting us make the item a one of a king could be to make it a structure! I've been considering it for a while, but usually a crystal orb doesn't.. move location in a tower. It's on a pedestal. MAYBE, make the pedestal on wheels so it can be brought to the court, but not carryable, yes?

I'd be up for this, having it static so it's always accessible.

My main issue with the orb is just any Joe Schmoe being able to buy a powerful magic artefact and reliably hunt people down. If doing so requires them to get into the Wiz Tower then that seems fine.

I agree with some bits @Useroth but I'd maintain that any price changes won't really help. The economy is so fucked at the mo that it'd just be meaningless. 30 mins/1 hour of grinding and you're pretty much set for anything.

PR #390, opened 2024-07-10
Vile prayers no longer set you on fire

Azarak:
## About The Pull Request
This was problematic for
1. People setting themselves on fire to run into vampires
2. People purposefully committing suicide while being captured with this
3. People having legitimate prayers to inhumen gods with the vile words included being punished


Teragenic:
I was gonna suggest making it only work in the Temple, but after reading the list of banned words. I don't see any reason not to just remove it yeah.

PR #393, opened 2024-07-10
Removes Inquisitor Antagonist Status

Stagrawl:
This literally just causes pointless antagonism for an objective that just genuinely isn't worth keeping in until properly reviewed.


Stagrawl:
This comes after another update specifically targeting Inquisitors who frag Divine Pantheon players.

This is genuinely starting to veer hard into a player issue line but also this objective is dumb in the current state of the game.

PR #396, opened 2024-07-10
Lowers small log stockpile price

Azarak:
## About The Pull Request
Lowers it from 3 to 2, adjusting other prices accordingly

Problem not exactly solved yet, as what is really the issue is how easy can non-lumberjacks jack wood (with stones or weapons)

Lowering it to 1 would be too harsh on the lumberjacks maybe? Maybe not? Maybe Steward can adjust it anyway


Elli-Skala:
I don't think this stops the problem at its source, In that it's just to easy to collect logs in mass even when the keep lowers it to 1 it's still brute forceable. Personality I think it may be better to make wood cutting much harder for people without the wood cutting skill seeing how alot of people doing it are adventures

PR #411, opened 2024-07-11
Removes cursed un-removable status from the Hostage mask

DiscordWizard:
## About The Pull Request

Removes the curse from the Hostage mask that makes it impossible to be removed

## Why It's Good For The Game

The cursed mask is some sort of reference, and the reasoning is "The King doesn't want people to recognise the nobles", but from playing with it all the mask does is get in the way of RP and only works in annoying the prisoner and the people who pardon the prisoner and then have no ability to remove the mask.

This turns the cursed mask into a shameful mask, so wearing it is something the king can force into someone under threat rather than an arbitrary, un-removable thing.


Gr33d-y:
why not make it so they are wearing the cursed collars instead? We could also make it so that anyone who the king gives the title of "the unchained" to will be able to take off their cursed item.

Useroth:
Maybe make code to make them lockable with keys.

DiscordWizard:
> Maybe make code to make them lockable with keys.

I could work on code for that, but until then, I feel like it should be disabled so we don't have to suffer eternal masking.

Drayco84:
I second replacing the mask with collars for hostages.

PR #413, opened 2024-07-11
Gives merchant 3 slots instead of one, also gives each merchant a little more starting coins in the shylock

zakanater19:
So the merchant won't be the sole trader anymore, there will be a trio of actual merchant's guild members, the shophands are still in the game

Also changes starting shylock coins from 22 to 50 per merchant


zakanater19:
https://github.com/Rotwood-Vale/Ratwood-Keep/pull/415

Jaaackx:
Bad idea and completely unnecessary.

Nobody's playing Merchant, and stripping the role of power/autonomy by throwing more people at it isn't gonna help. The Soilsons have the issue of nobody having authority and everyone trying to do their own stuff around each other and it's awful as result.

PR #414, opened 2024-07-11
Makes more slots for weaponsmiths, armorsmithes and masons, from 1 to 3

zakanater19:
Masons already had 3 total but 2 after roundstart, now all have 3 slots and 3 latejoin slots, for a max of 3 per job

### DIFF
--- a/code/modules/jobs/job_types/roguetown/yeomen/blacksmith.dm
+++ b/code/modules/jobs/job_types/roguetown/yeomen/blacksmith.dm
@@ -63,8 +63,8 @@
 	flag = WEAPONSMITH
 	department_flag = YEOMEN
 	faction = "Station"
-	total_positions = 1
-	spawn_positions = 1
+	total_positions = 3
+	spawn_positions = 3
 
 	allowed_races = RACES_TOLERATED_UP
 


LocalizedDownpour:
Would it be more feesable to just combine the Armorsmith and Weapon smith into a single role and keep the slots? So two Smiths with 3 in Blacksmithing/Weaponsmithing/Armorsmithing. This would fix like....75% of the Town Smith issues.

Alanii:
Masons are built in to have a specific number as they're "engineers" and provide mining and building for the entire town. They also have the map support (i.e. beds, equipment, etc).

Armor/Weapon smithes provide armor and weapons, with their 2 apprentices.

In a perfect world, it works well. 

I understand there is a large player count, so increasing the amount of town smiths is fine. However, if you commit to this, you are going to have to implement map support to support those numbers. You can not just "add more slots" assume that suddenly the issue will fix itself, because it will just cause more issues than solve.

Jaaackx:
> Would it be more feesable to just combine the Armorsmith and Weapon smith into a single role and keep the slots? So two Smiths with 3 in Blacksmithing/Weaponsmithing/Armorsmithing. This would fix like....75% of the Town Smith issues.

This would make sense, most of the line crowds one window anyway.

As with Merchant stuff, just opening more slots is a stupid idea.

zakanater19:
> Masons are built in to have a specific number as they're "engineers" and provide mining and building for the entire town. They also have the map support (i.e. beds, equipment, etc).
> 
> Armor/Weapon smithes provide armor and weapons, with their 2 apprentices.
> 
> In a perfect world, it works well.
> 
> I understand there is a large player count, so increasing the amount of town smiths is fine. However, if you commit to this, you are going to have to implement map support to support those numbers. You can not just "add more slots" assume that suddenly the issue will fix itself, because it will just cause more issues than solve.

And what would the map changes be in your opinion

john-pawson:
> > Masons are built in to have a specific number as they're "engineers" and provide mining and building for the entire town. They also have the map support (i.e. beds, equipment, etc).
> > Armor/Weapon smithes provide armor and weapons, with their 2 apprentices.
> > In a perfect world, it works well.
> > I understand there is a large player count, so increasing the amount of town smiths is fine. However, if you commit to this, you are going to have to implement map support to support those numbers. You can not just "add more slots" assume that suddenly the issue will fix itself, because it will just cause more issues than solve.
> 
> And what would the map changes be in your opinion

Well, some of the map changes needed would be:
- More windows, since a single line or two lines for 6 smiths would just make things as slow as they are.
- More anvils, forges, tools, etc.
- More space in general since the place will start to get crowded with 8 people running around (6 smiths and 2 apprentices)
- More furnaces since most of the time the smiths don't even have enough materials for their CURRENT orders. Now imagine having 3 ingots and 6 smiths running around. Most of them wouldn't even be smithing.

Now, this is unrelated to mapping, but:
If there ARE latejoin smiths and etc. The ones that latejoin will probably not even get work because, in comparison to the ones that were there since roundstart, they'll just have worse skills, wich means they'll have lesser quality, wich means they'll get ignored, and the more skilled two smiths will get most of the work done.
Also, latejoin smiths already exist by pilgrims.

zakanater19:
> they'll just have worse skills, wich means they'll have lesser quality, wich means they'll get ignored, and the more skilled two smiths will get most of the work done.

skills relating to quality was removed

john-pawson:
> skills relating to quality was removed, so it's all the same quality no matter skill level

Ah well, i didin't know that, my bad, but everything else still stands

zakanater19:
votes

PR #415, opened 2024-07-11
Gives merchant 3 slots instead of one, also gives each merchant a little more starting coins in the shylock

zakanater19:
From https://github.com/Rotwood-Vale/Ratwood-Keep/pull/413

So the merchant won't be the sole trader anymore, there will be a trio of actual merchant's guild members, the shophands are still in the game

Also changes starting shylock coins from 22 to 50 per merchant

### DIFF
--- a/code/modules/jobs/job_types/roguetown/yeomen/merchant.dm
+++ b/code/modules/jobs/job_types/roguetown/yeomen/merchant.dm
@@ -3,8 +3,8 @@
 	flag = MERCHANT
 	department_flag = YEOMEN
 	faction = "Station"
-	total_positions = 1
-	spawn_positions = 1
+	total_positions = 3
+	spawn_positions = 3
 	selection_color = JCOLOR_YEOMAN
 	allowed_races = RACES_TOLERATED_UP
 	tutorial = "You were born into wealth, learning from before you could talk about the basics of mathematics. Counting coins is a simple pleasure for any person, but you've made it an artform. These people are addicted to your wares and you are the literal beating heart of this economy: Dont let these filthy-covered troglodytes ever forget that."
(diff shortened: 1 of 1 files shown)


Gr33d-y:
Don't shophands already do this?

zakanater19:
> Don't shophands already do this?

the shophands are still in the game

coffeecoffeecoffee:
It's very much a major issue and the store has two sides and two counters which are never properly utilized. The area is clearly intended to be able to run two stores and the merchant line is often nearly as bad as the smith line. I think this is a good thing to try to begin tackling the issue of 160 fucking people trying to hog the same storefront.

LocalizedDownpour:
Two Merchants with Two Shophands seems a more fitting solution then three.

Alanii:
Similar to your previous "increase roles" PR.

If you are going to increase the slots, you will have to do a lot more than just increase slots. The Merchant is the owner of the "merchant guild" branch of the Town. Which means there is one merchant.

Simply increasing their numbers will lead to confusion. Just increase the shophands if you want more merchant guild members.

I suggest you attempt to look at how Lore, Map, and Roles work before deciding to open slots everywhere.

Jaaackx:
Bad idea and completely unnecessary.

Nobody's playing Merchant, and stripping the role of power/autonomy by throwing more people at it isn't gonna help. The Soilsons have the issue of nobody having authority and everyone trying to do their own stuff around each other and it's awful as result.

zakanater19:
> Bad idea and completely unnecessary.
> 
> Nobody's playing Merchant, and stripping the role of power/autonomy by throwing more people at it isn't gonna help. The Soilsons have the issue of nobody having authority and everyone trying to do their own stuff around each other and it's awful as result.
> 
> If anything, allow the Merchant to hire more Shophands.

Yeah you know what you are right, im thinking i should just rework shophands instead

PR #417, opened 2024-07-11
Keys overhaul & Map changes

Gyrandola:
## Keys overhaul & Map changes
From a single "garrison" key guards could use to access all garrison areas of the map, to:

- Bog guard:
   - Gatehouse
   - Barracks
   - Dungeon
   - Armory (Bog master)
  
- Town watch:
   - Barracks
   - Dungeon
   - Armory (Sheriff)
   - Sheriff's office (Sheriff)
   
- MAA
   - Gatehouse
   - Barracks
   - Armory (Captain & Knights)

<h3>Dungeon & Dungeoneer:</h3>

- Dungeoneer gets keys for all the aforementioned dungeons.
- The prison related area of the keep dungeon is now only accessible by the dungeoneer, bailif and councillors.
- The dungeoneer gets a whole new torture room just for him.
- The secret tunnel now connects the keep dungeon to the town watch one.

<h3>Mercenary quarters:</h3>

- Mercenary barracks are now completely separated from the town watch and all the doors come with locks.

<h3>Other changes:</h3>

- Steward gets a spare copy of all new keys.
- All the crates, wardrobes and cabinets in garrison areas now come with a lock.
- The steward's upper window has been replaced with a "wall window" to make it harder to break into.
- The bog barracks upper window has been replaced with a "wall window" for the same reason.
- ~~The top floor rooms in the manor now require the manor key and not the captain one.~~

## Screenshots

image
image
image
image
image
image

### DIFF
--- a/modular/code/game/objects/pillory.dm
+++ b/modular/code/game/objects/pillory.dm
@@ -87,7 +87,7 @@
 
 /obj/structure/pillory/proc/togglelatch(mob/living/user, silent)
 	user.changeNext_move(CLICK_CD_MELEE)
-	if(locked)
+	if(latched)
 		user.visible_message(span_warning("[user] unlatches [src]."), \
 			span_notice("I unlatch [src]."))
 		playsound(src, 'sound/foley/doors/lock.ogg', 100)


TheFenix323:
Could the servants/maids get keys for the upper rooms in the keep?
Since the rooms where added the servants/maids haven't been able to access those rooms.

CalamaBanana:
>The prison related area of the dungeon is now only accessible by the dungeoneer and the bailif (excluding master key).

This is not a good change gameplay wise. Too often the dungeoneer is missing, and no one would have access to the prisoners that way. Guards need to be able to store/handle prisoners without a dungeoneer, too. Unless you really just mean the keep dungeon cells, and not the town watch cells as well. Then it is perfectly fine in my opinion.

>The top floor rooms in the manor now require the manor key and not the captain one.

That was intentional to mark it as the Bailifs/Councillors private quarters, they have that key. By giving it to the manor key, it will be even harder to stop people from pretending it is general guest quarters.

CalamaBanana:
> I always did think it was a general guest area? I suppose I can revert the change but they always go completely unused.

No, it is the bailiffs office. He has the single bed room, his councilors have the double bed room there. It's where they spawn, and they all have the key to it. I do agree the room is weirdly placed, too far from where the bailiff could use it comfortably. But that's a bigger edit.

>The town cells can still be used, it's the keep's cells area that is restricted. 
Perhaps you could shift the entrance tunnel to the right of the underground torture room and shift the doors a bit, so it would be less of a hassle for normal guards to take prisoners to the town cells? My worry is mostly that the king orders someone put into a cell, and absolutely no one but he himself and his masterkey can get to them.
image

Jackofashepard:
So overall this was a pretty good commit, I'm impressed that you segregated the keys so well for all the different roles. I just have two suggestions. 

1. The boggurds really don't need to be locked out of their armory, they never loot it like the keepies do. This is mostly due to the fact that they don't even have anything to steal, its all iron. Now they just cant get their spears or stone axes.
2. Maybe a master key lite for roles like hand?

CalamaBanana:
If you end up making more adjustments to it, could you please move the ale cellar door to the upper or lower side of the wall again? Right now, one sees and lights up the entire room when opening it. Before, one could hide in the dark at the far side for maid RP. That was one of the last few good spots in the keep.

DavidAreYouSure:
The inquisitor also lost access to prisoners which is often a source of some good RP they're also called on frequently enough to torture prisoners for one reason or another that it makes sense for them to have it.

Alanii:
Should return the Steward key to hand.

No one but the steward can obtain the spare steward key or the vault key now.

So if there is no steward, no one can managed the economy if someone locks the nervemaster.

DorfGuy1:
> So overall this was a pretty good commit, I'm impressed that you segregated the keys so well for all the different roles. I just have two suggestions.
> 
>     1. The boggurds really don't need to be locked out of their armory, they never loot it like the keepies do. This is mostly due to the fact that they don't even have anything to steal, its all iron. Now they just cant get their spears or stone axes.
> 
>     2. Maybe a master key lite for roles like hand?

I support the boggard change, cause the one time someone did loot the armory, the otherboggards killed his ass.

CalamaBanana:
The steward's key being locked in a steward-keyed chest is good in theory, but a solo-clerk won't be able to get started without harassing the king for the master key first.

I mean, in 99% of rounds clerks get the key anyways, why not just start them with it already?

DorfGuy1:
Also giving only the bog master access to the armory is gonna mean the following: He's gonna leave it unlocked so people can use it,  and then if prisoners escape they can grab things. Also inquisitors should have access to prisoners cause RP

Leudoberct1:
Only the mortician seems to be unable to unlock the graveyard. It would be very much appreciated if Church roles would be able to unlock it. Morticians are not the only Necrans to perform burial rites.

PR #418, opened 2024-07-12
Allows inserting keys in the peddler

TheFenix323:
## About The Pull Request
Let's you put keys into the peddler.

## Why It's Good For The Game
It allows the Innkeeper to return the room keys to the peddler (saw that there was an issue that didn't let you insert them back in).


TheFenix323:
Would this be better?

PR #424, opened 2024-07-12
Makes town doctors good actually but also makes them woman too????

Gr33d-y:
Makes town doctors have master medicine at all ages instead of journeyman as an adult by default also ungenderlocks doctor because genderlocking this role makes absolutely no sense.

The *changes* go exactly as follows

Women can now be doctors.

Skills no longer get better with age, but are better than the adult stats were.
Medicine went from 3 to 5, sewing went from 2 to 3, and alchemy went from 0 to 3.

This makes it a nice middle ground between the adult skills, and the Old age skills across the board.


TheFenix323:
5 feels too high, what about 4 in medicine?
Also can you give them a dress if they are female?

Gr33d-y:
> 5 feels too high, what about 4 in medicine? Also can you give them a dress if they are female?

5 was already possible before this change by changing your age to old so I do not see how it changes much.

PR #439, opened 2024-07-14
Fixing short sprites

john-pawson:
## About The Pull Request

- Fixes the bug wich showed the legs shorter than they should be, and the missing leg for kobolds and verminvolk.
- Changes kobolds to use the anthro_small.dmi sprites, since they're the same and those aren't broken

## Why It's Good For The Game

I personaly consider the sprites working properly good for the game. Also, having 1 sprite for both races makes it easier to change the sprites on the future.


DeerJesus:
kobold and anthro sprites aren't the same - kobold sprites have slightly different bottoms of feet than volk

john-pawson:
> kobold and anthro sprites aren't the same - kobold sprites have slightly different bottoms of feet than volk

while yes i did notice it. I seriously don't think it matters that much.
anthro
koboldft
Top is volk, bottom is kobold. 
It also shouldn't break anything, just make the feet look slightly different. And considering how most people didin't notice the entire foot was missing, i don't think this should be a problem.

PR #441, opened 2024-07-14
Sheriff is no longer a noble

Gyrandola:
## About The Pull Request

He never should have been, it was an oversight on my part.


ChungusGamer666:
https://en.m.wikipedia.org/wiki/Reeve_(England)

Me when the Shire Reeve is not a reeve at all

PR #444, opened 2024-07-14
Fix Bath Swain Latejoining?

Stagrawl:
I'm useful.


coffeecoffeecoffee:
Maybe now there will actually be swains.

CalamaBanana:
Did this truly fix it? Never been able to latejoin as one today at least, even during lowpop where I was quite certain slots should have been free

Stagrawl:
this should have fixed it, but it's possible it's somehow fucked up elsewhere.

CalamaBanana:
All good, paid attention this round. Seems to work after all.

PR #446, opened 2024-07-15
Millstones now grind items into reagents for stamina - adds powder crafting recipe but the millstone still kinda works for it

Slorbin:
Powder can now made by the craft menu. MUCH less tedious than clicking the grain then the millstone then hunting for a new free space to click, given you can now just face the millstone and right click the craft button after your first craft to make quick powder. You can still do the old way too.

More importantly, at the cost of stamina, you can now grind reagents at a millstone in bulk - much faster than mortars but at a price. This will be pretty important for large-scale alchemy in the future. Enjoy!

### DIFF
--- a/code/modules/farming/produce.dm
+++ b/code/modules/farming/produce.dm
@@ -54,6 +54,7 @@
 	can_distill = TRUE
 	distill_reagent = /datum/reagent/consumable/ethanol/beer
 	distill_amt = 24
+	grind_results = list(/datum/reagent/floure = 10)
 
 /obj/item/reagent_containers/food/snacks/grown/oat
 	seed = /obj/item/seeds/wheat/oat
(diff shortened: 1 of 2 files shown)


Useroth:
Why the barrel removal with it?

DeerJesus:
> Why the barrel removal with it?

Oh, that's a leftover from an earlier version of the code. That's just a duplicate recipe for the barrel in the code I noticed needed to be removed - it's the TG recipe. You can still craft it ofc

DorfGuy1:
> > Why the barrel removal with it?
> 
> Oh, that's a leftover from an earlier version of the code. That's just a duplicate recipe for the barrel in the code I noticed needed to be removed - it's the TG recipe. You can still craft it ofc

Must be the TGwood recipe

DeerJesus:
you can now grind grains to get powder without the crafting menu as well.

PR #461, opened 2024-07-16
Moves the bog gate back to be between the bog village and the town, also connects the bog garrison and gate building

zakanater19:
## About The Pull Request

Moves the bog gate back, but the bog garrison building will be connected to the gate building, that is in the same place in the bog village, moves some of the unbreakable walls around the bog gate too, screenshots below

Z3
Screenshot 2024-07-16 170314

Z4
Screenshot 2024-07-16 170327

Me checking the gates work after relocation
Screenshot 2024-07-16 170708

## Why It's Good For The Game

Makes the bog village more dangerous and protected by the gates, although the bog guards will still have to patrol the village now they and the bog towners will encouter antags more often


TheFenix323:
This again, hopefuly it goes better in this server.
Also I would say to move the garrison to be behind the gate, would make more sense in my eyes,

deep-fried-faceplant:
Why would anyone willingly go to this area now? Even the people that spawn here are better off packing up and moving north of the gate. A better idea to make the village less safe is to remove the invincible walls from the rock wall surrounding it, and to add a second gate. Also keep in mind that antagonists are already extremely bloodthirsty with low quality of roleplay.

DeerJesus:
Cool, everyone in the bog now dies to fraggers

PR #464, opened 2024-07-17
Makes scrying orb more expensive

Azarak:
## About The Pull Request
Cost 200 from 120
Alternative to https://github.com/Rotwood-Vale/Ratwood-Keep/pull/388
I also have plans to modify the scrying orb a bit later


Clowntopia:
This is terrible for the merchant 

Your gonna over feed them and they gonna get a tummy ache

Issue #420, opened 2024-07-12
[REQUEST] Deintensification of the Half-Orc and Dwarf accents.


Frick-Frack:
## Is your feature request related to a problem? Please describe. 

Playing Half-orc (and by extension Dwarf) makes roleplay incredibly painful to do. 75% of the time You'll try to talk with someone only to have them be confused and fail to understand you at all. Dwarves as far as I'm aware, have it way worse. Their tics are just far too thick to communicate with and makes roleplay really hard to do, even among other dwarves. It's not hard to imagine why we don't have a lot of Half-orc and Dwarf mains when they have such a debilitating tic. 

## Describe the solution you'd like

I know we don't intend on removing tics and I think that's fine, but I really Dwarf and Half-Orc need their tics reworked so that it's possible to converse with people without them having to solve a Zodiac killer puzzle just to understand them. Just toning it down to a more readable level would be fine enough. Removing maybe half of the tic conversion from the list.

## Describe alternatives you've considered

Alternatively we could go the tiefling route and just completely redo the tics for these races. Though obviously that'd be a lot more work.

## Additional context

No other context.

Hmeister-real:
NO, THEY ARE FUNNY

Drayco84:
They're not very funny to people with English as a second language and even native speakers like myself who have a hard time understanding the super-thick Scottish accent. Thus, only a minority of players here find them funny. So, what joke value they do have is greatly diminished by practicality as we don't have several minutes to internally translate what exactly is being said.

juandexx:
Yea I would love if you can put an option to turn it off to people who are not native english speakers is REALLY DIFFICULT almost impossible to us to understand what are this people tryng to say, also the french elfish everything is really annoyng if You are making an effort to understand what people say to you in English and also to be able to express yourself. Imagine trying to decipher those strange accents in a text, the truth is that they are a headache and many times I cannot communicate with people because I understand shit, it would be nice if they added an option to deactivate it so that people who like this feature can continue enjoying it and those who don't can simply deactivate it so they can enjoy better gameplay and immersion for everyone.

Shaneyk:
To be honest reading through some of the dwarf accent file some of the word changes literally change the mean of what you said to just broken English it isn't even Scottish accent at the point. 

Example Fault = Blame
You are at fault = You are at blame 
Would be said correctly as You are TO blame but since you can't guide it like that what's the point.

Some of the words are so far out towards broken gibberish it's impossible to understand.

PR #470, opened 2024-07-17
Cleaning Up the Dwarf Accent (Update and rename dwarf_replacement.json to dwarfcleaner_replacement.…)

Shaneyk:
Cleaned it up a bit change some outputs. Change the name of the file as i saw in the species code you want to use original for different dwarf type.

Just need to change race accent "code/modules/mob/living/carbon/human/species_types/roguetown/dwarf/dwarfm.dm" if you choose to us it.

From 1450 Lines to about 570.


Shaneyk:
This following up on this https://github.com/Rotwood-Vale/Ratwood-Keep/issues/420

Squird5:
How about instead of replacing it with a slightly different name, you just edit the existing file

PR #471, opened 2024-07-17
Disables Torso

Azarak:
## About The Pull Request
Problematic class that I dont like

Players will soon be able to challenge themselves in another *special* way, that is also rewarding:tm:


atomicplan666:
Do people really abuse torso enough to remove it? god damn,,,

PR #472, opened 2024-07-17
Nudeshots

Gyrandola:
## About The Pull Request

title. Visible only when you're naked.

image


Alanii:
Maybe now we can actually drop PQ for people using those headshots that are 90% breast, and 10% head.

CalamaBanana:
Small nitpick, you listed the aspect ratio the wrong way. Already confused a few people. You wrote 4:3, when it should say 3:4.

PR #475, opened 2024-07-17
Charge and Sprint changes

Azarak:
## About The Pull Request
Change: Charging someone is now less consistent and modifiers such as distance / charging into someone's back affect it
Change: You can change sprint direction if you haven't moved a single tile
FIx: Sprinting no longer lets you sprint one extra tile in a different direction before stopping
Fix: You can no longer charge someone by moving into a different direction on the last step

New charge math:
It checks your "points" against enemy "points" where points are calculated as follows:
Points are equal to Strength + Con, divided by 2, rounded down (ie, 13str, 12 con => 12 points)
You loose 4 points for charging point blank
You loose 2 points for charging from close distance (1 to 2 tiles inbetween)
You gain 1 point for charging from far distance (5 + tiles inbetween)
You gain 2 points for charging into someone's back

The charger now also gets a random (+1, 0, -1) to points to make it less consistent

If you win, you knock down the opponent
If you loose, you get knocked down
If you tie, you both get knocked down

## Why It's Good For The Game
Fixes to unintended behaviours
Makes sprinting easier from an input perspective
Makes charges more balanced, with the point blank abuse cheese being punished by rough 4 points
Makes charges more interesting by making them better for charging into backs of people

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -182,15 +182,37 @@
 
 	if(m_intent == MOVE_INTENT_RUN && dir == get_dir(src, M))
 		if(isliving(M))
+			var/sprint_distance = sprinted_tiles
+			toggle_rogmove_intent(MOVE_INTENT_WALK, TRUE)
+
 			var/mob/living/L = M
-			if(STACON > L.STACON)
-				if(STASTR > L.STASTR)
-					L.Knockdown(1)
-				else
-					Knockdown(1)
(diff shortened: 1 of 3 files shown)


zakanater19:
"Points are equal to Strength + Con, divided by 2, rounded down (ie, 13str, 12 con => 12 points)
You loose 4 points for charging point blank"

So a knight would lose most of the time charging point blank into an MAA? that seems wrong

PR #477, opened 2024-07-17
Adds anti-lag railing to inn stairs

Coolgat3:
https://youtu.be/aBKEt3MhNMM?si=qoRescy0r9xt6t5m&t=52
image
The grave marker has edited description, from
"A simple marker honouring the departed.."
to
"A simple marker honouring those that fell.."

### DIFF
--- a/_maps/map_files/rockhill/rockhill.dmm
+++ b/_maps/map_files/rockhill/rockhill.dmm
@@ -3054,17 +3054,6 @@
 	},
 /turf/open/floor/rogue/ruinedwood/spiral,
 /area/rogue/under/town/basement)
-"dKK" = (
-/obj/structure/fluff/railing/wood{
-	dir = 4;
-	layer = 2.8;
-	pixel_x = 8
-	},
-/obj/structure/gravemarker,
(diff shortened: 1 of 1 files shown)


Drayco84:
**_SLAMS HANDS ON DESK._**
THANK YOU!!!
100% serious, 0% sarcastic here. THANK. YOU.

DeerJesus:
YES WHY DOES LEAVING THE BAR ALWAYS CAUSE MASSIVE LAG

zakanater19:
> YES WHY DOES LEAVING THE BAR ALWAYS CAUSE MASSIVE LAG

it's the music loading for new area, 90%

zakanater19:
could probably map the outside of the inn to be the same map node thing, then the lag will be when you go up/down stairs and hopefully be less noticeable

Alanii:
> could probably map the outside of the inn to be the same map node thing, then the lag will be when you go up/down stairs and hopefully be less noticeable

Move the inn area, and remove moat guard rails. People will somehow end up lagging into the sewers.

PR #486, opened 2024-07-18
Fix Dark elf's gender-swap job roundstart

SarmentiCampbell:
## About The Pull Request
Okay, so. After multiple minutes checking the code, Gender-swapped job/classes for drows seems to only work for latejoiners or adventurers/pilgrims, which meant that it was literally impossible for a female drows to be a queen or to be a roundstart veteran/sheriff/queen consorts

## Why It's Good For The Game
Kinda fix a issue, also i did test it several time and made sure that it didn't fuck anything up but i'll still like this to be test-merged and to get a idea on if something is wrong with it because i genuinely have no idea if this is bad or not (**_It did work tho_**)
image

### DIFF
--- a/code/controllers/subsystem/job.dm
+++ b/code/controllers/subsystem/job.dm
@@ -474,7 +474,7 @@ SUBSYSTEM_DEF(job)
 					JobDebug("DO incompatible with age, Player: [player], Job: [job.title]")
 					continue
 
-				if(length(job.allowed_sexes) && !(player.client.prefs.gender in job.allowed_sexes))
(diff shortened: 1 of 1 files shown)


ChungusGamer666:
this is probably a slop fix and not in line in how the allowed gender swaps are done elsewhere, please look at the latejoin code for reference

ChungusGamer666:
> This probably requires something like `/datum/species/proc/get_job_gender(gender)` and just use that, instead of slapping these extra checks in places

not sure why i did not think of that, but yes very much so

PR #487, opened 2024-07-18
Adds "Modern 2"

coffeecoffeecoffee:
## About The Pull Request
In #468, alongside a slight tweak to the pig's helm I heavily changed the Modern hairstyle once I saw that it's inexplicably bigger than other hairstyles and disproportionately bobbleheaded. This was reverted in #476 and after discussion about how this wasn't just a change that shaved a hairstyle down to fit helmets but an actual resprite and touch-up, I was told somebody actually complained about the change to "their" hairstyle, so this keeps the version I fixed up because I think it's a good sprite while keeping the old hydrocephalic version for that one person's hydrocephalic character.

dreamseeker_7dQHK4ZfwL

## Why It's Good For The Game
Having hairstyles that look like Megamind in certain directions is BAD, but apparently someone likes them. Learning how to add whole new hairstyles is GOOD. Keeps my improved version.

### DIFF
--- a/code/modules/mob/dead/new_player/sprite_accessory/hair.dm
+++ b/code/modules/mob/dead/new_player/sprite_accessory/hair.dm
@@ -406,7 +406,7 @@
 	icon_state = "modern"
 
 /datum/sprite_accessory/hair/head/modern2
-	name = "Modern 2"
+	name = "Modern (New)"
 	icon_state = "modern2"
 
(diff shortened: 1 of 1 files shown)


coffeecoffeecoffee:
Barbie comes in multiple Totally Stylelicious colors!

Untitled33456

Issue #474, opened 2024-07-17
[BUG] Posion arrow uses other poison arrows when crafting


Juubo:
## Describe the bug
when crafting a batch of poison arrows I had a stack of arrows at my feet and crafted my first arrow. After right clicking craft to repeat the recipe, I noticed the arrow moving around and none of the ones at my feet were getting poisoned.

in a later test, I had only one arrow out, I poisoned it, then I crafted a poison arrow again with no arrows out and seemingly poisoned it again.

Each time I poisoned arrows it took poison from my bucket but would prioritize the poison arrow. the poison arrow was not in my hand

## Round ID: 
Prior to Weed3280-1402.2.37

## Steps to Reproduce:
1. set a stack of arrows on the ground
2. prepare your poison source
3. craft a poison arrow
4. repeat
5. arrows will not be taken off the ground as they are supposed to.

## Additional context
most will just keep right clicking to craft poison arrows thinking they are doing a stack

DeerJesus:
this can be fixed pretty easily by making all regular arrows be arrows/iron instead of arrows/poison... however trying to ctrl f replace everything has this disgusting side effect of unwrapping the dmm

PR #490, opened 2024-07-18
fixes [BUG] Posion arrow uses other poison arrows when crafting #474

Slorbin:
yeah this should work


DeerJesus:
ermm i did something wronggg

PR #512, opened 2024-07-21
fixes #474 for real (OK YOU CAN MERGE NOW)

### LINKED ISSUE #474: [BUG] Posion arrow uses other poison arrows when crafting

Slorbin:
#474

### DIFF
--- a/code/modules/clothing/rogueclothes/quiver.dm
+++ b/code/modules/clothing/rogueclothes/quiver.dm
@@ -62,7 +62,7 @@
 /obj/item/quiver/arrows/Initialize()
 	..()
 	for(var/i in 1 to max_storage)
-		var/obj/item/ammo_casing/caseless/rogue/arrow/A = new()
+		var/obj/item/ammo_casing/caseless/rogue/arrow/iron/A = new()
 		arrows += A
 	update_icon()
 


DeerJesus:
HUH? WHAT? WHY DOES IT KEEP FUCKING UNWRAPPING?

DeerJesus:
THERE we go. FUCK git hooks

Issue #422, opened 2024-07-12
[BUG] You can't look at people while in chains


Hmeister-real:
## Describe the bug

While in chains, you cant turn yourself or examine anything

## Steps to Reproduce:

become chained/roped, try to turn or examine something

Drayco84:
I know this also happens when lying down, were you standing and trying to examine people while in chains?

Hmeister-real:
> I know this also happens when lying down, were you standing and trying to examine people while in chains?

both standing, and another time whilst laying

PR #526, opened 2024-07-22
Fixes some click inputs

Azarak:
## About The Pull Request
This fixes examines while cuffed, and maybe some other stuff to
It may break some things? But I dont think so


Hmeister-real:
fixes #422

PR #489, opened 2024-07-18
Moves Dark Elves and Tieflings to "RACES_SHUNNED"

Alanii:
# Why It's Good For The Game

Tieflings are literal demon spawn
- They look like demons, what more can you ask?

Dark Elves were creatures of zizo (supposedly)
 - Their stats were built for being a shunned race, not knights, or proper soldiers.
 - Dark Elven women were given stats that allowed them to robust other races pretty easily, even half-orcs. They were balanced around being able to have a chance in a fight when fighting people who accosted them for being a dark elf. 
 

Yes, I am aware I am going to get a lot of hatred, but until the lore makes these races less of a "shunned" race, this is the presumed ideal position these races will be.

PR #529, opened 2024-07-22
Drows and Tieflings are shunned

Azarak:
## About The Pull Request
Just like https://github.com/Rotwood-Vale/Ratwood-Keep/pull/489
If we're making a change on what people can play we can not undo it then do it again and fiddle like that

### DIFF
--- a/code/__DEFINES/roguetown.dm
+++ b/code/__DEFINES/roguetown.dm
@@ -34,8 +34,6 @@
 	/datum/species/dracon,
 
 #define RACES_TOLERATED \
-	/datum/species/elf/dark,\
-	/datum/species/tieberian,\
 	/datum/species/lizardfolk,\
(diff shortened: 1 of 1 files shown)


coffeecoffeecoffee:
Hell yeah, this is much better. I'm also totally game for reducing the advantages female dark elves get, since they're not going to be outright sunned given they're a fairly well-spoken race: They probably shouldn't be SO much stronger, even as a massive proponent of them, I'll admit.

Jul-Jul:
I said this in the Dev chat. Them being moved to shunned makes sense considering their stats (I know its a lore reason too). They were always meant to be be shunned and never within any powerful combative roles which was why drow were given high stats so they wouldn't get shit stomped all the time. Maybe in another PR should consider giving other shunned races some bonuses too.

PR #530, opened 2024-07-22
NPC ambushes by mobs made more likely to fire

zakanater19:
## About The Pull Request

Makes the trigger for spawning ambushes fire more often

## Why It's Good For The Game

Makes the bog more dangerous if you don't know how to avoid ambushes i guess, or if you are running from someone and don't have the time to move properly in the boglands


zakanater19:
Actually im retarded

PR #531, opened 2024-07-22
Ambiently spawns hostile ambush mobs in the boglands

zakanater19:
## About The Pull Request

instead of hostile mobs only being in the bog when triggered by an ambush, this will populate the bog with hostile mobs overtime, currently the values are set at a maximum of 60 spawned mobs, they will slowly trickle into existance as the round continues, you can read the commit changes above to see the exact values of other variables

## Why It's Good For The Game

Makes the bog more dangerous even to careful players

Me testing they spawned right using insane values to spawn them all at once (by all at once i meant i turned the max up to 300 and set the seconds to 1, and chance to spawn at 100)

Screenshot 2024-07-22 180720

Screenshot 2024-07-22 180751

### DIFF
--- a/code/game/area/roguetownareas.dm
+++ b/code/game/area/roguetownareas.dm
@@ -207,18 +207,28 @@ GLOBAL_LIST_INIT(roguetown_areas_typecache, typecacheof(/area/rogue/indoors/town
 
 	proc/spawn_and_continue()
 		if(current_spawned_mobs < max_spawned_mobs)
-			spawn_random_mobs()
+			spawn_random_mobs(3) // Specify the number of mobs to spawn
 		start_spawning()
 
-	proc/spawn_random_mobs()
+	proc/spawn_random_mobs(var/num_to_spawn)
 		var/spawn_chance = 100 // Adjust this value to control how often mobs spawn (default 20 say)
 		if(prob(spawn_chance) && current_spawned_mobs < max_spawned_mobs)
-			var/turf/spawn_turf = get_random_valid_turf()
-			if(spawn_turf)
-				var/mob_type = pickweight(ambush_mobs)
-				var/mob/new_mob = new mob_type(spawn_turf)
-				if(new_mob)
-					current_spawned_mobs++
+			var/turf/spawn_turf
+			var/mob_type
+			var/mob/new_mob
+			var/i = 0 // Initialize counter variable
+			while(i < num_to_spawn) // Use while loop to spawn mobs
+				if(current_spawned_mobs >= max_spawned_mobs)
+					break
+				spawn_turf = get_random_valid_turf()
+				if(spawn_turf)
+					mob_type = pickweight(ambush_mobs)
+					new_mob = new mob_type(spawn_turf)
+					if(new_mob)
+						current_spawned_mobs++
+				i++ // Increment counter
+					
+			
 
 	proc/get_random_valid_turf()
 		var/list/valid_turfs = list()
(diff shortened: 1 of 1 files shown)


zakanater19:
> I like the idea but holy moly that looks like a LOT of mobs.

Yes because i made it spawn 300 instead of 60, and then i made the chance per check 100%, and also made the check every second

zakanater19:
When i launched the server with the default config in this file there was no spawned mobs at roundstart yet, like i said the current values trickle them in over the course of the round, up to 60 maximum

CalamaBanana:
It would be nice if they would not spawn near player build structures, though. Would be weird to constantly have a skeleton pop up in your fortified settlement.

Or perhaps a simpler solution, only have them spawn in dark tiles? Would lead to no random mobs popping up in well lit settlements, and punish those living in shoddy hovels without lights.

zakanater19:
> It would be nice if they would not spawn near player build structures, though. Would be weird to constantly have a skeleton pop up in your fortified settlement.
> 
> Or perhaps a simpler solution, only have them spawn in dark tiles? Would lead to no random mobs popping up in well lit settlements, and punish those living in shoddy hovels without lights.

They can only spawn on dirt/grass/water tiles

CalamaBanana:
> > It would be nice if they would not spawn near player build structures, though. Would be weird to constantly have a skeleton pop up in your fortified settlement.
> > Or perhaps a simpler solution, only have them spawn in dark tiles? Would lead to no random mobs popping up in well lit settlements, and punish those living in shoddy hovels without lights.
> 
> They can only spawn on dirt/grass/water tiles

This means I need to pave the entire perimeter of my bog-house to not constantly have mobs interrupt my RP. That will just end up looking bad visually. There needs to be something to stop them from spawning close to players and/or their homes. Light is the easy solution would be my guess.

 (And you need a little dirt in your base for the farm, too)

burdwatcher:
I'd rather just have a lot more of them spawned in at round start randomly across the map. So once they're cleaned of the area you're working on, they'll only spawn via ambushes, We already have mob spawns, no need to double up on it. Would definitely be the easier solution in not having them randomly spawn too close to players too unless you want to put in a minimum distance to the nearest player which would probably be a lot more complicated.

zakanater19:
> unless you want to put in a minimum distance to the nearest player which would probably be a lot more complicated.

No it's quite simple, ill get around to adding a 10-20 tile radius around players so that mobs don't spawn ontop of players

CalamaBanana:
Checking for light might be simple on a first glance.
This one is used in the code for sneaking (turns you invisible in low light):
var/light_amount = T.get_lumcount()

In: https://github.com/Rotwood-Vale/Ratwood-Keep/blob/5766757dac9c61a0347e34451f73f796ea8c0ed3/code/modules/mob/mob_movement.dm#L594

zakanater19:
> Checking for light might be simple on a first glance. This one is used in the code for sneaking (T is the turf checked, used to turn sprites invisible in low light if sneaking): `var/light_amount = T.get_lumcount()`
> 
> In:
> 
> https://github.com/Rotwood-Vale/Ratwood-Keep/blob/5766757dac9c61a0347e34451f73f796ea8c0ed3/code/modules/mob/mob_movement.dm#L594

the light blocking spawning is simple too,  if(T.get_lumcount() > 0.2)

Just not sure if i want to add it, i probably will add it, just give me like 10 mins and ill add both into this PR

zakanater19:
Right, mobs won't spawn on light levels above 0.2, and a 15 tile radious around players

CalamaBanana:
> Right, mobs won't spawn on light levels above 0.2, and a 15 tile radious around players

Thanks!

How dark is it under leaves? Perhaps it might be good to spawn mobs at night only, so the few dark tile during day do not get mob-bombed. Or perhaps that could be fun to leave in, having large gathering of mobs in one place sometimes.

zakanater19:
> > Right, mobs won't spawn on light levels above 0.2, and a 15 tile radious around players
> 
> Thanks!
> 
> How dark is it under leaves? Perhaps it might be good to spawn mobs at night only, so the few dark tile during day do not get mob-bombed. Or perhaps that could be fun to leave in, having large gathering of mobs in one place sometimes.

Well the bog biome is pretty large, so you won't see groupings under any one tree since there are like 400+ trees in the bog? something like that, and a max cap currently on spawned mobs is 60, im still playing around locally with the numbers but as is in this PR it's not lagging the server

zakanater19:
Fixed some issues relating to the max mob spawning, it wasn't working right, now it works right, also the mob spawning rate was changed, they filter in slower now and it's less server performance heavy

zakanater19:
That should be final update, added a check so they cannot spawn within 10 tiles for an adventuer latejoin/roundstart marker, also lowered the max spawned mobs to 40 for performance reasons

zakanater19:
requested alternative https://github.com/Rotwood-Vale/Ratwood-Keep/pull/573

PR #533, opened 2024-07-22
Health potion buff

DuskLight9978:
## About The Pull Request
This requested PR allows healing potions to repair wounds such as fractured ribs, aortic dissections, etc. the potion heals 2 woundHP per .5 units imbibed. If you are fucked up you will need several units.

## Why It's Good For The Game
This is the requested alternative to miracles or surgeons without removing either. This also heals over time so it cannot be used to become the god of combat. This was requested by the host and they were kept in the full loop of development.


zakanater19:
I want to upvote but will this be the definate point of self heals outside of RP?

DuskLight9978:
> I want to upvote but will this be the definate point of self heals outside of RP?

This only buffs health potions as a reagent, the scarcity of said potions can still be altered by another PR.

PR #536, opened 2024-07-23
Migration system improvements

Azarak:
## About The Pull Request
Bandit waves are slightly more common
Bandit waves now start in the road to the bandit camp
All migrants are unset from the queue if they dont end up in a wave that started
Migration now gives text feedback on joining/leaving queue and prioritizing roles
Migrants can now have their titles be displayed on examines (This does not apply for adventurers and pilgrims)
Announcements (think King's) can now properly use migrant titles
Changes Heartfelt role degradation so it looses (prior > magus > lady) (previously (magus > knight > prior)

admininstation:
Migrant wave spawning shows up on message admins with a JMP topic
The status tab now shows a little status on migration system

### DIFF
--- a/code/datums/migrants/migrant_role.dm
+++ b/code/datums/migrants/migrant_role.dm
@@ -20,6 +20,7 @@
 	var/grant_lit_torch = FALSE
 	/// Whether to show wanderer examine like adventurers or pilgrims
 	var/show_wanderer_examine = TRUE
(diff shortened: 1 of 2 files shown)


zakanater19:
> Migrants can now have their titles be displayed on examines

Even for normal roles like cheesemaker ect? i always liked how if you didn't take towner nobody would know your trade

PR #540, opened 2024-07-23
Master Blacksmith up to modern code

Teragenic:
Makes master Blacksmith code not need to repeat for females.
Gives Master Blacksmith some skills from blacksmith.
Gives Master Blacksmith stats from normal Blacksmith.

Why it's good. 

It just gives Master Blacksmith normal blacksmith things.


Teragenic:
image

image

PR #541, opened 2024-07-23
Hair fix(very long with fringe only)

wow50000:
## About The Pull Request

uh, this probably makes vlongwfringe look less ass
image

turns this
image
into this

## Why It's Good For The Game
better hair=happier me


YennyWR:
I like this

PR #542, opened 2024-07-23
You can heat/cook stuff over a forge now, tell temperature with a thermoscope, grind a lot more stuff, and changes some recipes that were technically craftable to be more lorefriendly.

Slorbin:
A few chemical recipes turned out to be craftable, albeit it was really fucking hard to make them (fentanyl from heating space drugs and multiver from ash and salt): They have been given appropriate theme names and fentanyl actually does something.

It was also incredibly weird but possible trying to heat any liquid source to high temperatures using non-hearth methods, so I added the ability to heat reagents to greater temps on top of forges. 

I also changed 'glass pot' to 'metal pot' because it was meant to be a metal pot - you literally craft it out of metal. The guy who just edited it assumed everything that started with /glass/ in the code meant it had to be made of glass, ignoring shit like the wooden buckets... Check the code comments, they're very stupid.

You could technically make fentanyl by heating grinded dried swampweed. You can still do that but the drug has been renamed to 'Goblin' and gives you the appropriate moodlet (works better on the stress refactor tm).

You can now grind a lot more things that I meant to make grindable originally but didn't have the time to add. Ores and ingots and organs and such.

Most importantly, you can now make a thermoscope to detect the temperature of the reagents in a container. It's made by mining for Cinnabar, a relatively uncommon ore, grinding it into mercury, and crafting it with a glass bottle. It's the lowest level alchemy craft and gives level 0 high int people an in to start getting alchemy.

You can only use the thermoscope if you are at least a level 1 alchemist, and you can only use it with actual precision if you are a level 4 alchemist. It uses 'room temperature' to determine the heat so even then it's a bit hard to read given how lowtech it is.

### DIFF
--- a/code/modules/power/roguelighting.dm
+++ b/code/modules/power/roguelighting.dm
@@ -770,7 +770,7 @@
 						food = C
 			if(istype(attachment, /obj/item/reagent_containers/glass/pot))
 				if(attachment.reagents)
-					attachment.reagents.expose_temperature(400, 0.033)
+					attachment.reagents.expose_temperature(1000, 0.033)
 					if(attachment.reagents.chem_temp > 374)
 						boilloop.start()
 					else


TomGarou:
But isn't the 'glass' pot made out of stone and not metal currently? 
it's an accessible means for survivalists to cook water over hearths for many folks.
Unless it's *another* glass pot made out of metal, in which case both should be fixed?

DeerJesus:
> But isn't the 'glass' pot made out of stone and not metal currently? 
> 
> it's an accessible means for survivalists to cook water over hearths for many folks.
> 
> Unless it's *another* glass pot made out of metal, in which case both should be fixed? 

it's ambiguous, you can make it out of stone OR metal so I just renamed it Pot

DorfGuy1:
More actual alchemy stuff is nice. The pot thing is a nice flavor thing.

DorfGuy1:
Why'd ye remove the Refillable regent tag on the pot?

DeerJesus:
> I'm not sure about making Charcoal easier to make, or mixable at all. I like how deadly poison is, especially killer ice

I could comment out the recipe perhaps. I still think it would be nice to have SOME way to get poison out of one's system because of how absolutely DEVASTATING overdosing on swampweed is.

DeerJesus:
> I'm not sure about making Charcoal easier to make, or mixable at all. I like how deadly poison is, especially killer ice

What if I made Charcoal half as effective as it currently is, and made it require heat to make?

DeerJesus:
Hearths can now heat to maximum, though they're a third as fast as a forge

EzekyalLegit:
> > I'm not sure about making Charcoal easier to make, or mixable at all. I like how deadly poison is, especially killer ice
> 
> I could comment out the recipe perhaps. I still think it would be nice to have SOME way to get poison out of one's system because of how absolutely DEVASTATING overdosing on swampweed is.

We have leeches for this.

PR #543, opened 2024-07-23
Chat to 1000 (actually 200)!

Kepteyn:
## About The Pull Request

Simply increases the chat's maximum lines from 100 to 200.

## Why It's Good For The Game

Currently, with how short the chat box is, if you spend about thirty seconds in a busy bar or shift examine someone two or three times by accident, anything said before disappears. This simply increases it to a more reasonable size.

### DIFF
--- a/interface/skin.dmf
+++ b/interface/skin.dmf
@@ -120,7 +120,7 @@ window "outputwindow"
 		anchor2 = 100,100
 		background-color = #000000
 		is-default = true
-		max-lines = 1000
+		max-lines = 200
 		enable-http-images = true
 	elem "input"
 		type = INPUT


Kepteyn:
> 1000 is too much Plus I'm fairly sure it's laggy with the old chat

I would inclined to agree. I will change this soon.

coffeecoffeecoffee:
I think 1000 with some lag is better than 100 with less lag honestly, this is such a huge quality of life improvement for more RP-inclined folk and especially for dealing with ahelps as a player, holy fuck. 500 might be even better than either.

Kepteyn:
Value changed to 200 at the request of Azarak. Personally, I think 200 is a bit too low, but we will see how it goes!

kepesk1:
Thanks guys

PR #549, opened 2024-07-23
Some helmet rebalancing

coffeecoffeecoffee:
## About The Pull Request
Hey. Did you know the visored sallet mysteriously has a not-infuriating vision cone? It's the only one that lets you see 50% of the way around you out of all the face-protecting helmets in the game. This is supposedly because it's a "light" helmet, with less protection than it's "heavy" siblings like the pigface bascinet. Wanna know something else that's interesting?

Besides durability, these helmets actually all have the same protective values.

The only difference is that the "heavy" helmets with horrendous vision cones have double the durability at 400, while most of the non-face-protecting ones have 200. This, frankly, doesn't make any sense, and as good face-protecting helmets are virtually mandatory for combat-focused roles alongside neck protection, this forces players to play with 25% vision radius because... "skill issue", I guess? This increases the "default" vision radius of masked helms to 50% as a quality of life thing while preserving the balance of the helmets. It's still reduced from normal by a noticeable margin, just like a hood.

"But coffee!" you may ask, "What will become of the visored sallet, since it's one redeeming quality was being a visored helm with shit durability but decent vision!" The answer is I've given it "heavy" durability, making it an identical and equally viable choice. `The normal sallet as well, as it is just the visored sallet without the face visor` EDIT: no I did not, and it _is_ not

"But coffee!" you may ask, "30 or 40% seems more reasonable!" And I would say, I'd love to do this, but the values that seem to exist are "left", "right" and "behind". Coding entirely new vision values is far beyond my ability. If anyone can do it, please, stand up, it'd be even better, but I've tested this, it looks good, it feels good, and it's still a meaningful vision trade-off, it's just no longer ridiculous and annoying, especially for people meant to be standing guard in places. The protective values are all the same as before (that is to say, there was no difference between the two I'm mostly sure.)
dreamseeker_PlqdfizfBW

## Why It's Good For The Game
No more of everybody only using the same 2 fucking helmets that don't feel like absolute ass to use.


Teragenic:
There's nothing implying the Visored Sallet is lighter than the rest. The armor values are the same because that's just how armor works. Infact if you look at obj/item/clothing/head/roguetown/helmet

It's meant to be that way otherwise with how damage and AP works armor would be ignored / passthrough.

Sallet needs to be nerfed. It's one bar.

The reason we had the heavy helmets get more durability is because it's HEAVY. It's meant to be more protective. Most combat is durability based. If you make everything special nothing is. 

Visored helmets are now better in ALL ways now because they can be toggled. No reason to get heavy helmets without toggles.

Please remove 
> This is supposedly because it's a "light" helmet, with less protection than it's "heavy" siblings like the pigface bascinet. Wanna know something else that's interesting?
> Besides durability, these helmets actually all have the same protective values.

Because that's the point. If helmets acted like SS13 armor where damage goes through it would not make sense.

Teragenic:
Also what's your discord Dev-Chat is a better place for this

H0gberg:
the visored sallet does not protect your mouth, that is why it is better vision wise, there is balance to it.

coffeecoffeecoffee:
I think you guys answered your own question. The balance becomes "the visored sallet is a heavy helmet that costs one ingot, but physically leaves part of your face completely unprotected". This seems reasonable to me. However, if you guys just want it to be lower-durability, we can do that as well, I just thought it'd only be fair.

coffeecoffeecoffee:
After discussion in dev-general, people generally seem to feel the visored sallet should just straight-up be brought in line with other "heavy" helms with this change, and so it needs mouth protection. The sprite visibly doesn't HAVE face protection, so I've taken it upon myself to add it. I hope that visored sallet fans will be pleased with this slight addition and the improvement of mouth protection to it.

dreamseeker_qIxFyUIdib

I've slightly changed the inhand sprite to reflect this as well.

dreamseeker_JQRrJiGRYE

I REALLY HOPE THIS WORKS THE

coffeecoffeecoffee:
oops, that's very slightly the wrong sprite that slipped in, i was experimenting with removing the gigachad cheekbones from it. here's the original with the slats that I meant to put in, but we could use either.

dreamseeker_6RSHhGOuPe

john-pawson:
While i like this, isn't this just going to make all of the helmets basically exact copies of eachother? before it was nice to be able to choose to trade higher FOV for lower durability/protection with the visored sallet, but now it's just purely aesthetic (not including light helmets)

coffeecoffeecoffee:
> While i like this, isn't this just going to make all of the helmets basically exact copies of eachother? before it was nice to be able to choose to trade higher FOV for lower durability/protection with the visored sallet, but now it's just purely aesthetic (not including light helmets)

Pretty much, but that was _almost_ already the case for 99% of the helmets, with the visored sallet as the only odd one out for some strange reason. There are also, mind you, all of the light helmets, like the unvisored sallet, the plain bascinet and the skullcap, if you want full unimpeded vision at the cost of face protection.

Sarkness:
As the makers of the Patrons Helmets (Necra, Dendor, etc.)
The reason the helmets have such a shit FOV were because they were just "reskins" of the default templars helmets at the time, which were the bucket helmets.
Blackstone staff asked me to keep it similar and as such, did what I was told to do.

Honestly I was very tempted to increase it's FOV but as it is, being 50% is better than 25%, still

PR #555, opened 2024-07-24
Teifling skin tones

OozeSlimeJelly:
## About The Pull Request

Adds three new skin tones for tieflings with three ancestry names I found fitting, since like Castile they represent Iberian kingdoms before the unification of Spain. I assume that's the inspiration of where tieflings are from since they're "Tieberian" in the code and used to speak with a terrible Spanish accent.

The new ancestries are Leonese, Aragonese and Navarrese, and the new tones are white grey and purple respectively.

## Why It's Good For The Game
More customization and generally a more varied look to the tieflings in game. As they are now you only really see a few very similar looking types running around which stands out against the massive amount of diversity in color elsewhere. Even Orcs have blue and red. Tieflings have the least amount of color choices in game out of the more standard species iirc and if nothing else it hurts nothing and changes little while making a few people happy.

2a
3a
4a

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/roguetown/roguetown_species.dm
+++ b/code/modules/mob/living/carbon/human/species_types/roguetown/roguetown_species.dm
@@ -16,12 +16,12 @@
 
 /datum/species/proc/handle_speech(datum/source, mob/speech_args)
 	var/message = speech_args[SPEECH_MESSAGE]
+	
+	message = treat_message_accent(message, strings("accent_universal.json", "universal"), REGEX_FULLWORD)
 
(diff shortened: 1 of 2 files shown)


SayonaraEcho:
@OozeSlimeJelly, it's good to see man of culture. Anyway i have a question, why is there no blue and yellow-orange skin for tieflings? I think that would be more better.
image

OozeSlimeJelly:
> I dont like them as it'll make them less distinct visually, the purple one fits somewhat, but I think could use a little bit more saturation?
> 
> I'd like if this was just the purple one saturated slightly more

I get you, I'll close this and see if I can do one that's just a better purple.

PR #558, opened 2024-07-24
Nerfs healing potion blood volume regen

Azarak:
## About The Pull Request
Halves it, because one healing potion could potentially heal 4 people's amount of blood worth
This matters for vampires a lot, and it was rather cheesy

### DIFF
--- a/code/modules/roguetown/roguejobs/alchemist/reagents.dm
+++ b/code/modules/roguetown/roguejobs/alchemist/reagents.dm
@@ -10,10 +10,10 @@
 
 /datum/reagent/medicine/healthpot/on_mob_life(mob/living/carbon/M)
 	if(M.blood_volume < BLOOD_VOLUME_NORMAL)
-		M.blood_volume = min(M.blood_volume+50, BLOOD_VOLUME_MAXIMUM)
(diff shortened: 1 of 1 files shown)


DavidAreYouSure:
Water Buckets are just as good. 1 Vamp Spawn with a bucket can give infinity blood with little effort

coffeecoffeecoffee:
THANK GOD THIS INFINITE BLOODPIGGY STUFF IS GETTING FIXED. Please change it for water as well - I thank you for this!

MickDick1:
It's a pretty extreme nerf to everyone else that relies on it, not just vampires. It actually makes more sense to just succ someone by feeding them water than waste the healing potions. 

It will now be impossible to revive someone mundanely from an artery draining them as it will not refill blood fast enough to stabilize. Could we have it heal oxygen enough to stabilize someone below the minimum?

DorfGuy1:
So vampires, who until very recently did f all until the last 15 minutes, abused some thing, had their access to it removed, and now everyone else is being punished for it. Go figure.

DorfGuy1:
> Although you guys might be right because arteries do bleed out really fast...
> 
> I'll tackle this some other way

cool and good.

PR #561, opened 2024-07-24
Cuts the multi-tails and tentacles off Wild-kin and Wild Half-kin

Alanii:
## Removes Multi-Tails and Tentacles.

They don't fit.

I did not remove the sprites from the code. I may do that if it is requested, but could just leave it in. 

Maybe they'll be used for a new creature, race, or special event-type monster.

_cough_ Succubus Nine-Tail Fox spirits _cough_


Alanii:
> Remove the datum definitions and the sprites too, please

Okay, I'm about to play a game so I'll do it when I'm free in game or after the round.

PR #564, opened 2024-07-24
Drugs. Salt. Drugs and Salt.

Luctus-Comptus:
made ozium, moondust and salt craft able

### DIFF
--- a/code/modules/mob/living/living_movement.dm
+++ b/code/modules/mob/living/living_movement.dm
@@ -52,10 +52,11 @@
 			mod = CONFIG_GET(number/movedelay/run_delay)
 		if(MOVE_INTENT_SNEAK)
 			mod = 6
-	if(STASPD < 6)
-		mod = mod+1
-	if(STASPD > 14)
-		mod = mod-0.5
+
+	var/spdchange = (10-STASPD)*0.1
+	spdchange = clamp(spdchange, -0.5, 1)  //if this is not clamped, maniacs will run at unfathomable speed
+	mod = mod+spdchange
+	//maximum speed is achieved at 15spd, everything else results in insanity
 	add_movespeed_modifier(MOVESPEED_ID_MOB_WALK_RUN_CONFIG_SPEED, TRUE, 100, override = TRUE, multiplicative_slowdown = mod)
 
 /mob/living/proc/update_turf_movespeed(turf/open/T)


burdwatcher:
Just going to add a side note here that in striving to make this like your average ss13 code base where you can craft everything you also undermine the value of coin, sometimes it's better to have things uncraftable.

Moribun:
> Just going to add a side note here that in striving to make this like your average ss13 code base where you can craft everything you also undermine the value of coin, sometimes it's better to have things uncraftable.

these items are worth 5 mammoms when crafted. the value of what goes in is worth more than what comes out. if ya had looked at the sale value before talking.

Gr33d-y:
holy based. What skill does it take to craft?

burdwatcher:
> > Just going to add a side note here that in striving to make this like your average ss13 code base where you can craft everything you also undermine the value of coin, sometimes it's better to have things uncraftable.
> 
> these items are worth 5 mammoms when crafted. the value of what goes in is worth more than what comes out. if ya had looked at the sale value before talking.

That's not what I'm talking about. What I'm talking about is that in introducing more items to be crafted you're reducing the value of money as people can just work around buying it in the first place. Everyone is already covered in mammom, not everything has to be craftable.

Moribun:
> holy based. What skill does it take to craft?

alch novice for salt, alch journey for drugs.

Moribun:
> > > Just going to add a side note here that in striving to make this like your average ss13 code base where you can craft everything you also undermine the value of coin, sometimes it's better to have things uncraftable.
> > 
> > 
> > these items are worth 5 mammoms when crafted. the value of what goes in is worth more than what comes out. if ya had looked at the sale value before talking.
> 
> That's not what I'm talking about. What I'm talking about is that in introducing more items to be crafted you're reducing the value of money as people can just work around buying it in the first place. Everyone is already covered in mammom, not everything has to be craftable.

Im sorry should I further expand upon the fact this is a precursor to an advancement in alchemy I plan to do in the future when I add an alembic and this is used as base items in crafting new things? or would you rather pay 44 mammons per ozium for crafting an item worth maybe 15? your argument is moot.

Wildwestmars:
Hell yeah! Making alchemists feel more like alchemists, a rare skill that is under-utilized.
I hope you continue to add more crafting recipes like this in the future.

PR #565, opened 2024-07-24
Removes meta with commend someone

Azarak:
## About The Pull Request
Nuff said

### DIFF
--- a/code/modules/client/client_procs.dm
+++ b/code/modules/client/client_procs.dm
@@ -1135,8 +1135,15 @@ GLOBAL_LIST_EMPTY(external_rsc_urls)
 	var/list/selections = GLOB.character_ckey_list.Copy()
 	if(!selections.len)
 		return
-	var/selection = input(src,"Which Character?") as null|anything in sortList(selections)
+	var/selection
+	if(SSticker.current_state == GAME_STATE_FINISHED)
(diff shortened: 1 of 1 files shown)


glitchbow:
What is even the point of this? I don't think it's "meta" to be curious of what characters are currently in-round without clarifying what their roles are. SS13 fully let players ghost around, a list of names and nothing else seems blind enough. Announcements from the king and other major players are still visible from the start menu, is that too "meta"? Is roundstart listing off the names of who is applying for certain roles "meta"? It seems lame to implement something like this before even implementing a way to cryo out and leave the round/reopen a role to new players.

Also, this has been broken every time I use the text box. I tried to commend someone that I knew was playing and it still couldn't find them. :/ Tried their first name, last name, full name without caps, and full name perfectly spelled and it still didn't work. There's als the future possibility of people having duplicate character names and giving the wrong person a commend.

glitchbow:
> It's to prevent scanning people and antagonists. Names cannot be duplicate. This is not SS13 and ghosts are not privy to all information in the round I'll see about making this look for people easier so it's not case sensitive, etc.

I can understand preventing antag scanning, but I still think it would be nice to see names of nonantags, since Whom is virtually worthless and often can't even display the whole list of anonymous and nonanonymous names because of how populated the game is. Suffering from success...

PR #568, opened 2024-07-24
Fixes zombie mind unlink and burial rites

Azarak:
## About The Pull Request
I tested burial rites, but not the zombie mind linking stuff, 90% confidence it will work

### DIFF
--- a/code/modules/mob/living/carbon/spirit/spirit.dm
+++ b/code/modules/mob/living/carbon/spirit/spirit.dm
@@ -199,7 +199,7 @@
 				continue
 			if(pacify_coffin(stuffing, user, deep, give_pq = FALSE))
 				success = TRUE
-	// Success is actually the ckey of the last attacker so we can prevent PQ farming from fragging people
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/spirit/spirit.dm
```diff
@@ -182,19 +182,23 @@
...
 	if(deep)
 		for(var/atom/movable/stuffing in coffin)
 			if(isliving(stuffing) || istype(stuffing, /obj/item/bodypart/head))
 				continue
-			success ||= pacify_coffin(stuffing, user, deep, give_pq = FALSE)
+			if(pacify_coffin(stuffing, user, deep, give_pq = FALSE))
+				success = TRUE
 	// Success is actually the ckey of the last attacker so we can prevent PQ farming from fragging people
 	if(success && give_pq && user?.ckey && (user.ckey != success))
 		adjust_playerquality(give_pq, user.ckey)
```
DeltaFire15, line 204: Doesn't you changing the return value of `pacify_coffin()` / `success` brick this check ( `TRUE != some ckey` is always TRUE)

(hi hi)

PR #573, opened 2024-07-24
alternative for #531

zakanater19:
Does the same as https://github.com/Rotwood-Vale/Ratwood-Keep/pull/531

But does not spawn them directly into the bog biome area, but instead spawns mob spawners in that area and then spawns the mobs around those marker spawners, you can read the commits to see exactly what was changed

seems to have the same performance impact as the other method
Screenshot 2024-07-25 001723


zakanater19:
mob spawners now only spawn one mob per spawner and keep track of it, and the spawn waves are now RNG between 60 and 120 seconds, also the new maximum for mob spawned mobs is 30

zakanater19:
del to qdel also breaks everything

zakanater19:
Great now mobs dont spawn at all

zakanater19:
And now they do

PR #577, opened 2024-07-25
Changes to Nudist

dotCatshark:
## About The Pull Request

Changes nudist stats to make it more viable and hopefully doesn't die as soon as they get crit due to their total inability to wear armor. Changes nudist skills to be more in like with his survivalist nature (adds cooking and butchering and that's it), along with a slight buff ~~(potential high rolls)~~.

## Why It's Good For The Game

Nudist needed to be made an actual survivalist instead of a glorified combat class. This should help with that and make playing a weird naked dude in a swamp less annoying.


Hmeister-real:
Randomness is looking to be removed, so remove the picks. Also conflicts with #516 and it's nudist changes

dotCatshark:
> Also conflicts with #516 and it's nudist changes

Nudist changes made by #516 don't buff nudist con enough and add skills that aren't actually needed. Also, why increase the int debuff? It's the largest in the game as is iirc.

dotCatshark:
> > Nudist changes made by #516 don't buff nudist con enough and add skills that aren't actually needed. Also, why increase the int debuff? It's the largest in the game as is iirc.
> 
> I got screamed at for making the Con buff +3 by like, everybody + they should be needed since the ideal Nudist is basicly just a caveman and should live off the bog

What you're describing is your idea nudist. I've talked to a few players with very different ideas, including things like mood buffs for living in the bog in camps. Also, people screaming at you about the con buff seem to be forgetting about the fact that you cannot wear any kind of protection as nudist, so any crit is crippling, since people can just skullcrack you easily.

PR #580, opened 2024-07-25
Breaking grabs makes you resist them for 1.5 seconds

Azarak:
## About The Pull Request
Breaking from grabs makes you resist them for 1.5 seconds
You resist them automatically with no cost, the grabber still gets click cd

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -424,7 +424,7 @@
 				ContactContractDisease(D)
 
 		// Makes it so people who recently broke out of grabs cannot be grabbed again
-		if(TIMER_COOLDOWN_RUNNING(src, "broke_free") && M.stat == CONSCIOUS)
+		if(TIMER_COOLDOWN_RUNNING(M, "broke_free") && M.stat == CONSCIOUS)
 			M.visible_message(span_warning("[M] slips from [src]'s grip."), \
 					span_warning("I slip from [src]'s grab."))
 			log_combat(src, M, "tried grabbing", addition="passive grab")


coffeecoffeecoffee:
THANK GOD.

coffeecoffeecoffee:
> this is going to make restraining people an absolute last resort for me lmfao, all this will do is further reinforce the strength stat meta by making grappling pointless unless you have a significant strength difference.
> 
> automatically resisting grabs is fine I guess but its a skill issue if players cannot resist grabs adequately.
> 
> breaking from grabs making you resist them would be fine but 1.5 seconds is way too long, in that time I can pull out a sword and kill you instead, make it half a second, just to prevent grab spamming (which is a non issue as is anyway)

This is literally exactly the idea, because players are going around insta-winning fights in ways that can't be defended against unless you beat them in strength by constantly grabbing and instantly dropping and headslamming you. This blatantly shouldn't be possible or viable against armed opponents slashing you, but it works every single time, having been on both sides of it. It's ridiculous, it's not fun to be part of or to win with, and people should be getting hurt or disarmed in order to be restrained. So, good.

atomicplan666:
finally fuckinglee, this is going to change a lot of combat for the better...

Teragenic:
I have a question. Will it apply to all grabs? Or just the attacker? Multiple people should still be able to subdue an opponent. I like this PR because it stops chaingrabs, horrible system.

zakanater19:
Any chance the auto struggle could be combat mode only? i feel like if people "resist" bandits or whatever they are going to get fragged, seemingly with good reason to the fragger

Polotos:
Сделали бы просто парирование граба.

Teragenic:
> > I have a question. Will it apply to all grabs? Or just the attacker? Multiple people should still be able to subdue an opponent. I like this PR because it stops chaingrabs, horrible system.
> 
> It will stop all grabs

Might want to make it only against a certain person because multiple people should be able to overpower a person still.

HissyNays:
This seems like a considerable nerf to wearing people out by wrestling and pinning them.

n3tis:
ok are we now going to remove slipping from soaps or make it so that you can't slip again for 10 seconds if you slip once

PR #581, opened 2024-07-25
Makes a new crafting type for traps

Hmeister-real:
## About The Pull Request

 Mainly a temptplate for others to make new traps, but also makes the trapping skill have a real use. 
Now tested and fully works in game

## Why It's Good For The Game

unused skill is le bad, is now used

### DIFF
--- a/code/modules/roguetown/roguecrafting/trapping.dm
+++ b/code/modules/roguetown/roguecrafting/trapping.dm
@@ -8,7 +8,9 @@
 /datum/crafting_recipe/roguetown/trapping/mantrap
 	name = "mantrap"
 	result = /obj/item/restraints/legcuffs/beartrap
-	reqs = list(/obj/item/natural/fibers = 1,/obj/item/natural/thorn = 2,/obj/item/ingot/iron = 1)
+	reqs = list(/obj/item/grown/log/tree/small = 1,
+				/obj/item/natural/fibers = 2,
+				/obj/item/ingot/iron = 1)
 	verbage_simple = "puts together"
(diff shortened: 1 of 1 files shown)


Useroth:
I wonder if we should change the recipe.... now you can't just go to a smith and buy some of those really.

ThatReformedEngineer:
> I wonder if we should change the recipe.... now you can't just go to a smith and buy some of those really.

I agree with changing up this recipe, mostly because isn't the beartrap just metal in how it operates? I could see it maybe using a cog or gear or something, but thorns just seems arbitrary of an ingredient involved for the damage it causes.

Hmeister-real:
> > I wonder if we should change the recipe.... now you can't just go to a smith and buy some of those really.
> 
> I agree with changing up this recipe, mostly because isn't the beartrap just metal in how it operates? I could see it maybe using a cog or gear or something, but thorns just seems arbitrary of an ingredient involved for the damage it causes.

If it had to be smithed it would result in even less use, it already needs a hammer but if it requires blacksmithing and trapping it would be literally impossible to see outside of the town or in rare cases made by smiths in the bog smithy during high pop

Hmeister-real:
> I think it's bad to just have a skill that only faciliates 1 single recipe.
> 
> What if trapmaking was instead covered by engineering?

true, but then there would be even less people able to make them, and they would essentially never be seen. I want this to remain as a temptplate for more trapping stuff down the line, it's just limited to the mantraps while there aren't any other forms of traps.

Useroth:
We should approach this another time tbh

PR #593, opened 2024-07-26
Vampires can only slurp 9000 vitae from one person

Azarak:
## About The Pull Request
Fix to unhealthy bloodbag healing, and to discourage keeping one captive for 2 hours
Better than any bandaid like https://github.com/Rotwood-Vale/Ratwood-Keep/pull/589 I think


DorfGuy1:
Make it per person per vampire maybe? So they still need to expand but won't kill collaborating mortals

DavidAreYouSure:
How much is that compared to a normal humans entire blood supply?

ThatReformedEngineer:
> How much is that compared to a normal humans entire blood supply?

I believe the normal supply of human blood is something like 520? So it should be about roughly 17 feedings of blood give or take a few hundred blood units.

PR #595, opened 2024-07-26
Sheathes and genital descriptors

Azarak:
## About The Pull Request
Adds sheaths and genital descriptors, describing intimate parts if they're exposed
All animalistic penises now have sheathes, which they'll be hidden inside if they're not erect
And some tweaks to arousal and scon

Customizers for penises have changes, and **people will need to adjust their savefiles!!!**

### DIFF
--- a/code/modules/client/customizer/customizers/organ/genitals.dm
+++ b/code/modules/client/customizer/customizers/organ/genitals.dm
@@ -328,6 +328,17 @@
 		)
 	allows_accessory_color_customization = FALSE
 
+/datum/customizer/organ/vagina/human_anthro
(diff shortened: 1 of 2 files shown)


Hmeister-real:
merge NOW

coffeecoffeecoffee:
Finally, people can tease me about the small dicks I set on all of my characters.

Kepteyn:
A good addition, but not a massive fan of the arousal system. It would be nice to have a way to manually toggle it for those who prefer to use emotes as opposed to mechanics.

ToothyDeer:
could kobolds have more variety with genitals? like the rest wild-kin have, w/ knots and sheaths. kobolds w/ sheaths are hip and cool

PR #596, opened 2024-07-26
Changes the way round antags are picked

zakanater19:
## About The Pull Request

Currently major and minor antags work like this, they all roll depending on their number chance and one major antag wins and one minor antag wins, this PR will make it so they all roll on their own, so you could have extended, or you could have peasant rebels AND vamps in same round, same goes for minor antags like bandits or maniac

## Why It's Good For The Game

I guess specifically for minor antags this allows more to be created and put into the round

edit: forgot it also helps with meta i guess, since if i see bandit notice says no bandits, they it's maniac or aspirant


zakanater19:
closed cause i overlooked minor antags already roll seperately

PR #601, opened 2024-07-27
Adds Kettle Helm to armor crafting

Alanii:
## Makes Kettle Helm Craftable

That is all it does.

You may now create the kettle helm that is found in the keep armory at the anvil now.


Alanii:
Closed;
PR #569 Does this exact thing but adds two additional helmets

PR #607, opened 2024-07-27
Lowers inflated mercenary stats

Azarak:
## About The Pull Request
Lowers the stats of Desert Riders and Grenzelhofts to a net 8, matching the net stats of Man at Arms
The mercenaries used to be donator roles, and then also got stat inflated by Blackstone, to an offensive level

Desert Riders:
Str: 2 (unchanged)
End: 3- >2
Con 2- >1
Perc 2 -> 1
Speed 3- >2

Grenzelhoft
Str: 3 -> 2
End 2 -> 1
Con 3 - >2
Perc: 1 -> 2
Spd: 1 (unchanged)

Addendum: We need to restrict these with races more, and add "Sellswords" which are slightly worse but most races can play them, I think that would be good for racial identity and balance - something to do in the near future (I guess adventurers are sellswords? Eeeh)


DeesDeeDees:
Seconded for Sellswords DLC and trending towards Racial identities for the famous mercs.

Obviously, Adventurers can be mercenaries but it can be a wild crapshoot trying to hire them as-is. Dedicated sellswords you at least know they're in the market.

HissyNays:
I don't even play mercenaries but I'd say they don't deserve a nerf.

pilotmans:
Grenzlehofts are already racial locked to a certain degree

PR #609, opened 2024-07-27
Makes the Dying Station more immersive

Azarak:
## About The Pull Request
Now it's called "dye bin"
Now it needs 1 small log to build
Now it can be grabbed, pulled, thrown, manipulated around, and destroyed easier
Now it requires juicing berries into it to grant 3 uses of dying clothes
Dying clothes now requires a small do_after() timer of 3 SECONDS
Made all sorts of descriptions of the bin and usages of it much more fitting and immersive

dreammaker_ndXew3yeH0


DeerJesus:
This is good, not too hard so good

Kepteyn:
This is a good change, I think. That little bit more involvement in the process is nice.

PR #611, opened 2024-07-27
Adds short canid snouts to Tabaxi and Vulpkians

Azarak:
## About The Pull Request
They're rather fitting and I forgot to add them


Nonplayer775:
noice been waiting for this for a while, could also include the taj snout for tabaxi would be fitting

PR #617, opened 2024-07-28
The big MANIAC WONDER and HEART CODE rework. (+ Multi-Maniac fixes!)

zoktiik:
## About The Pull Request
I just spent a lot of time writing this code and can't be assed to summarize the whole diff at the moment. Nor do I have the time to test all of this at the moment. But I wanted to get the code out there, since pretty much all of it seems to work after some basic testing. Here's some bullet points.

- Wonder examine and processing gets a refactor. Only a maniac examining a wonder they created actually forces it to process. A single proc handles inscriptions of new hearts by both examine() and process().
- Hearts now support being etched by multiple maniacs, storing data for each antag datum ref by using them as keys in an assoc list. Each maniac will only ever see his own data when examining victims or their hearts.
- Hearts retain the functionality of only allowing one INRL key to be marked on them *per maniac*.
- Maniacs cannot mark each other's hearts with wonders, and are also physically unaffected by examining each other's wonders. (In the latter case a unique message is still shown.)
- Becoming a Maniac (midround) removes all traces of you ever being marked by a Wonder.
- As a victim, each new Maniac writing on your heart with his Wonder will stun you as if you had never been marked. You will not gain the client color or the stress event again, however.
- One Maniac TRIUMPHING makes him Trey Liam, cutting off all others. Any straggler maniacs will automatically die as if there was no Trey Liam landmark to spawn on.
- Maniacs constructing the fourth wonder now alert all other maniacs. ~~(I have not tested 4th wonder construction yet with this added code.)~~ **TESTED** with this recent patch.
- Admin stuff. Logging and examine messages mostly. **Messages admins** when a Maniac builds a Wonder, finds the final heart code, or prays the Sum.
- Added a few seemingly-missing istype helper procs, even if I only used one of them in this PR.
- Changes the Wonder naming scheme to put the key_name second (i.e. "Wonder INRL" instead of "INRL Wonder".)

**Update August 8:** 
- All issues related to screenshake should be fixed. Screenshake will not persist after deactivating combat mode (pre-agony), waking up (TRIUMPHED sequence), getting culled by another Maniac's victory, or getting your maniac antag removed. It will persist only if you are an active maniac on the 4th wonder, and that is it.
- Getting culled as a losing Maniac deletes all your wonders, clears your screenshake animation, and removes `TRAIT_SCREENSHAKE` and `TRAIT_SCHIZO_AMBIENCE` from your mob.
- Trey's cyberdeck headset is now see-through. You will be able to see the ending room from the start (during your sleep), and without needing to take off the headset during the window that you're not stunned for.

Iterating through the carbon list is gonna have to stay, as coal as it is. The issue I thought was attributed to it (Trey permastun) wasn't an issue on its part at all (since that also exists in Blackstone's code).

~~Issue I found while testing: Trey seems to be permastunned after waking up from sleep, at least in my test round. Not 100% sure why that happens but I'll need to investigate a bit more.~~ Right this actually seems intentional. 

Putting my to-do list of things I want to implement here. 
- [x] ~~make sure the maniac datum we're referencing in heart and examine code hasn't been QDELETED (like if the owning character had its Maniac status removed via Traitor Panel). If it is, clean both it and its data from the heart.~~ This is unnecessary due to the qdeletion evidently causing the reference to not exist anymore, even in the heart, which means that the code iterating through maniac antag datums inside hearts automatically ignores the invalid references.
- [ ] add an admin secret to prevent a Maniac victory from ending the round (it will still cull all other maniacs and prevent them from waking up).
- [ ] add an admin secret to toggle GLOB.maniac_highlander between TRUE and FALSE, to allow cycles of maniacs.
- [ ] track all maniac datums in a global list on creation and removal. make the carbon_list for loops iterate through this instead.
- [ ] add a "do_not_spam" variable to /mob that suppresses a) dumping mob memory to chat on mob login, and b) generic sound messages when mob is unconscious i.e. "You can almost hear something...". Set this to TRUE for the Trey Liam mob before the mind transfer.
- [ ]  Maniac-marked client colour/stress event should decay after a 5-10m timer. "My sanity returns to me." Timer & effect renew with a new maniac's inscryption.
- [ ]  Maybe also implement some spooky ambient noises while your marked state is at full effect. 
- [ ] *Maybe* add Psychotic as an adminbussable antagonist datum, giving you the Maniac's process() without the pursuer hallucinations. Because I'm a bit annoyed with people thinking that getting marked by a wonder == mini-dreamer psychosis. Cringe.
- [ ] Add a proc for forcing an object-say hallucination on a maniac/psychotic antagonist datum (or just on a player!), as a more immersive in-world alternative to subtle PM. 
- [ ] *maybe* clear chat when a maniac mind gets transferred to Trey Liam?
- [ ] *maybe* write an adminbus proc to mark a person's heart with a maniac datum's key...
- [ ] remove Trey's spells before he becomes Trey

## Why It's Good For The Game
Fixes a really bad issue with wonders. Also makes rounds with more than one maniac work properly. Each maniac is essentially isolated from the others, and each heart stores the maniac's own data for his instance of the antag datum.


DeerJesus:
'One Maniac TRIUMPHING makes him Trey Liam, cutting off all others. Any straggler maniacs will automatically die as if there was no Trey Liam landmark to spawn on.'

This might sound stupid but... Since only a maximum of one maniac tends to spawn per round how who we just enforce only one maniac spawning per round like how only one VL can spawn? Makes more lore sense and probably makes more player sense

ChungusGamer666:
Mostly good changes, but I don't think anyone should be able to be marked by multiple maniacs and this was a VERY intentional decision on my end.

ChungusGamer666:
Non-maniacs forcing other people to look at a wonder was intentional too, not sure why you are changing that at all.

ChungusGamer666:
For the to-do changes:
The client color should decay, but making the negative mood event is *integral* as a game mechanic, it is meant to be a permanent minor handicap to any of the victims of the maniac. Removing it is super lame because it is very soulful that your character gets permenanently traumatized by wonders.

ChungusGamer666:
Everything is said is mostly nitpicks and I don't care too much, the intent of making multiple maniacs work in a round (for the sake of adminbus and succession) is noble. 
Thanks for answering my feedback, the last thing I have to say is, if you want to follow my original intent with each victim only being able to be marked once, let maniacs know when someone was marked but it was done by someone else on examine - Both the heart examine and human examine.
In a round of multiple maniacs, this could possibly lead to fun intrigue and infighting between them, as they are not supposed to cooperate in this hypothetical scenario.

PR #618, opened 2024-07-28
Adds another gatemaster slot

wow50000:
## About The Pull Request
So, gatemaster... the role where you sit at a desk for three hours and then listen to retards telling you "open". You wouldnt even be able to pay me to do that IRL. So to slopfix this, I am adding a second gatemaster slot to keep em company so they can take turns resting and manning the gate or having these things called "break times." THIS PR IS NOT INTENDED FOR THE SECOND GATEMASTER TO MAN THE BOG GATE.

## Why It's Good For The Game
gatemasters cant catch a break and no MAA wants to man the gate at all. This PR was poorly thought out and has not been tested so issues may arise. This should make gatekeeper a bit less of a shit job since there are two of them to share the burden. Yes I am aware that this is not a perfect fix, yes I believe that this is probably a good thing.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/garrison/gatemaster.dm
+++ b/code/modules/jobs/job_types/roguetown/garrison/gatemaster.dm
@@ -3,8 +3,8 @@
 	flag = GATEMASTER
 	department_flag = GARRISON
 	faction = "Station"
-	total_positions = 1
-	spawn_positions = 1
+	total_positions = 2
+	spawn_positions = 2
 	allowed_sexes = list(MALE)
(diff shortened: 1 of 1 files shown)


Elli-Skala:
I would be very surprised if there was more then one spot for the gate master to actually spawn

SixplyDev:
Honestly the second gatemaster should just be assigned or start at the bog. The bog gets and see a whole lot more action + a lot of the time a gatemaster is missing due to bogmen needing to do stuff elsewhere leading to break-ins

Useroth:
We'll be reworking the gate a bit

PR #619, opened 2024-07-28
New round announcements

Useroth:
Hopefully it will work... It should.


Useroth:
Shameless self-merge because it works

PR #625, opened 2024-07-30
Perception Buff for melee combat

ddsaw:
## About The Pull Request

This PR will buff the PER stats usefulness in melee combat. It does this by shifting a little bit of accuracy from skill to per aswell as tying the accuracy increase and critchance increase of aimed mode to PER.
It also ties stab and cut bonus accuracy to per.
The mathematics work out to be about the same at 10 per and journeyman skill with values below being more heavily punished and values above gaining more value, especially on aimed mode.

The actual aimed mode critchance increase is 1% per point of per. Note that STR increases critchance innately by 1% per point due to how damage factors into critchance, and it does so in every mode.

Examples for how this change numerically affects things:
Big Winners: 14 Per Rangers stabbing you in aimed mode. They gain 4% crit and about 20% accuracy.
Big Loser: 8 Per Knight stabbing you in aim mode. He loses 2% crit and 15% accuracy.
## Why It's Good For The Game
Perception is considered one of the most useless stats alongside int. Combatwise, It affects almost nothing except bow damage.
It's out of combat utilities are limited to look around times.
Even with this I think perception is still a significantly worse stat than str, and it could use a few more out of combat utilities.

## Additional Considerations (please give thoughts)
Currently, critical misses are a pure fortune thing. And as long as you do not critically miss, any missed precise bodypart hit is automatically converted to a chest hit, making aiming chest totally pointless(might aswell go for the head..).
I considered checking for a miss instead of always rerolling to chest, but the checkmiss function is always called before a given hit so it may lead to jank behaviour (needs to be checked thoroughly first).

I tested the changes locally and confirmed it worked on my machine(tm), but as I can't expose the chances directly and they change with every single hit it's hard for me to say it's 100% correct code. But considering the simplicity of the changes I am fairly confident. Please do check though.

### DIFF
--- a/code/modules/surgery/bodyparts/bodypart_wounds.dm
+++ b/code/modules/surgery/bodyparts/bodypart_wounds.dm
@@ -207,7 +207,7 @@
 			if((bclass in GLOB.artery_strong_bclasses) && istype(user.rmb_intent, /datum/rmb_intent/strong))
 				used += 10
 			else if(istype(user.rmb_intent, /datum/rmb_intent/aimed))
-				used += 10
+				used += user.STAPER
 		if(prob(used))
 			attempted_wounds += /datum/wound/artery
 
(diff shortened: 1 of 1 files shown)


H0gberg:
very good pr, making unused stats useful in a somewhat realistic manner, speed merge plz

EzekyalLegit:
thres a massive flaw with this in that accuracy rolls are already way too easy, people are able to consistently stab foot and eyes and nearly never miss.

if this pr is going to make everyone with 11+ per even MORE accurate its making an existing provlem worse.

we should be makong current levels of accuracy harder to reach

PR #628, opened 2024-07-30
Narcolepsy and Moon Dust tweaks

Azarak:
## About The Pull Request
Moon Dust now reduces sleepiness, so you can force people awake with it
Moon Dust is more responsive to work with Narcoleptic, making sure the trigger gets skipped
Narcoleptic now has a little bit more drownsiness time and time between triggers
Also makes Greedy have the shittier See Prices trait


DeerJesus:
Does it remove "Tired" for triumphant sleeps?

Koteshkin:
I dont know if its merged, but currently narcolepsy triggers for converted vampires even when they have insomnia. 
image

PR #631, opened 2024-07-30
Grab breaking free re-balance

Azarak:
## About The Pull Request
Resisting out of grabs has been re-balanced. It uses 2 components which is 1. determining the chance 2. clamping to a minimum chance

Previously these looked like this:
1. The roll
50%
Str: -+10% per diff
prone: -20% (5% less per wrestling diff)
aggro: -20% (5% less per wrestling diff)

2. the minimums:
aggro: 50% +5%/wrestling diff
passive: 70% + 5%/wrestling diff

If you look at this closely.. you'll notice... the roll modifiers dont matter AT ALL in most cases (since it'll be clamped to minimum chance), and you could be an expert wrestler with 15 strength, and people would still have 50% chance to break free (and you can break every 2 seconds)

And if you would be fighting against an expert wrestler, no matter if he has 5 STR or 15 STR, he'll have 70% chance to break free

These were changed to work like this:
Base chance: 50%
Str: -+7% per diff
Prone: -15% lowered by 4% per wrestling diff
Aggro grab: -15% lowered by 4% per wrestling diff

The minimum roll values are:
Passive grab: 40%
Aggro grab: 25%

This means being a non wrestler and being aggressively grabbed while prone, will end up with you having 20% clamped to 25%
And if you have 3 skills of wrestling above your opponent, your chance is only reduced by mere 6%, resulting in 44%, almost twice

That is with the same level of strength being in playing field, each one will grant a 7% in either way

Open to feedback and discussion

**Update: I updated the base chance to 60% from 50%**

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -1107,7 +1107,7 @@
 	. = TRUE
 
 	var/wrestling_diff = 0
-	var/resist_chance = 50
+	var/resist_chance = 60
 	var/mob/living/L = pulledby
 
 	if(mind)


MefAnneFeminene:
FINALLY!!!!! THANK GOD!!!!

PR #635, opened 2024-07-30
Manual Genital Arousal

Kepteyn:
## About The Pull Request

This PR adds the ability to manually set the erection state of male genitals via sexcon, next to speed and force, which is set to natural (matching arousal level) by default. This option is hidden if the player lacks the suitable organ (penis).

Untitled

## Why It's Good For The Game

With the welcome addition of genital states, the ability to set it manually was conspiciously lacking, requiring players to use mechanics even if they dislike using the system to make their sprite match the described state. This PR will allow emote-preferring players to have their sprite match the scene without using mechanical sex. It also allows for arousal due to non-physical causes (i.e. morning wood or simply seeing someone naked), or remaining unerect despite of physical contact (such as being a slow grower, unattracted to your partner or simply erectile dysfunction.) 

While many scenes will not need this, many others will benefit from it, and the addition of a setting that can be ignored if irrelevant is a non-inconvenience.

### DIFF
--- a/code/datums/sexcon/sexcon.dm
+++ b/code/datums/sexcon/sexcon.dm
@@ -434,7 +434,7 @@
 			dat += "</tr><tr>"
 
 	dat += "</tr></table>"
-	var/datum/browser/popup = new(user, "sexcon", "<center>Sate Desire</center>", 470, 540)
+	var/datum/browser/popup = new(user, "sexcon", "<center>Sate Desire</center>", 490, 550)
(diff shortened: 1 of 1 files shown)


Red-Honk:
This would've been comical in a scene, when the partner of my character did such a bad job (on purpose) to lose his erection after the first bust due to lack of foreplay arousal.

Gar-Delin:
Would be cool to add a female version (when they're dripping)

PR #638, opened 2024-07-31
Bandits gain 2 athletics and swimming

MickDick1:
## About The Pull Request

Bandits get 2 skill level in swimming and athletics.

## Why It's Good For The Game

It seems like a glaring oversight that the dedicated swamp trotters that chase people down and all that stuff have 0 bodily conditioning and 0 ability to swim through the murk that's everywhere in their environment.

So, this makes them more logically able to operate in their environment.


Elli-Skala:
Surprised they don't got pickpocket

PR #639, opened 2024-07-31
takes -1 con from smith and adds it to int

wow50000:
## About The Pull Request
removes one point of con from smith and adds it to int

## Why It's Good For The Game
since humen female got one point of int removed and it got added to con im gonna put this here so i can actually hit my +2 int threshold.


wow50000:
also tempted to take a point of end and add it to str for personal statmaxxing reasons so female humen int is at base 11 so they cant roll below 10 str because they dont really need that many points in end anyways

PR #640, opened 2024-07-31
Rebalances smith so they can do their job better and gives more leeway for other races

wow50000:
## About The Pull Request
takes one point of con and end and puts it into str and int.

## Why It's Good For The Game
This lets -1 races hit the base 11 threshold allowing them to roll a min of 10 str which is what is needed to avoid getting stammed out while smithing, If this gets too op I am not against removing some of the athletics and weapons skills. Also takes a point of con to put into int because blacksmiths have no use for con and also a lot of races with +2 int got nerfed into having +1 so this lets them hit the base 12 threshold of int allowing them to roll 11-13 int. These changes were imagined with humen female in mind but also shouldnt fuck any of the other races smith gameplay over and should actually be an elf blacksmith buff as well as other weaker races.


wow50000:
I am also willing to remove a lot of their combat stats if i could add traps 1 so they can make mantraps that are metal and only really keep maces 2 and nerf fistfighting and wrestling to 2

PR #646, opened 2024-07-31
Special Traits

Azarak:
## About The Pull Request
A lot of these made by Steiner, big thanks to him!

Use "Be Special" in your prefs to roll for a special trait, and when you spawn, you'll Be Special, which could be something negative, positive or neutral

## Why It's Good For The Game

Adds more randomization and spices up the rounds with interesting changes, that add RP hooks straight from the start of the round, it is also fully opt-in

### DIFF
--- a/code/_globalvars/special_traits.dm
+++ b/code/_globalvars/special_traits.dm
@@ -28,6 +28,20 @@ GLOBAL_LIST_INIT(special_traits, build_special_traits())
 	if(special.req_text)
 		to_chat(user, span_boldwarning("Requirements: [special.req_text]"))
 
+/proc/apply_job_prefs_special(mob/living/carbon/human/character, client/player)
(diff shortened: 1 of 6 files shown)


DorfGuy1:
The randomly giving a beggar 20 strength endurance and Constitution might be a bit much. Maybe 15? Or even 16?

CalamaBanana:
After rolling a special, clicking the special button should tell you again what you rolled. 

I rolled a special at roundstart, but had to leave to get food. An hour later and after rejoining the server, I have no idea what I rolled. Would be neat to be able to click it again and be reminded.

wow50000:
the special trait that gives someone crafting skills can lead to issues where bandits just rob smiths of their tools and leave a smith essentially fucked for the round. not a fan of that one

PR #647, opened 2024-07-31
City Walls

Gyrandola:
## About The Pull Request

Adds walls to the city, making the two entry points major chokepoints the garrison can use to properly defend the town against external threats. (One can still enter by breaking into the sewers)

## Why It's Good For The Game

The town in its current state is impossible to defend, since anyone can just leap over the moat at any given location. Towners obviously don't feel secure and the garrison is helpless against major attacks from vampires and the likes.

image

image

image

image

image


Alanii:
If you let me, I would like to expand the town to the west.

This way I can have the main road to merchant properly open. Or you could. The town walls create a huge bottle neck at the moment.

zakanater19:
Nice

Useroth:
I think we won't surround the town in walls for now

PR #648, opened 2024-08-01
Apothecary towner

TheFenix323:
## About The Pull Request
Adds the Apothecary as a towner subclass, they start with expert alchemy and starting tools.
(Open to changes if the role feels underwhelming)

## Why It's Good For The Game
Has been requested and will allow players who want to try alchemy not need to go around town looting what they need to start.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/pilgrim/apothecary.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/pilgrim/apothecary.dm
@@ -1,6 +1,6 @@
 /datum/advclass/apothecary
 	name = "Apothecary"
-	tutorial = "You are an Alchemist, go make the deadliest poison or the best potions in the kingdom"
+	tutorial = "You are an Alchemist, with your tools and knowledge you are the most capable of making remedies for your fellow townsfolk's ailments, or creating drugs to help withstand the gloom this town carries."
 	allowed_sexes = list(MALE, FEMALE)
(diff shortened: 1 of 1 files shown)


zakanater19:
Nice

TheFenix323:
> Change carpentry to 1 Add cooking 1 Change polearm skill to knife skill
> 
> Description could use some work. Like.. something about making remedies for fellow towners ailments and drugs to keep spirits high in the dark times.

Requested changes made.

Elli-Skala:
Half the potions use fish parts to make them, maybe a *low* level of fishing would be a good idea in the case they don't have a dedicated fisher so they can make do but not super well

TheFenix323:
> I'm currently wondering whether I want to make put alchemy more into the Barber Surgeon, since they dont have all that much to do other than just to join the church and help em

We could make it a subclass

EzekyalLegit:
> I'm currently wondering whether I want to make put alchemy more into the Barber Surgeon, since they dont have all that much to do other than just to join the church and help em

I disagree. The barber lugs around a heap of tools and has no room for alchemy. When i spawn as barber surgeon to do alchemy I straight up ditch the tools. 

Also barber surgeon is a huge job you can give cosmetic surgery to outlaws, you can provide sex change operations to whores, you can work for the bog gate, you can find a lot of work. Alchemist needs to be working in a garden and making a workshop from round start. No time for healing folk.

These two jobs are too much work to combine.

PR #655, opened 2024-08-01
New Tiefling Colors

Gar-Delin:
-New Tiefling colors:
Mephistopheles (Blue),
Zariel (Yellow),
Dispater (Purple).


OozeSlimeJelly:
> -New Tiefling colors: Mephistopheles (Blue), Zariel (Yellow), Dispater (Purple).

I tried to do the same thing and was told they might make the tieflings less distinct, though maybe they'll like this one better. I'm still going to try one that's just blue/purple because that's what the dev seems to want to add.

PR #670, opened 2024-08-02
Treasure Hunter rework

Hmeister-real:
## About The Pull Request

Plucked from #520 and fixed up. Treasure Hunter is a really shitty and boring role, now its the new Whip/Flail adventurer class! No other adv has it other than Witchhunter, so now that isn't the case. 

## Why It's Good For The Game

because having a boring, useless adventurer class as rare as Treasure hunter is stupid. Also literally indiana jones

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/combat/rare/treasurehunter.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/combat/rare/treasurehunter.dm
@@ -15,7 +15,7 @@
 	shirt = /obj/item/clothing/suit/roguetown/armor/leather/vest
 	backr = /obj/item/storage/backpack/rogue/satchel
 	belt = /obj/item/storage/belt/rogue/leather
-	backpack_contents = list(/obj/item/bait = 1)
+	backpack_contents = list(/obj/item/bait = 1, /obj/item/flashlight/flare/torch/metal = 1)
 	gloves = /obj/item/clothing/gloves/roguetown/leather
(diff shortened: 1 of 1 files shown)


ddsaw:
Move down whips to 3, there is no adventurer class with expert (melee) combat skills as of right now and it sure as hell should not be treasurehunter. Whips are also likely to be the best weapon type in the game right now, with flails being so strong.

Also hand out some more low level misc non combat skills in order to clearly define a combat secondary class, as their stat total and traits are mediocre.
FYI: Likely hold off on big and impressive changes beyond stat adjustments until the rituals magic rework hits, we will likely have a big adventurer balance pass then.

ddsaw:
> > Move down whips to 3, there is no adventurer class with expert (melee) combat skills as of right now and it sure as hell should not be treasurehunter. Whips are also likely to be the best weapon type in the game right now, with flails being so strong.
> > Also hand out some more low level misc non combat skills in order to clearly define a combat secondary class, as their stat total and traits are mediocre. FYI: Likely hold off on big and impressive changes beyond stat adjustments until the rituals magic rework hits, we will likely have a big adventurer balance pass then.
> 
> Amazon with 5 spear skill

True. And a fuckup/oversight by whatever idiot coded amazon. Not really something we should design around. It's very likely getting removed, infact I already have a PR up to disable Amazon on account of it basically being barbarian(we already disabled sorceress for the same reason) and being so undermaintained that we let 5 spearskill go through for so long.

PR #672, opened 2024-08-02
Gives hostages collars instead of the cursed mask

wow50000:
## About The Pull Request
See title

## Why It's Good For The Game
The cursed mask sucks, It hides the face and is an old reference to something I could care less about. Also back when it was implemented it was implied the prisoner had the mask to hide their nobility but everyone already knows hostages now are nobles.


wow50000:
havent tested this, may break everything.

PR #673, opened 2024-08-02
Lists a keyring's keys on examine

zoktiik:
## About The Pull Request
I'm tired of not being able to actually see what keys are on my damn keyring. This should perhaps fix that. I haven't tested the change yet in-game, thus the draft PR status.

## Why It's Good For The Game
You can now see what keys are on your keyring at a glance.

### DIFF
--- a/code/game/objects/items/rogueitems/keyrings.dm
+++ b/code/game/objects/items/rogueitems/keyrings.dm
@@ -21,6 +21,7 @@
 			addtoring(new X())
 			keys -= X
 	update_icon()
+	update_desc()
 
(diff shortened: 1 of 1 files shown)


n3tis:
You know, it would be cool if this was paired with a function to remove keys, so you don't have to remove them one by one. Like, adding a "(Remove)" hypertext to the side of the listed key.

n3tis:
> > You know, it would be cool if this was paired with a function to remove keys, so you don't have to remove them one by one. Like, adding a "(Remove)" hypertext to the side of the listed key.
> 
> This is planned.

Damn... *kisses you*

PR #674, opened 2024-08-02
Resident Keys, Towner Class Info & Town Expansion

Azarak:
## About The Pull Request
Towners and Migrants know what classes others are picking
Towner residential homes no longer have loose keys on them, instead you can claim one of them, first come first serve
There's also homes which are reserved for a specific class, such as doctor, blacksmith, fisherman etc. which have facilities to fit their class
Reworked south-east part of the town, which now includes:
1 residential home with a cellar
4 other residential homes homes
1 doctor home
1 blacksmith home
1 cheesemaker kitchen

Added a fisherman hut by the east lake
Added a miner hut in the town mountain
Added a woodcutter hut in the basin, south and a little west of the town

### DIFF
--- a/_maps/map_files/rockhill/rockhill.dmm
+++ b/_maps/map_files/rockhill/rockhill.dmm
@@ -253,6 +253,18 @@
 	},
 /turf/open/floor/rogue/greenstone,
 /area/rogue/under/town/sewer)
+"aoV" = (
+/obj/structure/fluff/railing/wood{
+	dir = 1;
+	layer = 2.7;
+	pixel_x = 1;
+	pixel_y = 7
+	},
+/obj/machinery/light/rogue/torchholder{
+	dir = 4
+	},
+/turf/open/floor/rogue/ruinedwood,
+/area/rogue/outdoors/mountains)
 "aoY" = (
 /obj/structure/bed/rogue/inn/hay,
 /turf/open/floor/rogue/ruinedwood{
@@ -282,6 +294,13 @@
 	},
 /turf/open/floor/rogue/blocks,
 /area/rogue/indoors/town/manor)
+"ask" = (
+/obj/structure/fluff/railing/wood{
+	dir = 4
+	},
+/obj/structure/fluff/dryingrack,
+/turf/open/floor/rogue/dirt/road,
+/area/rogue/outdoors/rtfield)
 "asx" = (
 /obj/structure/mineral_door/wood/deadbolt{
 	dir = 8;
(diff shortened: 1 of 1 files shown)


Drayco84:
Any changes to the Seamster Townie spawn?

code/game/objects/structures/mineral_doors.dm
```diff
@@ -771,3 +836,56 @@
...
+/obj/structure/mineral_door/wood/towner/hunter
+	resident_advclass = /datum/advclass/hunter
+	lockid = "towner_hunter"
+
+/obj/structure/mineral_door/wood/towner/miner
+	resident_advclass = /datum/advclass/miner
+	lockid = "towner_miner"
+
+/obj/structure/mineral_door/wood/towner/farmer
+	resident_advclass = /datum/advclass/miner
```
CalamaBanana, line 874: Might have the wrong class here.

CalamaBanana:
- Would be nice if anyone could claim a normal home after 1h if no towner staked their claim by then. A shame to let all those neat homes go completely unused during lowpop. There are far fewer towner players than homes most of the time. (Especially as you removed the bolt-door houses)

- Design wise, I would recommend removing the walkway from the normal homes. Do they really need to be so easy to break in? We all know what kind of RP they will be used for, and not like there are any valuables that would make them a target for anything but to disrupt RP. Have people break down doors if they want to break in, no need to present opportunities on a platter like this.

- Having entire homes visible through a single window makes them unusable for 90% of home-RP, sadly. Wish we had a "curtains" object, honestly.

CalamaBanana:
I also noticed you removed the SCOM near the southern plaza. Would be nice if we could have it back, that was the only truly public SCOM in the town, I believe.

CalamaBanana:
What about a "home permit" item that anyone can use to claim a home? Available an hour after roundstart, so towners get first pick, but latejoining bards can find their love-shack no one else needed. Perhaps available at the stockpile machines?

CalamaBanana:
> > What about a "home permit" item that anyone can use to claim a home? Available an hour after roundstart, so towners get first pick, but latejoining bards can find their love-shack no one else needed. Perhaps available at the stockpile machines?
> 
> I don't like this. Townies get homes because they canonically live in the town for years. I'll make a deadbolt home, and later I want to add another one to another section of the map ||there's also two on rooftops if you look around||

Even if not available as easily as I'd wish, there should be some way to hand them out. Purely mechanically, if the king decides to grant the small smithy to someone, he should be able to have some way to create the magic keys for the stranger. (Just like the steward can hand out keys to nearly any other place)

PR #675, opened 2024-08-03
(Bugfix) Makes sneaking serverside

zakanater19:
## About The Pull Request

Makes sneaking invisibility serverside instead of clientside because there was a bug constantly happening where you would appear invisible but other people would see you as visible, really kills sneaking for me because you never know if you can actually sneak past that guard or if he will just see you slowly moving around like a spastic not realizing the guard can see you

## Why It's Good For The Game

QOL for sneaking, this should also fix the issue of sneak fucking up by looking around on a tile, or moving from one tile to another also breaking sneaking sometimes

### DIFF
--- a/code/modules/mob/mob_movement.dm
+++ b/code/modules/mob/mob_movement.dm
@@ -589,29 +589,33 @@
 /mob/proc/update_sneak_invis(reset = FALSE)
 	return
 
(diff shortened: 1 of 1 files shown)


zakanater19:
mfw i keep doing retarded mistakes, ill make a new PR

PR #695, opened 2024-08-04
Nix the Nudist class

zoktiik:
## About The Pull Request
Disables the Nudist from being a pickable class.

## Why It's Good For The Game
The existence of nudist as a class causes far more animosity between players than is necessary, due to the nudist trait mechanically blocking you from wearing clothes, thus giving you an entirely OOC excuse to argue about why you should avoid consequences for being nude. It's far better as something you deliberately roleplay when nudism is widely shunned.

A lot of the roleplay happens from behind the bog gates, and being blocked due a mechanical barrier preventing you from becoming more socially acceptable is not a good thing. Especially when Nudist in its current state doesn't do anything to add to the gameplay of being out in the bog, seeing as you're not even a survivalist. Just get rid of it instead of forcing players to engage in this sort of conflict.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/special/nakedandafraid.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/special/nakedandafraid.dm
@@ -6,7 +6,7 @@
 	allowed_races = RACES_ALL_KINDS
 	outfit = /datum/outfit/job/roguetown/adventurer/nudist
 	traits_applied = list(TRAIT_NUDIST)
-	category_tags = list(CTAG_PILGRIM)
+	category_tags = list(CTAG_DISABLED)
 
 /datum/outfit/job/roguetown/adventurer/nudist
 	allowed_patrons = list(/datum/patron/divine/dendor)


Hmeister-real:
NUDIST IS GOOD FUCK YOU!!!

X404ErrorX:
If you dont like it dont play it then, why remove it for everyone alse? the rol is fun and a challange thats the entire point of it. this pr is dumb.

Kepteyn:
I dunno, I don't see why some people get so upset about seeing a naked person on an MRP/ERP furry server. Just ignore them, or bully them, and the nudists should just suck it up. If it's that much of a problem to people, then 1) you can **choose** to not be a nudist, nobody is forcing you to pick it, or 2) don't play on an ERP furry server, naked people will exist.

Alanii:
The nudist being its own class is made to allow people who role play as "nudists" to be mechanically forced to be a nudist if they selected it. This is to prevent people who want to play as a nudist and then get bored later on and stop being a nudist. If you select the nudist class, you already understand the consequences of being a nudist.

If players have a problem with it, they can handle it in the game. They do not need to make PR's that remove the class just because it makes them personally uncomfortable.

AGlassDragon:
Do not advocate for the removal of content you disapprove of without providing an alternative to take its place. Do not subtract from the player experience without adding to it to counterbalance the loss. It establishes a bad precedent to allow this sort of thing.

It's as simple as making a PR to change the Nudist to a Survivalist, removing the offending trait while keeping the class fantasy in place with similar starting skills and origins. Removing it wholesale should not be entertained, especially under such reasoning as has been put forwards.

coffeecoffeecoffee:
This class is so actively detrimental for the immersion this server so tightly tries to hold to that the fact this was ever a class is actually lunacy to me. Just strip down normally for fuck's sake.

pilotmans:
> This class is so actively detrimental for the immersion this server so tightly tries to hold to that the fact this was ever a class is actually lunacy to me. Just strip down normally for fuck's sake.

see that pure immersion

ChungusGamer666:
Lamest excuse for removing nudists I have ever seen.

Conflict with a completely valid IC reason supported by hame mechanics is good, actually! It's almost like nudists are supposed to be wildmen that can't integrate into society... 🤯

PR #696, opened 2024-08-04
Remove hammer req for crafting improvised weapons

zoktiik:
## About The Pull Request
Requested by ezekyal on Discord. Simply removes the hammer as a tool requirement for crafting the goedendag, improvised billhook, and peasant war flail.

## Why It's Good For The Game
From what it looks like: the need for a hammer to craft these makes them more difficult to make than would be necessary for an improvised weapon.


n3tis:
Based... Sir...

PR #701, opened 2024-08-05
Gives half elves blunted elf ears

lolzers:
## About The Pull Request

Gives half-elves the blunted elf ears instead of having no ears at all. 

## Why It's Good For The Game

Makes half-elves distinctive in their own right instead of being human with slightly different stats.


lolzers:
Screenshot 2024-08-04 185921
Bald helf-elf squire to showcase the ears

PR #705, opened 2024-08-05
키치 Kitchen remodel

Alanii:
## Why It's Good For The Game

- Reorganizes servant quarters and kitchen.
- Adds a storage room.
- Adds bars and table in main hall
   - Goal is to help guide keep roles from rushing the kitchen like a bunch of goblins and eating all the ingredients
   - If it continues to happen, I'll add a new key that keeps the storage locked say for servants, hand, master key, and queen.
- Adds chairs for maids to sit daintily and wait to be called to their obese king's and serve them. ( ͡° ͜ʖ ͡°)
- Moves vomitorium to the kitchen to easily access food stuffs.
  - No, men at arms do not _need access_ the keep has servants for a reason.
  - Yes, shylock exists outside of kitchen.

image

Oh yeah I am an awful person too, I put a dumb half cloak in hand room. heh.

### DIFF
--- a/_maps/map_files/rockhill/rockhill.dmm
+++ b/_maps/map_files/rockhill/rockhill.dmm
@@ -1641,7 +1641,7 @@
 	icon_state = "wcv";
 	locked = 1;
 	lockid = "manor";
-	name = "Storageroom"
+	name = "Servant's Room"
 	},
 /turf/open/floor/rogue/wood,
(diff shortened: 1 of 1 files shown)


Cyprex:
Could you push a commit to this? MapDiffBot isn't running, so I can't actually see what was changed!

Alanii:
duplicate

PR #708, opened 2024-08-05
Removes half elf night vision eyes

Azarak:
## About The Pull Request
Was never meant to be like this, night vision is strong, and is a special now


Drayco84:
Night vision is so damn strong because base night vision is so godawful.
Also, WHY don't Tabaxi, WHO ARE FRIGGIN' CATS, have nightvision?!

PR #717, opened 2024-08-05
Women at Arms and Inquisitorettes

Kepteyn:
## About The Pull Request

Unlocks the Man at Arms, Gatemaster, Sheriff, Captain, Veteran and Inquisitor to female characters.

## Why It's Good For The Game

Roles have been steadily becoming available to female characters for a while now, and this continues the trend. Given the abundance of roles, such as knights, that are no longer gender-locked and the relative gender equality in-game, it feels more abitrary and disjointed to have certain roles remain locked when comparable roles are not.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/church/puritan.dm
+++ b/code/modules/jobs/job_types/roguetown/church/puritan.dm
@@ -5,7 +5,7 @@
 	faction = "Station"
 	total_positions = 1
 	spawn_positions = 1
-	allowed_sexes = list(MALE)
+	allowed_sexes = list(MALE, FEMALE)
 	allowed_races = RACES_TOLERATED_UP
 	allowed_patrons = list(
 		/datum/patron/old_god,
--- a/code/modules/jobs/job_types/roguetown/garrison/gatemaster.dm
+++ b/code/modules/jobs/job_types/roguetown/garrison/gatemaster.dm
@@ -5,7 +5,7 @@
 	faction = "Station"
 	total_positions = 1
 	spawn_positions = 1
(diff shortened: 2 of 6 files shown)


hugmeorelse:
> I'm sorry, but could you please give a slightly better reason for this change, other than just "well the trend is going this way"?

I don't really care one way or another, but I appreciate people who can bring this sort of thing up civilly.

Kepteyn:
> I'm sorry, but could you please give a slightly better reason for this change, other than just "well the trend is going this way"?

Certainly, though I did provide more reasons than just 'well the trend is this'. The setting in-game is one of gender equality. There is almost no gender discrimination, beyond that some roles arbitrarily prevent female characters from beginning in it while other similar roles prevent them. If women can be knights, templars, and bog masters, why not gate masters, men at arms or sheriffs? If a woman can be the priestess in charge of the religion of an entire city, why not the extremist that investigates heretics on behalf of said priestess? 

There is a great deal of character freedom, and if a twink ratboy can be a man at arms, a butch lupian woman who is a veteran of wartime would almost certainly should be eligible, given the aformentioned lack of gender-discrimination in-game beyond the king always being male. The restriction on female character is one that does not make much sense in the context of ratwood - inquisitors maybe, if a lore reason why is created.

I mention that it is trending this way because, even if it is not this PR, these roles would almost inevitably be un-genderlocked down the line anyway.

Kepteyn:
> I think we could use bringing back some gender distinctiviness and keeping the couple locks we've still got

I would strongly disagree. I do think having character creation affect what roles you can play is good, but I think that should be done in species/gods and traits, not in gender.

PR #719, opened 2024-08-06
Heat Treating plate armor to blue it

Slorbin:
You can now heat treat plate armor to 'heat blue' it! Only works with plate.

This is done by case hardening it. This was done in medieval times if you wanted a really damn nice set of armor. It was usually done for extra durability, but also was done to get a nice cool blue, black, gold, or shiny color. You basically added extra carbon to the outside and heat it hot.

I was going to have more colors since there's a lot more reddish browns you can make by heat treating, and even some pretty purples, but I just stuck with the upper shades of blue since they look more realistic.

Screenshot 2024-08-06 011210
Screenshot 2024-08-06 011353
Screenshot 2024-08-06 011531
Screenshot 2024-08-06 011700
Screenshot 2024-08-06 011829
Screenshot 2024-08-06 011951
image

They would also sometimes blacken armor by burning it with linseed oil, so that's an option.

It's a very expensive craft to make the heat-treatment forge it can be done at, and each singular piece of armor consumes one coal per case hardening. Not cheap like berry dyes, it's only meant for the truly RICHEST of knights.

### DIFF
--- a/code/modules/roguetown/roguejobs/blacksmith/treatingfurnace.dm
+++ b/code/modules/roguetown/roguejobs/blacksmith/treatingfurnace.dm
@@ -23,7 +23,7 @@
 			)
 	var/static/list/selectable_colors = list(
   		"Scrape Off Coating" = "#ffffff",
-		"Burn With Oil" = "#414143",
+		"Burn With Oil" = "#5f5f5f",
 		"Low" = "#f9d083",
 		"Medium" = "#6f6398",
 		"Medium-Hot" = "#6c7ca8",


n3tis:
that looks cool but I'd respect if you made actual documentation of how it works

all I know is coal

DeerJesus:
> that looks cool but I'd respect if you made actual documentation of how it works
> 
> all I know is coal

In game it's pretty similar to the dye bin: add coal, add item, select temperature, and heat treat.

kepesk1:
> that last one is way too black, would make it insanely difficult to fight them in armor like that in the bog or even over mud tiles

there is already jet black armor ingame, don’t crush a good pr when its entirely fine

DeerJesus:
> that last one is way too black, would make it insanely difficult to fight them in armor like that in the bog or even over mud tiles

It is pretty back, I can lighten it when I get home, but it's not blacker than our current black.

DeerJesus:
image
 i lightened it

PR #720, opened 2024-08-06
Adds more items to the bandit sacrifice list and refactor (zoktiik)

MickDick1:
## About The Pull Request

Adds:
riddle of steel
emerald choker
desert rider amulet
merchant amulet
rockhill crown
paintings
-
To the list of things that can be sacrificed at the bandit idol.

## Why It's Good For The Game

I think the only things I need to justify here are the crown and the paintings.
If the bandits actually manage to get the crown, and feed it to the idol. Well. I guess bandits win. Besides killing the local government off, that's peak matthios right there with sacrificing the object of their hatred. It just makes sense to let them destroy it. If the town screwed up that badly they deserve it.

The paintings are actually kinda valuable hovering around 30-40 mammon give or take. And it's a trope with raiders stealing paintings in media. So why not?

### DIFF
--- a/code/game/objects/structures/fluff.dm
+++ b/code/game/objects/structures/fluff.dm
@@ -954,7 +954,7 @@
 		/obj/item/clothing/neck/roguetown/shalal,
 		/obj/item/clothing/neck/roguetown/horus,
 		/obj/item/rogue/painting,
-		/obj/item/clothing/head/roguetown/crown/serpcrown
+		/obj/item/clothing/head/roguetown/crown/serpcrown,
 		/obj/item/clothing/head/roguetown/vampire
 	)
 


Heheniy:
why not just let them sacrifice anything that is coin or worth more then, lets say, 10

MickDick1:
> why not just let them sacrifice anything that is coin or worth more then, lets say, 10

A lot of items should have a proper value assigned but don't. I'm not against this implementation exactly. Just working with how it was originally done.
Also, it's just kinda weird to feed the idol fur/hides and stuff which would fall under that category of 10 mammons or more. You could just hunting-max and feed it animal skins to fulfill the quota.

While mining personally by bandits is kinda cringe it's at least forcing them to take control of high resource value sites and likely come into conflict with other players.

radarseas:
personally, I'd give a unique bad omen if they get the crown and feed the idol it... at that point, the town 1000% deserves being flexed on even harder if they fuckup **_that_** badly.

PR #721, opened 2024-08-06
Adds some overflow cells to the keep dungeon

zakanater19:
## About The Pull Request

More simple cells for prisoner overflow, they are not hostage cells like for the upper classes but are just basic lower dungeon cells, giving the garrison and keep more cells for common criminals to rot awaiting trial, i have tested all the cell buttons and doors and they work

## Why It's Good For The Game

It's just a nice place to leave prisoners where they both cannot scream and shit up the throne rooms diaglog, and also where the dungeoneer can keep longterm prisoners

Screenshot 2024-08-06 120420


zakanater19:
hm... it's changing things it should not, i will try this PR again once the server files are not borked lel

PR #727, opened 2024-08-06
The other bandit hideout PR

MickDick1:
## About The Pull Request

bandit hideout changes

Several things done in the previous PR were added but I will list everything anyways:

12 steaks in the food chest.
cobblestone roads.
8 old keys.
The chests in the map old and new now can be locked by the old keys.
Made a second prison cell.
Expanded the sleeping area for the bandit dorms.
Added a roofed spot near the idol with 2 furnaces, a pair of tongs, and a skullcandle.
Added a shack for private interactions.
More braziers so things aren't pitch black.
One chest + key next to the exit for prisoner transports.
Completed the palisade.
Lots more parchment for sending out ransom letters.
2 more wood shields on the rack.

## Why It's Good For The Game

The steaks are for food to stave off early starvation + allow them to hunt using bait, given that bandits cannot be ambushed, there's no chance to hunt 'conventionally'. No more farming is required right away, but still very beneficial.
Roads because every round the moment it rains it's permanently muddied up and you are slowed down so painfully it's just hell.
8 keys so the SINGULAR key is not constantly snatched and gone forever by some nimrod + we need spares for other tasks
Lockable chests are necessary to move prisoners into the hideout without them being able to learn the path back/break free in an unlocked chest to be a shitter
Second prison cell is necessary for the scale of management we sometimes have to deal with in general.
More room and beds for the potentially 20+ bandits swarming in to sleep.
Furnaces near the idol is good for smelting loot for sacrificing conveniently. They ought to have these set up already rather than hunting for coal every round, in retrospect.
Private shack. Private interactions.
The road and inner area was painfully pitch dark before. Now you can actually see.
Completed palisade for more convenient management of prisoners really + defense.
Parchment = ransom letters
Wooden shields are pretty important in certain cases... mostly copied it from the previous PR just cause.

I would have turned off the zones but I failed to discover the lever that needs pulling to achieve this for the screenshot.

I was already planning on making this and just happened to do it at the same time as the last guy. Guh.

### DIFF
--- a/_maps/map_files/generic/CentCom.dmm
+++ b/_maps/map_files/generic/CentCom.dmm
@@ -1652,13 +1652,6 @@
 	icon_state = "chess"
 	},
 /area/rogue/indoors)
-"oV" = (
-/obj/structure/fluff/railing/fence{
-	dir = 1;
-	icon_state = "fence"
-	},
-/turf/open/floor/rogue/dirt/road,
-/area/rogue/outdoors/woods)
 "pb" = (
 /turf/open/floor/rogue/ruinedwood{
 	dir = 1;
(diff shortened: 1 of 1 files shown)


MickDick1:
Remade it in SDMM cus I was told to.

MickDick1:
bandit hideout changes 2

PR #728, opened 2024-08-06
Court Mage can no longer be Zizoid

Azarak:
## About The Pull Request
I actually missed that and merged it, I only wanted adventurer mages to get this

### DIFF
--- a/code/modules/jobs/job_types/roguetown/courtier/magician.dm
+++ b/code/modules/jobs/job_types/roguetown/courtier/magician.dm
@@ -21,7 +21,7 @@
 	max_pq = null
 
 /datum/outfit/job/roguetown/magician
-	allowed_patrons = list(/datum/patron/divine/noc, /datum/patron/inhumen/zizo)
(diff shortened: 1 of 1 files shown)


DeerJesus:
Why shouldn't a court mage be zizoid?

Elli-Skala:
Why It's Good For The Game

Gr33d-y:
> Why shouldn't a court mage be zizoid?

A lot of trusted roles in the keep can't be inhumen, it only makes sense the court mage can't be inhumen as well. Knights can't be, the phys can't be, the captain can't be. This was the outlier.

radarseas:
court mage is working off the king's paycheck for their research... its one of the few roles which does make sense, to be a zizolite power hungry megalomaniac that worships a god that wholeheartly approves of power over reason. Especally with the entire court to hide behind this and their funding.

I don't think we should be axing potental RP fronts that make sense and bring interesting conflict into the court, especally ones that have some mechanical backing to base more RP off of, like the inquisitor.

Useroth:
I think this should stay unless people are massive shitters

PR #729, opened 2024-08-06
Baggin

Luctus-Comptus:
added bedroll. intermediate between nicebed and shitbed. portable. skincraft (no skill needed so you can level it). middle-mouse to deploy and roll up. takes 2 hide and 1 rope at drying rack. if you make fun of the sprite fuck you make a better one.


Alanii:
Should show images so we can make fun of the sprite.

Elli-Skala:
this was good :(

PR #735, opened 2024-08-07
Pet your local pit volf

Heheniy:
## About The Pull Request

Adds unique faction that makes keep pit volfs frienly to you
Currently this faction have:
-pit volfs (obviously)
-king
-dungeoneer
-guard captain
-hand (with chance 1/2)
-knights (with chance 1/3)

Also captain and knights finally gets dungeon keys

## Why It's Good For The Game

Makes pit volfs more unique then just angry volfs in a pit.
More possible interesting interaction with the pit.
Things from keep now can be safely retrieved without killing all volfs.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/garrison/dungeoneer.dm
+++ b/code/modules/jobs/job_types/roguetown/garrison/dungeoneer.dm
@@ -48,6 +48,6 @@
 	H.change_stat("constituion", 2)
 	H.dna.species.soundpack_m = new /datum/voicepack/male/warrior()
 	H.verbs |= /mob/living/carbon/human/proc/torture_victim
-	H.faction = list("pit_wolf_friend")
+	H.faction += "pit_wolf_friend"
 	ADD_TRAIT(H, TRAIT_MEDIUMARMOR, TRAIT_GENERIC)
 	ADD_TRAIT(H, TRAIT_STEELHEARTED, TRAIT_GENERIC)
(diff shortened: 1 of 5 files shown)


Teragenic:
Hand, Sheriff, Prince, and Princess(Same role)

I'd also recommend maybe the squires too on RNG.

Teragenic:
Being outlawed should also remove you from this faction.

Heheniy:
> Hand, Sheriff, Prince, and Princess(Same role)
> 
> I'd also recommend maybe the squires too on RNG.

Added hand
Sheriff most time spends in town so volfs wont know him well enough to not attack
Prince also wont spent much time in dungeon with pit volfs 



> Being outlawed should also remove you from this faction.

Those are volfs, they dont care about outlaws

ikrarkjr:
None, this is bad design as it makes it annoying to pit keep people, who often get pitted the most.
Make it werewolves and king and nothing else.

DeerJesus:
> I don't know about the job setting this to be your only faction. It feels like that's going to mess with things if we decide to use more factions in the future. Also don't quite like that it's RNG with some jobs whether the pit volfs like you or not, it should be a purely binary thing that is predictable and expected.
> 
> You should probably change all instances of `H.faction = list("pit_wolf_friend")` to `H.faction += "pit_wolf_friend"` so that it just adds on the faction instead of replacing the mob's faction list. Also, add it to the fourth pit wolf in the far corner.

On top of this, the king should be able to be fed to wolves. Ideally, change it to dungeoneer only.

john-pawson:
> None, this is bad design as it makes it annoying to pit keep people, who often get pitted the most. Make it werewolves and king and nothing else.

> On top of this, the king should be able to be fed to wolves. Ideally, change it to dungeoneer only.

Why not just... keep it like it is right now and choose *any* other execution method for the 6 roles that know the wolves? 
Just literally anything else than using the wolves.
Removing the roles decreases the amount of unique interactions with them and makes it so nobody can retrieve the things if the dungeoneer is dead or busy.

PR #736, opened 2024-08-07
Low pressure at latejoin boat fix

Heheniy:
## About The Pull Request

Fixes "low pressure" near latejoin boat. Someone used lavaland tiles there.

Also removes sewage dead end at river side
image
image


Drayco84:
Still some there near the stairs into and out of the town.

PR #737, opened 2024-08-07
Increase of saiga import price

Heheniy:
## About The Pull Request

Increases price to import saiga from 100 to 400

## Why It's Good For The Game

Who had a thought that ability to buy 400hp live shield that removes your slowdown and can itself deal 60-90 dmg per kick for 100 mammon was a good idea?

Go ask farmers for tamed saigas


DorfGuy1:
Does your saddled mount attack things you do?

Useroth:
I think it's fine as it is, I barely see anyone buy them even at current price

PR #740, opened 2024-08-07
Accents II CORE

Squird5:
#315 Reboot

## About The Pull Request

A little more accent variety among races to make them less homogeneous ((I hated 100% of elves being Russian)). Also tried to make the formatting more consistent between the JSONs.

Also removed the majority of the non-English words from most lists (Except German), making them more oriented around the sounds they make than simply tossing in foreign words.

## Why It's Good For The Game

More variety in speech patterns is always good. Also, all elves speaking demi-Russian would throw me completely out of character every time I saw "Privyet" or "Da" and I would stop thinking of my own character and instead of my friend's Pathfinder character.

## REFERENCE GUIDE

Accents are now broken into four parts: **Full**, **Start**, **End**, and **Syllable**.

**Full** replaces only entire key words with their counterpart. Any additional characters will prevent it from triggering.
`"watch": "look"` will only apply to "**Watch**", changing it to "**Look**", but variants like "**Watch**ed" or "**Watch**es" will **not** change

**Start** will replace words that start with the key word, regardless of the following characters.
`"watch": "look"` will end up allowing **Watch**ing and **Watch**ed to convert to **Look**ing and **Look**ed

**End** is the opposite of Start. It will apply to characters only at the end of a word. Useful for consistently dropping a letter at the end that would be messy if applied as a Syllable.
`"ing": "en'"` will end up allowing Watch**ing** and Slapp**ing** to convert to Watch**en'** and Slapp**en'**

**Syllable** will apply to any instance of the characters in the text, so use it carefully.
`"a": "aye"` would end up being painful, changing "H**a**ve **a** nice d**a**y" to "H**aye**ve **aye** nice d**aye**y"

### Notes
**Do not** use escape characters for things like apostrophes. `\'` It will work in some cases, but not all, resulting in demihuman accents working fine while the exact same accent list on Humans doesn't.

The **order of your replacers matters**. While having them in alphabetical order might look nice, it can result in unwanted outcomes. If a "c" becomes a "k", and a "k" becomes "kh", words like "Change" turn into "Khhange".

**Always test your work.**

## Change List

* REGEX fixed to allow for different types of replacers, while working reliably and multiple times in a line of text.
* Reformatted Universal without changes
* Reformatted German without changes, adding the new regex fields
* Reformatted Anglish without changes, adding the new regex fields
* Reformatted Middlespeak without changes, adding the new regex fields
* Reformatted Dwarf without changes, adding the new regex fields
* Reformatted Proper without changes, adding the new regex fields
[...]

PR #741, opened 2024-08-07
Accents II Tabaxi

Squird5:
Reliant on #740 

Adds a new accent for Tabaxi


Squird5:
Looks like I need to wait for #740 to be merged before I try to post any derivatives of it

PR #744, opened 2024-08-07
Changes Templar helmet FOV, removes them from heavy category

Leudoberct1:
This PR changes the FOV restrictions of the Templar patron-specific helmets to only block the behind, the same as the visored sallet does. To balance it out, they're also removed from the 'heavy' category, and thus lose the added protection from crits that such helmets would offer.

The reason for this change is that these helmets are not adjustable at all, unlike the Knight helmet- and when your job is to guard a place, that becomes pretty difficult to do when you can hardly see a thing around you- and it kind of sucks to take off these awesome looking helmets 90% of the time. A lot of people that I know play Templars have expressed the same sentiment of wishing that the view in them wasn't so restrictive.

This has been tested on my private repo with no errors.

### DIFF
--- a/code/modules/clothing/rogueclothes/hats.dm
+++ b/code/modules/clothing/rogueclothes/hats.dm
@@ -523,6 +523,46 @@
 					var/mob/living/carbon/H = user
 					H.update_inv_head()
 		user.update_fov_angles()
+		
+/obj/item/clothing/head/roguetown/helmet/astratahelm
+	name = "astrata helmet"
+	desc = "Headwear commonly worn by Templars in service to Astrata. The firstborn child's light will forever shine on within its crest."
+	icon_state = "astratahelm"
+	item_state = "astratahelm"
+	emote_environment = 3
+	flags_inv = HIDEEARS|HIDEFACE|HIDEHAIR
+	block2add = FOV_BEHIND
+	smeltresult = /obj/item/ingot/steel
+
+/obj/item/clothing/head/roguetown/helmet/nochelm
+	name = "noc helmet"
+	desc = "Hearwear commonly worn by Templars in service to Noc. Without the night there can be no day, without Noc there can be no light in the dark hours."
+	icon_state = "nochelm"
+	item_state = "nochelm"
+	emote_environment = 3
+	flags_inv = HIDEEARS|HIDEFACE|HIDEHAIR
+	block2add = FOV_BEHIND
+	smeltresult = /obj/item/ingot/steel
+
+/obj/item/clothing/head/roguetown/helmet/necrahelm
+	name = "necra helmet"
+	desc = "Headwear commonly worn by Templars in service to Necra. Let its skeletal features remind you of the only thing which is guaranteed in life. You will die."
+	icon_state = "necrahelm"
+	item_state = "necrahelm"
+	emote_environment = 3
+	flags_inv = HIDEEARS|HIDEFACE|HIDEHAIR
+	block2add = FOV_BEHIND
+	smeltresult = /obj/item/ingot/steel
+
+/obj/item/clothing/head/roguetown/helmet/dendorhelm
+	name = "dendor helmet"
+	desc = "Headwear commonly worn by Templars in service to Dendor. Its protrusions almost resemble branches. Take root in the earth, and you will never be moved."
+	icon_state = "dendorhelm"
+	item_state = "dendorhelm"
+	emote_environment = 3
+	flags_inv = HIDEEARS|HIDEFACE|HIDEHAIR
+	block2add = FOV_BEHIND
+	smeltresult = /obj/item/ingot/steel
 
 /obj/item/clothing/head/roguetown/helmet/heavy
 	name = "barbute"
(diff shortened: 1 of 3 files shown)


Teragenic:
I have a question. Why would you then ever order a Templar Helm over a Visored Sallet? Or hell. This PR also nerfs Templar Helm health to 200.

Essentially you get a themed NO ADJUST Visored Sallet with mouth protection.

It costs the same amount of metal, but now it doesn't have the Crit resistance.

You could get a Knight Helmet for the same amount of steel, it includes Crit resistance AND has more health than a Visored helmet.

Teragenic:
Also I think all helmets protect from hits. The heavy Category just adds protection to SMASH.

Essentially this PR trades vision for pick, smash vulnerability and less health.

Leudoberct1:
I've never once seen someone order a Templar helmet at the smithy to begin with. Yes, this puts it basically level with a visored sallet, minus the adjust. That's pretty much what I've already said.

You can get a visored sallet for the same amount of steel as a Knight helmet, too. Knight helmets also have adjust. I feel like I'm repeating myself to you here, but the entire tradeoff with heavy helms is that they restruct FOV to 90 degrees. In the case of Knigbt helms and the like, you can adjust them to alleviate this. You can't do this with Templar helms, and this leads lots of Templars to remove or swap their helms so they can actually see what the hell is happening in the Church they're meant to be guarding.

If I kept them heavy and removed the FOV reduction, you've now suddenly got a heavy helmet with all of the bonuses and none of the drawbacks. They're being moved out of heavy for the sake of balance.

pilotmans:
templars have great parry yet are given helmets to encourage tanking hits by blocking their vision whilst wearing mere chainmaille, this is a change for the better.

Leudoberct1:
Always found it weird myself, getting a kite shield yet being unable to use it to parry half the time due to the lack of vision.

DelineFortune:
Different players different player styles 
Yes, its hard to protect the church because you can see almost nothing and supposed to be equipped 
But also having a shield is not only one playerstyle that supposed to be (its not cool to wear a helmet and shield 24/7 as a female that i am, because people will think that its another arcade server with 18 years old one-man-armory catgirl capitans)

Id like to offer second option. You can adjust your templar helmet with a delay, to turn it into sentinel mode (you are loosening your internal helmet's plate, to get less protection but more vision, or you are hardening it to get more covered zones and less vision).

Dunno about changing the sprite, maybe description only. Would be a good feature to -not let players know- that you can see them because your helmet in sentinel mode. Or maybe you can't, because its in defence now.

Also, having 400 hp helmet is important in low pop as templar, because you can win some time to yell to non combat churchlings to run away. With 200 hp helmets and 2+ enemies you will lose after 2-3 mace/flail hits.

Leudoberct1:
Giving the helmet the ability to swap between heavy and non-heavy is something that would exist only in the  Templar helmets, and is pretty unfair. It would make them a pretty meta item to have.

pilotmans:
> Different players different player styles 
> 
> Yes, its hard to protect the church because you can see almost nothing and supposed to be always fully equipped while guarding.
> 
> But also having a shield is not only one playerstyle that supposed to be (its not cool to wear a helmet and shield 24/7 as a female that i am, because people will think that its another arcade server with 18 years old one-man-armory catgirl capitans)
> 
> 
> 
> Id like to offer second option. You can adjust your templar helmet with a delay, to turn it into sentinel mode (you are loosening your internal helmet's plate, to get less protection but more vision, or you are hardening it to get more covered zones and less vision).
> 
> 
> 
> Dunno about changing the sprite, maybe description only. Would be a good feature to -not let players know- that you can see them because your helmet in sentinel mode. Or maybe you can't, because its in defence now.
> 
> 
> 
> Also, having 400 hp helmet is important in low pop as templar, because you can win some time to yell to non combat churchlings to run away. With 200 hp helmets and 2+ enemies you will lose after 2-3 mace/flail hits.

As I said before you're wearing a hauberk and a Gambison 
, it doesn't matter how strong your helmet is when people can full force stab you through your chainmaile with either daggers or pole-arms

Altangy:
As an avid templar player, there is a time to use the low FOV helm. If I want a sallet we templars usually have enough coin to replace our entire armor set and the priest to bug for more coin otherwise

Leudoberct1:
The point is that the low FOV helmet is not suited to the Templar loadout or job. People say you can just get a sallet, but to me, your iconic Templar helmet should be suitable for your job as a Templar. To me, if you REALLY want the low FOV, you can likewise just go get a Knight helmet.

Altangy:
It is suitable to templars. Like everything though it is a tool with its uses. The helm has carried me through a number of fights already and sat in my pack for the more casual scuffles. Personal preference but I can fight just as well with or without the lack of sight.

Leudoberct1:
I am also an avid templar player- to the point where I've played it every single day for the past two weeks. It can be pretty damn crippling not having any decent line of sight, especially in any larger fight. I believe it is FAR more suitable for the helmet to be more useful in general situations, especially given the standard loadout of a Templar. A low FOV helm is useful for people with the armour to tank hits, due to not being able to parry most attacks. You do not start with heavy armour. You start with a shield, which is again useless if you can't see people to parry with. I believe it is far, far, far better to have the higher FOV and therefore be more compatible with the standard Templar loadout. You shouldn't NEED to have to go and bug the smiths each round to feel combat viable.

Leudoberct1:
Screenshot taken from a round. Templars all abandoned their helmets because they couldn't fight in them. https://gyazo.com/9622b6ebc05ee72d36ca06e159e33159

PR #746, opened 2024-08-08
IRS

Heheniy:
## About The Pull Request

Tax evasion abuse by depositing amount of coins which would have tax value <1 mammon is no more.
Minimum tax (untill you are noble) is always 1 mammon.
Also noble trait wasnt removing your taxation, only message about it- it is fixed now too

### DIFF
--- a/code/controllers/subsystem/rogue/treasury.dm
+++ b/code/controllers/subsystem/rogue/treasury.dm
@@ -140,10 +140,10 @@ SUBSYSTEM_DEF(treasury)
         // Player received money
         if(source)
             send_ooc_note("<b>The Bank:</b> You received money. ([source])", name = name)
-            log_to_steward("+[amt] from treasury to [name] ([source])")
+            log_to_steward("[amt] from treasury to [name] ([source])")
         else
             send_ooc_note("<b>The Bank:</b> You received money.", name = name)
-            log_to_steward("+[amt] from treasury to [name]")
+            log_to_steward("[amt] from treasury to [name]")
     else
         // Player was fined
         if(source)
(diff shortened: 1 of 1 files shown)


zakanater19:
> On the Discord I already gave a reason for why there shouldn't always be the minimum one-mammon tax, but I'll reiterate here. For one, deliberate tax evasion is at least semi-obvious on the Nerve Master, and you as the Steward have the tools to punish that directly with fines, or optionally through IC confrontation. Also, it's a bit weird if the machine whose sound implies that you're inserting coins one-at-a-time just eats a single zenny that you put in.
> 
> The noble trait oversight is my fault lol, that was definitely one of those late-night PRs.

in my experience by the time you check the nerve master logs and see someone evading taxes, they have already cleared out their account so there is nothing to fine, best you can do is go ask the baliff or king to make them an outlaw

zakanater19:
> > best you can do is go ask the baliff or king to make them an outlaw
> 
> Conflict good. Probably shouldn't go all the way with making them an outlaw, and making every single tax evader an outlaw first to ask questions would be VERY impractical. Telling the watch to apprehend the guy if you see them would probably work a lot better. A good number of the tax evaders with deep pockets are rooted-down anyways, such as the Merchant. For them, stuttering deposits likely also becomes impractical and sticks out even more than it otherwise usually would. Things are fine as they are.

Making them an outlaw will make the guards apprehend them, going around trying to tell every guard IC just wastes time, not every SCOM message will be heard by the guards, making someone an outlaw will pressure them to either flee or surrender, your solution does nothing but frustrate and annoy

DeerJesus:
Noooo... Noooo. This is realllly stupid.

PR #748, opened 2024-08-08
Treasure hunter actual luck

Heheniy:
## About The Pull Request

At the treasure hunter rework pr, when was asked to add luck to treasure hunter, author added 1 fortune (even miner has 2) for cost of 1 speed, which is almost nothing and still doesnt make rare(!) class anyhow unique.
Now treasure hunter has random 5-10 luck buff, allowing him actually do his job (hunt treasures) more efficient. 

## Why It's Good For The Game

Rare anventurer class now actually not useless

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/combat/rare/treasurehunter.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/combat/rare/treasurehunter.dm
@@ -49,4 +49,4 @@
 		H.change_stat("intelligence", 1)
 		H.change_stat("speed", 3) 
 		H.change_stat("endurance", -2)
-		H.STALUC += rand(5,10)
+		H.change_stat("fortune", rand(5,10))


Hmeister-real:
thats a /lot/ and brings the treasure hunters stat count from 5, all the way up to a maximum 16 if rolled correctly which is... something. I guess this could work but it's really powerful, and if they mine they could essentially guarantee gems and gold

ikrarkjr:
This PR fucking blows treasure hunters should not be infinitely better than the LEGENDARY miner highroll pilgrim class.

Lhars:
> thats a /lot/ and brings the treasure hunters stat count from 5, all the way up to a maximum 16 if rolled correctly which is... something. I guess this could work but it's really powerful, and if they mine they could essentially guarantee gems and gold

Actually, mining drops are a flat roll based off of the type of turf you're mining. Fortune does nothing in this regard.

All this PR will do is give the treasure hunter a disgustingly high critical chance.

PR #756, opened 2024-08-09
Feint Mini Rework (Now scales with Int and a bit with SPD)

ddsaw:
## About The Pull Request

Reworks feint to be based on int difference and to a lesser degree spd difference. Changes the impact of feint intent to instead grant a longer immobilize effect (still barely notable but very hard to test on a local server, its still too quick to tab between two clients and see if I can move)
The actual number changes:

SKILL IMPACT 10% or 15% if feint intent -> always 15%
INT IMPACT Singular check to see if the opponent has more int -> int difference *10
SPD IMPACT None -> 5% per difference
Feint Intent -> Increases Impact of Skill difference -> Increases duration of Immobilize

Please doublecheck the code, but I ran it locally and tested it and it seemed to work as intended.

## Why It's Good For The Game

Feint is a mostly unused mechanic, this gives it some use in int builds, and to a degree SPD builds. Simultaneously counters powerful intdump builds like orc barbarians by letting reasonably clever players reliably feint them.

This is technically a Buff to things like Knight, but let's be real they would defeat your parry anyways, and people spamming feints at them in mobs will be a bigger issue than any boon this gives to them.
And feinting has an extremely punishing hardcounter in the form of dodgeintent anyways.
Still, I clamped the chance to 90% to keep a bare minimum of risk.

One of the few funny consequences is that old mages and the archivist will be able to reliably feint most players, but I doubt this is going to be any real issue. Chalk it up to the cunning trick of an actual genius.

### DIFF
--- a/code/game/objects/items/rogueweapons/rmb_intents.dm
+++ b/code/game/objects/items/rogueweapons/rmb_intents.dm
@@ -26,13 +26,12 @@
 				I = L.get_active_held_item()
 				if(I?.associated_skill)
 					theirskill = L.mind.get_skill_level(I.associated_skill)
-		if(ourskill > theirskill)
-			if(istype(user.rmb_intent, /datum/rmb_intent/feint))
-				perc += (ourskill - theirskill)*15
-			else
-				perc += (ourskill - theirskill)*10
-	if(user.STAINT < L.STAINT)
-		perc -= 15
+		perc += (ourskill - theirskill)*15 	//skill is of the essence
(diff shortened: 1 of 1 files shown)


H0gberg:
very good pr, feints were already pretty strong, this only grows that niche a bit more, I think it makes sense if you fight against a bumbling idiot character there should be a way to exploit their weaknesses

H0gberg:
Though there should also be some impact that one's perception has as well

PR #757, opened 2024-08-09
Adds some overflow cells to the keep dungeon

zakanater19:
## About The Pull Request

More simple cells for prisoner overflow, they are not hostage cells like for the upper classes but are just basic lower dungeon cells, giving the garrison and keep more cells for common criminals to rot awaiting trial, i have tested all the cell buttons and doors and they work

## Why It's Good For The Game

It's just a nice place to leave prisoners where they both cannot scream and shit up the throne rooms diaglog, and also where the dungeoneer can keep longterm prisoners

Screenshot 2024-08-09 191315

Screenshot 2024-08-09 191303


Hmeister-real:
they look too nice, each cell shouldn't have its own candles, and it needs to be more grimdark

zakanater19:
> they look too nice, each cell shouldn't have its own candles, and it needs to be more grimdark

Yeah ill get around to making them look better eventually (soon maybe)

hugmeorelse:
Server is no longer grimdark, rather dark fantasy. So, candles are fine, but I think it'll look better anyways being a tad more sparse.

Also, instead of these many individual cells, I recommend having one big group cell with the dungeon door that has the visor, so prisoners can't see out and only people checking in the cell can see in. Have a few smaller solitary cells that's tiny for punishment too. I feel like that would be far more interesting and get far more use than many individual cells that will more than likely go empty.

PR #758, opened 2024-08-09
You can no longer eep in armor

GuppyLaxx:
Heavy armors no longer let you sleep. I tried to cover all the heavy armor I know, including chainmail. Leather is excluded. I also included the crown because I find it fitting.

image

This is part of a wider range of changes mostly targetted at Knights. They will be unable to sleep in their own, heavy armor. They will also need someone else to help them take off and put on their FULL plate of armor, unless they want to sit and wait *minutes* to wear it on their own. Squires will be given the ability to clean, polish, and repair the armor and weapons.

Also, now adds a suggested new special, makes you unable to sleep unless you are in bed, nude. It's got a decently low weight.

image

### DIFF
--- a/code/__DEFINES/traits.dm
+++ b/code/__DEFINES/traits.dm
@@ -72,6 +72,7 @@
 #define TRAIT_NORUN "Decayed Flesh"
 #define TRAIT_PUNISHMENT_CURSE "PunishmentCurse"
 #define TRAIT_LEPROSY "Leprosy"
+#define TRAIT_NUDE_SLEEPER "Nude Sleeper"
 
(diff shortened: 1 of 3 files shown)


DeerJesus:
You can sleep in chainmail. Source:

Dull-man-builds-himself-a-chainmail-blanket-3-1536x1421

DeerJesus:
Also for shits and giggles you should add a 'special' trait with this PR that requires you to sleep in the nude

zakanater19:
> Also for shits and giggles you should add a 'special' trait with this PR that requires you to sleep in the nude

Lol good idea

PR #764, opened 2024-08-10
Adds a delay for taking off armor, heavy nerf to Knights

GuppyLaxx:
Clothing(Technically all items) can now have a delay to self de-equipping, or even be entirely unequippable on your own. 
Armor now has a delay to being taken off.
PLATE armor has a HUGE penalty, 2 minutes to be taken off on your own. It takes someone else 6 seconds to take it off you instead.

image
*(This is for items you cannot take off yourself. Plate is actually set to the earlier stated penalties. There are no items that you cannot take off yourself aside from NO DROP ones currently. Maybe change the prisoner mask?)*

Also re-introduced text when putting on items, and also added it to taking stuff off. It's only for things that take longer than a 1.5 seconds.

This is the 2nd part of my KNIGHT CHANGES. Ideally my next PR will be some decent changes to Squire, where they'll get the ability to clean up and polish full plate and weapons to give them some temporary buffs or something. I want Knights to have down time from their armor, and for Squires to be more useful. This also gives these two roles a dynamic of their own.


X404ErrorX:
From 80 to1200 to equip and unequip lmao

H0gberg:
giving squire a tiny bit of armorsmithing (one point in it) would be the equivalent of them 'buffing' out any issues with armor

DorfGuy1:
going to make a lot of tedium. The Squire thing is cool, but this will not be. I love the attempt at realism and faithfulness to medieval shit though. I really do.

ApplePlastic:
I do not like tedium for the sake of adding tedium

radarseas:
antagonists should get some lee-way with a unique trait to bypass this since they don't usually have squires. especally bandit hedge knights.

_Trust me, its not going to be fun as the only bandit and a hedge knight spawn to be waiting almost an entire day almost to remove your armor so you can sleep only to get tired, again._

HissyNays:
absolute coal. This is actually headache inducing change. Delete this.

PR #769, opened 2024-08-11
Fixed: swords PR, now with 25% more swords

H0gberg:
First PR with Azarak's help

## About The Pull Request

Look at the closed sword PR for further info


pilotmans:
Looks great, love the selection and having a steel messer is also pretty good

ddsaw:
The Messers are the same as bastard swords, except slightly better at chopping. Cute, I guess. Not very notable.

The Estoc would finally provide a decent armor piercing sword option, doing much to narrow the gap between Swords and other more useful weapon types such as maces, flails and polearms.
I would almost consider giving it a long windup/cd (longer than dagger) "pick" intent on 2h, to signify a halfsword thrust.
Infact a bunch of swords really deserve something like that, to give them at least some way of fighting armor, so it might be out of scope for this PR.

I expect a lot of Estoc wielders if this goes through, but I think that is just a symptom of all other swords being really bad once opponents wear any armor.
It's still much worse than any good mace or polearm, without even speaking of the sheer brutality of flails.

PR #770, opened 2024-08-11
Fixes Hand allowing shunned races

Hmeister-real:
## About The Pull Request

nuf' said. Kobold Hand was fun while it lasted, but man was it stupid

## Why It's Good For The Game

Kobold being royalty is bad for so many reasons, especially one as important and commanding as the Hand

### DIFF
--- a/code/modules/jobs/job_types/roguetown/nobility/hand.dm
+++ b/code/modules/jobs/job_types/roguetown/nobility/hand.dm
@@ -6,7 +6,7 @@
 	total_positions = 1
 	spawn_positions = 1
 
-	allowed_races = RACES_SHUNNED_UP
+	allowed_races = RACES_TOLERATED_UP
 	allowed_sexes = list(MALE, FEMALE)
 	outfit = /datum/outfit/job/roguetown/hand
 	display_order = JDO_HAND


Takoros:
Kobolds already are restricted by half of the roles, dont take one more.

Hmeister-real:
> Kobolds already are restricted by half of the roles, dont take one more.

sorry Takoros but being a noble is bad... i do like your bog master PR though, hopefully it gets merged with this to balance things out

pilotmans:
Agree with this, I should not be able to either easily kill the hand to steal their keys or deal with the fact that somehow a kobold outranks me, it's fun but also rather annoying to role play with especially if they have the whole (kobolds are stupid) shtick

Takoros:
> sorry Takoros but being a noble is bad... i do like your bog master PR though, hopefully it gets merged with this to balance things out

Well.. If my PR is merged and this one too.. it might.. be okay i guess.. but if only this one, it will be sad.

LocalizedDownpour:
Hand allowing any Race should be allowed as it's basically the Kings chosen friend. The King can 'choose' a shunned person if they want to. Also I don't think the Hand is mechanically a noble.

PR #777, opened 2024-08-12
Silver rings

Heheniy:
## About The Pull Request

Steel gem rigns now requires silver 

## Why It's Good For The Game
image


Elli-Skala:
I'm fairly sure the code comment was from a different vision. Silver is vastly VASTELY more rare then gold in our current state which I don't imagine the code comment accounted for

BladeburstNINJA:
What's funny is that silver tends to drop along with gems, making silver gemrings pretty much guaranteed if you locate silver ore.

PR #788, opened 2024-08-14
Allows Half-Kin access to all Penis types

EzekyalLegit:
## About The Pull Request
Currently half-kin only get human or canine penis options even though they have all the animal feature options of anthros.

This gives them all penis options.

## Why It's Good For The Game
More customization for players. Allows halfkin besides canine to have appropriate genitals.


Red-Honk:
I see this as a total win.

PR #790, opened 2024-08-14
Pirate and Smuggler + Changes to a few things

SarmentiCampbell:
## About The Pull Request
Added Smugglers and Pirates from #205 (restricted to axian only, hah)
allowed witch hunters to be female, just like the inquisitors
buffed legendary fishers(why were they weaker and shittier than normal fisher?) while buffing the spawn rate of grubs
reverted the removal of eating rare fish giving you buffs, now giving you a reason to cook them and sell them rather than just selling them to the merchants.


DorfGuy1:
Are we just not getting the harbormaster, ever? I think it'd be really cool to have a more 'old school ss13 cargo' type thing. And more people with actual jobs in town that are not just garrison troops would be cool.

PR #798, opened 2024-08-15
Restricts the Royal Family to Human and Half-Human only.

GuppyLaxx:
The extreme variance in race between Royal Families is extremely jarring at times, and entirely ignored about 90% of the time.

While it can make a good plot hook, it's such an extreme difference every single time that it leads to nobody wanting to go through the effort of coming up with excuses and playing out why the Human King and Human Queen have a bunch of dog people as their children/successors.

The discrepancies that can play out right now are way too extreme and at times, same-y to really be able to play this out in fun ways. Calling out the princess because she is a cat and the rest of her family is very different is fun once. It's not fun when the next rounds, the same Royal Family happens in terms of races, or even players, and you have to decide between de-railing the whole round to bring up the differences or be entirely ignorant.

I think that letting Half-Humans still be able to pick these roles gives a fun angle to play these scenarios, still lets people have fun creating a character they want for these roles, and gives much more plausible explanations between WHY there's a half-kin in the family while the rest are all Humans.

Though I don't think this is THE best option, or end all be all, I think that as for now it's a good quick change that helps shake up roleplay while much more in-depth systems are worked on in terms of handling families and race between them.

### DIFF
--- a/code/__DEFINES/roguetown.dm
+++ b/code/__DEFINES/roguetown.dm
@@ -138,6 +138,13 @@
 	/datum/species/anthromorph,\
 	/datum/species/demihuman,\
 )
+
+#define HUMANS_AND_HALFS list(\
+	/datum/species/human/northern, \
+	/datum/species/human/halfelf, \
+	/datum/species/demihuman, \
+	)
+
 #define ALL_CLERIC_PATRONS list(/datum/patron/divine/astrata, /datum/patron/divine/noc, /datum/patron/divine/dendor, /datum/patron/divine/necra, /datum/patron/divine/pestra)
 
 #define ALL_ACOLYTE_PATRONS list(/datum/patron/divine/astrata, /datum/patron/divine/noc, /datum/patron/divine/dendor, /datum/patron/divine/pestra)
--- a/code/modules/jobs/job_types/roguetown/nobility/lady.dm
+++ b/code/modules/jobs/job_types/roguetown/nobility/lady.dm
@@ -7,7 +7,7 @@
 	spawn_positions = 1
 
 	allowed_sexes = list(FEMALE)
-	allowed_races = RACES_TOLERATED_UP
+	allowed_races = HUMANS_AND_HALFS
 	tutorial = "Picked out of your political value rather than likely any form of love, you have become the King's most trusted confidant and likely friend throughout your marriage. Your loyalty and, perhaps, love; will be tested this day. For the daggers that threaten your beloved are as equally pointed at your own throat."
 
 	spells = list(/obj/effect/proc_holder/spell/self/convertrole/servant)
--- a/code/modules/jobs/job_types/roguetown/nobility/lord.dm
+++ b/code/modules/jobs/job_types/roguetown/nobility/lord.dm
@@ -10,7 +10,7 @@ GLOBAL_LIST_EMPTY(lord_titles)
 	total_positions = 0
 	spawn_positions = 1
 	selection_color = JCOLOR_NOBLE
-	allowed_races = RACES_TOLERATED_UP
+	allowed_races = HUMANS_AND_HALFS
 	allowed_sexes = list(MALE)
 
 	spells = list(
--- a/code/modules/jobs/job_types/roguetown/youngfolk/prince.dm
+++ b/code/modules/jobs/job_types/roguetown/youngfolk/prince.dm
@@ -7,7 +7,7 @@
 	total_positions = 2
 	spawn_positions = 2
 	f_title = "Princess"
-	allowed_races = RACES_TOLERATED_UP //Maybe a system to force-pick lineage based on king and queen should be implemented.
+	allowed_races = HUMANS_AND_HALFS //Maybe a system to force-pick lineage based on king and queen should be implemented.
 	allowed_sexes = list(MALE, FEMALE)
 	allowed_ages = list(AGE_ADULT)
 


MefAnneFeminene:
Do not merge this

Squird5:
Human King and Human Queen
Dogboy Prince
You Just Know

```#define HUMANS_AND_HALFS list(\
	/datum/species/human/northern, \
	/datum/species/human/halfelf, \
	/datum/species/demihuman, \
	)

EmilyFontaine:
I've been playing Queen Consort a lot lately, and I've been attempting to do what the comments suggest.

However, most Prince/Princess characters simply refuse to accept it, or get offended by the notion that they're not your children. Most rounds start off with the Prince/Princess scuttling over and going "Hello, mother!" to the queen, without giving these same considerations mentioned here any thought. The cat's already out of the bag at this point, so it's on the queen to choose whether to play along with it, or cut the imaginary umbilical cord and tell them that they're a bastard. Most players that I've done this with feel offended, to the point where it feels like they get slighted on an OOC level, which leaves a bad taste in my mouth.

I sometimes see the comparison that we're meant to be actors on a stage, and playing the royal family feels like being on a stage with 4 people with 4 different scripts. I understand that RP on a server like this is largely improv, but the rules, and setting and the hierarchy of the roles establish a basis, and playing the royal family currently feels like you're playing without a basis, except that you know that the king is the head of the house.

It's really tiring to have to do all these mental gymnastics at the start of every round, when there's no way to agree on these things on an OOC level. We want to handle everything IC here, and I think that's great, but handling this IC feels atrocious every time, because of what I stated before - people seem to take genuine, OOC issue with it, and I don't have the ability to tell them that it's not because I dislike their character, I'm just trying to take a different route with mine. "Just pretend" isn't good enough when everyone's imagination is clashing.

DeerJesus:
Dwarves have a good reason to be king
Elves have a good reason to be king
Aasimar have a good reason to be king
This is stupid

SixplyDev:
> Do not merge this.
> 
> Rockhill is an intentionally semi-multicultural society, as outcasts of other societies end up there. Humans also only make up a minority of the nation's population so it wouldn't make sense for only Humans to be royalty. There is no real tangible reason to make only Humans and half-kin as royalty, unless you wanna further shorten the playerbase of royalty, and ruin the lore while you're at it.
> 
> Other races existing in the royal family can so easily be explained as having been married in, in 90% of cases, a royal family will marry their daughters, sons, etc. Into another royal family, and with a world that has such a diverse selection of races such as Ratwood, it is kind of impossible that there would only be Human nobility, unless all the Royalty are inbred which would be funny.
> 
> So it's either:
> 
> * Inbred Human-only Nobility
> * Married-in nobility with the other races

The uhhh royal families of Europe were like historically comically horifically inbred to shit. The Hapsburgs even have a fucking jawline named after them because they were so inbred. An inbred royal family makes a ton of sense

Red-Honk:
Have you fellas forgotten that Kingdoms would give away their own child for the other one to raise as a sign of trust and goodwill?

Alanii:
# Suggestion
Restricting the roles will just bring more people to complain and whine. It would be better to implement a system that checks the race of the king and queen and lock the prince/princess to specific racial groups based on that when the game starts.

If no one is available for that race, then the heir player can late-join. This way they can come up with a story explaining why they are the heir.

This way the heirs are typically some-what making sense. It also allows some people who roll for the heir role to have a higher chance due to the king/queen race.


## Regarding the comments of people going "heirs dont need to be related"
The original intent behind the "heir" was to be biological children of the King and Queen. Sure, its an attempt to justify why the Human King, and wild kin mother have a dark elf son, but it just doesn't work. Players need to be mechanically supported when role playing. You can not expect every single person to go "Ah yes the tiefling prince, clearly they are born of the elf mother and human king" or make up a story of "they're adopted" or some other nonsense. When role playing both the systems in place (code) and the game need to support the role play. =

You can't expect everyone to 'pretend' when things just don't really make sense, otherwise you risk people being pulled out of the game and acting out in odd ways that do not feel immersive to the setting. There needs to be actual mechanical support behind much of the role play done. This way the game feels more immersive, and you have less players thinking it is okay to do ridiculious acts as the game brings the player into their character and into the game's atmosphere.

CalamaBanana:
The more fun solution would be to leave King and Queen as not race restricted, yet prince/princess will always be forced to be the appropriate mix of their parents. 

Examples:
Elf King, Human Queen -> Prince/Princess will be forced half-elf, skintone meeting both parents in the middle, hair/eye colour randomly picked from parents.
Wildkin King, Human Queen -> P/P will be forced half-kin with visual attachments matching the King.


With 10+ princess players signing up every round, I am sure at least two of those would stick around even if it means they end up with slightly randomized features instead of playing that same OC that was princess 50 rounds in a row.

Keazra:
An idea: if the royal family is meant to be a single bloodline or mildly intermixed at best.. perhaps non-humans / non-demihumans can become royal family, but they are "Wards" and not "Princes/Princesses"? Maybe Prince (Ward) or Princess (Ward) could be appended to their title, to denote they're from another bloodline.

DeerJesus:
> # Suggestion
> 
> Restricting the roles will just bring more people to complain and whine. It would be better to implement a system that checks the race of the king and queen and lock the prince/princess to specific racial groups based on that when the game starts.
> 
> If no one is available for that race, then the heir player can late-join. This way they can come up with a story explaining why they are the heir.
> 
> This way the heirs are typically some-what making sense. It also allows some people who roll for the heir role to have a higher chance due to the king/queen race.
> ## Regarding the comments of people going "heirs dont need to be related"
> 
> The original intent behind the "heir" was to be biological children of the King and Queen. Sure, its an attempt to justify why the Human King, and wild kin mother have a dark elf son, but it just doesn't work. Players need to be mechanically supported when role playing. You can not expect every single person to go "Ah yes the tiefling prince, clearly they are born of the elf mother and human king" or make up a story of "they're adopted" or some other nonsense. When role playing both the systems in place (code) and the game need to support the role play. =
> 
> You can't expect everyone to 'pretend' when things just don't really make sense, otherwise you risk people being pulled out of the game and acting out in odd ways that do not feel immersive to the setting. There needs to be actual mechanical support behind much of the role play done. This way the game feels more immersive, and you have less players thinking it is okay to do ridiculious acts as the game brings the player into their character and into the game's atmosphere.

While this would be complicated, it would be pretty neat.

Each race can be prince if either of their parents is that race

Half elves can come if there is one human/aasimar/dwarf (elf tall gene overtakes dwarf small gene) and one elf parent

Halfkin come from human/aasimar and wildkin/vulpakinin/lupian

We could even have unique fucked up special rolls too, like, if a dwarf and a wildkin are married a verminvolk can roll prince, but not latejoin, same with dwarf+sissean/dracon and kobolds

wow50000:
queen should be allowed to be any race except shunned and maybe tolerated

ThinkArbies:
Allow elves, dwarves, aasimar and half-kin to be part of the royal family, too and this PR is golden.

Any less than this and you'll see a massive dip in _competent_ royal family players.

Useroth:
Not going to be the case for time being

MefAnneFeminene:
> half kin, as if they are not constantly the lowest quality players.

Even me?

PR #799, opened 2024-08-16
Makes jester str, luc and int rolling a curve

Slorbin:
currently, jester stat rolling is all sorts of messed up.  you have... way too high a chance to roll a 20 or a one. a 20 in strength, no one can beat you. a 1, you can't actually play the game at all. 20 in int, you can learn anything. 1 in int - you can't actually craft a single damn thing, which is annoying if you're trying to make funny shit for japes.

this changes it to follow this chart **(MINUS FOUR)**. subtract 4 from this chart (0 counts as a 1) and that will be the probability

4-chart

### DIFF
--- a/code/modules/jobs/job_types/roguetown/courtier/jester.dm
+++ b/code/modules/jobs/job_types/roguetown/courtier/jester.dm
@@ -42,9 +42,9 @@
 		H.mind.adjust_skillrank(/datum/skill/misc/stealing, 5, TRUE)
 		H.mind.adjust_skillrank(/datum/skill/misc/climbing, 4, TRUE)
 		H.mind.adjust_skillrank(/datum/skill/misc/music, 2, TRUE)
-		H.STASTR = clamp(roll(4d6-4), 1, 20)
-		H.STAINT = clamp(roll(4d6-4), 1, 20)
-		H.STALUC = clamp(roll(4d6-4), 1, 20)
+		H.STASTR = clamp(roll("4d6"-4), 1, 20)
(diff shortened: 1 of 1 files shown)


YeCrowbarLamb:
this pr has screwed jesters out of their traits and combat music, please unmerge this

ChungusGamer666:
this pr takes a lot of the uniqueness about jester as a role, which is sad since it is a one or two persons in a round thing

DeerJesus:
> this pr takes a lot of the uniqueness about jester as a role, which is sad since it is a one or two persons in a round thing

It's still there and possible, and the jester still gets the wildly unique traits of high jumping and the unique roleplay position they get.

I kindof do wish I could change the distribution curve to be a bit more lenient and allow ones and twenties more than I set it- though I'm not sure of a very eloquent way to do that other than doing a pick() and a ton of numbers.

ChungusGamer666:
i believe there is a nornal distribution function in the code that would work better than this, just use it with something around 5 as the standard deviation apply round on top and you should be able to get reasonable stat distributions

ChungusGamer666:
Anyways not an awful pr i get why this was made tbh but i have a soft spot for how ridiculous jesters were on blackstone

DeerJesus:
> Anyways not an awful pr i get why this was made tbh but i have a soft spot for how ridiculous jesters were on blackstone

Yeah I like to play jester and I liked it a lot before the sleep update. Post sleep update, it makes doing any cool shit as jester like, take multiple IRL hours and it gets - rough to put it likely. I'll keep looking around for that normal distribution function and see if I can make it a curve that makes it a little better.

PR #801, opened 2024-08-16
shoves trapmaking into engineering

Slorbin:
what bona fide retard made a skill that has a single fucking use and that is to make a single fucking item


Elli-Skala:
Apart of me feels like this should be more crafting then engineering as this has given *alot* of roles the ability to make trap doors and passages which was a fairly rare skill before

Murderfoxen:
It has potential as a stand-alone skill for rangers/hunters. It just doesn't have any recipes or solid implementation.

Useroth:
We'll end up reworking it in a different vein

PR #809, opened 2024-08-18
Lets you change your own PQ (debug)

Hmeister-real:
## About The Pull Request

great for debugging and testing on a localhost, if actual admins abuse this everything on the server is logged so it can be seen

### DIFF
--- a/code/modules/admin/playerquality.dm
+++ b/code/modules/admin/playerquality.dm
@@ -181,8 +181,6 @@
 		if(!selection)
 			return
 		theykey = selection
-	if(theykey == ckey)
-		to_chat(src, span_boldwarning("That's you!"))
 		return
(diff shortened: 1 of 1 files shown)


Useroth:
On localhost you can literally edit it in your json file

john-pawson:
> On localhost you can literally edit it in your json file

Why not just make it more convinient to add so you don't have to get out of the game and edit a .json file every single time you want to change your PQ? Admins abusing it is clearly not going to happen since the clear logs. (And it's not like PQ has that many benefits for admins to be abusing it)

Not counting times when someone wants to do testing on a server with multiple people and not everyone has access to the .json file

So... why not? It's already coded, all you have to do is accept the 2 line change.

PR #815, opened 2024-08-19
Surrendering changes. No stripping/equipping people in Combat Mode.

GuppyLaxx:
This PR changes Surrendering to now require confirmation through a pop up.

Once you surrender, you are applied 2 debuffs. One lasts 60 seconds and makes you unable to use your hands properly, it also applies a slowdown and DISABLES COMBAT MODE.

The other debuff is a 30 second stun/knockdown that can be shaken to shorten it, AND for the entire 30 seconds you gain Crit Resistance.

The PR also changes stripping/chaining to make it so you cannot do it to people in combat mode UNLESS they are down on the ground, stunned, etc. And surrendering makes it so people strip/chain you faster.

Main points of this PR are to trade off surrendering into a much more mechanically based feature where you are ENTIRELY CRIPPLED, but get a bonus to stay alive. Depending on feedback, I might make it help you survive more things and the like, but the main gist is to encourage people to kidnap/arrest more anybody who arrests. Even if the person is injured or caused trouble earlier, the fact they are so crippled while surrendering helps make way for even those less inclined to take live prisoners to actually roleplay. It also makes killing someone who surrenders VERY purposeful, as you very literally have to sit there beating the crap out of them to make sure they're dead.


CalamaBanana:
Do not readd the popup please, the time it takes to slowly hover your mouse there and hit yes is the time it takes to get clicked to death.

When I initially "enabled" surrendering code wise, it was already in the code, just the hotkey was broken. One could always right click combat mode to surrender even before that, but no one even knew it existed. What I am saying is, no one surrendered by accident. There is no need for a popup, a popup would only lead to delaying most surrender attempts till is too late.

If you truly believe some people need that safety, please add a toggle to the options tab where the popup can be enabled/disabled.

MickDick1:
I think the timers are inflated but otherwise im on board.

CalamaBanana:
Another small addition that would be neat:
Allow people in combat mode to be chained standing, if they are aggro-grabbed. Having to push people down for it just feels a little gamey, and does encourage unnecessary violence to keep someone down who tries resisting too much.

DeerJesus:
Can you still go combat mode when chained?

DeerJesus:
this also prevents resisting during sex if you surrender.

Squird5:
The crit resistance would be nice, as I have multiple times been crit after surrendering and died shortly after. Also a way to auto extinguish fires as well, as currently surrendering prevents resisting, so if a cleric sees you surrender and sets you on fire anyways, you just die

K4rlox:
i believe surrendering should temporarily stop bleeding until you are unstunned and give a flat 50% damage reduction from everything for the following reason


there are certain people that would hit a person that surrendered and kill them before their superior could tell them to not do that

PR #829, opened 2024-08-21
Adjusts Steel Flail Damage

EzekyalLegit:
## About The Pull Request

Reduces steel flails smash damage from 45 to 30 by lowering its its base damage from 30 to 20. Also takes the swing delay off of peasant war flail.

## Why It's Good For The Game

Steel flail does DOUBLE the damage of iron flail currently, and its smash intent increases damage by another 50%:
Iron 15 (22.5 when smashing)
steel 30 (45 when smashing)

This is a massive leap no other weapon takes. 

Steel flail is a cheap weapon that costs 1 steel but with smash intent it is dealing 45 damage on hit. This is more than you can deal with any weapon asides from legendary unique swords. Along with 80 damage penetration you are dealing full or nearly full damage through any armor in the game.  

This massively over tuned damage number is causing steel flails to critical hit super consistently, since critical hits are based on how much damage is dealt. This is how a single player with a flail can defeat an entire army of players by dashing to each and breaking their spine or leg in a single hit. Something you do see often. 

This change would result in steel flail in one hand doing 30 damage on smash and 20 on strike, which is still more damage than most weapons can do in one hand. Even with this change flails are still easily the best weapon just by less of a margin.

As well the peasant war flail is a weapon that nobody uses, even bandits drop it when they spawn and ask to borrow somebodies knife. This is because it takes two hands to do the same thing other flails do in one hand. Its smash attack has reach but also has a delay making it impossible to hit moving targets. So I took off the delay in hopes people will use it.

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/flail.dm
+++ b/code/game/objects/items/rogueweapons/melee/flail.dm
@@ -166,7 +166,7 @@
 
 /obj/item/rogueweapon/flail/peasantwarflail
 	force = 10
-	force_wielded = 35
+	force_wielded = 25
 	possible_item_intents = list(/datum/intent/flail/strike)
 	gripped_intents = list(/datum/intent/flail/strikerange, /datum/intent/flail/strike/smashrange)
 	name = "peasant war flail"


EzekyalLegit:
Its worth noting people who frequent this games combat agree even after this significant damage nerf steel flail will remain the best weapon in the game.

hugmeorelse:
It still needs a bigger nerf.

HissyNays:
Overnerf. Rather than nerfing base damage, nerf the damage multiplier on the smash intent. I believe steel flail should do 30 damage on basic attacks like steel mace, there's less AP on it.

The iron flail on the other hand needs a bit of a buff being at jokingly small 15 base damage.

MickDick1:
I pretty much approve these changes without much complaint.

I really appreciate peasant war flail getting the smash delay removed I was thinking about that too. It has like 5 different horrible stats holding it back from being worth a damn and swing delay was the major dead weight.

EzekyalLegit:
> 15 and 10 damage isn't a giant change balance wise it still does like a third your health in 1 hit and doesn't change Crit chance or AP two of its biggest upsides

crit chance is based on how much damage you deal

PR #833, opened 2024-08-22
We dont like real gems here

Heheniy:
## About The Pull Request
Replaces real names of gems with their fictional analogs where it was missed
Golden gemerald ring receipt now gives scomm gemerald ring instead of basic gemerald ring with no mechanics

## Why It's Good For The Game
Dublicated items, one with mechanics and one without them, 'removed'
Using real names of things in this game is bad? Idk

### DIFF
--- a/code/modules/roguetown/roguejobs/blacksmith/anvil_recipes/valuables.dm
+++ b/code/modules/roguetown/roguejobs/blacksmith/anvil_recipes/valuables.dm
@@ -60,7 +60,7 @@
 	name = "Gemerald Ring (+1 Gemerald)"
 	req_bar = /obj/item/ingot/gold
 	additional_items = list(/obj/item/roguegem/green)	
-	created_item = /obj/item/scomstone
+	created_item = /obj/item/clothing/ring/emerald
 	i_type = "Valuables"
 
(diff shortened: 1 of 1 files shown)


DeerJesus:
doesn't this make it to where there are now two 'gemerald rings'

Heheniy:
they literally have same functional

i dont even know why there are two rings with same fnctional

DeerJesus:
> they literally have same functional
> 
> i dont even know why there are two rings with same fnctional

no no one ring (the emerald ring) is a SCOMstone and acts as a scom. the other (the gemerald ring) is just a regular ass ring. right??

Heheniy:
yeah you are right

DeerJesus:
based making them craftable, but that turns into more of a balance issue, and also there is *still* the steel gemerald ring. perhaps rename it to SCOMstone ring or something. or 'enchanted gemerald ring' or somethingf.

CalamaBanana:
Turning a normal smithed gemerald ring into a scomstone ring could be a recipe for high level alchemy instead, in my opinion. They are a very powerful item, and making them that easily smithable can be dangerous. You don't want to end up with "Vamp in maint"-spam on the scom network every round.

PR #835, opened 2024-08-22
Adds some stairs in the merchants so he can more easily access his balloon from the store front

zakanater19:
## About The Pull Request

Adds some stairs to ease the pain of hauling shit up to the balloon

## Why It's Good For The Game

Playing merchant is already quite a pain in the ass, this will lessen the pain slightly

https://github.com/user-attachments/assets/ca280093-6053-49a8-a435-ec5bae337d79

Screenshot 2024-08-22 223153

Screenshot 2024-08-22 223211


zakanater19:
Added a door and moved some things around


https://github.com/user-attachments/assets/09b70642-19ce-4d12-9d6f-dc32f5532064


Screenshot 2024-08-22 224246

CalamaBanana:
If you could go even further, and remove the old staircase, that would be great. Or at least shuffle that area around that that door there is closer to the desk.

If one locks the door there, one can not hear it getting broken in when standing at the merchant's desk. You can have twenty armed men break that door down and you'll never know.

PR #842, opened 2024-08-24
Gives a debuff upon dying. You can no longer be revived if you have this debuff.

GuppyLaxx:
This PR finally gives death some meaning!

You get two lives.
One you get to live wild and free!
The other you live crippled, and scared.

Upon dying, you now receive a permanent -2 debuff to all stats. The debuff also prevents you from being revived again. Period.
I was going to make the debuff stack in-case people figured out a way around this, but, the code for status effects affecting stats is DOGSHIT so I skipped on it. I might come back to this and make it stack.


Great-Beebo:
Exactly what this server needs for people to at least be even be a little scared of death. I can't count how many times I've seen people get up for the third or fourth time.

JGlitch:
Stop coding here.

Alanii:
> Give Tabaxi 9, it'll be funny.

Sounds dumb as shit. But it would be a funny quirk built into their race. Would be hilarious having someone going "I have 5 lives left"

Hmeister-real:
Good shit, even kept my -2 debuff idea. Finally death has some meaning. 

Only thing I can suggest is removing the randomised removal of a random stat when you're revived, that we currently use

DorfGuy1:
> Good shit, even kept my -2 debuff idea. Finally death has some meaning.
> 
> Only thing I can suggest is removing the randomised removal of a random stat when you're revived, that we currently use

One or the other would be nice. If you want to do a 'softer' version, just have it be a debuff that lasts a day or two. That prevents revival. 
If this goes though my one suggestion is to just give it a try and then put it up to player vote.

Takoros:
Death is punitive enough, remember, the players that dies already dont play for some time until raised and healed properly. If you really want to add a slight more punitive thing, only add like -1 to two random stats. -2 to all stat is really dumb

SixplyDev:
-2 to every single stat upon dying feels horrendously punishing considering everyone dies differently. It seems stupid to me that someone who got randomly fragged now has -2 to every stat after being saved or someone who was mid-combat and then gets beheaded while standing now has -2 to every stat. On the other hand, this also penalizes potential roleplay of a secret revival and then them coming back stronger than ever and other roleplay like that. I think a temporary debuff rather than permanent would be better or a -2 to a random stat. I do think death needs to be more punishing, however but I think for right now, this is a bit too far of a step in that direction considering the atmosphere of the server.

ikrarkjr:
Make it a status effect. Starts at -10 to all stats and goes down by 1 every 3 minutes.
Doesn't include intelligence or fortune.

Hmeister-real:
MERGE THIS ALREADY!!!

zakanater19:
based

PR #845, opened 2024-08-24
Fixes blocked door in the graveyard

Leudoberct1:
Door had a stone wall blocking it in the crypt, this is now fixed.

Picture: https://gyazo.com/747e99fda450d274145a0e3eee493f2a


Leudoberct1:
!bugfix

PR #818, opened 2024-08-19
Pickpocketing risk adjustments

CalamaBanana:
## About The Pull Request

Pickpocketing always felt massively overtuned considering to the amount of bad RP it creates, mostly just one player having a number go up, and the other one having their round messed up without much counterplay. Bonus points for it being coded in a way that there was 0% failure chance to be discovered while sprinting up and down the main road while MMB spamming random people. The "Someone tried pickpocketing me!" warning was impossible to trigger for rogues.

Changes:
- Added a very small delay to pickpocketing. No more complete drive-by pickpocketing.
- Shows "Someone tried pickpocketing me!" warning on attempted pickocketing fail, not just critical fail. No more endlessly spamming till one has a success.
- Shows "X fails to pickpocket Y!" to the crowd on a critical fail. Think of it like a failed attempt to snatch a purse, eyes will be pulled towards the thief.
- Adjusted critical fail chance. Before it was impossible to happen to rogues, now only legendary pickpockets will never critfail.

## Why It's Good For The Game

Pickpocketing was in most cases a completely risk free 0 RP method to severely mess with another player's round with usually very little gain for the thief (other than a number going up). With this change, pickpockets will hopefully be more targeted in their attempts, and not just randomly fish everyone who walks past. 

Pickpocketing can still be just as rewarding as before, but now it actually has some risk to it.

Perhaps people will dare to RP in town a bit more, without constantly being in combat mode and a wall to their back.

(Small note, I did not change much in the code, just looks that way because I had to add indentation to that entire block for the delay to be added)

PR #847, opened 2024-08-24
Revert "Pickpocketing risk adjustments"

Hmeister-real:
Reverts #818 because it ruined pickpocketing, and needs a different change that isn't... This.

PS: Once I can (in two weeks or so) Il revert all the pickpocket changes since blackstone, and then fix the whitelist making neck armour unable to be pickpocketed, and bags/pouches have to use a knife to steal them and some other stuff

### DIFF
--- a/code/_onclick/other_mobs.dm
+++ b/code/_onclick/other_mobs.dm
@@ -384,67 +384,63 @@
 					var/list/stealpos = list()
 					var/list/mobsbehind = list()
 					var/exp_to_gain = STAINT
-					to_chat(src, span_notice("I try to steal from [V]..."))
-					if(do_after(src, 5, target = V, progress = 0))
-						if(stealroll > targetperception)
-						//TODO add exp here
-							// RATWOOD MODULAR START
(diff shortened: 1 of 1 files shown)


DorfGuy1:
Fair enough. Still in the firm opinion that beggars have it too easy. And it's fairly one sided. unless they go out of their way to fuck up, or someone notices a silent dissappearing bag, it's totally one sided, no rp thing.  There's no town guard chasing beggars around because they thieved, no hand cutting. A fraction of a second of wind up, maybe, and the possiblity of a message. As it stands there is very little risk for high reward, and seeing beggars walking around in armor and with swords is cringe, cause if the guard was RPing properly...yada yada yada. Honestly pickpocketing didn't add any rp before, so we don't have to hurry to revert this. Beggars and rouges can do something else for a few rounds.

CalamaBanana:
You did see my response that the 5 in the code is ticks, not seconds as you assumed, right?

Useroth:
I think we'll leave it for now

PR #850, opened 2024-08-25
[RU-ROTWOOD ONLY] Trying to prevent players from entering a role by roundstart if the value of available seats per role was 0 in the managejobslots parameter (Ruzone testmerge)

ghost:
It's just a test, maybe. But it works!


zakanater19:
What does this do?

PR #853, opened 2024-08-26
Removes Axian lock from pirate class.

Coolgat3:
## About The Pull Request

This reverts Axian fanboyism

## Why It's Good For The Game

Because I don't see how in the fuck in a world filled with water where we live on a literal fucking island only anthro sharks can be pirates. The race lock should have never been a thing, let me enjoy this class without having to create a shark character. We haven't seen these classes at all thanks to the fact nobody wants to play Axian.

This racelock is just stupid, please remove it. Not only sharks can swim or operate a ship.

If this is about lore, then I am sorry, this lore is stupid, it makes absolutely 0 sense that only axians are comitting organised crime on the seas.


Coolgat3:
> I believe just removing the line of code for allowed races will just break the class, pretty sure it needs to be
> 
> `allowed_races = RACES_ALL_KINDS`

done

Quink-Fix:
Should remove it from the Smuggler class too.

PR #854, opened 2024-08-26
Removes race lock from Smuggler

Coolgat3:
## About The Pull Request

Sorry that it's in 2 PRs but I forgot that smuggler also was added and these are web edits I don't really know how to connect them, my laptop is too shit for doing this properly..

## Why It's Good For The Game

Same story as with pirate, not only axians can swim or operate ships, seriously, please.


CalamaBanana:
This race-lock was far more confusing to me than the pirate one in the first place.
Their special ability is running endlessly. Which does not sound like a fish thing to do in the first place.

Coolgat3:
> > This race-lock was far more confusing to me than the pirate one in the first place. Their special ability is running endlessly. Which does not sound like a fish thing to do in the first place.
> 
> I believe it is so you can swim upstream without being Stam drown by the river

no that has no correlation to running with good swimming skill you can walk upstream without running

PR #855, opened 2024-08-26
Changes estoc penfactor to 60

Leudoberct1:
This brings it up to just above the penetration factor as a spear thrust. The reasons for this are as follows:

At 40 penfactor, the estoc does not actually penetrate armour whatsoever. This is kind of lame for a weapon whose whole point is that it's meant to be able to penetrate armour. While this may be a bit better penetration than the spear, the spear is an iron weapon, whereas the estoc weapon requires 2 steel bars, and should thus reflect the material investment.  This is still not anywhere near the insane penetration levels of a flail- you are not going to be critting anyone through their armour with this, but you are at least able to soften them up before you do, which is quite needed on a weapon whose whole point is to be used against armour.

In addition, a spear allows you to have this penetration factor whilst also holding a shield, and also having increased range against your enemy. WIth the estoc, you have a higher material cost, need to twohand the weapon, and have to be face to face with your enemy.

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/swords.dm
+++ b/code/game/objects/items/rogueweapons/melee/swords.dm
@@ -507,7 +507,7 @@
 
 /datum/intent/sword/thrust/estoc // Around 12 or so dmg gets through armor, making it weaker than spears, but more portable
 	clickcd = 10
-	penfactor = 50
+	penfactor = 60
 
 /obj/item/rogueweapon/sword/cutlass
 	name = "cutlass"


Leudoberct1:
I invite you to go on your local repo and test it with pen factors of both 50 and 60. 30 damage is not getting through, whatsoever. The full danage does not penetrate. You are not going to cause any crits whatsoever.

Leudoberct1:
That is unfortunately not how armour penetration works. It doesn't check for what armour it is, it's just a flat thing. What makes a difference is the armour health- until the armour is broken, you're still not doing any serious damage with the estoc even at 60ap. Higher health armours like full plate will take a lot longer to break, and thus take a lot longer before you're doing anything like critting someone. 60ap just lets you soften them up a bit- it is by no means a 'beat armour for free' thing like a flail, which takes LESS steel.

PR #725, opened 2024-08-06
Keep map tweaks

Heheniy:
## About The Pull Request

Overall just adds more light to keep yard, 3rd mage tower floor, 2nd keep floor near king's room, basemant.

image
image
image
image
image

Also adds more doors leading inside first floor walls, alongside with stone windows for archers, making first floor walls not complete useless. Makes east wall first floor 2 tiles wide instead of 1. Also replaces all wooden wall windows with stone ones (cant see windows on screenshots because SDMM shows them just as walls, but ~every 8th wall is actually a window).

image
image
image
image

Changes to small cell? room? in the basemant.

image

Breaking into vault is now a little harder (not just 1 wall)

image

Steward got more spare keys

image

## Why It's Good For The Game

image

PR #858, opened 2024-08-26
Keep map tweaks

Heheniy:
## About The Pull Request

Re-made #725 since map got broken
Overall everything same but small changes here and there


CalamaBanana:
While you are touching keep stuff, can you please swap the barrel in the barrel niche at the kitchen counter to a full one? It is just painful to have to spend clicking the well for ten minutes each and every round as servant.

Red-Honk:
> While you are touching keep stuff, can you please swap the barrel in the barrel niche at the kitchen counter to a full one? It is just painful clicking the well for ten minutes each and every round as servant.

Just go outside of the keep, jump in the moat, go east or west, there should be a river, and that river is a near insta refill on a bucket or pot, so, just grab someone and tell them to grab a pot or bucket while your character gets one or the other and speed fill it up.

PR #871, opened 2024-08-28
Inquisitor supermax prison + Confessor

Menpbel:
## About The Pull Request

what I did:
- confessor added, has 3st expert maces and a cudgel and can wear medium armour(he's the inquisitor's subordinate) he also gets a cool unused helmet in his bedroom with shit vision.

- buffed inquisitor's medical abilities so he can do more surgery on people, (I mean we can just force players to cut themselves and to stich themselves up to level surgery before operating someone (that's retarded)

- replaced the shitty two rooms with a inquisitorial dungeon, I believe it should be like a place that has always been from the Inquisiton because Psydon was once the leading religion and the Inquisition is Psydonite so it makes sense that the dungeon is an old facility that has always been used by the Inquisiton way before the ten.
z3:
image
z2:
image
z1:
image

- distances the inquisitor from the church, he should work for kingsfield and for what's left of a Psydonian church (I we're thinking of making something like the bandit's matthios to change confessions for armour and stuff but that should be up to debate because we won't like a inquisitor rushing kidnap everyone and making them confess)

- added music at the inquisiton's dungeon.

## Why It's Good For The Game

Right now the what the Inquisitor does is wander the streets ask people and wait for the priest/priestess bring them a heretic or something to do. Now, with the help of a buddy the Inquisitor can now get violent and forcibly arrest people he suspects without having to hire a mercenary that might just turn on him and ditch him (happend to me once)

The Dungeon is a pretty good addition , I'm open for criticism though.

P.D: There's a lot of commits because I had no idea how to use github, if there's a inconsistent indentation i'm sorry.

### DIFF
--- a/rockhill.dmm
+++ b/rockhill.dmm
@@ -745,7 +745,7 @@
 	locked = 1
 	},
 /turf/open/floor/rogue/blocks{
-	icon_state = "paving"
+	icon_state = "newstone2"
 	},
 /area/rogue/under/town/inquisition)
 "aOn" = (
@@ -1093,10 +1093,9 @@
 /turf/open/floor/rogue/church,
 /area/rogue/indoors/town/church)
 "bgd" = (
-/turf/open/floor/rogue/blocks{
-	icon_state = "bluestone"
-	},
-/area/rogue/under/town/inquisition)
+/obj/structure/flora/roguegrass/water,
+/turf/open/water/cleanshallow,
+/area/rogue/outdoors/beach)
 "bgI" = (
 /obj/item/roguebin/water,
 /turf/open/floor/rogue/blocks,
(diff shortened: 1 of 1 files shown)


Menpbel:
I missed a door for some reasson.

Heheniy:
this is ugly as fuck
image

Cyprex:
Inquisitors role is largely to mess with random people, maim then and quite possibly kill them. I think it's a crappy role, largely based around RRing people. I really don't think it should be expanded upon.

Menpbel:
ok, I remaded somethings because it was indeed ugly as shit and now there's a path into the beach
z2: 
image
z1: 
image
now there's a pit you can put heretics at, maybe the dungeoner should have something like it.

thebigtony:
What about a heretic prisoner role? They'd be inhumen worshipers that start branded as heretics. That way the inquisitor has someone on hand to mess with. And if the inquisitor does arrest someone their victim won't get as lonely and bored by themselves in the inquisitors dungeon.

Menpbel:
I don't know about a heretic prisoner role, there's already a heretic special and being a roundstart heretic prisoner would be shit. There's two people that can interact with you. 

Also I've got a question should we rename confessor to something else?

Menpbel:
i'm gonna rethink this

PR #876, opened 2024-08-30
Misc sprite bundle.

AberrantQuerist:
## About The Pull Request
A small batch of seven new sprites to hold over until I can get the other half of the crops done.
Included are resprites of the following: Mincemeat, eggtoast, peppersteak, spicedeggs, westernspice, easternspice, and a palette shift for cloth to be more in-line with the bundled appearance. 

## Why It's Good For The Game
For all the food items, they use placeholder /tg/ sprites. I referenced and/or copied parts of existing items that are used in the crafting of these to make the transition not quite so abrupt. It's a transformation of an object, not an entirely new one- at least as I want to see it.
Westernspice is modeled after rosemary and easternspice is modeled after ginger. I couldn't really tell what either of them were supposed to be, so these are my best guesses that fit within the universe. 
Cloth is very bright white generally. This might adjust the greyscale to be a bit darker than whatever colour it's supposed to be, but I think it looks better in the long run.
image


Cyprex:
btw cloth is greyscaled, there's a bandit subtype that makes use of this, /obj/item/natural/cloth/bandit
How does that look with this update?

aberrantQuesrist:
> btw cloth is greyscaled, there's a bandit subtype that makes use of this, /obj/item/natural/cloth/bandit How does that look with this update?

Screenshot 2024-08-30 122547
Sorry for the delay, I updated BYOND and it stopped compiling.

As you can see, it just looks a bit darker. Cloth still inherits the general colour from clothing it's ripped from, even side-by-side it doesn't look too far off. Freshly sewed cloth looks like it can actually be taken from and put back into the cloth bundle. Bandit cloth looks fine and vibrant still, without being a glowing beacon with how bright the previous cloth was.

aberrantQuesrist:
Despite the common request, I'm not touching fibres. Their hue can be a different person and/or PR's mess. I will, however, fix a weird thing with wells.

PR #878, opened 2024-08-30
Fills main keep and church water barrel and fixes old reagent/unit mixup in the code.

CalamaBanana:
Fixed: 
- Main water barrel of the keep now spawns fully filled
- Public water barrel of the church now spawns fully filled
- Barrels were only going to 100u at roundstart, not 300u. Old code had barrels filled to 300 reagents, which only corresponds to 100u. Now random barrels go from 0u to 300u instead, and full barrels spawn at 300u, not 100u
- Filled barrels spawn with a specific sprite, no longer random. Water with white band, ale with golden band.

Added:
- Wine barrel added to code while I was in the area. So far not used anywhere, but maybe mappers want it at some point.


aberrantQuesrist:
Maybe stick a wine barrel or two in the keep cellar?

CalamaBanana:
> Maybe stick a wine barrel or two in the keep cellar?

Thought about it, but did decide to only fix things with this PR. Adding a wine barrel does remove a tiny bit of merchant interaction for the keep.

CalamaBanana:
> Okay but isn't starting with 100 good for balance? These shouldn't last forever, you should need to refill them soon enough if several people use them.

The problem is the main barrel at the keep started completely empty, and the few others were randomized from 0 to 100. If the main barrel were just at 100, it would be empty after a minute as well, due to everyone in the keep starting thirsty. It is just boring having to spend the first 10 minutes of each round clicking well and barrel in solitude.

aberrantQuesrist:
Filling water is the least enjoyable part of maidwork. This is just pure QoL.

PR #879, opened 2024-08-30
Removes a boring trait that does absolutely nothing to half the roles

Coolgat3:
## About The Pull Request

Removes the no pouch trait.

## Why It's Good For The Game

This trait is equivalent to getting absolutely nothing, no negative, no positive, it's just so damn boring all you want to do when you get it is beggar suicide to get something actually fun even if more crippling. Every single guard role doesn't give a shit about this trait, there's I'm sure even more roles that don't get pouches, the only situation where this does anything to you is if you choose to play bogspawn role, and even then, you just cut 3 small logs and the guards will let you through anyway. Please remove it, it's boring as hell.

Oh no, my 7 zennies, whatever will I do.

### DIFF
--- a/code/datums/special_traits/traits/traits.dm
+++ b/code/datums/special_traits/traits/traits.dm
@@ -340,21 +340,6 @@
 	character.change_stat("speed", -2)
 	character.change_stat("intelligence", -4)	
 
-/datum/special_trait/nopouch
-	name = "No Pouch"
-	greet_text = span_boldwarning("I lost my pouch recently, I'm without a zenny..")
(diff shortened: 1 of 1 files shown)


zakanater19:
it also blacks out the neck slot and still shows you as having a pouch on your sprite

radarseas:
this trait is also completely nulled out as vampire lord/bandit and dose nothing to werewolves, hard agree, kill it.

Coolgat3:
Can we please merge this already i'm getting no pouch trait literally every 2nd round at this point.

PR #885, opened 2024-09-01
"Reworks" Grabs. Makes them properly use stats and wrestling skill

GuppyLaxx:
This PR changes the math on pretty much all grab related actions to properly use stats and skills.
Pinning is no longer a roll, and instead an instant but lesser stunning action, meant to hold someone down while others help you, or to keep someone VERY down if you are much stronger than them.
Grabbing can now be parried and actually uses stamina properly. The grabber uses more stamina than people resisting out of grabs.
No more wall stuns, no more throwing up Z-levels unless they're laying down.
A couple other changes overall!
Also buffs monk wrestling by 1.

This PR should hopefully make grab combat less of a nightmare. It was tested in a private server with some other players. This is the last iteration of my changes after testing. Always up to changing more and making the math different!

### DIFF
--- a/code/game/objects/items/rogueitems/ropechainbola.dm
+++ b/code/game/objects/items/rogueitems/ropechainbola.dm
@@ -152,8 +152,8 @@
 	w_class = WEIGHT_CLASS_SMALL
 	throw_speed = 1
 	throw_range = 3
-	breakouttime = 1 MINUTES
-	slipouttime = 5 MINUTES
+	breakouttime = 10 SECONDS
+	slipouttime = 2 MINUTES
 	cuffsound = 'sound/blank.ogg'
 	possible_item_intents = list(/datum/intent/tie, /datum/intent/whips)
 	firefuel = null
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -323,7 +323,7 @@
 		return FALSE
 
 /obj/structure
-	var/breakoutextra = 30 SECONDS
+	var/breakoutextra = 10 SECONDS
 
 /mob/living/carbon/resist_buckle()
 	if(restrained())
@@ -385,7 +385,7 @@
 
 /mob/living/carbon/proc/cuff_resist(obj/item/I, breakouttime = 600, cuff_break = 0)
 	if(I.item_flags & BEING_REMOVED)
-		to_chat(src, span_warning("You're already attempting to remove [I]!"))
+		to_chat(src, span_warning("I'm already trying to get out of \the [I]\s!"))
 		return
 	I.item_flags |= BEING_REMOVED
 	breakouttime = I.slipouttime
@@ -395,18 +395,18 @@
 	if(STASTR > 15 || (mind && mind.has_antag_datum(/datum/antagonist/zombie)) )
 		cuff_break = INSTANT_CUFFBREAK
 	if(!cuff_break)
-		to_chat(src, span_notice("I attempt to remove [I]..."))
-		if(do_after(src, breakouttime, 0, target = src))
+		to_chat(src, span_notice("I try to get out of \the [I]\s..."))
+		if(move_after(src, breakouttime, 0, target = src))
 			clear_cuffs(I, cuff_break)
 		else
-			to_chat(src, span_danger("I fail to remove [I]!"))
+			to_chat(src, span_danger("I fail to get out of \the [I]\s!"))
 
 	else if(cuff_break == FAST_CUFFBREAK)
-		to_chat(src, span_notice("I attempt to break [I]..."))
-		if(do_after(src, breakouttime, 0, target = src))
+		to_chat(src, span_notice("I attempt to break \the [I]\s..."))
+		if(move_after(src, breakouttime, 0, target = src))
 			clear_cuffs(I, cuff_break)
 		else
-			to_chat(src, span_danger("I fail to break [I]!"))
+			to_chat(src, span_danger("I fail to break \the [I]\s!"))
 
 	else if(cuff_break == INSTANT_CUFFBREAK)
 		clear_cuffs(I, cuff_break)
--- a/code/modules/mob/living/grabbing.dm
+++ b/code/modules/mob/living/grabbing.dm
@@ -139,7 +139,7 @@
 	if(M.surrendering)
 		combat_modifier = 2
 
-	if(M.handcuffed)
+	if(M.restrained())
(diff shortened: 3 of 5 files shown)


Jackofashepard:
I think the numbers need a little adjustment. As a journeyman i tried to grab someone with no wrestling skill and had about a 50% chance to tighten my grip. Maybe raise it a bit?

Red-Honk:
Finally, something that is actually balanced, and not bias to make wresting a nightmare to even execute like the last one that over punishes the grabber, now with pins being briefer than extremely long like last time, a group will no doubt be needed to chain someone that isn't all that new to not just grab the leg and toss away to cancel the chaining, not to mention being unable to just pin and let them stay there for a good while, but making it instant does deal with my issue of costing stamina.

My curiosity is what will now happen if someone is tossed into a wall?; Do they get damaged and stay up or do they fall down as before but without the stun like when getting kicked against it? (And oh boy, did I not enjoy being tossed for tiles away like a ragdoll because of a half-orc)

Being unable to throw folks up a Z-level if they aren't prone?; That's a rather clever one, allowing most of the time only consensual parties to do it, as well as not allowing those that toss folk up to start instantly on an even footing with those on a higher level.

Takoros:
Perfect

HissyNays:
Here we go again...
First of all, calling it rebalance is dishonest. It's a nerf.

>grabbing now uses more stamina than the person resisting, is now stopped by parrying, no more wallstuns.
So essentially, you should never grab now.

This is same as last pr but with potentially even bigger nerfs.
I haven't tried this yet but I can already see it's gonna be terrible.
Why can't you just learn to counter it and use it your self?

HissyNays:
Yeah I tried using grappling in combat, it's effectively useless now. Please revert this PR.
There are some good changes like being unable to toss people up to level above while they stand. But overall. It's a soft removal of wrestling as combat skill.

DorfGuy1:
My main frustration with wrestling and grabbing is that it seems like a: Its ALL strength based, the skill not playing much of a role in holding or breaking the grabs and then there's horrendous flat percentage chances to break or grab. I'm pretty majorly skill issued on grabs though, does resisting like, have a cooldown shared with clicking or something?

Jackofashepard:
Another feedback. The tackling is a bit scuffed now, there seems to be a "stun"/retry timer on it now if you fail which makes it a bit too punishing since it takes quite a long time. Perhaps lower it a bit?

HissyNays:
>Prior to this PR, wrestling and grabs were entirely Stregnth reliant, with only RESISTING using wrestling skill.

Untrue. In the past I was able to outwrestle people who were stronger than me in same roles. Namely knight and watchman.

HissyNays:
Playing some more I had to find out hard way grabs are still not entirely useless but most useless.
There's a desperate need to roll back wallstun removal.
Hard grabs feel unreasonably hard to do now. Given hard it can be to get first grab, the hard grab shouldn't be this difficult to get.

pilotmans:
> Finally, something that is actually balanced, and not bias to make wresting a nightmare to even execute like the last one that over punishes the grabber, now with pins being briefer than extremely long like last time, a group will no doubt be needed to chain someone that isn't all that new to not just grab the leg and toss away to cancel the chaining, not to mention being unable to just pin and let them stay there for a good while, but making it instant does deal with my issue of costing stamina.
> 
> My curiosity is what will now happen if someone is tossed into a wall?; Do they get damaged and stay up or do they fall down as before but without the stun like when getting kicked against it? (And oh boy, did I not enjoy being tossed for tiles away like a ragdoll because of a half-orc)
> 
> Being unable to throw folks up a Z-level if they aren't prone?; That's a rather clever one, allowing most of the time only consensual parties to do it, as well as not allowing those that toss folk up to start instantly on an even footing with those on a higher level.

from what i've seen you can't throw people who are alive up onto walls even if they consent.

ToothyDeer:
This PR has broken so many things, you can not grapple people who are first off PARALYZED. They will still wrestle out of it, or armless. With mobs, when your blade becomes stuck, its now a low chance that you can even pull it out, so if you lose your sword into a minotaur, good luck getting it out; its just not very fun. People laying on the ground who WANT to be grabbed can no longer be grabbed half the time, and by the time you POSSIBLY grab them, you stam yourself out. 

This PR should be disabled and reverted until further improvements made on it, like making half of these's take effect in combat mode only.

pilotmans:
> This PR has broken so many things, you can not grapple people who are first off PARALYZED. They will still wrestle out of it, or armless. With mobs, when your blade becomes stuck, its now a low chance that you can even pull it out, so if you lose your sword into a minotaur, good luck getting it out; its just not very fun. People laying on the ground who WANT to be grabbed can no longer be grabbed half the time, and by the time you POSSIBLY grab them, you stam yourself out. 
> 
> 
> 
> 
> 
> 
> 
> This PR should be disabled and reverted until further improvements made on it, like making half of these's take effect in combat mode only.
> 
> 
> 
> Edit 1: Wall-throwing / stunning should also be re implemented, I've seen genuinely amazing (non-cheese) wall throws clutch up situations. I've never seen it abused / bugged before, but I just don't see it too often in general.

Wall Throwing being removed is so annoying and I'm not sure if it's even possible to throw people up onto walls most of the time now cause of this pr, it also fucks up trying to save people cause grabbing is longer dependent on combat mode meaning people who are dieing on the ground will actively resist a fireman carry which is really dumb.

Red-Honk:
Just don't try to tighten the grip on them if you want to fireman carry them, get a firm grip and put them on the shoulder, and guess what happens when that is done, that's right, if the str is high enough you can even run with them on normal speed.

HissyNays:
So far, I found ways around this PR to still be able to wrestle people. It doesn't however change the fact how annoying and frustrating many changes are.
Like there's no point in trying to hard grab some one, you'll just repeatedly stun your self and run out of stamina, it's a waste of time and stamina both. Only regular grabs seem to be viable againt some one who isn't down or really weak.
Can we just go back already? This has been tested plenty enough.

pilotmans:
> So far, I found ways around this PR to still be able to wrestle people. It doesn't however change the fact how annoying and frustrating many changes are.
> 
> Like there's no point in trying to hard grab some one, you'll just repeatedly stun your self and run out of stamina, it's a waste of time and stamina both. Only regular grabs seem to be viable againt some one who isn't down or really weak.
> 
> Can we just go back already? This has been tested plenty enough.

Agreed this just made wrestling pointless

katsukai:
These changes are perfect, don't listen to the non-combat whiners.

PR #891, opened 2024-09-01
[RU ONLY] Map change

ghost:
Totaly new view on old map


Coolgat3:
what? Are you on the right github?

LeShprotina:
If you don't know ruzone are actually working with your server. But we have other testmerges and type of rp.

PR #892, opened 2024-09-01
Improves Merchant's shop QoL and security

CalamaBanana:
## Why It's Good For The Game

Merchant is another one of those massively underplayed roles, mostly because it was such a designated victim role. Each round is a race against time to find a carpenter before the first adventurers arrive at town to break through the paper thin windows and loot the shop. This PR does both improve the security of the shop while also making some QoL adjustments for both customer and merchant. 

## About The Pull Request

Changes made:

**Downstairs:**
image

1. A small niche for customers, so they do not clog up the streets. Peddler can be loaded with full vision of the area.
2. Changed the carpet to one that is not a a Gizan IRL reference
3. Paper, feather and hermes closer together
4. Removed the redundant second staircase, and reclaimed some breathing room there

**Upstairs:**
image

5. A knoife. Not much, but at least something for the shophands to grab when they get roundstart rushed by beggars
6. Easy access windows are gone
7. Easy access pipe to window is gone
8. Easy access area changed to cosy balcony. Can be climbed just as before, yet secured with a door instead of a paper window

**Roof:**
No changes

(Things not reflected in the screenshots yet: Merchant/Shophand does actually still have spawn points on the ground floor. Cosy balcony does not use roof shingles as floor tiles)

### DIFF
--- a/_maps/map_files/rockhill/rockhill.dmm
+++ b/_maps/map_files/rockhill/rockhill.dmm
@@ -184,18 +184,19 @@
 /obj/structure/chair/bench/ultimacouch,
 /turf/open/floor/rogue/wood,
 /area/rogue/indoors/town/manor)
-"aic" = (
-/obj/structure/mineral_door/bars{
-	lockid = "graveyard"
-	},
-/turf/open/floor/rogue/cobblerock,
-/area/rogue/outdoors/town)
+"ahL" = (
+/turf/open/floor/rogue/grass,
+/area/rogue/indoors/shelter/mountains)
 "ait" = (
 /obj/effect/decal/cleanable/blood/drip,
 /turf/open/floor/rogue/blocks{
 	icon_state = "paving"
 	},
 /area/rogue/under/town/basement)
+"aix" = (
+/obj/structure/flora/roguegrass/bush,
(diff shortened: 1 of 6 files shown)


NikolayByondPlayer:
I would also decrease the whole shop and nearby clerk's areas. 
Right now they make one giant unused house that stretches for the whole town.

CalamaBanana:
> I would also decrease the whole shop and nearby clerk's areas. Right now they make one giant unused house that stretches for the whole town.

I thought about shaving a little from the bottom of it to add a few more run down homes, but decided against it as it would be a bit out of scope. Larger map changes would need a lot more adjustment to the entire area around. I did not trust myself to go that far just yet, so I kept within the confines of what we already have.

NikolayByondPlayer:
> > I would also decrease the whole shop and nearby clerk's areas. Right now they make one giant unused house that stretches for the whole town.
> 
> I thought about shaving a little from the bottom of it to add a few more run down homes, but decided against it as it would be a bit out of scope. Larger map changes would need a lot more adjustment to the entire area around. I did not trust myself to go that far just yet, so I kept within the confines of what we already have.

Yeah, agreed

PR #893, opened 2024-09-01
Inquisitor New Job Description 

Gyrandola:
## About The Pull Request

Changes the Inquisitor's job description to something that makes sense.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/church/puritan.dm
+++ b/code/modules/jobs/job_types/roguetown/church/puritan.dm
@@ -11,7 +11,7 @@
 		/datum/patron/old_god,
 		ALL_DIVINE_PATRONS,
 	) //gets set to old god.
-	tutorial = "“Still clinging to the fading cult of Psydon, The Inquisition has sent you to Rotwood on a quest to root out the blasphemous and the cursed of these lands. Guided by religious fervor and extreme paranoia, you’re determined to extract confessions from the sinful, hoping that your uneasy alliance with the Church of the Ten holds.”"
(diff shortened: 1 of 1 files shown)


Maromar15910:
I like it over all, but it feels like it pressures the inquisitor into a personality type. Less 'inquisitors are typically this' and more 'you ARE this', which can limit rp or verbally slap people around for not being the buzz lightyear meme.

SixplyDev:
Frankly, I don't know how to feel about this change. I do think that making the inquisitor's description clearer or more in-line with the current lore that we have is needed, but I also think that this would only make treatment of inquisitors worse than it already is. This implies that it's a religious nutjob that's probably a hermit or some crap, not an educated and powerful man that's there to carry out a noble duty. 

I would much rather see this transformed into implying that the inquisitor is a learned man that's from a much more powerful but mysterious authority. I would also like to see it more imply/say why the inquisitor was hired even though they believe that the church has been taken over by heretical pagans. Overall, I do not think that this change would benefit the inquisitor and I believe that it would in fact harm the inquisitor and how people treat the role.

SixplyDev:
> > **[SixplyDev](/SixplyDev) ** commented [Sep 3, 2024](#issuecomment-2325548551) •
> 
> Pretty much all you wrote is your headcanon

Psydon patron description literally says that you believe the church was taken over by heretical pagans 💀 but I would genuinely like to see something more in line with a man with authority rather than a hermit from a fading cult

Cyprex:
Very mixed feelings on this as well, @GuppyLaxx what do you think?

Keyo360:
As a repeating Inquisitor player, Inquisitors are already looked down upon incredibly in game even if the Kingsfield Queen has sent them for whatever purposes she might have. This change will only amplify that and completely destroy the chance of an Inquisitor to do much any round due to the lack of authority and standing.

PR #894, opened 2024-09-02
Expands Examine

GuppyLaxx:
Examine now takes into account where exactly you're looking at on people, and also if you're looking at something on you, your hands, or inside your bags.

Removes automatically facing whatever you examine aswell.

dreamseeker_bktkHX5ZLc
dreamseeker_VucVFH8F79


Coolgat3:
omg i can be a fucking pervert and stare at people's ass hell yeah

hugmeorelse:
What if you make it a perception check? :^) To see where they're looking at directly?

Hmeister-real:
DO NOT REMOVE FACING WHAT YOU EXAMINE, PLEASE. I use it all the time to look in a certain direction, or to turn around and look at someone in one motion

radarseas:
Leave turning as you examine, its a suble but **EXTREMELY** _useful que for you've caught someone's eye_ and they're trying to check if _something is off with you_. It can absolutely change the game despite how subtle it is. In fact it can be enough to give you that early que to move faster/hide away and not get ganged on.

Ragnuv:
> Leave turning as you examine, its a suble but **EXTREMELY** _useful que for you've caught someone's eye_ and they're trying to check if _something is off with you_. It can absolutely change the game despite how subtle it is. In fact it can be enough to give you that early que to move faster/hide away and not get ganged on.

THIS, definitively this, something that would also be interesting, and also add a use for the perception stat which is borderline useless as of now, would be determining rather you turn or not on examining something depending on your perception stat (above 10, you don't turn, and below 10 you turn)

HissyNays:
Ehhh not a fan of examine not making you turn towards thing you examine. You can now even examine some one behind you without turning. It ends up feeling like game is glitched and thus makes the game less immersive.

EzekyalLegit:
Guys you can turn to face a direction by just holding CTRL and hitting any WASD direction

Cyprex:
Is this good to merge?

PR #897, opened 2024-09-02
Randomly generates caverns in the Z1 bog

zakanater19:
## About The Pull Request

So instead of mapping it i figured it would be easier to just have it randomly generate inside the new area, you can take alook at the files changed to see what items currently spawn in there, another thing to note is that inside the area it replaces the rockwalls with /turf/closed/mineral/random/rogue/high so it's very good for mining also

and as you can see in the video is does not put too much strain on the server during startup when it is doing all this

i also mapped in some areas where you can travel down into the caverns

## Why It's Good For The Game

I think it's cool xD

https://github.com/user-attachments/assets/a5796c6d-e94b-4081-bf43-a89a88342360

Screenshot 2024-09-02 104357

Screenshot 2024-09-02 104410

### DIFF
--- a/code/game/area/roguetownareas.dm
+++ b/code/game/area/roguetownareas.dm
@@ -788,6 +788,7 @@ GLOBAL_LIST_INIT(roguetown_areas_typecache, typecacheof(/area/rogue/indoors/town
     var/list/spawnables = list(
     // Not rare objects
     list(/obj/structure/spider/stickyweb, 0.45),
+    list(/obj/structure/flora/rogueshroom, 0.30),
     list(/obj/structure/glowshroom, 0.20),
     list(/obj/effect/spawner/lootdrop/roguetown/dungeon/misc, 0.085),
     


Coolgat3:
I dunnnnnno, is this supposed to replace the undersea completely? It kinda feels like the worst part of the underworld (getting lost in a cringe maze) is replacing the undersea

zakanater19:
> I dunnnnnno, is this supposed to replace the undersea completely? It kinda feels like the worst part of the underworld (getting lost in a cringe maze) is replacing the undersea

no it's not replacing anything but empty rock currently, the two pictures above show that so the lowest is Z2, i included it to show the areas where i placed some openings into Z1

The top image is showing Z1 with the areas adjusted to spawn these caverns in, you can resize the area to whatever you want or place this area on any Z level to get the same random genning

zakanater19:
Also this isn't related to being dead or the underworld

DorfGuy1:
i like this. More place for people or make their littleforts.

zakanater19:
Now spawns mushroom trees like DF does, also you can now cut the stumps too before they were broken and you couldn't attack the shroom stumps, now you can

CalamaBanana:
Wandered through it yesterday, some thoughts:

- You could have the general size of the caverns be more dynamic, to have some underground biomes that are really just tight hallways.  
- Someone made a bunch of neat mushroom sprites in #dev-spriting. Could be a good opportunity to add them in as floor deco down there till they are used for the alchemy system. Some could give a differently colored glow for even more ambiente. Perhaps even use them like grass to trigger ambushes?
- A specific flora that grows only on tiles with no roof above, so it is easier to spot exits
- A few loose stones on the ground for silly people that accidentally drop their torch (me)
- (Delete the Z2 tunnels and replace it with this already, I love this so much more.)

Useroth:
Add the rest in another PR

PR #910, opened 2024-09-03
Allows you to ferment potatoes into vodka

Mantweerp:
## About The Pull Request

Allows you to ferment potatoes into vodka (and makes it's color be clear, instead of the comical blue which it was before)

## Why It's Good For The Game

Not many things can be done with potaters, and I found it weird that I could not ferment them when I tried to do so yesterday (it also makes historical sense for vodka to exist)

### DIFF
--- a/code/modules/reagents/chemistry/reagents/alcohol_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/alcohol_reagents.dm
@@ -225,9 +225,9 @@ All effects don't start immediately, but rather get worse over time; the rate is
 /datum/reagent/consumable/ethanol/vodka
 	name = "Vodka"
 	description = "Number one drink AND fueling choice for Russians worldwide."
-	color = "#0064C8" // rgb: 0, 100, 200
+	color = "#AAAAAA77" // rgb: 170, 170, 170, 77 (alpha) (clear like water)
 	boozepwr = 65
-	taste_description = "grain alcohol"
+	taste_description = "vodka"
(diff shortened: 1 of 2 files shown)


LeShprotina:
The Belarusian people are proud of you.

code/modules/reagents/chemistry/reagents/alcohol_reagents.dm
```diff
@@ -225,9 +225,9 @@ All effects don't start immediately, but rather get worse over time; the rate is
 /datum/reagent/consumable/ethanol/vodka
 	name = "Vodka"
 	description = "Number one drink AND fueling choice for Russians worldwide."
-	color = "#0064C8" // rgb: 0, 100, 200
+	color = "#AAAAAA77" // rgb: 170, 170, 170, 77 (alpha) (clear like water)
 	boozepwr = 65
-	taste_description = "grain alcohol"
+	taste_description = "vodka"
```
Cyprex, line 230: Consider adding a hydration_factor, sake, ale and beer use 10. What do you think would be appropriate for vodka? Default is 5.

Genevesh: Ten as all others. Sake and vodka are pretty similar, and if we're being awfully technical, I'd say most drinks make you a little thirsty, so if anything, their hydration should be lowered to default.

DorfGuy1:
Ale is set to be the same power level as Vodka. Jack it up a little?

PR #914, opened 2024-09-04
Mercenary house, open for business

Rayforx:
## About The Pull Request
This adds a more open space for the mercenary building, so it can be completely opened up for other people to get in and hire them
It adds the sellsword a key (the peasant of merc) so they can organize or use the area to get hired.
Adds two more rooms in the mercenary place, and a mini-dungeon (for your bounty)

<THIS PR IS NOT TESTED YET, I WILL TEST IT IN A FEW DAYS>

## Why It's Good For The Game
More places to roleplay. It also makes the awesome large building that the mercenary has, most likely more used. Nobody knew what was going on in there, and most people assumed that it was part of the watch.
It also gives a few lockable deadbolted rooms

### DIFF
--- a/code/modules/jobs/job_types/roguetown/garrison/veteran.dm
+++ b/code/modules/jobs/job_types/roguetown/garrison/veteran.dm
@@ -50,7 +50,7 @@
 	beltl = /obj/item/rogueweapon/sword/sabre
 	beltr = /obj/item/keyring/veteran
 	backr = /obj/item/storage/backpack/rogue/satchel/black
-	backpack_contents = list(/obj/item/rogueweapon/huntingknife/idagger/steel/special = 1, /obj/item/signal_horn = 1, /obj/item/roguekey/mercenary = 1)
+	backpack_contents = list(/obj/item/rogueweapon/huntingknife/idagger/steel/special = 1, /obj/item/signal_horn = 1)
 	if(H.mind)
 		H.mind.adjust_skillrank(/datum/skill/combat/swords, 6, TRUE)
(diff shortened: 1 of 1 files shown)


john-pawson:
Could you maybe post some screenshots of the changed version of the map? It usually helps on the mapping PRs

Rayforx:
> Could you maybe post some screenshots of the changed version of the map? It usually helps on the mapping PRs

When I test it in a few days, else you can go to >checks >map difference

RadiantFlash7:
> No to the veteran getting a key. The veteran serves the keep primarily and doesn't act like an actual mercenary, never seen them actually ask for money from anyone. They currently don't have a key to the guild hall and i see no reason why they should get one aside from thrm being labelled as a merc in the code which seems out of place to begin with.
> 
> edit: in their description it says they traded adventure for stable pay, meaning they may have once been a mercenary and gave that up for serving the crown, thats my interpretation anyway and seems to be what most players do.

In truth. I'm also somewhat skeptical of the veteran getting a key, as one of the veteran mains. But at the same time, VETERANS ARE STILL IN THE MERCENARY CATEGORY 😿 
It makes it a pain in the ass to get paid in tound, and if the keep refuses- it would be nice to be able to fall back into the merc guild.

radarseas:
considering the merc guild training rooftop never sees use, it would be nice for vet to get a mercenary key, riders and grezels don't really want to use it hardly ever at the moment because the point of their jobs is to be hired, not to train and nobody else with the stats can use it to train people.

PR #915, opened 2024-09-04
Bog ranger choice

Rayforx:
## About The Pull Request
It makes the bog ranger a choice, instead of a randomly selected
It still has the same limit of only being >3 players
The first to choose the 3 rangers gets the role
Change the name of the bog crossbowman to ranger, since it fits and females can get it

+ Removed females being forced into certain hairstyles, a bit weird to have your hair changed as a vulp
+ Added a cloak to ranger for STYLE while they go alone in the wilderness
+ Removed a bug where the crossbowman would get LEGENDARY SKILLS since they got the skill points from both the bog guard and the crossbowman skills

## Why It's Good For The Game
Not having to play a ranger, if you do not want to for the shift
Having the choice to be something, so your character can be known as a skilled bowman

### DIFF
--- a/code/modules/jobs/job_types/roguetown/garrison/bogguard.dm
+++ b/code/modules/jobs/job_types/roguetown/garrison/bogguard.dm
@@ -50,7 +50,7 @@
 	name = "Bog Ranger"
 
 /datum/advclass/bogguard
-	name = "bog guard"
+	name = "Bog Guard"
 	allowed_sexes = list(MALE, FEMALE)
 	allowed_races = RACES_VERY_SHUNNED_UP
 	outfit = /datum/outfit/job/roguetown/bogguardsman
@@ -59,7 +59,7 @@
 	category_tags = list(CTAG_BOG)
 
 /datum/advclass/bogranger
-	name = "bog ranger"
+	name = "Bog Ranger"
 	allowed_sexes = list(MALE, FEMALE)
 	allowed_races = RACES_VERY_SHUNNED_UP
 	outfit = /datum/outfit/job/roguetown/ranger


RadiantFlash7:
Oh, Yes please. Theres nothing worse then joining a role and having to HOPE for the 35% chance of getting what you want, if you want your character to be an archer.

wow50000:
could you do the same with town guards too ?

Rayforx:
> could you do the same with town guards too ?

It would have to be in another PR, but yes

code/modules/jobs/job_types/roguetown/garrison/bogguard.dm
```diff
@@ -65,22 +80,24 @@
...
+
+/datum/outfit/job/roguetown/ranger/pre_equip(mob/living/carbon/human/H)
+	. = ..()
+	backr = /obj/item/storage/backpack/rogue/satchel
+	backl = /obj/item/gun/ballistic/revolver/grenadelauncher/bow
+	beltr = /obj/item/quiver/arrows //replaces sword
+	wrists = /obj/item/clothing/wrists/roguetown/bracers/leather
+	backpack_contents = list(/obj/item/rogueweapon/huntingknife/idagger/steel = 1, /obj/item/clothing/cloak/raincloak/brown = 1 )
+	ADD_TRAIT(H, TRAIT_MEDIUMARMOR, TRAIT_GENERIC)
+
```
RadiantFlash7, line 92: Missing 
if(H.min)
assign_skills(h)

RadiantFlash7:
Currently test merged on the server, ranger is missing the two lines that call the skill proc. Currently testing it, and no skills at all.

Rayforx:
The bug was fixed

HissyNays:
Yes please, more of this. Town guard needs it as stated above it. Good work.

CalamaBanana:
Please do capitalize the name of the subclasses, noticed that in the steward's console. Looked a bit ugly with them being the only ones in lower case.

(Also, just personal preference, maybe not "Ranger"? The word Ranger invites people to think they should be ranging in the bog instead of guarding the gate. But I don't have any better idea myself.)

Rayforx:
> Please do capitalize the name of the subclasses, noticed that in the steward's console. Looked a bit ugly with them being the only ones in lower case.
> 
> (Also, just personal preference, maybe not "Ranger"? The word Ranger invites people to think they should be ranging in the bog instead of guarding the gate. But I don't have any better idea myself.)

I will see about capitalizing them

But in terms of name, then I feel that it fits well with calling them ranger since the bog master still has to scout the bog/ check for tunnels/ and perform other scout duties, which makes sense to have a *ranger* do it

Rayforx:
image
Picture of changes

CalamaBanana:
> Picture of changes

Thanks! Looks much cleaner.

PR #916, opened 2024-09-04
Buffs noble pilgrims, removes sexism

wow50000:
## About The Pull Request
Removes sexism from noble code, now both genders get the same skills and stat modifiers, buffed their combat skills to 2(high civilian level) with the exception of polearms IF OLD, which is 3. This is because old man with a staff is cool as fuck. Also they shouldn't be able to hold any of the heavier polearms with old debuffs and -1 str stacked on top of that unless they game it like horc or something. And lastly made it all races because they are a foreign noble. Slopfix for until someone makes a drow noble class or a horc noble class.

## Why It's Good For The Game
Noble was kinda a dogshit class, still should be after these changes with no useful skills and no unique stats (just checked and they get better bows than warriors 50% of the time... oops. Still shouldn't be an issue because warrior is melee coded anyways) other than being a lootbag and having noble blood. Still vastly inferior to other pilgrim classes in terms of combat so shouldn't be an issue.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/pilgrim/noble.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/pilgrim/noble.dm
@@ -4,7 +4,7 @@
 	you are encouraged to make your own story. But with wealth, come the poor, ready to pilfer you \
 	of your hard earned (inherited) Mammons, so tread lightly unless you want to meet a grizzly end."
 	allowed_sexes = list(MALE, FEMALE)
-	allowed_races = RACES_ALL_KINDS
+	allowed_races = RACES_VERY_SHUNNED_UP
 	outfit = /datum/outfit/job/roguetown/adventurer/noble
 	traits_applied = list(TRAIT_NOBLE)
 	category_tags = list(CTAG_PILGRIM)


LeShprotina:
Finally, good noble.
But what about items and clothing?

wow50000:
> Finally, good noble. But what about items and clothing?

lol
lmao even

CalamaBanana:
Create equality by giving both the domestic skills. 

While you are touching it, do slip in silken stockings for the female one, Black or white ones would me my choice.

SixplyDev:
Not a fan of allowing shunned races to be nobles but other than that,  looks good

radarseas:
not too much a fan of shunned races being nobles. _But, I suppose maybe, as long as people don't game it for horc strength?_

But otherwise needed changes, female nobles weren't able to do anything against tackles from anyone but mages and maybe rogues sometimes. Let alone elf female nobles whom basically lost any fight against you the second you got a grab on them. They still suck but -2 strength on top of racial modifiers is too overkill.

Any chance of giving them value apprasial? _pretty controversal...

But it gives them something in the long run_ to pull their weight with, especally if it comes to trading with the merchant or whatever. It also makes the class more unique than the rest of the pilgrim roles since noble-blooded and fancy gear is all it has going for it at the moment, _not counting the miserable stats and high chance of dying off the bat._

Cyprex:
Sure, whatever.

PR #917, opened 2024-09-04
Re stat amazon to be Hellion

wow50000:
## About The Pull Request
Ah shit here we go again, made amazon closer to their male counterpart while also attempting to give them their own identity of skirmishing and polearms essentially turning them into a faster, tougher more retarded grenzelhoft merc with less skills and training. Also made them have chainkini only, because it's funny. I also forgot to account for gender stats such male getting +1 con -1 int and females getting -1 str and +1 spd so this technically makes them a 13 speed class and 11 str class, probably won't be an issue lmao. Also removed the 5 polearms bullshit and left it at 3

## Why It's Good For The Game
https://www.youtube.com/watch?v=5XyZsiedLXs
image


helper-2:
Only *good* thing you did here is removing spawn option chance for them, everything else feels like you made another warrior

wow50000:
> Only _good_ thing you did here is removing spawn option chance for them, everything else feels like you made another warrior

they have the same stat point balance as barbarian with 8 total points into combat, and aren't able to wear medium or heavy armour and have critical resistance. I could make them tougher and stronger than warriors but I honestly don't know where to take points off other than tanking int further which would lead to horc amazons being incredibly busted.

CalamaBanana:
Amazons are known as legendary archers most of all. Even if not spawning with a bow, let them keep high archery.

PR #919, opened 2024-09-04
Allows repairing damaged and broken doors

CalamaBanana:
image

- If a door is damaged, it can be healed for half its health by using ingredient one. Repairing a damaged (not broken) door restores half health, so a heavily damaged door might need to be repaired twice.
- If a door is broken, it can be repaired by first using ingredient one, then ingredient two.

- Locks are preserved this way. (Be smart, don't be on the wrong side of a locked door when you repair it)
- You need at least 1 level in the corresponding skill to start repairs. 
- It takes 30 seconds divided by skill level for repairing each stage. Means a novice carpenter will stand there for a minute in total to repair a door. This was done to intentionally prevent combat repairs, and to make higher skill levels more relevant. A normal carpenter will barely take 15 seconds or less to fix it.

The costs to repair:
- Wooden door / deadbolt door / tavern door: 2 small logs, carpentry
- Reinforced door (With the sight slit): 1 small log and 1 iron ingot, carpentry
- Stone door: 2 stone, masonry
- Iron Door (Graveyard gate): 2 iron, blacksmithing

(If I missed a door, please do let me know)

### DIFF
--- a/code/game/objects/structures/mineral_doors.dm
+++ b/code/game/objects/structures/mineral_doors.dm
@@ -59,6 +59,12 @@
 	damage_deflection = 10
 	leanable = TRUE
 
+	var/repairable = FALSE
+	var/repair_state = 0
+	var/obj/item/repair_cost_first = null
+	var/obj/item/repair_cost_second = null	
+	var/repair_skill = null
+
 /obj/structure/mineral_door/proc/try_award_resident_key(mob/user)
 	if(!grant_resident_key)
 		return FALSE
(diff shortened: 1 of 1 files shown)


SayonaraEcho:
When windows will be repairable too?

CalamaBanana:
> When windows will be repairable too?

Use two stones and build a stone wall instead.

More seriously, I held off on it for now because people were talking about glassblowing a while ago. If nothing moves in that direction, I might just add an importable and expensive sheet of glass to the merchant's machine.

PR #921, opened 2024-09-04
Grants shophands conditional access to their key, and cosmetic adjustments

CalamaBanana:
## About The Pull Request

They key was removed from them to make the merchant more relevant, but in the end it only made the shophands more unplayable. In 90% of cases when one latejoins as shophand, the shop is either looted and/or the merchant dead. Which made the role of shophand completely unplayable, making latejoining a pure gamble.

Changes made:
- If there is no merchant, if the merchant is dead or if the merchant is disconnected, a thrifty shophand can find a spare key at the Goldface. This interaction is only possible once per round.
- Male shophand now gets the same undershirt as the merchant, to bring his appearance more in line with company policy. He no longer shows up to work wearing pajamas.
- Also removes the 50% chance to meme humen's into being Gizan as merchant. We don't do it to any other races. Merchant's skill has been renamed away from Gizan ages ago anyways.

### DIFF
--- a/code/modules/roguetown/roguemachine/merchant.dm
+++ b/code/modules/roguetown/roguemachine/merchant.dm
@@ -122,7 +122,8 @@
 	var/budget = 0
 	var/upgrade_flags
 	var/current_cat = "1"
-	var/hidden_key = TRUE
+	//Whether or not the hidden key is still present- Shophands can take it.
+	var/hidden_key_present = TRUE
 
 /obj/structure/roguemachine/merchantvend/Initialize()
 	. = ..()
@@ -138,7 +139,7 @@
 
 /obj/structure/roguemachine/merchantvend/attack_right(mob/user)
 	if(user.mind.assigned_role == "Shophand")
-		if(hidden_key)		
+		if(hidden_key_present)		
 			for(var/mob/living/carbon/human/boss in GLOB.human_list)		
 				if(boss.mind)
 					if(boss.mind.assigned_role == "Merchant")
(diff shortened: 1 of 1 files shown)


zakanater19:
Excellent

code/modules/roguetown/roguemachine/merchant.dm
```diff
@@ -122,6 +122,7 @@
 	var/budget = 0
 	var/upgrade_flags
 	var/current_cat = "1"
+	var/hidden_key = TRUE
```
Cyprex, line 125:
```suggestion
	//Whether or not the hidden key is still present- Shophands can take it.
	var/hidden_key_present = TRUE
```
There's better ways to put this, but yeah.

code/modules/roguetown/roguemachine/merchant.dm
```diff
@@ -135,6 +136,28 @@
 	set_light(1, 1, "#1b7bf1")
 	add_overlay(mutable_appearance(icon, "vendor-merch"))
 
+/obj/structure/roguemachine/merchantvend/attack_right(mob/user)
+	if(user.mind.assigned_role == "Shophand")
+		if(hidden_key)		
+			for(var/mob/living/carbon/human/boss in GLOB.human_list)		
+				if(boss.mind)
+					if(boss.mind.assigned_role == "Merchant")
```
Cyprex, line 144: I hate this loop, I would really hope there's a list for this but I'm not going to ask you to find it.

code/modules/roguetown/roguemachine/merchant.dm
CalamaBanana: Alright, changed the name around eveywhere

PR #926, opened 2024-09-05
Silver Related fixes

Jackofashepard:
## About The Pull Request

Silver ore is now actually spawned like all the other ores. Slightly more common than gold
Elf Sabre now smelts into silver which it is also made of
Silver ore no longer spawns as a gem drop (gems will be slightly more common as a result)

## Why It's Good For The Game

For some reason silver which has less value and use than gold is insanely rare. I found the reason is that it was never introduced as a ore. Someone simply added silver ore into the list used to drop gems making it one of the rarest "gems" you can find.

### DIFF
--- a/code/modules/roguetown/roguejobs/blacksmith/anvil_recipes/weapons.dm
+++ b/code/modules/roguetown/roguejobs/blacksmith/anvil_recipes/weapons.dm
@@ -40,12 +40,12 @@
 	created_item = /obj/item/rogueweapon/huntingknife/idagger/steel
 	i_type = "Weapons"
 
-/datum/anvil_recipe/weapons/sidagger
+/*/datum/anvil_recipe/weapons/sidagger
 	name = "Dagger"
 	req_bar = /obj/item/ingot/silver
 	created_item = /obj/item/rogueweapon/huntingknife/idagger/silver
 	i_type = "Weapons"
-
+*/
 /datum/anvil_recipe/weapons/iflail
 	name = "Flail"
 	req_bar = /obj/item/ingot/iron


Elli-Skala:
> For some reason silver which has less value and use than gold is insanely rare.

Really its just antag balance in the end due to silver being hard counters to both vampires and wearwolfs

annikaRU:
> > For some reason silver which has less value and use than gold is insanely rare.
> 
> Really its just antag balance in the end due to silver being hard counters to both vampires and wearwolfs

Note that the effects of silver on werewolfs and vampires were drastically reduced so the rarity serves no real purpose anymore

radarseas:
> Note that the effects of silver on werewolfs and vampires were drastically reduced so the rarity serves no real purpose anymore

despite absolutely loving silver weaponry even as an avid player of vampires beforehand that complained a TON of about miracles, I'd much prefer a silver weapon I can notice, outrun or stun the user of, than deal with miracles halfway accross the screen. I don't think the status quo should be making it useless either, silver SHOULD be rare and absolutely terrifying for paranormal threats.

Elli-Skala:
Also if we're making it spawn it's likely we would want to add it to the stock pile but if we add it to the stock pile we have to deal with the fact then silver could be ordered en mass

annikaRU:
> Also if we're making it spawn it's likely we would want to add it to the stock pile but if we add it to the stock pile we have to deal with the fact then silver could be ordered en mass

It already spawned, but as a gem drop, this just moves it towards a mineral drop,

ikrarkjr:
"For some reason silver which has less value and use than gold is insanely rare. "
If you don't know the reason why the hell are you making PRs? The reason is because it's extremely overpowered as a weapon versus antags. I put it exactly where it is because it's exactly where it should be.

annikaRU:
> "For some reason silver which has less value and use than gold is insanely rare. " If you don't know the reason why the hell are you making PRs? The reason is because it's extremely overpowered as a weapon versus antags. I put it exactly where it is because it's exactly where it should be.
> 
> Edit: "Silver related fixes" Fixing what? What is this fixing?



> "For some reason silver which has less value and use than gold is insanely rare. " If you don't know the reason why the hell are you making PRs? The reason is because it's extremely overpowered as a weapon versus antags. I put it exactly where it is because it's exactly where it should be.
> 
> Edit: "Silver related fixes" Fixing what? What is this fixing?

If you read what was changed you would learn what is being fixed, regardless silver has less of an effect on antags now, there were PRs merged a bit ago that lessened its effects on vampires and I am uncertain about werewolfs silver isnt as overpowered as it used be


elvish sabre now actually is made of silver and not steel

radarseas:
counterpoint to even calling it OP vs werewolves, they should be fighting people with silver in numbers anyway, they can literally oneshot delimb you through fullplate.

Cyprex (changes requested):
No. Silver is supposed to be rare, that's why it's in the gem drop table in the first place. Smelting elven daggers+sabers we can talk about, however- how rare are these weapons?

Jackofashepard:
> No. Silver is supposed to be rare, that's why it's in the gem drop table in the first place. Smelting elven daggers+sabers we can talk about, however- how rare are these weapons?

So firstly may I ask why? To my understanding its useful against some antags but its not OP. Value wise its also worth less than gold which is why i opted to make it slightly more common than gold which, to be fair, is still pretty rare especially when you consider all ores will be "rarer" now due to a new possibility.

Secondly these weapons are pretty rare. As far as i know only some adventurer starts with elf race get them but is there any problem with smelting these down? Isnt the entire supposed point of having silver rare to avoid having these weapons? If someone wants to smelt it down and make something else i honestly dont see any reason to stop them.

Cyprex:
> Secondly these weapons are pretty rare. 

Yes, and they should stay that way! Hence I don't want this! You do sometimes see silver ore- last round had two just laying in the smithy!

annikaRU:
> > Secondly these weapons are pretty rare.
> 
> Yes, and they should stay that way! Hence I don't want this! You do sometimes see silver ore- last round had two just laying in the smithy!

The weapons are rare, and the effects of silver on antags aren't that great anymore, silver shouldn't be in the gem drop table considering  that is isn't that good anymore, a silver dagger used to be scary against a werewolf, it isn't now

Jackofashepard:
I have investigated. As far as I can see silver inherently has 0 effect on antags. There is only one wildly OP silver item which can effect antags which I have now simply removed from the crafting recipe list.

PR #860, opened 2024-08-27
Update legendary pilgrim classes

ikrarkjr:
## About The Pull Request

All legendary pilgrim roles have been updated. Whether due to deprecation, or just a balance reshuffle. About half have been outright buffed. 

Roll chance for all legendary pilgrim roles have been adjusted based on combat prowess. Most likely are the weakest, least likely are the strongest.

Legendary acolyte was in the code as a file but it was blank. Now it's not! It has been added as a new legendary class.

## Why It's Good For The Game

Legendary fisher was literally straight up worse than normal fisher lmfao. Someone has to do this stuff.
You now actually FEEL legendary when you're a legendary class. No more "I'm slightly better, whoopie. Kinda nice."

PR #927, opened 2024-09-05
Gives polearms 2 to every towner role

wow50000:
## About The Pull Request
gives polearms 2 to every towner and some yeomen

## Why It's Good For The Game
man(I think its a man anyways) on discord told me to
image
the non joke reason is so that towners can be drafted into peasant armies


challade:
Seems like powercreep.

wow50000:
its 20 one line meme changes

NikolayByondPlayer:
what

NikolayByondPlayer:
what

DorfGuy1:
Nah nah. Like it was common for people to be made to go out and practice things like that, or archery, in case of needing to be raised as a levy, at least in england. Also can you fix the "master" classes? They almost unversally worse than their normal  pilgrim counterparts, like terribly so, in terms of skills. Miner is the egregious example. No masonry, no and horror of horrors No Athletics.

TheFenix323:
>Also can you fix the "master" classes? They almost unversally worse than their normal pilgrim counterparts, like terribly so, in terms of skills. Miner is the egregious example. No masonry, no and horror of horrors No Athletics.
Seems like #860 is already fixing this

PR #928, opened 2024-09-05
Extends the keep wall up to Z5

zakanater19:
## About The Pull Request

Replaces the outer roof area with higher walls, there are windows but in the map editor they just look like normal stone walls

## Why It's Good For The Game

Makes the keep more secure

Screenshot 2024-09-05 092802

### DIFF
--- a/_maps/map_files/rockhill/rockhill.dmm
+++ b/_maps/map_files/rockhill/rockhill.dmm
@@ -5346,6 +5346,10 @@
 /obj/machinery/light/rogue/torchholder{
 	pixel_y = 26
 	},
+/obj/structure/mineral_door/wood{
+	locked = 1;
+	lockid = "walls"
+	},
 /turf/open/floor/rogue/ruinedwood,
 /area/rogue/outdoors/town)
 "fNc" = (
@@ -13814,6 +13818,10 @@
 /obj/structure/chair/bench,
 /turf/open/floor/rogue/grass,
 /area/rogue/outdoors/rtfield)
+"pdO" = (
+/obj/machinery/light/rogue/firebowl,
+/turf/open/floor/rogue/cobble,
+/area/rogue/outdoors/town)
 "pel" = (
 /obj/structure/fluff/railing/border{
 	dir = 6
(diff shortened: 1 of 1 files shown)


zakanater19:
added a keep wall door up there too

Heheniy:
keep should not be inpenetrable

Hmeister-real:
Bad, literally no way to get in now except for the entrance or the underground

CalamaBanana:
> keep should not be inpenetrable

Walls can be broken easily enough. Quickly even with a small group. The keep should be more impenetrable than the damn farm and withstand a small siege, the rogue on the roof meta was just so damn overplayed.

SixplyDev:
Frankly, I like this change because the Keep will now be a lot more secure than it was previously. It was annoying as fuck seeing someone skybridge over to the Keep and escape easy as shit. This will make it a lot harder now to do that as well as allow us to repair holes in the Keep's wall if someone manages to break in. Win all around if you ask me.

MefAnneFeminene:
People downvoting this are idiots. This is an amazing change.

radarseas:
as much as I hate to say it... I think the upper part of the keep being vulnerable is fine, for mechanical reasons plus the **round's sake** over roleplay. _literally just shoot people with your armory bows and gatekeeper's crossbow, or even ask the watch for theirs, they get PLENTY in their armory._


Revolutions need a way to hit the keep while their numbers are around. you shouldn't be able to stop people getting in nearly forever by using 6 rocks to make 3 walls. (grill from river, watch tunnel).

_Until we have more threat to the keep's expert soldiers than bored adventurers joining revs with a king that's reasonable anyway_ or bandits high on moondust, I think giving variety to break ins is fine. Yes it should be a fortress.

But it shouldn't be foolproof neither and it'll be impossible to kidnap nobles if people can camp the likely exits.. That's considering the keep will even do hostage RP instead of fragging them over it.

Unless we get a faster way to break stone walls... via picks or whatever, currently you'll have an entire army waiting at the other side with how loud and noisy it is.

PR #934, opened 2024-09-06
Keeps the bandits out of the dungeon for the first 30 minutes of the round

zakanater19:
## About The Pull Request

When a player becomes a bandit, they get a dungeon death timer component added to them.
If the bandit enters the dungeon area (/area/rogue/under/cave/spider), they receive a warning message.
The bandit has 30 seconds to leave the dungeon area.
If they don't leave within 30 seconds, their character is gibbed.
After 30 minutes from the start of the round, the restriction is removed, allowing them to enter the dungeon safely.

## Why It's Good For The Game

Apparently the bandits would clean the dungeon out first, so, yeah, here is the bandaid for now

Screenshot 2024-09-06 122906

### DIFF
--- a/code/game/objects/structures/traveltile.dm
+++ b/code/game/objects/structures/traveltile.dm
@@ -138,13 +138,64 @@
 /obj/structure/fluff/traveltile/vampire
 
 /obj/structure/fluff/traveltile/dungeon
-	name = "gate"
-	desc = "This gate's enveloping darkness is so opressive you dread to step through it."
-	icon = 'icons/roguetown/misc/portal.dmi'
-	icon_state = "portal"
-	density = FALSE
-	anchored = TRUE
-	max_integrity = 0
-	bound_width = 96
-	appearance_flags = NONE
-	opacity = FALSE
+    name = "gate"
+    desc = "This gate's enveloping darkness is so oppressive you dread to step through it."
+    icon = 'icons/roguetown/misc/portal.dmi'
+    icon_state = "portal"
+    density = FALSE
+    anchored = TRUE
+    max_integrity = 0
+    bound_width = 96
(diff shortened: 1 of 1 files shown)


DelineFortune:
they can just camp outside dungeon then

zakanater19:
> they can just camp outside dungeon then still doing nothing

so you are saying the bandits will be robbing people

SylvinSeik:
Outright gibbing them feels comical levels of silly, even for a band-aid.  Would it not be practical to just prevent them from going through the travel tiles until the 30 minutes are up?

zakanater19:
> Outright gibbing them feels comical levels of silly, even for a band-aid. Would it not be practical to just prevent them from going through the travel tiles until the 30 minutes are up?

i did try for about half an hour to block them from using the traveltile gate and then when it started to piss me off i settled on this solution instead

radarseas:
Why not just teleport them outside of the gate after like 3 seconds instead. _Gibbing them utterly obliterates_ any chances of revival for such a vague message and makes absolutely, _no sense at all, as based as the idea is I will admit._

radarseas:
honestly dungeon in general should have a grace period so people can't rush it for _**steel fullplate less than 30 minutes**_ in anyway, _that's without including bandits._ Especally that the skele gear buffs give them tons of metal equipment instead of something sane like leather armor like bandits, wolf helmets, gambesons, etc.

zakanater19:
> Honestly, you also could add a trait similar to the bandit hideout one. Have only adventures be able to initially go into the dungeon unless they take you along. Adventurers having a limited use for once, perhaps even getting hired to go adventuring

Yeah maybe, if someone is willing to PR that is obviously a better solution

PR #936, opened 2024-09-06
Add support for the Auxtools Debugger

Cyprex:
## About The Pull Request

This PR adds support for the auxtools debugger. All that a user needs to do is have the DM Langserver client extension installed(which comes with the recommended extensions) and run with debugging.

A reference I used a lot in making this: https://github.com/SpaceManiac/SpacemanDMM/wiki/Setting-up-Debugging

I've elected to add /world/Genesis() to enable debugging as soon as possible, which allows debugging early init. This is taken directly from tg.

A note on SpacemanDMM: I've only enabled the debugger here. We should enable the (local on the users PC, NOT on github!) linter part of it too, but that first requires dealing with a ton of linter issues, largely from old tgcode.
Documentation for its config: https://github.com/SpaceManiac/SpacemanDMM/blob/master/CONFIGURING.md

There is one thing that needs addressing either before or shortly after this is merged- and that is explaining coders why their game suddenly froze(The debugger defaults to breaking execution on runtimes).

### DIFF
--- a/modular/ze_genesis_call/genesis_call.dm
+++ b/modular/ze_genesis_call/genesis_call.dm
@@ -1,3 +1,5 @@
+//Shamelessly stolen from https://github.com/tgstation/tgstation/pull/74808 by Dominion
+
 /*
  * BYOND loves to tell you about its loving spouse /global
  * But it's actually having a sexy an affair with /static


Cyprex:
This is good to merge, I've put up a basic explanation on Discord, and I'll copy it here too:

A debugger is a tool that can stop execution at any point, and allow you to inspect the state of the program. For example you could ask it to stop execution in the middle of a proc you just added, and then you could inspect the value of local variables, arguments etc. You can also step through code line-by-line to see exactly what is going on- this is incredibly useful when you just wrote code and it's not doing what it's supposed to do.

Here's a more elaborate explanation, but note that not all of this applies to us: https://www.youtube.com/watch?v=XmpIBsnc3xU&t=96s

Importantly, and the issue here is that by default the debugger breaks on runtimes too. So if you've never used the debugger and then you get a runtime in local testing the game freezes(you're supposed to look at the debugger, the game freezing is the whole point- we just stopped all code execution, of course it's gonna freeze!). You can just hit F5 in VSCode to continue, but you ought to look at the runtime that made it freeze- and if that's your code do fix it!

Cyprex:
If someone has a better explanation of what a debugger is do let me know!

Cyprex:
Neon said I could selfmerge

PR #938, opened 2024-09-07
Fix some linter issues found via Opendream

Cyprex:
## About The Pull Request

This fixes various issues found via the OD linter. There should be no changes to in-game behavior, as far as I can tell.

I also snuck in some explanatory comments here and there.

### DIFF
--- a/code/datums/migrants/migrant_role.dm
+++ b/code/datums/migrants/migrant_role.dm
@@ -18,7 +18,7 @@
 	var/greet_text
 	/// Whether to grant a lit torch upon spawn
 	var/grant_lit_torch = FALSE
-	/// Whether to use a job title like "wandering Human"
+	//Whether to display "wandering Human" rather than their job title
 	var/show_wanderer_examine = TRUE
 	var/advjob_examine = TRUE
 	var/banned_leprosy = TRUE


code/_compile_options.dm
```diff
@@ -77,3 +77,8 @@
 // A reasonable number of maximum overlays an object needs
 // If you think you need more, rethink it
 #define MAX_ATOM_OVERLAYS 100
+
+//Disable unimplemented proc/var warnings
+#ifdef OPENDREAM
+#pragma UnimplementedAccess disable
+#endif
```
Cyprex, line 84: There are around ~200 of these, very annoying- these just signify that opendream does not implement a given byond builtin

code/datums/components/construction.dm
```diff
@@ -87,8 +87,9 @@
 				. = user.transferItemToLoc(I, parent)
 
 			// Using stacks
-			else if(istype(I, /obj/item/stack))
-				. = I.use_tool(parent, user, 0, volume=50, amount=current_step["amount"])
+			else 
+				if(istype(I, /obj/item/stack))
+					. = I.use_tool(parent, user, 0, volume=50, amount=current_step["amount"])
```
Cyprex, line 92: else if isn't really valid in a switch statement, see also:
https://github.com/tgstation/tgstation/pull/67036/files#r873321614

This may actually affect us- This is some general tg construction code, I'm unsure if we utilize this.

code/modules/antagonists/blob/powers.dm
```diff
@@ -248,7 +248,7 @@
 
 /mob/camera/blob/verb/expand_blob_power()
 	set category = "Blob"
-	set name = "Expand/Attack Blob ([BLOB_SPREAD_COST])"
+	set name = "Expand/Attack Blob (4)"
```
Cyprex, line 251: This kinda sucks(verb names need to be compile time constants, and must not use string interpolation like this does) but nobody will adjust this define, we don't even use blob so I suppose it's okay.

code/modules/mob/living/carbon/human/species_types/roguetown/roguetown_species.dm
```diff
@@ -18,7 +18,7 @@
 #define REGEX_ENDWORD 3
 #define REGEX_ANY 4
 
-/datum/species/proc/handle_speech(datum/source, mob/speech_args)
+/datum/species/proc/handle_speech(datum/source, list/speech_args)
```
Cyprex, line 21: This is used like a list here, is declared as a list in most declarations of handle_speech and is actually a lust(I checked via the debugger. I didn't check the two preceding cases in tgcode but I'm pretty sure the same applies.

PR #939, opened 2024-09-07
Disallows Nobles and Garrison members from becoming Peasant Rebels

Cyprex:
## About The Pull Request

This PR fixes old, broken code that was supposed to stop nobles & garrison members from becoming peasant rebels. I believe Aspirant is entirely separate from Peasant Rebels.

Whilst this does fix old code this is effectively a balance change, and as such I put this in its own PR(It came up during Opendream lints). I'm more than open to discussion on this, I'm not sure what the right balance is myself!

Note that these roles are:
```
GLOBAL_LIST_INIT(garrison_positions, list(
	"Watchman",
	"Bog Guard",
	"Bog Master",
	"Man at Arms",
	"Dungeoneer",
	"Gatemaster",
	"Town Sheriff",
))

GLOBAL_LIST_INIT(noble_positions, list(
	"King",
	"Queen Consort",
	"Prince",
	"Guard Lieutenant",
	"Royal Marshal",
	"Councillor",
	"Steward",
	"Knight",
))
```
Note the presence of bog guard, and the absence of hand

Also: I didn't actually test this, it's a bit of a pain to test locally.

Do not merge until I have adjusted the list of roles this PR affects

### DIFF
--- a/code/modules/antagonists/roguetown/villain/aspirant.dm
+++ b/code/modules/antagonists/roguetown/villain/aspirant.dm
@@ -37,7 +37,7 @@
 /datum/antagonist/prebel/can_be_owned(datum/mind/new_owner)
 	. = ..()
 	if(.)
-		if(!new_owner.assigned_role in GLOB.noble_positions || !new_owner.assigned_role in GLOB.garrison_positions)
+		if(!(new_owner.assigned_role in GLOB.noble_positions) || !(new_owner.assigned_role in GLOB.garrison_positions))
 			return FALSE
 
 /datum/antagonist/aspirant/on_gain()


MefAnneFeminene:
Why bogguard? Bogguard SHOULD be allowed to join the revolt. They're literally peasants

DelineFortune:
"Watchman", - literally towners but with uniform, not even close to court as mma's
"Bog Guard", - full of shunned races, penal platoon, mentally same as bandits but on king's side

Elli-Skala:
Doesn't this open a weird mechanic where you can "mind shield" people via force enlisting the entire town to the guard?

UntoldTactics:
> Doesn't this open a weird mechanic where you can "mind shield" people via force enlisting the entire town to the guard?

No. Only adventurers and pilgrims can be recruited via ANY OF THE RECRUIT SPELLS (e.g recruit town guard, recruit bog guard, recruit acolyte, recruit templar)

RadiantFlash7:
I'm of similar mind personally about the bog guard. If theres any garrison members that should be allowed to join a rebellion, it's the ones furthest from town, without much oversight and often ignored/neglected by the keep.

Worth noting that this change keeps the veteran able to become a peasant rebel.

Cyprex:
Honestly just give me a list of roles that should/should not be able to become peasant rebels and I'll add them.
So, so far we have: No veterans
Add bog guards
What about watchmen & bog masters?

radarseas:
I think honestly, the entirety of the bog garrison should be able to join including the bog master, maybe watchmen, in the case the king doesn't pay them. They are still conscripts in a way and if it comes to enforcing street executions or terrible laws or without any proper pay, it makes sense they'd start to think for themselves and join those oppressed.

Also it gives some interesting potental in the case people don't immedately field-execute suspected and actual rebels by decapitation or dragging them to the keep to volf pit them immedately since you can sneak fellow revs out of the dungeon.

Just because people _can join_ a revolt doesn't mean they will. We have different standards to where the change was put in for.

radarseas:
although honestly, only the current ruler should be immune to joining revs and there should be a grace period before you can accept/decline to prevent mid-typing recruitment/denial.

often most non-recruitable roles will have reasoning to be rebels or heirs onboard with the rebellion/wannabie aspirants, might as well just give them the status and ability to let rebels actually win.

Keazra:
Lord knows I never get fucking paid as the Watch. Maybe I WILL revolt!!

PR #947, opened 2024-09-09
Map fixes

Heheniy:
## About The Pull Request

Removes tree walls around the map, placed a bunch of single trees instead of deleted ones
Replaced bones at south-west bog mine with proper ones, that can be passed
Fixes double closet at hand's room
Fixed torches inside walls where i managed to find them
Added lights to admin test area
Fixed drawer at 3rd fancy inn room
Moved shylock near smiths to be between peddlers
Made merchant's goldface look less out of place
Tweaked some standing fires at inn bath to be passable
Standartized stone keep walls on 1st floor
Deleted this 
image

## Why It's Good For The Game
Stupid tree walls is no good

### DIFF
--- a/_maps/map_files/templates/smalldungeons/dungeon1.dmm
+++ b/_maps/map_files/templates/smalldungeons/dungeon1.dmm
@@ -810,7 +810,6 @@
 /area/rogue/under/cave/spider)
 "Op" = (
 /obj/machinery/light/rogue/forge,
-/obj/machinery/light/rogue/forge,
 /turf/open/floor/rogue/tile/masonic/inverted,
 /area/rogue/under/cave/spider)
 "Oq" = (


Cyprex:
I like the treewalls, but the rest of this PR is good. They make the areas they're in hard to navigate, and really incentivize you to get out your axe.

CalamaBanana:
Tree walls are important to break up the openness of the bog at least a little. They are important for juking as well, the bog should not be just neatly trimmed city park. To be honest, there should be even more tree walls, to turn the bog into a proper maze.

Red-Honk:
Tree walls should stay, interesting combat situations happen when one party gets cornered there or uses the place to hide, or even makes an omega tree house, it's quite amusing really.

And when it comes to the rest, well, why not.

Cyprex:
Good changes, but don't change the trees!

Heheniy:
There will still be stupid tree walls because of random tree spawn

Jaaackx (changes requested):
Not a fan of the 'tree wall' removal, everything else here is good.

Heheniy:
artificial bullshit like this was removed
image



there still will be random tree walls as trees have random spawn

Cyprex:
Tree maze is good

PR #948, opened 2024-09-09
Miracles while gagged fix

Heheniy:
## About The Pull Request

Being able to cast miracles and invisibility while gagged fixed


Cyprex:
Have you tested this? Does it try to whisper something with this?

Heheniy:
it doesnt

PR #952, opened 2024-09-09
Added bog armory keys to boggurds

Jackofashepard:
Title says it all

### DIFF
--- a/code/game/objects/items/rogueitems/keyrings.dm
+++ b/code/game/objects/items/rogueitems/keyrings.dm
@@ -137,7 +137,7 @@
 	keys = list(/obj/item/roguekey/councillor_rooms, /obj/item/roguekey/keep_barracks, /obj/item/roguekey/keep_dungeon,  /obj/item/roguekey/manor)
 
 /obj/item/keyring/bog_guard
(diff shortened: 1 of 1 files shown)


Cyprex:
Can be merged in 21h

Keazra:
There should be, like, one key somewhere roundstart for the bogguards to grab in the event of no bogmaster.

SayonaraEcho:
>  It's a massive treasure trove of equipment for literally anyone who wants it and giving a key to every single bogman is a bad idea because of this. It'd also be easy to ambush a single person and get keys to the whole armory and everything in it. It doesn't make sense to make an exception for the bogmen when every other armory is locked behind the garrison captain's keys. I don't think this would be helpful to the Bog Guard and I think it would simply cause more issues with the well-stocked armory they already have.

Do you know that if there's no bog master then bog guards can just easily break the armoury's door in roundstart? There's no point for it. It just takes a little more time to get the gear than just opening the damn door in one second. 
If you're talking about bandits or thieves, they can get into the barracks without keys with ease too, because no one is in them, since the bog guards are sitting on the gate at all times. Bandits usually break a reinforced window on the third floor and climb inside and then go down into the basement to the armoury.

Cyprex:
If this ends up being problematic we'll just revert it.

PR #956, opened 2024-09-10
Changes default voice pack's most used sounds/emotes

Coolgat3:
## About The Pull Request

This changes the default voice pack's laugh, screaming, pain and rage sounds, to use the commoner voice from mordhau
(Someone is going to say copyright issues, our knight and barbarian [also king is mordhau Cruel, blackstone/roguetown had teenagers using Young voice lines from mordhau aswell] voice packs are from mordhau and have been that since roguetown, all the way through blackstone, and still are.)

## Why It's Good For The Game

Because right now everyone sounds like they're some horny hobo crackhead whenever they laugh, this also makes most of the voices uniform, ideally in the future I'd like to add pickable voicepacks somehow

https://github.com/user-attachments/assets/2b397341-1c58-4d9c-98d4-cdb8b880731c


Coolgat3:
> as much as I hate seeing the hoohhoeheheheee go, this is tenfold better. The current laughs ingame would ONLY fit the jester but I think even they have their own kind of laughing.

thank you for the kind words C: honestly i think the old laugh maybe it would fit bandits too in a way or adult age wizards.

BlackjackIsTired:
Love these. Need more.

PR #957, opened 2024-09-10
increase paper price

Cyprex:
## About The Pull Request

This needs to be increased to stop an exploit from selling paper, but I'm not sure what the proper price would be- the code for this is surprisingly involved. I hope 16 for 8 paper isn't too much!

### DIFF
--- a/code/modules/cargo/packsrogue/tools.dm
+++ b/code/modules/cargo/packsrogue/tools.dm
@@ -43,7 +43,7 @@
 
 /datum/supply_pack/rogue/tools/paper
 	name = "Paper"
-	cost = 3
+	cost = 16
 	contains = list(
 					/obj/item/paper/scroll,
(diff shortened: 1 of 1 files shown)


TuskTusk:
I want to pitch in on this and say that mass-buying Paper and selling it to the navigator has been a long-standing 'exploit' lazy or lowpop Merchants have done to rapidly multiply money. It was more xhausting before the Navigator stairs was mvoed, but now you just do a couple of runs and generate several hundred mammons every Navigator drop.

I agree it needs to be changed, but someone needs to check out the math.

Inveldy:
Make it fifteen mammon so that you can still, technically, profit off of paper sales if you're desperate enough, but it takes over ten times as long due to going from fifteen in profit to one.

Cyprex:
Absolutely not, no exploits

PR #960, opened 2024-09-10
Nerfs Dendor as a patron and changes kneestinger spell

GuppyLaxx:
This PR removes dendor's kneestinger immunity and the conjure kneestingers spell

Instead, dendorites get a trait that lets them walk through weepvines easier, and the spell is replaced with one that makes a wall of weepvines infront of them in a 3x1 row.

Also, this PR adds the evil tree into the spawning pool of the bog

### DIFF
--- a/code/__DEFINES/traits.dm
+++ b/code/__DEFINES/traits.dm
@@ -47,12 +47,13 @@
 // PATRON GOD TRAITS
 #define TRAIT_ROT_EATER "Blessing of Pestra" //can eat rotten food
 #define TRAIT_ORGAN_EATER "Blessing of Graggar" //can eat organs
-#define TRAIT_KNEESTINGER_IMMUNITY "Blessing of Dendor"
+#define TRAIT_VINE_WALKER "Blessing of Dendor"
 #define TRAIT_SOUL_EXAMINE "Blessing of Necra" //can check bodies to see if they have departed
 #define TRAIT_CRACKHEAD "Blessing of Baotha" //will never overdose
 #define TRAIT_COMMIE "Blessing of Matthios" //recognized by bandits as an ally
 #define TRAIT_CHOSEN "Astrata's Chosen"
 
+#define TRAIT_KNEESTINGER_IMMUNITY "Kneestinger Immunity"
 #define TRAIT_BASHDOORS "bashdoors"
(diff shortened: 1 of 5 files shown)


CalamaBanana:
This is not the way to go about this. The kneestinger spell should just be changed to be less combat-castable with either a long warmup where one needs to remain standing still or growth time for the seedlings on the ground. Lighting your Dendorite commune with kneestingers is great, using them in your defenses is great. (Not like kneestingers can't be easily removed if one brings a stone knife or a torch)

Besides, there are Dendorites that are not clerics. Why would you make the Dendor-perk apply to something that only clerics have now? Let us keep the immunity against kneestingers. No reason to rip that out completely. Kneestingers exist on the map.

radarseas:
I think a better idea would be keep the kneestinger spell but rework it so you can cast a growth at a spot instead, with a small windup time and a brief grace period once its placed, so you can still use it for offensive or defensive utility like blocking off an exit, just not for an immedate win stun.

That and cheapen it down to compensate for a nerf, kneestingers should still stay, I hate to say it, they are a staple of fighting dendorites in combat.

DorfGuy1:
I've only ever seen weepvines ONCE. Don't nerf Dendorites for the sins of a few people that should have not been able to play clerics anyway. The Druids especially don't deserve it. If I'm not mistaken they still gain devotion slower from a previous kneestinger nerf.

SixplyDev:
I frankly think this is an incredibly based change and should cut down on Dendor patron abuse. Thank you

DorfGuy1:
Is that charge time in seconds or what, because it's either two or twenty seconds, either way, wtf. If it takes that long it'll barely be usable. Summoning vines to entangle your enemy is a dangerously based idea for a Dendor spell though.

CalamaBanana:
If you rip out the kneestinger spell, there is no reason to remove the immunity alongside it. There are not-Cleric Dendorites playing it for the vibe, no reason to not have them get any benefits from it anymore. 

Kneestingers spawn in the wild, it's a cute little thematic benefit for Dendorites to have a very very subtle advantage when sproinking through the bog. No reason to not have both vine and stinger immunity.

(Once more, unlimited adventurers are the problem tempting to make the game worse for everyone else. The current kneestinger spell would have been perfectly fine if it were druid only.)

PR #962, opened 2024-09-11
Removes Bandit Love Potion, replaces it with 2 Firebombs

GuppyLaxx:
What it says on the title.

The love potions were a f***ing mistake.

### DIFF
--- a/code/game/objects/structures/fluff.dm
+++ b/code/game/objects/structures/fluff.dm
@@ -991,7 +991,6 @@
 							I += /obj/item/reagent_containers/powder/moondust
 							I += /obj/item/reagent_containers/powder/moondust
 							I += /obj/item/bomb
-							I += /obj/item/bomb
 						if(3)
 							I += /obj/item/clothing/suit/roguetown/armor/plate/scale
 						if(4)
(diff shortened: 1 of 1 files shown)


EzekyalLegit:
I think the interaction between bandit and love potion WAS interesting. I wonder if it could be changed so its 10% chance a love potion and 90% a firebomb.

EzekyalLegit:
One example of the value of love potion for bandits was how it could simulate stockholm syndrome in a captive worker. Usually captives often betray the camp with a reckless abandon, losing their life to lash out at the bandits. With this potion the captive slave was able to be given a firm reason to obey the camp and actually play along, Instead of trying to behead sleeping bandits or sneaking out to lead the keeps army into the camp.

The bandits just DONT have the manpower to corral slaves properly so this potion was a nice IC workaround.

radarseas:
why not push them up with the matthios dragonscale necklace reward instead? they DO lead to interesting gameplay but it was far too early.

Maybe put firebombs at T3... I think bandits luring the keep's army out and just absolutely critting all of them by spamming firebombs is urm... a little bit of a balance issue for something you can do early into the round.

kepesk1:
this guy makes some of the goofiest PRs

wow50000:
I understand why you are removing them but dont replace them with fire bombs lmao

Takoros:
love potion are cool, dont remove that

DorfGuy1:
And he merged it anyway.

PR #941, opened 2024-09-07
Milking and Saiga Fixes

TheFenix323:
## About The Pull Request
Milking always worked in an odd manner, you needed to break away from the progress bar to milk the animal, now you need to wait for the progress bar to finish.
This also fixes Saigas breeding spawning aggressive Saigas instead of tamed ones

## Why It's Good For The Game
Fixes milking and Butchers getting killed by Saigas.
fixes #812

PR #964, opened 2024-09-11
Rewrites cow milking & fixes grown-up Saigas being aggressive

Cyprex:
## About The Pull Request

This fixes milking(you had to cancel it due to an incorrect ! in front of a do_after) as well as a problem where mobs, when growing up into their adult forms, did not properly transfer their tamed status. i. e. adult saigas grown from young ones were always aggressive.

This is a reimplementation of #941


Cyprex:
We'res seriously understaffed, Neon is busy so I'm selfmerging this

PR #965, opened 2024-09-11
Templar Buff

Jackofashepard:
Simply buffs templars to be more of a threat, especially with the new grab changes.

The Templars get more skills like improved swords, polearms, whips/flails. Also added a one more level of medicine so they are more useful than just knight light.

The church has become kinda a joke lately. If the king wants to reject the church its pretty easy to do so with no consequences, this will at least make sure that if the king or any other party in the game decides to assault the church they will need to be well prepared.


Jackofashepard:
For anyone wondering i balanced this against Keep. The templar is basically a knight light with less equipment and skills. They still probably wont survive a full keep onslaught as the keep has the veteran, captain, 2 knights and quite a few men at arms (and even mercs).

radarseas:
while I get the intent, remember that flails can stunlock super hard, there's three of these going around, they HAVE miracles to heal out of being paralyised, undead antagonists are already well, well, well countered by just churning them and the last, most important point is that flails can stunlock through armor.

These guys CAN, get fullplate reilably if a dendorite paladin/druid/cleric/templar farms and sells off to stockpile or even a coat of arms, they also start with heavy helmets.

the church should still have to rely on some outside help for defending themselves, they shouldn't be impossible to fight for undead antagonists without cheese tactics, ranged magic spam and crossbows either.

Cyprex:
This is probably too much

radarseas:
> I would (personally) like to see them also be given expert axes, the skill is rare and not overturned. Admittedly I simply enjoy using axes.

axes don't really seem, thematic though... plus I believe giving them weapons that specialise in delimbing people when they hold the cards of healing and surgery isn't a good idea.

also remember vampires, when they existed there was like, no hope if you lost a limb because miracles = fire = vamp disguise fails = valid... yeeeah.

Jackofashepard:
Closed since inquis got goons

PR #966, opened 2024-09-11
In-Game Preferences

Gyrandola:
## About The Pull Request
image

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -652,6 +652,7 @@ GLOBAL_LIST_EMPTY(chosen_names)
 		dat += "<a href='?_src_=prefs;preference=save'>Save</a><br>"
 		dat += "<a href='?_src_=prefs;preference=load'>Undo</a><br>"
 
+	var/mob/dead/new_player/N = user
 	// well.... one empty slot here for something I suppose lol
 	dat += "<table width='100%'>"
 	dat += "<tr>"
@@ -661,7 +662,6 @@ GLOBAL_LIST_EMPTY(chosen_names)
 	dat += 	"</td>"
 	dat += "<td width='33%' align='center'>"
 	dat += "<a href='?_src_=prefs;preference=bespecial'><b>[next_special_trait ? "<font color='red'>SPECIAL</font>" : "Be Special"]</b></a><BR>"
-	var/mob/dead/new_player/N = user
 	if(istype(N))
 		if(SSticker.current_state <= GAME_STATE_PREGAME)
 			switch(N.ready)


Cyprex:
You need to add to the PR description that this lets you open the preferences menu at any time.

Also, I suspect there are code crimes where your character(in the preferences menu) and your character(in-game, your actual mob) aren't properly separated. The way this should work: These are entirely separate, and options from the prefs menu are copied over onto the actual mob when you spawn in.

But you know what? If that's the case we'll simply find out the hard way, we can squash those issues, or revert this change if it becomes too bad.

code/modules/client/preferences.dm
```diff
@@ -688,7 +689,7 @@ GLOBAL_LIST_EMPTY(chosen_names)
 		dat = list("<center>REGISTER!</center>")
 
 	winshow(user, "preferencess_window", TRUE)
-	var/datum/browser/popup = new(user, "preferences_browser", "<div align='center'>[used_title]</div>")
+	var/datum/browser/noclose/popup = new(user, "preferences_browser", "<div align='center'>[used_title]</div>")
```
Cyprex, line 692: I'm not sure about this naming, I thought this was window that you can't close.

code/modules/client/preferences.dm
```diff
@@ -661,18 +658,22 @@ GLOBAL_LIST_EMPTY(chosen_names)
...
-		if(!is_active_migrant())
-			dat += "<a href='byond://?src=[REF(N)];late_join=1'>JOINLATE</a>"
+	var/mob/dead/new_player/N = user
+	if(istype(N))
+		if(SSticker.current_state <= GAME_STATE_PREGAME)
+			switch(N.ready)
+				if(PLAYER_NOT_READY)
+					dat += "<b>UNREADY</b> <a href='byond://?src=[REF(N)];ready=[PLAYER_READY_TO_PLAY]'>READY</a>"
+				if(PLAYER_READY_TO_PLAY)
+					dat += "<a href='byond://?src=[REF(N)];ready=[PLAYER_NOT_READY]'>UNREADY</a> <b>READY</b>"
```
Cyprex, line 668: I just want to say here that people shit on tgui a lot but this isn't exactly pretty either.

code/modules/client/preferences.dm
```diff
@@ -661,18 +658,22 @@ GLOBAL_LIST_EMPTY(chosen_names)
 	dat += "<td width='33%' align='left'></td>"
 	dat += "<td width='33%' align='center'>"
 	dat += "<a href='?_src_=prefs;preference=bespecial'><b>[next_special_trait ? "<font color='red'>SPECIAL</font>" : "Be Special"]</b></a><BR>"
-	if(SSticker.current_state <= GAME_STATE_PREGAME)
-		switch(N.ready)
```
Cyprex, line 665: For anyone looking at this- the only actual changes here are this and the preceding line, and last two line of this block(675 & 676). The rest is indented one level deeper than before because it's now inside an if(). This is fine.

Cyprex: Honestly this should be factored out into its own function for readability, I think.

code/modules/client/preferences.dm
Cyprex: "notclosing" might work

Cyprex:
That shouldn't be hard to fix

pilotmans:
Finally

Cleric2145:
> If you can change defiant mid-round with this we need to remove that ngl

As mentioned above, the changes (other than keybinds) are applied at spawn, so changing defiant/appearance/headshot/etc. shouldn't be 'hot-dropped' if changed during the round. Anything related specifically to the character is locked for that life and the settings in this menu shouldn't be getting checked until the character is respawned (i.e. dead to the underworld, not revived), things like keybinds should be able to take effect immediately. This is the case with almost every SS13 codebase I've played on, and should be the case here as well.

Cyprex:
Yes, that's how it **should** be. I simply don't trust contributors to have actually done this properly!

Koteshkin:
Im not sure if this PR cause this, but nothing else seems to cause that from logic. 
Lobby screen with class readied now stucks on the screen roundstart and doesnt close after you move as it was before.

Cleric2145:
> Im not sure if this PR cause this, but nothing else seems to cause that from logic. Lobby screen with class readied now stucks on the screen roundstart and doesnt close after you move as it was before.

I've experienced this at least once. It may well be related, as I was able to fix it by clicking the new Prefs option to open the char window and then closing it.

Gyrandola:
Not really sure what to do here. It seems to work perfectly fine.

babamandaa:
does this mean you can turn defiant on and off during gameplay? i'm honestly okay with that if that's the case.

Gyrandola:
> does this mean you can turn defiant on and off during gameplay? i'm honestly okay with that if that's the case.

No. That would be abusable.
The preferences save for next round/next life

Koteshkin:
It still causes roundstart ready menu to dont close after round has started.

Gyrandola:
I cant replicate it at all

> It still causes roundstart ready menu to dont close after round has started.

Nakalon (approved):
tried and tested

PR #967, opened 2024-09-11
Fixes Severe Roundstart Delay

Gyrandola:
## About The Pull Request

> so apparently when the lord spawns in, it fucking loops through the ENTIRE WORLD to FIND the LORD and assign them as the LORD for the ROUND.
> 
> tahts stupid as fuck. it assigns the lord directly now

### DIFF
--- a/code/controllers/subsystem/ticker.dm
+++ b/code/controllers/subsystem/ticker.dm
@@ -538,14 +538,6 @@ SUBSYSTEM_DEF(ticker)
 			player.new_player_panel()
 		CHECK_TICK
 
-/datum/controller/subsystem/ticker/proc/select_ruler()
-	for(var/mob/living/carbon/human/K in world)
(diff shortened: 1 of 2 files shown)


Cyprex:
testmerged

TuskTusk:
Anything that helps with delays like this is great. My question would be, why did they code it in like that, was it to prevent some rare fringe-scenario?

Cyprex:
Probably because they thought it'd only loop over humans in world instead of everyone. I don't think there's anything smart or deliberate here.

BlueMemesauce:
> Anything that helps with delays like this is great. My question would be, why did they code it in like that, was it to prevent some rare fringe-scenario?

Originally there was a feature that the Queen would become ruler if there was no King at roundstart. I removed that in https://github.com/Blackstone-SS13/BLACKSTONE/pull/689/
image

PR #969, opened 2024-09-12
Removes gibbing from the gate. Makes it chop off limbs.

GuppyLaxx:
Gate now chops off random limbs and crushes the sh*t out of people instead of just straight up gibbing them.
Can decapitate and spill out their organs.


K4rlox:
good, i was sick of bailiffs being thirsty to gib people using the gate

radarseas:
blessed PR, unrevivable circumstances should be reserved to divine punishment, antagonists, the lava mines because lava just makes sense, removable means or difficult to pull measures.

Especally when we have the wolf pit for executing people and gibbing them with a chance to fight or live.

PR #971, opened 2024-09-12
REMOVES THE ZIZO ZIZZLE!!!!!!!!!!

GuppyLaxx:
This has bothered me since the DAY F***ING ONE.
Christ what a dumb, simple fix.

### DIFF
--- a/code/datums/wounds/types/special.dm
+++ b/code/datums/wounds/types/special.dm
@@ -50,6 +50,7 @@
 	can_sew = FALSE
 	can_cauterize = FALSE
 	critical = TRUE
+	var/do_blinding = TRUE
 
 /datum/wound/facial/eyes/can_apply_to_mob(mob/living/affected)
 	. = ..()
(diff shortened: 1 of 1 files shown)


Hmeister-real:
it's iconic...

code/datums/wounds/types/special.dm
```diff
@@ -94,6 +94,11 @@
 /datum/wound/facial/eyes/right/permanent
 	whp = null
 	woundpain = 0
+	sound_effect = null
+
+/datum/wound/facial/eyes/right/permanent/on_mob_gain(mob/living/affected)
+	if(mob_overlay)
+		affected.update_damage_overlays()
```
Cyprex, line 101: Shouldn't this change be added to a parent type?

Cyprex: This looks vaguely like it should be handled on /datum/wound or something, I'm not sure what makes sense here!

radarseas:
Zízo

ChungusGamer666:
soul removal

PR #972, opened 2024-09-12
REALLY horrible and stupid and insane fix for the area intro lag

GuppyLaxx:
This PR makes it so the area intro sound plays extremely quietly upon roundstart to pre-load it and HOPEFULLY avoid any lag later in the round

It's a really, really fucking bad and horrible band-aid fix.


pilotmans:
I mean it works but that doesn't change the fact that this is the equivalent of using duct tape to fix a helicopters broken rotor blades

PR #973, opened 2024-09-12
Removes Excidium FROM THE MAP

GuppyLaxx:
Just removes the stupid bounty machines that destroy heads from the map. Everything else is in-tact

### DIFF
--- a/_maps/map_files/rockhill/rockhill.dmm
+++ b/_maps/map_files/rockhill/rockhill.dmm
@@ -10152,10 +10152,6 @@
 /obj/structure/rack/rogue/shelf,
 /turf/open/floor/rogue/hexstone,
 /area/rogue/indoors)
-"lji" = (
-/obj/structure/roguemachine/bounty,
-/turf/open/floor/rogue/cobblerock,
-/area/rogue/outdoors/rtfield)
 "ljm" = (
 /obj/structure/mineral_door/wood/donjon/stone{
 	locked = 1;
@@ -10770,7 +10766,6 @@
 	dir = 1
 	},
 /obj/effect/decal/cleanable/blood/old,
-/obj/structure/roguemachine/bounty,
 /turf/open/floor/rogue/blocks,
 /area/rogue/outdoors/town)
 "lSf" = (
@@ -284635,7 +284630,7 @@ wQY
 qvR
 qvR
 wSn
-lji
+fxx
 fxx
 fxx
 rZt


Heheniy:
makes hunted trair useless

Pascal125:
Just make it so it doesn't accept heads that do not have bounties 4head.

radarseas:
I think they're fine honestly, this is a bit too hugboxxy, the entire idea of having bounties to MAKE sure that people stay dead brings some conflict where antagonists wouldn't otherwise.

Social deception and paranoia being at play leaves a much, darker setting, also you can just check who set a bounty anyway and outlaw them if its for a stupid reason that breaks the law.

Why not instead just make it so the owner can cancel the bounty as well as the current ruler and add one to the keep instead? If the king allowed such machines they should have sway over them.

SixplyDev:
Why would we ever remove these things? They're great for actually redeeming bounties and crap from people and allow for retribution without directly going and killing people

SayonaraEcho:
Erm... strange PR. Maybe a joke. I don't believe that PR-maker is telling us about bounties when there's in-game hunted trait and mercs jobs.

Koteshkin:
Oh no, my fragile snowflake with rich inner world and complex drama lore nobody cares about just died and beheaded and its head fed to the Excidium, so now I cant be ressurected or would be forced to go with ugly skeleton head, that breaks my vulnerable  soul. Time to remove Excidium.

Bruh, I hope it will never be merged.

pilotmans:
Why is this even a PR? Just make it not destroy heads that aren't bountied

radarseas:
> Just make it not destroy heads that aren't bountied

if someone uses it, they **WANT** you gone, its good to have final death in a way, since decapitation isn't like the keep gate where you can just _click once and obliterate someone_ and probably not even be noticed as the one doing it. _If you're slashing a dead body at its throat, its pretty obvious you want them to stay dead and people will intervene._

Its significantly less scummy than just de-organing the head with a blade so they're blind and brainless, since this is pretty loud, obvious and _in a publically noticable place._

wow50000:
ffffucking finally

Elli-Skala:
The main issue with the bounties right now is there isn't a way to get rid of them. If jhon soilson puts a 1000 bounty on the queen the queen can't ever pay it off its always sitting there forever

Alanii:
Before removing, actually get the replacement in the game first.

Removing and then 'eventually' getting around to its replacement is not a good idea.

Just tells me no one plans to actually replace it with something and would rather remove a feature of the game.

PR #974, opened 2024-09-12
Physician fixes/buff?

ChuChu42:
## About The Pull Request
image
adds: organ chest with 1 of every organ, but 2 eyes and 2 tounges
adds: leech chest with 10 leeches for potionmaking
adds: SCOM+Vomitorium
adds: dedicated surgery bed, instead of the physician's own bed, this one looks more sterile
adds:a chest with 2 forceps and 1 speculum in the case physician forgets his tools inside someone or gets called to the court mid surgery

makes the physician farm 2 tiles wider so more crops may be grown more effectively

also physician now gets a key to the dungeon surgery/torture room, not the dungeon itself

## Why It's Good For The Game

Gives physician more depth

physician can now unlock the torture/surgery room as intended

theres now a more sterile bed for the physician to do surgery on, not on his own bed

chest of organs, only 1 per however, still, useful and makes sense for the royal surgeon to have some spares, including 1 of every wooden prosthetic
chest of leeches, self explanatory
slight redesign of the room

Scom and vomitorium, so the physician can now buy alchemy supply, despite leeches and potion supply not being in vomitorium yet

Mail machine, the number is ALWAYS going to be 12

### DIFF
--- a/_maps/map_files/rockhill/rockhill.dmm
+++ b/_maps/map_files/rockhill/rockhill.dmm
@@ -889,7 +889,14 @@
 /area/rogue/under/town/basement)
 "aZj" = (
 /obj/structure/table/vtable,
-/obj/item/clothing/mask/cigarette/pipe/westman,
+/obj/item/reagent_containers/glass/bottle/rogue,
+/obj/item/reagent_containers/glass/bottle/rogue{
+	pixel_x = 5
+	},
+/obj/item/reagent_containers/glass/bottle/rogue,
+/obj/item/reagent_containers/glass/bottle/rogue{
+	pixel_x = -5
+	},
 /turf/open/floor/rogue/church,
 /area/rogue/indoors/town/manor)
 "aZG" = (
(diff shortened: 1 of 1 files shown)


K4rlox:
> why not change the objects the cautery can be used on?

already resolved dw

K4rlox:
> The cautery is able to be used on any source of fire though?

already handled 😭

Heheniy:
why chests locked to armory id and not physician

K4rlox:
> why chests locked to armory id and not physician

physician can unlock/lock them

i just copied the plant chest

_maps/map_files/rockhill/rockhill.dmm
```diff
@@ -4041,6 +4055,24 @@
 /obj/structure/mineral_door/swing_door,
 /turf/open/floor/rogue/dirt/road,
 /area/rogue/indoors)
+"eqT" = (
+/obj/structure/closet/crate/chest{
+	lockid = "keep_armory"
```
Cyprex, line 4060: I don#t think the physicia ncould open that?

_maps/map_files/rockhill/rockhill.dmm
```diff
@@ -6261,16 +6293,12 @@
...
-/obj/item/paper,
-/obj/item/paper,
-/obj/item/paper,
-/obj/item/paper,
-/obj/item/paper,
-/obj/item/paper,
-/obj/item/paper,
-/obj/item/paper,
+/obj/structure/closet/crate/chest{
+	lockid = "keep_armory"
```
Cyprex, line 6297: same as above

K4rlox: whats your discord, ill ping you when its fixed

K4rlox: resolved but needs re-review

PR #976, opened 2024-09-12
Changes the satchel slots to 4

john-pawson:
## About The Pull Request

Changes the satchels slots to 5, and it's maximum weight to 14 (Max 4 normal items and 1 small)

## Why It's Good For The Game

The satchel (currently) has 3 slots and 21 max weight, it's clearly intended to be a more versatile and lighter version of the backpack, but it usually fails at that since it's low capacity. It even has a max weight of 21, which is 7 normal items max.
This hopefully will make carrying things from one place to the other less of a trouble without a backpack, and help with making carrying more than three things slightly less annoying.

### DIFF
--- a/code/modules/clothing/rogueclothes/storage.dm
+++ b/code/modules/clothing/rogueclothes/storage.dm
@@ -215,9 +215,9 @@
 	. = ..()
 	var/datum/component/storage/STR = GetComponent(/datum/component/storage)
 	if(STR)
-		STR.max_combined_w_class = 14
+		STR.max_combined_w_class = 18
 		STR.max_w_class = WEIGHT_CLASS_NORMAL
-		STR.max_items = 5
+		STR.max_items = 4
 
 /obj/item/storage/backpack/rogue/attack_right(mob/user)
 	var/datum/component/storage/CP = GetComponent(/datum/component/storage)


Coolgat3:
I disagree tbh. The fun of having only 6 easily accesible slots is making you have to think about what you carry. As a bandit player this will buff me massively, I will be able to carry needle, cloth, red, moondust, mantrap, chain, and still have 2 free, easy access slots. A bit meh if you ask me. God forbid 2 satchel gamers having half of a backpack or i think more on their shoulders..

john-pawson:
> I disagree tbh. The fun of having only 6 easily accesible slots is making you have to think about what you carry. As a bandit player this will buff me massively, I will be able to carry needle, cloth, red, moondust, mantrap, chain, crossbow, which is my entire prefered kit that carried me a bunch of times and still have 2 free, easy access slots. A bit meh if you ask me. God forbid 2 satchel gamers having half of a backpack or i think more on their shoulders..

Well you've got to understand that's kind of the point, the satchel has a ridiculous amount of storage and for it to take an entire back slot. It should at least let you carry more than a pouch usually does, if you're going double satchel that means that you have no back slots, which means no weapons in there, no backpack, no shields, no space for a decent weapon if you're already carrying a quiver and a crossbow (Crossbow doesn't fit on satchels btw, as it is bulky weight) if anything it's rewarding them for giving up both of their back slots with decent capacity at the penalty of not being able to carry a weapon, which if you're going for a storagemaxxing build then fuck yeah they should get more slots.

If you had to choose between being able to carry 6 items by giving **both** of your back slots up, but they're easy to access.
Or have 14 slots by giving up a single back slot which is slightly harder to access (depends mostly on the speed of the player).
Which one would you choose? 
Better yet, pouches exist, which means you have the option to just wear a pouch around your belt and neck and **not** have to give up any back slots for almost the same carrying capacity!

Bandits **should** get that storage because it means that now you get to carry more than 6 things if you'd like. It also means that fights are going to be more diverse since people will now have the option to carry more than 3 things around to use on a fight with a satchel.

This, i should also point out, applies to everyone else! Now everyone can afford to carry an extra bottle of red, or a mantrap, or some cloth, or an extra needle, or some moondust around without worrying of not being able to pickup something just because "satchel too small, too bad, next time get a backpack"
Now you can buy more than three things to from the merchant without having to worry of buying a pouch to make it fit.
Now surgeons can go ahead and have a satchel with their 5 tools for easy access instead of having to take out the backpack every single time.
Now hunters can take most of the animal they killed without having to worry about sorting it between the pouch, the cloak, the belt, the satchel, and maybe having to carry one or two hides on their hands just because they chose a satchel over a backpack.
Now you can go ahead and pick up that **one** extra piece of equipment before going to an advenutre because you have valued the quick access of the satchel instead of a backpack.

You're already sacrificing two inventory slots for weapons or other things for what? About the same carrying capacity as a pouch? Six slots?? This makes it usable, it makes it an option, it makes _not_ almost objectively worse than a backpack.

Red-Honk:
John I love the idea, I even up voted, but I think you miss understand the satchel, you can quite literally just put things in a pouch and then put that same pouch in the satchel, it'll take more time to pull the item out, but you'll fit more, also, you'll have to pull the pouch out to put things in it, it's made so, it's more cumbersome to use, hence unable to exploit that extra space that easily, so, perhaps keep that in mind while developing possible updates on it.

john-pawson:
> John I love the idea, I even up voted, but I think you miss understand the satchel, you can quite literally just put things in a pouch and then put that same pouch in the satchel, it'll take more time to pull the item out, but you'll fit more, also, you'll have to pull the pouch out to put things in it, it's made so, it's more cumbersome to use, hence unable to exploit that extra space that easily, so, perhaps keep that in mind while developing possible updates on it.

If you mean to tell me that you can put pouches inside the satchel to gain more space then i've got to tell you that yes, i did keep the pouches in mind, this was already something that happened with satchels before and now you can only fit a single more pouch on the satchel since the weight of a pouch is 3 and the new max of the satchel is 14, leaving you with a total of 14 slots if you put 4 pouches inside of a satchel (Same as a backpack, would you look at that), or just 10 if you decie to put normal items inside the pouches, since pouches only fit 2 normal items or 3 small ones, so the amount of space won from pouches inside satchels isn't that different from satchels as they were before.

If anything if you filled a satchel with pouches it'd end up being a worse backpack, since now you have to take out a pouch to put things or take things out of it, and then there's the tedium of having to open every single pouch just to check what's inside, not being able to differentiate one pouch from another, etc.

There's also the fact that you still have to go ahead and get 3-4 pouches if you want to go ahead and actually do use that extra space. Or 6-8 if you're trying to go ahead and double satchel it.

By putting pouches inside of the satchel you're sacrificing the easy access of the satchel for more space, which i think is intended, and completely normal for someone to do. This even applies to the backpack!

Am i misunderstanding this and is it something else that i'm misunderstanding about the satchel, or did you mean that?

PR #981, opened 2024-09-13
Add Tracy .so + enabling define

Cyprex:
## About The Pull Request
My power level has reached new orbits.

Anyways DO NOT MERGE THIS. This is for testmerging only, and will enable Tracy. Be mindful of the disk usage, I estimate around 100 GB/round. They will be put into data/profiler/(lotsofnumbers).utracy
You'll need to run [rtracy](https://github.com/Dimach/rtracy) against these, and then connect with tracy directly or run capture.exe(from tracy 0.11.0) against these. But frankly nobody but me needs to know that I just wanted to put it here.

Be mindful of the disk usage this will create. Also, this might not work because the .so was compiled on another debian system(Server is also debian, but could be an older version.


Cyprex:
Yes, it's a BYOND feature. I am, however, rather curious how much time we spend on maptick- but tracy should show that. There are other ways to find out, byond reports this in: https://www.byond.com/docs/ref/#/world/var/map_cpu .There's a bunch of code attached to this in modern tg but we might not have any of that.

Cyprex:
@Moribun Honestly I feel like we can just leave this open for convenience- To be clear when this is testmerged it will generate a massive file each round(~100 GB), and if that's not dealt with the server can quickly run out of disk space, so this should only be merged as needed.

PR #984, opened 2024-09-14
Chasuble and Stole are priestess compatible

Evie7056:
## About The Pull Request

Priest additional cloak clothes are lacking any female sprites which is making them appear as invisible onmob. This is PR is kinda fixes that by giving this clothing pieces the same as male sprites. That serves as a band-aid fix until someone fill look at chasuble as it looks 'meh' on fem statics by being a little too chunky.

image
image
In game screenshots for chasuble and stole on elf character.

Keep in mind that the same problem will _probably_ occur for dwarfs and any onther small species as well, but they are out of the scope of this PR as this is simple use of male sprites on females. 

## Why It's Good For The Game

Invisible sprites are bad :(


Cyprex:
The way this was originally done is so lazy, can't other races just kill the priest and put on his outfit?

Evie7056:
> The way this was originally done is so lazy, can't other races just kill the priest and put on his outfit?

image
image

Well tbf the only small race that _can_ be possibly a priest is dwarf, so sprite issue isn't really an excuse to make priest's stuff so snowflakey exclusive by **one single person** in the round.
I don't really think removing restriction for the robes gives much for the round. I don't think King can make a fully functional priest either.

PR #991, opened 2024-09-14
Reinforced armor diversity.

Red-Honk:
# **REINFORCED ARMOR DIVERSITY!**

## About The Pull Request

The pull aims to change varies armor pieces to have a purpose, they are as follows: (Unless you wish to read the code)

Breastplate - +1 Iron - INTEGRITY: 250 - Chest/stomach protection. (200 -> 250; Because it lacks groin protection and being costly yet only being as good as a chainmail shirt and even being beaten by studded leather which is 300.)

Bascinet Helmet - INTEGRITY: 200 - Face/skull/ear protection, 180 degrees of vision. (200 - > 235 due to the vision penalty could've chosen to remove the vision option but that'll make it like any other helmet.)

Brayette - INTEGRITY: 325 - Groin protection. (Same cost as chain chausses, and only covers the groin, so, why not make it good at it to protect the family jewels?; 200 -> 325.)

Bucket helmet +1 Steel - INTEGRITY: 435 - whole-head. (400 -> 435 If you can't flip it for vision, you may as well make it tough.)

Full-Plate Armour - +3 Steel - INTEGRITY: 520 - Torso/arm/leg protection - Heavy armor. (500 -> 520 a tickle better for the cost.)

Hauberk - +1 Steel - INTEGRITY: 220 - Chest/stomach/groin/arm/leg protection. (200 -> 220 a tickle better for the cost.)

Horned Cap - INTEGRITY: 225 - Face/skull protection. (200 -> 225 same cost, less protection it requires some kind of upside)
Winged Cap - INTEGRITY: 225 - Face/skull protection. (200 -> 225 same cost, less protection it requires some kind of upside)

Sallet visored - +1 Steel - INTEGRITY: 275 - Flipped whole-head except mouth and go down to 180 degree vision, unflipped normal vision however eye/nose exposed. (200 -> 275 if it's gonna cost more but protect way less compared to a regular helmet, it may as well be tougher, even if it doesn't reach the heights of a Knight/Pigmask, why isn't it 300?; Because it only blinds your backside hence why it was balanced with cost and vision in mind.)

## Why It's Good For The Game

It gives more purpose to helmet that could be viewed as straight downgrades: more exposed places, less vision, extra ingots that don't change any integrity, this aims to carve those issues and give some armor pieces some kind of purpose, be it by integrity or being able to sleep in them, most chain mail armor pieces can be slept in (except coif), some helmets are able to take an extra hit or two besides looking dashing and having a clear weakspot without coifs, and what used to be downgrades such as the Bucket that didn't have a visor to flip for vision was given extra durability, this PR aims to bring more choice diversity in armor instead of going for the optimal choice every time.

[TL;DR: Armor changes, some slightly better due to weaknesses with no upsides while same cost, now those have purpose, also new descriptions on some items.]

### DIFF
--- a/code/modules/clothing/rogueclothes/pants.dm
+++ b/code/modules/clothing/rogueclothes/pants.dm
@@ -143,7 +143,7 @@
 	gender = PLURAL
 	icon_state = "chain_bootyshorts"
 	item_state = "chain_bootyshorts"
-	max_integrity = 325 //I want to give it 325 INTEGRITY hope this doesn't break it.
+	max_integrity = 325
 	sewrepair = FALSE
 	armor = list("blunt" = 70, "slash" = 100, "stab" = 80, "bullet" = 50, "laser" = 0,"energy" = 0, "bomb" = 0, "bio" = 0, "rad" = 0, "fire" = 0, "acid" = 0)
 	prevent_crits = list(BCLASS_CUT, BCLASS_STAB, BCLASS_CHOP, BCLASS_BLUNT)


DorfGuy1:
We should really remove the no sleep in armor thing. It just means your not gonna be able to sleep off firedamage well. Which seems to be more damaging lately. Are you going to make the helms fireproof?

PR #992, opened 2024-09-15
Travel tiles only visible to trait holders

GuppyLaxx:
This PR makes travel tiles only visible to people who can go through them/have gone through them.
Also changes a bit of the logic behind spotting people to learn the path and removes the dead end text. 

Affects vampire lords and bandit travel tiles.


radarseas:
speedmerge, blessed PR

PR #994, opened 2024-09-15
WIP thievery changes (DO NOT MERGE OR TM)

Hmeister-real:
## About The Pull Request

Removes previous thievery changes (only stealing from behind, combat mode nullifying stealing, pickpocket delay) and replaces it with a weight system that calculates the chance to steal on the weight of the thing being stolen (e.g. stealing a pouch is easier than stealing a sword off someones belt). Planning on trying to add a check for holding a knife while stealing a bag, but nothing coded yet.

## Why It's Good For The Game

good, but WIP and last tested didn't work.

### DIFF
--- a/code/_onclick/other_mobs.dm
+++ b/code/_onclick/other_mobs.dm
@@ -387,20 +387,13 @@
 					to_chat(src, span_notice("I try to steal from [V]..."))
 					if(do_after(src, 5, target = V, progress = 0))
 						if(stealroll > targetperception)
-						//TODO add exp here
-							// RATWOOD MODULAR START
-							if(V.cmode)
-								to_chat(src, "<span class='warning'>[V] is alert. I can't pickpocket them like this.</span>")
-								return
-							// RATWOOD MODULAR END
 							if(U.get_active_held_item())
 								to_chat(src, span_warning("I can't pickpocket while my hand is full!"))
 								return
 							if(!(zone_selected in stealablezones))
 								to_chat(src, span_warning("What am I going to steal from there?"))
 								return
-							mobsbehind |= cone(V, list(turn(V.dir, 180)), list(src))
-							if(mobsbehind.Find(src))
+							if(mobsbehind.Find(src) || V.eyesclosed)
 								switch(U.zone_selected)
 									if("chest")
 										if (V.get_item_by_slot(SLOT_BACK_L))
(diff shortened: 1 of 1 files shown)


CalamaBanana:
By rolling back those changes this way, Rogues will once more have a 0% chance of ever alerting their victim, even on failure. There would be 0 risk to casually click people while running past, fishing for success. Not like the victim would ever know. The victim needs to be alerted when the thief can not overcome the victim's perception. 

You are also only sending the failure messages from heavy objects into the log, not the victims chat.

Hmeister-real:
> By rolling back those changes this way, Rogues will once more have a 0% chance of ever alerting their victim, even on failure. There would be 0 risk to casually click people while running past, fishing for success. Not like the victim would ever know. The victim needs to be alerted when the thief can not overcome the victim's perception.
You are meant to be weary of people and watch your items, if a thief steals from you and you don't notice thats the whole idea, you're meant to be careful around the streets and be careful around strangers
 
> You are also only sending the failure messages from heavy objects into the log, not the victims chat.
intended? thieves should be able to tell if they are able to steal something without being punished for it, either ways it's a 50/50 chance to steal from back l and r, and belt l and r so it would mean a 50% chance to fail if the object you randomly chose is too heavy

CalamaBanana:
> You are meant to be weary of people and watch your items, if a thief steals from you and you don't notice thats the whole idea, you're meant to be careful around the streets and be careful around strangers

In your code there is 0 risk for Rogues pressing the pickpocket button. Address that point please, why do you want Rogues to never even show the "Xy attempted to pickpocket me" on a failure? You can walk up and down the main street clicking everyone till you get a success and sprint away. That does not make for engaging RP for the victims.

What do I do if I notice my item stolen? The main thief class is also the main murder-hobo class skill/stat wise. You can not challenge Rogues as towner in the slightest. I'd have to find more than one guard, as a single one would just be murdered as well. And by that time the Rogue is nowhere to be found. You can't be careful around strangers either, as they just have to sprint down main street and click everyone they see half a second before popping up on my screen.

The main pickpockets were supposed to beggars, maids and shophands. Classes that can be punished for it.

With your changes, the only way to protect oneself against pickpockets boils right back down to "don't RP in public in town."

Hmeister-real:
no longer working on this

PR #996, opened 2024-09-15
[TEST-MERGE] Zombies are once again player-controlled

Gyrandola:
## About The Pull Request

Zombies were made NPCs on Blackstone due to constant griefing.

I think our higher standards allow us to give players more freedom and punish the griefers, not like a player can be any more aggressive than a npc zombie anyway.

This also lets them keep their gear when resurrecting, because why would they be dropping it.

 It's probably worth a TM to see how it goes.

### DIFF
--- a/code/controllers/subsystem/rogue/roguemachine.dm
+++ b/code/controllers/subsystem/rogue/roguemachine.dm
@@ -51,7 +51,7 @@ PROCESSING_SUBSYSTEM_DEF(roguemachine)
 	var/static/list/safe_areas = typecacheof(list(\
 		/area/rogue/outdoors/town,\
 		/area/rogue/indoors/town,\
-		/area/rogue/under/town,\
+		/* /area/rogue/under/town,\ */  //to allow zombies to raise in the sewers
 		/area/rogue/under/town/basement,\
 		/area/rogue/under/town/caverogue,\
 	))


radarseas:
remember, tackles with zombies can be histerically deadly. NPCs only know punching... player zombies, oh boy...

then again, I think there should be a "you've played for at least a day or two" minimal requirement before turning into a player zed, so people can't drown themselves instantly to play zombie and also to give town/bandits a grace period before sentient zombies crop up. Unless its via bite infection.

Coolgat3:
Just ban people who suicide just to be zombie. Dont ban people who suicide to instantly go to spirit world though. Simple solution really. I fully support this it makes zombies dangerous again and lets us hear their insanely good combat music again

K4rlox:
im pretty sure blackstone also made it so zombies are FORCED to bite people near them automatically, even other zombies because people used to be friendly zombies

radarseas:
personally I'd do something like lesser vamps where you need to bite the living to stay in control of your zombified body or get ghosted, to encourage hunting down the living and going out like well, a zombie.

HissyNays:
WHAT A BASED PR. You sir have restored a great and fun part of the game. Thank you. Zombies are now actually dangerous like before.

katsukai:
Consider, removing climbing on trees, sprinting and add the decaying flesh trait because they do not drop their neck or head armor upon turning. Also enable custom emotes, because there is lot potential to be a spooky zombie with RP.

radarseas:
> Consider, removing climbing on trees, sprinting and add the decaying flesh trait because they do not drop their neck or head armor upon turning. Also enable custom emotes, because there is lot potential to be a spooky zombie with RP.

I think climbing is fine, however zombies should take longer to climb than the living so you can still out-climb them, just not easily escape and snipe them over and over, unless they have _expert or above climbing_, rogues should remain a fucking menace when they turn, _since turning a rogue isn't easy at all as most people decap them_ or they'd outdodge you.

radarseas:
zeds need a check for autobite so that they don't bite through gear they shouldn't, like say... a steel mask, a knight helmet _**AND**_ a bervor all at once.

PR #997, opened 2024-09-15
[TEST-MERGE] Reinforced Bog Gate

Gyrandola:
## About The Pull Request

Turns all the bog gate glass windows into proper stone ones. Not really worthy of a screenshot.

## Why It's Good For The Game

Makes the bog gate NOT a joke. It's meant to keep the baddies out, not a simple obstacle you can skip by jumping on the second floor. Hopefully it also makes people more willing to play bog guard.


Cyprex:
And then what if bogguards are absent on lowpop?

PR #998, opened 2024-09-16
Slight weapon and buckler changes

john-pawson:
## About The Pull Request

List of changes: 

1. Charging a flail smash now drains energy ro keep charged. With standard 10 END you should be able to keep a flail charged for ~10 seconds.
2. Combat knife is now able to stab.
3. Combat knife is now called combat knife instead of just knife.
4. Combat knife is now called combat knife on the smithing list instead of hunting knife.
5. Rapier cut now only deals 80% damage instead of 50% damage.
6. Buckler now properly uses armorsmithing when being crafted.
7. Buckler now needs 1 steel bar instead of 2.
8. Buckler now needs higher skill to craft since it is now cheaper on material.
9. Combat knife now needs 1 steel bar instead of 2.
10. Combat knife now needs higher skill to craft since it is now cheaper on material.
11. Added a missing / on the buckler code.
## Why It's Good For The Game

1. This is mostly prevent people from charging the flail smash for unlimited time. Should prevent a bit smash spamming (hopefully) now that it actually consumes stamina. Shouldn't change much for anyone that doesn't charge the hit more than they should.
2. It's a combat knife, of course it should be able to stab. It's based ot the "THIS is a knife" meme, and the only reason it didn't have a stab option (i think, correct me if i'm wrong) is because it's based off the cleaver code-wise, and a cleaver cannot stab.
3. Self explanatory.
4. Self explanatory.
5. It's a rapier, it's got an edge throughout most of the blade, it should still be able to cut decently, not as good as a regular sword, but it should still do some damage. Now it should deal 14 damage at standard strength (slightly less than a dagger's cut)
6. All other shields use armorsmithing, why wouldn't the buckler? Known issue, now fixed.
7. You mean to tell me that the buckler shield needs the same amount of material as a tower shield? It's a small shield used for skill over flat defense, it's already underused as it is, no need to make it unreasonably expensive on material.
8. Self explanatory.
9. Why would a knife half the size of a sword take twice as much steel to make? The weapon doesn't even get used that often probably because of the high cost and low damage.
10. Self explanatory.
11. This probably does nothing, but i just HAD to fix it. It haunted me. And i _hated_ it.

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/swords.dm
+++ b/code/game/objects/items/rogueweapons/melee/swords.dm
@@ -475,7 +475,7 @@
 
 /datum/intent/sword/cut/rapier
 	clickcd = 10
-	damfactor = 0.8
+	damfactor = 0.7
 
 /datum/intent/sword/thrust/rapier
 	clickcd = 8


Red-Honk:
5. Rapier cut did 0.5 damage for a reason, it's supposed to be bad at cutting, it's a rapier and great at stabbing with quick Quick r', it has a thin blade afterall, not to mention it's a balancing act since it has defense stat 9, in one-hand no less, that's its appeal.
6. The buckler is a combat shield, it works differently than the others, why wouldn't it? (Game most likely views it as a weapon.)
7. The buckler shield is better than a kiteshield if the user is very skilled in the weapon they hold in the other hand.
Yes, the combat knife was basically a reskin/coded version of a clever but better at damage.

Red-Honk:
5. Rapier cut did 0.5 damage for a reason, it's supposed to be bad at cutting, it's a rapier and great at stabbing with quick recovery, it has a thin blade after all, not to mention it's a balancing act since it has defense stat 9, in one-hand no less, that's its appeal.
6. The buckler is a combat shield, it works differently than the others, why wouldn't it? (Game most likely views it as a weapon.)
7. The buckler shield is better than a kite-shield if the user is very skilled in the weapon they hold in the other hand.
Yes, the combat knife was basically a reskin/coded version of a clever but better at damage.

john-pawson:
5. It still is bad at cutting with a 0.8 damage, as bad as cutting as a spear but with even less damage, this just makes it bad at cutting instead of EXTREMELY bad at cutting, the damage is still bad and it is still a mainly stabbing weapon, just that now you can cut and actually deal some damage instead of tickling your opponent.
6. It's a known issue, all shields are made with armorsmithing because it's metal designed to take blows and lessen the damage taken, a weaponsmith knows how to make a blade not dull while hitting something and make it comfortable to wield, an armorsmith knows how to make a plate not bend when getting hit and how to shape it to redirect the force.
7. I know, but it's still ridiculous that it has quarter of the size and the same material requirements, i already made it harder to smith so only a proper armorsmith or learned apprentice can make it, it's just that now the material cost makes sense.

Coolgat3:
I know it's irrelevant because sadly maniac is disabled but I always imagined he woul be stabbing people while wearing a hooded cloak and the knife not being able to stab is a bit wild i guess its cause its a version of the cleaver cool pr

Red-Honk:
John wins, flawless argument, Alright, aye, I'm convinced.

PR #999, opened 2024-09-16
Makes Rockhill Great Again!

Alanii:
## Why It's Good For The Game

Players have begun to make pickaxes and other crude items to bypass the Bog-Gate entirely by simply building around it. This alleviates the issue by forcing players to either confront the Bog-Guard when attempting to bypass the gate, or use a more undetectable method.

Effectively just makes it so when players want to bypass the Gate, they will do so in an obvious manner through mining directly to the left of the Bog Gate, mining to the rooftops of the Bog Gate house, or actually take effort as they need to utilize the bog mines that are under the bog to completely bypass the Gate undetected.

Additionally pulls PR #997

### DIFF
--- a/_maps/map_files/rockhill/rockhill.dmm
+++ b/_maps/map_files/rockhill/rockhill.dmm
@@ -15823,7 +15823,7 @@
 /turf/open/floor/rogue/ruinedwood/spiral,
 /area/rogue/indoors)
 "rrn" = (
-/turf/closed/wall/mineral/rogue/stone/window/blue_moss,
+/turf/closed/wall/mineral/rogue/stone/window/moss,
 /area/rogue/indoors)
 "rrA" = (
 /obj/effect/decal/cleanable/cobweb,
@@ -388370,7 +388370,7 @@ uLB
 uLB
 uLB
 qrn
-rrn
+rDt
 eFy
 uuE
 uuE
@@ -389174,7 +389174,7 @@ uLB
 uLB
 uLB
 pRP
-hHI
+pAV
 eFy
 uuE
 uuE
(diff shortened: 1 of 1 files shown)


radarseas:
also if you have decent swimming stat, just swim up the river, there's absolutely nothing they can do about it.

SixplyDev:
This PR should hopefully encourage more people to play Bog Guard and reinforce the security of the gate. As it stands, the gate is far too easy to break into/bypass its current state and this should hopefully fix that.

DorfGuy1:
Whenever i dig under there, I dig like, a huge twenty five block long autism tunnel. Nobody's ever noticed it when I did do it either.

DorfGuy1:
Though, make it more difficult, because when I play boggard recently, I basically never see the bandits until they've already dug into town.

Alanii:
> Whenever i dig under there, I dig like, a huge twenty five block long autism tunnel. Nobody's ever noticed it when I did do it either.

The idea is to incentivise people to cross the underground river and dig through the under-mines that are under the bog. 

Elsewise, they'll be digging the usualy place going to the roof of gatehouse or directly to the left of the gate house.

This PR is mostly to prevent the meme of "bandits sneaking into town by just digging around the gate" and the adventurers who fire ball the bog guard for demanding tolls and breaking in through the windows or digging a hole around like a gremlin.

People who want to bypass the Bog-Gate should do so in a more hidden method or very loud method. Not autism tunnel build on the Z levels directly right of the Bog-Gate.

It may seem "dumb" or "bad idea" but most players are unable to think methodically and just go "unga tunnel" because they pulled the map up on SDMM and know the optimal route of bypassing the gate.

PR #1002, opened 2024-09-16
Iron mace for war cleric

SylvinSeik:
## About The Pull Request

War cleric gets an iron mace instead of a steel one. Kind of like most other common adventurers.

## Why It's Good For The Game

I really hope I shouldn't have to explain why a miracle spamming adventurer class with EXPERT MACES shouldn't get a 40 damage 80 AP weapon on spawn, but regardless.  Also, the "everyone else gets iron weapons" PR went through so clearly iron weapons are the way to go here.


DorfGuy1:
I guess I always chose the life cleric, cause heavy healer, but I swear I chose it once or twice. I thought they got an iron mace just like the life one did.

PR #1005, opened 2024-09-16
Drug rebalance

SylvinSeik:
## About The Pull Request

Tries to mix up the dusts in the game to be a bit more nuanced and have use-cases.  Note that this PR DOES NOT TOUCH ALCHEMY RECIPES OR WHAT DUSTS/AMOUNT OF DUSTS ARE GIVEN TO BANDITS. NO VILLAIN NERF HERE, AND ALCHEMISTS CAN STILL PUSH THEIR STOCK.

Spice loses its 3 endurance buff but gains 5 intelligence buff. Now costs 20 from the merchant, or 20-35 from purity.  Meant for crafters, not mages, so theres still the big fortune debuff which may or may not inhibit them in any meaningful way.

Ozium speed malus from -99 (always sets you to 1) to -5 (usually halving your speed, or putting you at the movespeed minimum). Gives a +2 perception bonus, because if you're going to be stuck in one spot for awhile, you might as well notice the little things.  Still retains the NOPAIN trait.  Price is now 5 from merchant (btw, now the merchant can sell it), or 6-15 from purity.  It's intended to be the common street drug, used by lost souls to cope with the pain of living in Rockhill.

Moondust gives +4 spd/end and -3 per, as well as a 20 second crash debuff after its effects wear off, which reduce speed and end by 3.
Moondust_purest gives +5 spd/end for -2 perception, with the same crash debuff.

## Why It's Good For The Game

There is no use-case for spice. There is no use-case for ozium. When was the last time you even SAW ozium?  Why are these dusts so expensive?  Moondust has zero penalties, on top of being extremely cheap.  This PR addresses all these issues.

### DIFF
--- a/code/datums/status_effects/rogue/debuff.dm
+++ b/code/datums/status_effects/rogue/debuff.dm
@@ -50,7 +50,6 @@
 
 ////////////////////
 
-//Attempting a moondust crash effect
 /datum/status_effect/debuff/moondust_crash
 	id = "moondust_crash"
 	alert_type = /atom/movable/screen/alert/status_effect/debuff/moondust_crash


DorfGuy1:
can confirm, only time i saw ozium is when someone really wanted to rp being a drugggy. Also if it increases perception, but nerfs speed really badly, nobody is gonna use it, cause archers want to run away from people anyway.

aberrantQuesrist:
Love the Spice changes. Maybe Ozium should give +Endurance rather than +perception? I still don't see it being used much in that.

But man, that Moondust nerf would HURT. Regular moondust hardly lasts very long anyways, I don't know it's warranted. The difference between normal and pure moondust is the length of the effect whilst having the same. I'm not so sure I like that. 
If you must nerf it, maybe make it +4 speed, +3 endurance, -2 perception, and no doubling for purest?

SylvinSeik:
> But man, that Moondust nerf would HURT.

That's kind of the point.  Moondusts main appeal isn't the attribute boost, it's the other effect where it just reduces delays for you like your attack delays.  I'm not against discussion and tweaking on the numbers though. Everyone is invited to ping Triyarg on the discord to discuss them.

About ozium still not being used: I don't expect it to be great. It might see some actual use for its capabilities, but I expect it to be more used for the druggie flaw, and I'm okay with that.  Might give it bonus endurance instead of/in addition to the perception. Everything is up for debate.

Coolgat3:
Do not nerf moondust it lasts 30 seconds.

Jackofashepard:
Looks good. I'd just say half the change on moondust, a nice compromise

K4rlox:
Spice spawns in the court magician's basement, i would say spice IS a mage drug, NOT a crafter drug

if it was INTENDED FOR CRAFTING
court magician wouldnt get it

SylvinSeik:
Moondust stat buffs moved back to be more like how it was(save for keeping a perception malus), even though I didn't touch the halved attack cooldown shit anyways, w/e.
Moondust crash added, which is just a -3 spd/end after it wears off, for 20 seconds (compared to the 30 or 40 seconds of the moondust buff depending on what type you take).

DorfGuy1:
Moondust too expensive now.

PR #1008, opened 2024-09-17
Adds PSYDON to the all divine def

Jackofashepard:
## About The Pull Request

Seeing as my last PR was simply closed with 0 explanation I made it yet again. How about we actually look at what this PR does before closing yeah? 
ALL_DIVINE_PATRONS is applied to Knight, captain, gatemaster and bog master. 
There's more but they actually dont count because they get auto set to respective patrons such as priest force locked to astarta.

## Why It's Good For The Game

I think its cool having more roles be able to be psydonites. The king can literally be a psydonite but for some reason a good chunk of his most important retinue cannot. Not to mention the bog and gate masters as well.

### DIFF
--- a/code/__DEFINES/roguetown.dm
+++ b/code/__DEFINES/roguetown.dm
@@ -142,7 +142,7 @@
 
 #define ALL_ACOLYTE_PATRONS list(/datum/patron/divine/astrata, /datum/patron/divine/noc, /datum/patron/divine/dendor, /datum/patron/divine/pestra)
 
-#define ALL_DIVINE_PATRONS list(/datum/patron/divine/astrata, /datum/patron/divine/noc, /datum/patron/divine/dendor, /datum/patron/divine/abyssor, /datum/patron/divine/ravox, /datum/patron/divine/necra, /datum/patron/divine/xylix, /datum/patron/divine/pestra, /datum/patron/divine/malum, /datum/patron/divine/eora, /datum/patrol/old_god)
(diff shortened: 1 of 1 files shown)


Red-Honk:
I think that a better way around this is to just simply, allow those roles to also be able to be accepted, even if they're thinking of the old-ways of doing things, instead of. . . you know, literally making no sense by adding Psydon to a list that doesn't make sense for him.

john-pawson:
> > I think that a better way around this is to just simply, allow those roles to also be able to be accepted, even if they're thinking of the old-ways of doing things, instead of. . . you know, literally making no sense by adding Psydon to a list that doesn't make sense for him.
> 
> There's two ways. Simply removing the ALL DIVINE def since its not really used as is but this also allows inhumen in all these roles. Other option is... to manually make a list on each of these roles with psydon included. Its just easier to edit the DEF. Of course i could just rename it to something like ALL_NOT_INHUMEN but in that case ALL_DIVINE already fits

Well i think the main problem with adding psydon to the ALL_DIVINE def is that psydon isn't a divine god

There's a grand total of 7 roles that use the ALL_DIVINE def so manually adding psydon as an option like the inquisitor already does shouldn't be much of a problem, you could also go ahead and make a second def called ALL_NOT_INHUMEN too since it's not exactly very hard to do. Removing the ALL_DIVINE def or modifying it so it includes non-divine gods woudln't make any sense in general.

Jackofashepard:
Coling, need to fix the branch

PR #1012, opened 2024-09-18
New attempt at removing Bandit Seelies

TheFenix323:
Now it ask the client what race they are selecting in the lobby.

## About The Pull Request
Adds new logic to check if the player is chosing to be a Seelie wile rolling for bandit.

## Why It's Good For The Game
Fixes Seelies being able to spawn as Bandits.

### DIFF
--- a/code/game/gamemodes/game_mode.dm
+++ b/code/game/gamemodes/game_mode.dm
@@ -391,7 +391,7 @@
 			continue
 		if(is_total_antag_banned(player.ckey))
 			continue
-		if(role == ROLE_BANDIT && isseelie(player.client.prefs.pref_species.type))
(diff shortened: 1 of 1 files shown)


code/game/gamemodes/game_mode.dm
```diff
@@ -393,6 +391,8 @@
 			continue
 		if(is_total_antag_banned(player.ckey))
 			continue
+		if(role == ROLE_BANDIT && isseelie(player.client.prefs.pref_species.type))
```
Cyprex, line 394:
```suggestion
		if(role == ROLE_BANDIT && isseelie(player.client?.prefs?.pref_species?.type))
```
If we runtime here Antag selection will break- so we have to be defensive

TheFenix323: Got it

PR #1019, opened 2024-09-19
Removes random from bandit armor

Heheniy:
## About The Pull Request
Removes random from bandit armor spawn, giving them 100% gamberson, studded leather armor, gloves, leather bracers and chain coif

## Why It's Good For The Game
Bandits have no decent starting armor anyway and only 23% chance for any under-armor


radarseas:
yeah, honestly I dunno why they get worse armor than a gloom ranger for their body but absolutely spot-on-good decent guard-tier armor for their heads and necks.

Currently to kill a bandit, just stab them in the legs and they can't do anything about it really, meanwhile their head armor usually gets stripped long before its broken.

That or just cast lightning because lack of gloves = stun, into fireball to finish them with an insta-crit.

PR #1020, opened 2024-09-19
Snout mask 2 electronic boogalo

Heheniy:
## About The Pull Request

Additional to steel snout mask, iron one is added

pr was made based on the original before it was merged so changelog includes changes from there too


Heheniy:
> icon conflit sorry :((

resolved

PR #1021, opened 2024-09-19
Blademasters to arms! 

Red-Honk:
(Gives purposes to weapons, fixes typos/bugs, adds new verbs and a tiny bit more description.)

# Blademasters to arms!

## About The Pull Request
[Swords]
Change things called strikes to cuts because that's what they do. (Also makes it easier for staff to tell lethals from less-lethal hits)
Add an extra verb besides bashes and clubs to include strikes.
Iron messer was given 125 integrity.
Sabre/saber integrity was given 230 integrity due to the thicker blade.
Rapiers were given 215 integrity due to the thing being made for parrying.
The cutlass was given a purpose, its integrity has become 220 with the def of a rapier, now feeling more like a making it an inbetween of a sabre/saber and a rapier minus the swift balance and quick swinging recoveries.
Elvish saber has been put to 205 integrity instead of the normal sabre's 230 with the logic of, elven blades might be sharper but they're also thinner.

[Ammo]
Makes arrows not do 30 damage on cut/stab intent, instead doing 7.

[Special]: Adding more verbs.

[Polearms]: Fixing typos and adding more verbs.

[Knives]:
Fixed that stupid cut examine: desc = "A simple iron dagger favored as a fallback weapon for archers and crossbowmen. Just as likely to be in the hands on an assassin or rogue too."
added more verbs

[Blunt]: More verbs, extra desc' text.

## Why It's Good For The Game

Because more weapons would be given their own little purpose instead of just being same'ish or only worse in all aspects such as the cutless, this would no doubt provide more variety and for others not to feel bad because they aren't making the optimal choice every time.

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/blunt.dm
+++ b/code/game/objects/items/rogueweapons/melee/blunt.dm
@@ -103,7 +103,7 @@
 /datum/intent/mace/strike
 	name = "strike"
 	blade_class = BCLASS_BLUNT
-	attack_verb = list("strikes", "hits")
+	attack_verb = list("strikes", "whacks")
 	hitsound = list('sound/combat/hits/blunt/metalblunt (1).ogg', 'sound/combat/hits/blunt/metalblunt (2).ogg', 'sound/combat/hits/blunt/metalblunt (3).ogg')
 	chargetime = 0
 	penfactor = 15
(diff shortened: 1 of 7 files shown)


pilotmans:
thanks for adding this https://www.youtube.com/watch?v=vVOfCEyRIT4
really helps... i think?

PR #1023, opened 2024-09-19
Servant QoL buff

Heheniy:
## About The Pull Request
Gives servants 2lvl of crafting, 1lvl of crafting and fishing and also 1 speed (king want us to move FASTER)


CalamaBanana:
Please lets not normalize maids ripping up the lawn and training yard to start farms in the keep instead of interacting with the real farm. They are maids, not farmers. If people want to play farmers, they should pick the classes made for it. If the keep has no food and the real farmers are not farming, the steward can always import it for cheap.

If you give maids farming skill, it will be expected of them to farm. I rather not.

Keazra:
Honestly all we need is just one level of crafting. The farming is pretty easy to earn on your own. And please don't make fishermen even more useless than they currently are.

Frienderanian:
> Please lets not normalize maids ripping up the lawn and training yard to start farms in the keep instead of interacting with the real farm. They are maids, not farmers. If people want to play farmers, they should pick the classes made for it. If the keep has no food and the real farmers are not farming, the steward can always import it for cheap.
> 
> If you give maids farming skill, it will be expected of them to farm. I rather not.

Seriously: THIS. Please please PLEASE do NOT give maids farming skill. Maids should not be farming EVER. It should never be encouraged or required. The Maids should be cooking, cleaning and socializing, not farming. The only issue right now is that berries and wine are only accessible by farming, but that can be fixed by just adding (non-poison) berries to the stockpile or spawning the Keep with wine to begin with/having maids go buy wine from the Merchant. 
I am okay with giving them SOME crafting skill though, I am sick of losing buckets and not being able to make more lol even if it's super easy to get crafting 1.

Keazra:
Tending a small garden should be okay. Making an entire field, no.

SixplyDev:
The fact servants are effectively forced to farm and shit just to keep the place stocked is an issue, I would say no farming skills personally but other than that seems fine.

PR #1030, opened 2024-09-20
Craftable arming jackets + fem sprite for it

SylvinSeik:
## About The Pull Request

Arming jackets now have a sprite for females, because right now they just don't.
Arming jackets can now be crafted.  Identical cost to a gambeson, since it is completely identical to a gambeson stat/coverage wise.

## Why It's Good For The Game

Women not having sprites for a clothing they can wear makes them appear nude. That's bad.
Craftable arming jackets because why not, They are currently only available from from spawning as the Hand or Marshal, despite not actually being any special.  Can add a silk cost to the recipe if we want to keep it as a "higher end" piece.


AbuObrej:
I don't see any reason to actually add it as a craftable item since it really is just meant to be a unique looking gambeson for the marshal and hand.
Also, the south and north sprites are shifted dowards 1 pixel and the side sprites leave two pixel of the female sprite uncovered on the breasts.
image

For one last thing, you have to add sleves under armor_sleeves.dmi specific to females

SylvinSeik:
I toyed around with it to make sure it looked right in game, but I suppose I'm blind, I'll look into it, thank you.
Isn't the breast sticking out covered by the dambeson_f_boob, however?

SylvinSeik:
Added sleeves. Attached a poorly made image showing that the sprites do in fact cover all the skin.  Once more, I based it on the shape of the gambeson, so there SHOULDN'T have been any issues at all.
ARMINGJACKET

Cyprex:
Conflicts, good to merge after conflicts are resolved

PR #1033, opened 2024-09-21
Chain ropes, grab stam consumption and verb readability

Red-Honk:
# Chain ropes, grab stamina consumption and verb readability.

## About The Pull Request

[Verb for readability in combat]:
Removes the punch's strike verb. (To make it easier to tell from blunt attacks such as maces and the like)
Removes the shove's push verb. (You aren't pushing at that point if someone gets moved against their will)
The whip punishing intent has had its verb changed to punish-strikes instead of just strikes.

[Grabs]:
Releasedrain from 10 to 8.
Miss cost from grabs from 8 to 6.5.
Changed the "I must be above them" to "I must be on top of them."

[Ropes and chains]:
Rope's throw damages gets lowered from 5 to 3, however its throw range goes from 3 to 5.
Renames the second strike intent to lash and adds the slow cumbersome function of a lash with 2 range, 1.5 second charge up attack with no early release that deals 13 blunt damage with armor penetration of 15 and dulling bashchops and 3 seconds of recovery, using its new /datum/intent/whips/iron_chain mixed with whips and flails skills and now added 1 weapon defense. (This attack doesn't apply any effects, it just deals enough damage to slow down the opponent's movement, just like any other weapon, this might not work if the user is too weak to deal above 10 damage with it)

## Why It's Good For The Game

The easier it is to read the chatbox, the more one will know how to react to the given situation.

It addresses the issues by slightly reducing stamina drain from grapples that punished players too harshly compared to any other method was both easier to perform and safer to do so.

Gives diversity to ropes and chains, now both having a reason to exist, alongside making catching folk that flee a possible action, with the new description on the newly refined second strike "A rather slow ranged lash used for slowing down opponents for a tie up."

[Changes mid PR]:
punish-strikes verb changed to punishes instead.
Whips can no longer be put on the belt slot due to that interaction being very buggy.

### DIFF
--- a/code/game/objects/items/rogueitems/ropechainbola.dm
+++ b/code/game/objects/items/rogueitems/ropechainbola.dm
@@ -153,7 +153,6 @@
 /datum/intent/whips/iron_chain
 	name = "chain-lash"
 	desc = "A rather slow ranged lash used for slowing down opponents for a tie up."
-	force = 13
 	blade_class = BCLASS_BLUNT
 	attack_verb = list("chain-whips", "chain-lashes")
 	hitsound = list('sound/combat/hits/blunt/flailhit.ogg')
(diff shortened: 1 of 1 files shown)


Red-Honk:
> fix compilation errors

Fixed.

PR #1034, opened 2024-09-22
Fablefield just golaird wave as final step

EzekyalLegit:
## About The Pull Request
Makes a final step of the fablefield wave thats just the Goliard.

## Why It's Good For The Game
For the solo clowns in the audience.

EDIT: Lowered strength requirement for the repenta en since the inquisitor is often too feeble to carry his own whip.


EzekyalLegit:
Added lower strength requirement for repenta en on request since a large selection of inquisitors don't have the strength to use their own whip, particularly females.

PR #1035, opened 2024-09-22
Affordable Moondust Prices

AberrantQuerist:
## About The Pull Request
Changes Moondust prices to be more affordable again.

## Why It's Good For The Game
With the drug adjustment PR, moondust saw a host of nerfs with both stats and the added crash. On top of this, its price was raised to be comparable with Spice pre-rework, now the most expensive drug.
This makes it more economic to obtain whilst still leaving Ozium as the go-to for druggies. Moondust now has a downside, the price doesn't need to reflect its pre-rework quality.

### DIFF
--- a/code/modules/cargo/packsrogue/luxury.dm
+++ b/code/modules/cargo/packsrogue/luxury.dm
@@ -22,7 +22,7 @@
 
 /datum/supply_pack/rogue/luxury/moondust
 	name = "Moon Dust"
-	cost = 30
+	cost = 8
 	contains = list(/obj/item/reagent_containers/powder/moondust)
 
 /datum/supply_pack/rogue/luxury/spice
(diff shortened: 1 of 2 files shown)


aberrantQuesrist:
It now is set to 8 from the merchant, and ranging from 13 to 26 from Purity.

SylvinSeik:
Just a note: The price was raised so high due to it being (rather easily) craftable, while also being the best drug in the game.
It still IS the best drug in the game, and lowering its price so much to the point where alchemists are no longer even wanted is something I wanted to avoid, with the intention that you found someone specialized in such a trade to get it for cheap.

aberrantQuesrist:
> Just a note: The price was raised so high due to it being (rather easily) craftable, while also being the best drug in the game. It still IS the best drug in the game, and lowering its price so much to the point where alchemists are no longer even wanted is something I wanted to avoid, with the intention that you found someone specialized in such a trade to get it for cheap.

In all my time here I have seen an alchemist actually make moondust to sell ONCE, and I have seen no change in behaviour since after the price change. Even so, if an alchemist gets a setup and can produce it in bulk for cheaper, there's still incentive for a player to go to the alchemist rather than merchant. Maybe we could allow alchemists to create purest moondust if it's still a concern as to their worth?

K4rlox:
moondust is easy to massproduce by apotheocatry, farmers and physician

Unneeded
image

PR #1036, opened 2024-09-22
Allows shunned races in the position of Head Butler/Maid

AberrantQuerist:
## About The Pull Request
As stated in title, shunned races are now eligible for Head Butler/Maid position.

## Why It's Good For The Game
If shunned races are allowed to be Hand, why should not a position of ultimate servitude not be fit for them? This (aside from Hand) would be representation of the highest they might ever climb and the closest to respect they would ever be afforded.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/courtier/butler.dm
+++ b/code/modules/jobs/job_types/roguetown/courtier/butler.dm
@@ -7,7 +7,7 @@
 	total_positions = 1
 	spawn_positions = 1
 
-	allowed_races = RACES_TOLERATED_UP
+	allowed_races = RACES_SHUNNED_UP
 	allowed_ages = list(AGE_MIDDLEAGED, AGE_OLD)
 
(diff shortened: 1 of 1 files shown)


Jackofashepard:
I'd say it'd be better to just remove shunned races from hand. Which staff has just been too lazy to do

aberrantQuesrist:
> I'd say it'd be better to just remove shunned races from hand. Which staff has just been too lazy to do

I think being a servant that leads other servants offered as the 'best case' for shunned positions makes a lot of sense in-universe, personally.

SixplyDev:
Why...? This is just a horrid idea, the majordomo of a major household like the royal household would NOT be someone that's a shunned race. This is the man in charge of the wine cellar, the servants, a man that's _almost_ like family. Hell, this is the man that would've been referred to by his first name rather than his last name like all the other servants. If anything, it should be a respected race.

PR #1039, opened 2024-09-23
Miner rebalance

wow50000:
## About The Pull Request
Removes useless polearms 3 nerfed maces to 2, adds a lamptern for QOL and removes int negative and adds one endurance

## Why It's Good For The Game
Miner polearm 3 is retarded(they dont use it), having miners have to juggle a torch was retarded(literally why ?) -1 int was also retarded due to miners actually needing to have common sense to identify ores and stuff as well as being a towner they are not expected to be dumber than combat specialists such as warrior. Reduced maces to 2 solely because all towners are meant to have combat skills max out at 2. Gave them +1 end as a nice treat because all miner mains I know are based.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/pilgrim/miner.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/pilgrim/miner.dm
@@ -41,7 +41,7 @@
 		H.mind.adjust_skillrank(/datum/skill/craft/masonry, 1, TRUE)
 		H.mind.adjust_skillrank(/datum/skill/misc/medicine, 1, TRUE)
 		H.mind.adjust_skillrank(/datum/skill/labor/mining, 4, TRUE)
-		H.change_stat("strength", 3)
+		H.change_stat("strength", 3) 
 		H.change_stat("endurance", 2)
-		H.change_stat("constitution", 2)
+		H.change_stat("constitution", 2) //highest paying towner role as well as very fucking hard labour
 		H.change_stat("fortune", 2)


john-pawson:
I like the changes, but you've got to realize you can't give a role + 2 STR, + 2 END, +2 CON and +2 FRT with no stat debuffs, if you're going to be removing the INT and buffing the END, maybe lower their SPD or their CON since miners don't really use any of those two while mining or on their jobs overall.
Just to give an example of how high these stats are, the closest would be the sheriff or bogmaster with +3 STR, +2 END, +2 CON, and they're supposed to be some of the toughest around the town.

Heheniy:
pickaxe uses mace skill so it is justified for them to have higher skill of it

BladeburstNINJA:
As a miner main. Polearm nerf is fine. However, lowering maces is pretty rough. Not only, as was pointed out that pickaxes rely on mace skill now, but miners are probably the towner role who end up isolated the most. A miner going out to the lava mines is likely to encounter any assortment of bandits, fauna, vampires, werewolves, etc. Even town mine puts you away from any avenue of traffic, and any threat that occurs there can leave miners singled off and killed with no one ever finding their body.

wow50000:
> I like the changes, but you've got to realize you can't give a role + 2 STR, + 2 END, +2 CON and +2 FRT with no stat debuffs, if you're going to be removing the INT and buffing the END, maybe lower their SPD or their CON since miners don't really use any of those two while mining or on their jobs overall. Just to give an example of how high these stats are, the closest would be the sheriff or bogmaster with +3 STR, +2 END, +2 CON, and they're supposed to be some of the toughest around the town.

Miners are arguably the toughest towner role and can afford that, access to good food and lots of hard work would build up toughness and good stats, as for bogmaster he has also really fucking high skills, good starting equipment and armour proficiencies as well as the authority to use it. Keep in mind training any skill past 2 is already hard requiring you to learn from the veteran or a knight.



> pickaxe uses mace skill so it is justified for them to have higher skill of it

yes, miners have maces 2 which was established to be the high grade civilian level of maces, anything higher would be professional grade skill, I would not be opposed to a pickaxe buff but miners are not supposed to be a combat role



> As a miner main. Polearm nerf is fine. However, lowering maces is pretty rough. Not only, as was pointed out that pickaxes rely on mace skill now, but miners are probably the towner role who end up isolated the most. A miner going out to the lava mines is likely to encounter any assortment of bandits, fauna, vampires, werewolves, etc. Even town mine puts you away from any avenue of traffic, and any threat that occurs there can leave miners singled off and killed with no one ever finding their body.

If you are going to the lava mines, a place very dangerous, you should hire mercs and adventurers. The people whose entire job is fighting and protecting people for pay. As for being fragged in the town mines, that sounds like griefing.

john-pawson:
> Miners are arguably the toughest towner role and can afford that, access to good food and lots of hard work would build up toughness and good stats, as for bogmaster he has also really fucking high skills, good starting equipment and armour proficiencies as well as the authority to use it. Keep in mind training any skill past 2 is already hard requiring you to learn from the veteran or a knight. Miners should be using STR, END, SPD, and LCK to do their job and it makes sense for them to have con. STR to hit rocks less, END to hit rocks more, SPD to haul rocks and ores faster and LCK to get more ores and of course CON is because they don't have any armour proficiencies nor does it make sense for them to go around armoured. In fact I may buff their CON even more due to this fact.

What i meant to say is that you can't just give the miners the best stats just because they can "use it", by that logic we should give every other towner high CON too because they don't have armor proficiencies, and every towner high SPD too for the soilsons to deliver crops faster, woodworkers and carpenters to haul wood faster, blacksmiths to get ingots faster, and the townspeople to walk faster in general.

- The STR is gained by hitting rocks with a pickaxe all day and it's used and needed to mine efficiently, and it's alright for him to have high STR.

- The END is used since mining is exhausting work and it's built after days of restlesss works, and it's alright for him to have high END

- The FRT is used to have higher ore drops which is crucial, and it's alright for him to have high FRT.

- The SPD doesn't make sense. You're usually stuck in a single place hitting rocks with a pickaxe and the most you'll move is two steps ahead whenever you finish mining the rock you had in front of you, besides, "hauling ores" is more of a reason to have high STR not SPD. Something that trains SPD for example is the hunter, probably having to be fast to avoid animal attacks and be able to chase them when injured.

- The INT doesn't make sense, since if you trained them right probably a dog could be able to differintiate ores by just how they look, you could teach it to a kid if you wanted to, it doesn't take exactly that much brainpower. And mining isn't exactly a complex job so much so that the ones doing the job need to be intelligent in any way, if anything most people that have brawn over brains probably end up being miners.

- The CON isn't used in mining at all, yes i do get it that they're supposed to have a lot of food, but you can't just say "Yeah they eat a lot so they're definitely able to withstand a lot more damage" the soilsons probably have the best diet around considering that they have every type of food on their farm but you don't see them with 2 CON. And then what about all of the damage to their lungs and probably sickness that they get from the various fumes and dust that comes with mining normally leaving them relatively weak? What about the fact that they don't exactly do anything to get hurt and actually be able to train to withstand the damage?

What about the fact that you've quite litterally given them the exact same stats as one of the strongest roles which (from a balance standpoint) isn't exactly a good idea to make available to everyone? Yes the bogmaster might have higher skills but that doesn't mean you can make everyone as strong as him stat-wise just because he has higher skills and theoretically is stronger than you, what i meant is you're making a PILGRIM ROLE that anyone can choose as strong as a LEADER ROLE.
For example woodcutters could almost use the exact same stats because of the same reasons (excluding FRT), but you don't exactly see them walking with the best stats, they have 3 STR and some CON because 3 STR was already more than enough of a stat buff. They could use END? Yes. They could use SPD? Yes. But they don't get them.
On another note, the miners were already the strongest of all the pilgrims stat-wise, they don't need a buff.

PR #1040, opened 2024-09-23
Forgotten lupian ears

Heheniy:
## About The Pull Request
Gives access for lupians to 3 types of ears that were kinda forgotten about (1 literally is named "dog, long")


Coolgat3:
> they weren't given those ears because they're explicitly not dogs but i guess if you've got a 'tangled' or 'bastardized' pack ancestry it makes sense
> 
> just expect me to call you a mutt and a degenerate

like as if you'll notice the 3 pixel difference tbh

PR #760, opened 2024-08-09
Dungeons

zakanater19:
## About The Pull Request

Moves the gateway from inside the gate to the left of the bog village into the bog, it's far away from the bog gate

Currently it loads one of two dungeon map files, but they are the same right now, the loot spawns are randomized but in the same locations, at the end of the dungeon is where the good loot is, and it's full of hostile mobs

## Why It's Good For The Game

Gives something for players to do who want to go fight PVE battles and get loot (adventurer players)

Showing the mobs&drops while aghosting (this was filmed when it wasn't complete)

https://github.com/user-attachments/assets/a1ce3d4d-92a3-4679-b573-b339f6be4ec0

Me getting rekt trying to show off

https://github.com/user-attachments/assets/b1228997-5eb1-4e2d-a77d-83aef76e0e3f

And now showing gate location

Screenshot 2024-08-10 003427

And the dungeon so far, it's pretty basic but mappers and such can build upon it
Screenshot 2024-08-12 100612

PR #1042, opened 2024-09-23
AI NPC FIX - Dungeon Revert

Gyrandola:
## About The Pull Request

I don't know exactly why, but the entire PR breaks the AI completely. I tried ONLY reverting the zmanager changes and the fix was partial. So the entire PR must be reverted and whoever wants to add the Dungeon back must first figure out whats wrong with it.

BEFORE:
https://github.com/user-attachments/assets/7b70e4ba-f711-49fa-8d75-508a5205b72e

AFTER:
https://github.com/user-attachments/assets/56fb447e-3002-4049-a8fb-6396e219c3c6

The PR in question:
https://github.com/Rotwood-Vale/Ratwood-Keep/pull/760


radarseas:
I have a sneaky suspicion its breaking the AI _just because the place has so many active NPCs at once._

If we left the dungeon without any NPCs, _just as a test_, obviously not great for usual gameplay, I think it might just be that. Worth a testmerge PR over anyway.

If anything it probably needs spawners to place down NPCs as players go through it, _instead of having 40+ skeletons and godknows what else active at once._

HissyNays:
Honestly just remove dungeon. 90% of the time it only serves as a looting stage for bandits. It takes more from games than gives.

zakanater19:
yeah i also can't figure out why it busts the AI, just nuke it

radarseas:
> Honestly just remove dungeon. 90% of the time it only serves as a looting stage for bandits. It takes more from game than gives.

then again, that.

PR #1046, opened 2024-09-24
Blob -> Fleshmass, "new" antag

ikrarkjr:
## About The Pull Request

I have no idea what I'm doing please help me.

## Why It's Good For The Game

WE NEED ANTAG VARIETY PLEEEEEAAAAAAASSSSEEEEEEEE.


ikrarkjr:
> this is probably a lot more work than you might think.

I never said it would be done in the next months!

PR #1049, opened 2024-09-26
Marriage 2: Electric Boogalo

Heheniy:
## About The Pull Request

Allows same-sex marriage
It is still up to priest to decide marry those people or no

## Why It's Good For The Game
More possible interactions always good


JGlitch:
Could potentially initiate conflict and in-fighting, I like it.

Cyprex:
The original code for this is so crap, ugh.

PR #1052, opened 2024-09-26
No rot on table

Heheniy:
## About The Pull Request

Food will no longer rot while on table

## Why It's Good For The Game

No need to always hide your food in chest, now can actually serve feast hall without any fear that in a few minutes all your heavenly food will just spoil


aberrantQuesrist:
Thank you so much for this.

PR #1060, opened 2024-09-29
Ratwood is dead

ikrarkjr:
Play on Azure Peak


CalamaBanana:
Please do not change the soap recipe's difficulty to trivial. I intentionally put it that high to avoid gamers spamming them for combat use everywhere in the bog. Your craftdiff 1 means every character with 11 int can make it without even having to skillup.

ikrarkjr:
> Please do not change the soap recipe's difficulty to trivial. I intentionally put it that high to avoid gamers spamming them for combat use everywhere in the bog. Your craftdiff 1 means every character with 11 int can make it without even having to skillup.

Soap was nerfed to have a 20% chance to slip (unless that was reverted) and only costs 10 mammon at the merchant. Most pilgrims and adventurers can just afford soap spam with just bringing in some hides and fur, or mining a bit, or cutting wood a bit. 

If someone wants to do that, they can, but they have a vast majority chance of their soap doing nothing while the opponent attacks and kills them.

Issue #945, opened 2024-09-08
[BUG] Steel messers give iron in an ore furnance


pilotmans:
#round ID: Weed719-14.14.05

Problem: When a Steel Messer is smelted in an ore furnace, it smelts into an iron ingot instead of a steel one.

If i had to guess this is probably from re-used code from the iron messer.

Nakalon:
This can be closed!

PR #1065, opened 2024-10-01
Steel messer gives you steel when smelted

zoktiik:
## About The Pull Request
#945  I'm not putting in effort

## Why It's Good For The Game
oversight fix


radarseas:
it just makes sense.

PR #1068, opened 2024-10-01
Trays no longer spill open containers' reagents (#1007)

### LINKED ISSUE #1007: [BUG] Mugs placed on wooden trays lose their contents as if placed within an inventory
## Describe the bug

See title. Trays should not be spilling the drinks they contain, it's half the point of using them.

## Round ID:

N/A, didn't save it, sorry.

## Steps to Reproduce:

1. Fill up cups with any liquid. 
2. Place onto a wooden tray.
3. Wait a bit.
4. Empty the tray onto some surface and notice that all the cups are empty.

zoktiik:
## About The Pull Request
Fixes #1007.
Trays will uniquely refuse to spill any liquids from "`spillable`" reagent_containers, such as cups and bottles, that are put on them. For reference, liquid "spilling" (deleting 3 units of each present reagent) is generally supposed to occur:
- when putting an open r-container into a bag
- when grabbing a bag with open r-containers,
- when moving around with a bag that has open r-containers

Trays will now ignore all 3 of these cases. If you want to make certain things like throwing a tray cause it to drop & spill r-containers, be my guest. But that's out-of-scope for this PR.

## Why It's Good For The Game
Using trays for carrying around open beverage containers both makes sense, and is good for roleplay. The tray becomes substantially more interesting of a storage container with this PR.

### DIFF
--- a/code/datums/components/storage/storage.dm
+++ b/code/datums/components/storage/storage.dm
@@ -72,6 +72,9 @@
 	//Vrell - Used for repair bypass clicks
 	var/being_repaired = FALSE
 
+	var/spills_on_move = TRUE 					//Reagent containers inside this will not spill if the storage moves. (For trays.)
+	//roguespill.dm additionally handles spilling on a containing bag getting equipped - change the code if you want something else to ignore that behavior
+
 /datum/component/storage/Initialize(datum/component/storage/concrete/master)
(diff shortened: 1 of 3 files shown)


Keazra:
I want to love trays so much and this will finally make them great. Ty.

PR #1070, opened 2024-10-01
Ratwood is dead

ikrarkjr:
Play on Azure Peak


hugmeorelse:
I requested this. Should help make towner more interesting to play as.

PR #1073, opened 2024-10-03
Astratan Preist's Purging Flames: Solar Smite

Squird5:
I made a thing

## About The Pull Request

Priest now has a Not-Tier 5 spell "**Solar Smite**".

Solar Smite deals the most damage ~~during the day~~ inside the church, scaling down ~~at dusk and dawn and least at night~~ outside the church.
Solar Smite will gib undead, ~~except at night.~~ if they are inside the church.
Solar Smite requires a charge up to cast, costs 100 devotion and has a ~~60~~ 90 second cooldown.

## Why It's Good For The Game

Priests are a high priority target, but have little way to directly fight. Now the Priest can smite a single dangerous target semi-reliably, or at least criple one to buy time for the Templar to engage, or to escape. <sub>~~Or whiff their Smite entirely and get beheaded~~</sub>

https://github.com/user-attachments/assets/2108140f-c738-415b-9285-fd93cda25d33

https://github.com/user-attachments/assets/cf72304d-418a-46d9-9a0c-6d80082b1417

https://github.com/user-attachments/assets/df5f3e24-207e-4fd0-a625-bcacee3090aa


radarseas:
an interesting combat spell, although I'd weaken it slightly outside of the church, not too much of a fan of the priest potentally being able to use this offensively to effectively sniper-firebomb people wherever they can see with laser-deadset accuracy in comparison.

I'd still make it gib undead, is vampire lord protected though? I know they don't roll at all but I think gibbing a vampire lord off-screen during the day as funny as it is, might be a **little too much**.

Especally that like all miracles you could probably do this hidden behind a window, in a closet or in the shadows and just keep doing it. Where nobody can figure out where you are.

NikolayByondPlayer:
A bit to OP?

Heheniy:
one click to win is coal

SayonaraEcho:
Too OP, consider your PR because it's very powerful indeed even for the Priest.

Hmeister-real:
i like this, the Priest feels like a pushover. However, i think the BURN!! at the end is unneccesary, just keep it as ASTRATA SMITE YOU!!!

Squird5:
> Although I'd weaken it slightly outside of the church, not too much of a fan of the priest potentally being able to use this offensively to effectively sniper-firebomb people wherever they can see with laser-deadset accuracy.
> 
> I'd still make it gib undead, is vampire lord protected though? I know they don't roll at all but I think gibbing a vampire lord off-screen during the day as funny as it is, might be a **little too much**.

Vampire is currently disabled, but yes, it might not always be. This does give me an idea however. Instead of scaling damage based on time of day, I could have it scale based on being within the church instead, THEN check if it's an undead, so undead only get gibbed inside the church.
Potential downside to that however is that only a portion of the church is actually the AREA "Church", while the upstairs and rooms are "Town". That's the same reason why announcements don't work while upstairs.

> Especally that like all miracles you could probably do this hidden behind a window, in a closet or in the shadows and just keep doing it. Where nobody can figure out where you are.

You can't offscreen it since the range is only 8, and you also double exclamation point **YELL!!** what you are doing, so everyone within two or three screens can hear it, above and below

----
As current, it applies a number of fire stacks, as well as a chunk of burn damage. You extinguish 5 fire stacks per attempt, so 15 fire stacks take 3 attempts to fully extinguish. You also lose 0.1 firestack per second passively. The knockdown also does not stun, so moving and extinguishing can prevent all firestack damage. The goblin in the video had to be struck with smite four times before it stopped standing back up.

The smite can also be whiffed. If it is fired without being on a target, it expends devotion and the cooldown without applying any blast or damage.

Additionally, my experience is a fight involving an unarmored target generally lasts less than twenty seconds, while the cooldown on Smite is currently 60 seconds.

radarseas:
> Vampire is currently disabled, but yes, it might not always be. This does give me an idea however. Instead of scaling damage based on time of day, I could have it scale based on being within the church instead, THEN check if it's an undead, so undead only get gibbed inside the church. Potential downside to that however is that only a portion of the church is actually the AREA "Church", while the upstairs and rooms are "Town". That's the same reason why announcements don't work while upstairs.

Honestly though, I think that downside is fine its good enough for ambushing the priest in the quieter parts of the church which is honestly fine balance wise. There should be vulnerable spots even if its a weird code oversight on the mapping part. Plus it prevents sniping with gibs from above and forces the priest to take risk-reward factor of fighting with their flock thematically vs survival.

radarseas:
maybe axe the knockdown outside of the church for **definite**, you're pretty much guarrenteed to be fucked if someone casts this on you while trying to run away from combat because you'll just be stunned, floored and burning fast to crit. Especally if you can't get into the mire to get the fire off in the terrorbog as say, a lone bandit.

In the church though its fine, the _house always wins._

Squird5:
Fireball has already been nerfed and I'm going to be totally reworking it after I finish the explosions WIP, which I have on the side while I look into the Crits and Hits as Useroth asked.

And why the fuck shouldn't a powerful magical priest of THE GODDESS OF THE SUN not be able to BURN someone? It applies RAW FIRESTACKS not even a hotspot like Fireball currently does. It does not infinitely scale.

CloverIsLucky:
Reopen the PR!! ✊✊✊✊

PR #1080, opened 2024-10-06
Armory Reinforcement

Alanii:
## About The Pull Request

Due to individuals being incapable of tact, the Armories throughout rockhill and their associated barracks' are receiving dire reinforcement. 

Stone windows are replacing the old creaky wooden windows throughout the Bog Barracks and Watchman Head Quarters.

## Why It's Good For The Game

It isn't good or bad for the game. It should fixes the problem of those incapable of actually role playing of rushing the Armories for materials 10 minutes into a round.

### DIFF
--- a/_maps/map_files/rockhill/rockhill.dmm
+++ b/_maps/map_files/rockhill/rockhill.dmm
@@ -4831,12 +4831,6 @@
 	},
 /turf/open/floor/rogue/tile,
 /area/rogue/indoors)
-"fhH" = (
-/obj/structure/roguewindow/openclose{
-	dir = 8
-	},
-/turf/open/floor/rogue/wood,
-/area/rogue/indoors/town/garrison)
 "fhJ" = (
 /obj/item/trash/applecore{
 	pixel_x = -7;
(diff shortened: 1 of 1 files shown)


NikolayByondPlayer:
Wasn't armory moved to the gatehouse?

radarseas:
to be honest, the bog armory should be lessened in unused smeltable stuff too like a dozen pitchforks no sane individual would even pick up to fight with when you start with a sword that can lop off limbs.

the pitchforks are only ever missing if they've been smelted, not for use ever, period. They only encourage theft.

Jaaackx (changes requested):
Bog:
image
Still very vulnerable from these windows.

image
Also here which is the main source of entry from my experience.

image
Also the roof. It's insanely easy to get up there regardless of your climbing skill because there's almost always a tree branch leading onto it from the back.

I think it'd be better to reduce the number of windows in stupid spots (e.g. remove them from the roof, from the room with the bows and quivers) and then replace all the others with reinforced windows. **Stone windows are fine on the ground floor but they need swapping for something weaker on the floor above.**

I'm happy with the Watchman armoury changes, makes sense that's more fortified in the Town.

PR #1083, opened 2024-10-07
Whips blunt damage and punish damage modifier to 0.5

MickDick1:
## About The Pull Request

Changes all whip intents to blunt damage.
Changes punish intent to halve its damage.

## Why It's Good For The Game

Whips delimbing is whack.
Whips embedding is whack.
Whips being turbo lethal, a melee weapon with 3 range and 40 AP, is whack.
Punish intent no longer defaults to massive hematomas as a result of the base damage being too high on regular whip.

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/flail.dm
+++ b/code/game/objects/items/rogueweapons/melee/flail.dm
@@ -102,11 +102,11 @@
 	penfactor = 10
 	reach = 2
 	icon_state = "inlash"
-	item_d_type = "slash"
+	item_d_type = "blunt"
 
 /datum/intent/whip/crack
 	name = "crack"
-	blade_class = BCLASS_CUT
+	blade_class = BCLASS_BLUNT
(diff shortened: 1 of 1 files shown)


katsukai:
I think these changes are warranted since it is still quite formidable weapon that hits from far and you cannot dodge or parry the attacks from distance.

Keazra:
To non-robust players like me, someone with a whip is basically unassailable if I'm not covered head to toe in armor. The nerf is needed.

radarseas:
needed honestly, it already still has 40% armor peircing, insane range, decent swing time, in the case of a normal whip it FITS INTO A BACKPACK which is arguably more than it ever should have considering most weaponry stab intents only have 20% armor peircing. _(seriously what the **fuck?**)_


blessed PR.

Useroth:
Could also split whips into its own combat skill rather than shared with flails.
And give it only to the dungeoneer and inquisitor (and the odd migrant/adventurer ones)

PR #1087, opened 2024-10-09
Make rivers move everything

Cyprex:
## About The Pull Request

Makes rivers move not just mobs, but items, structures, and so on. Note that conveyormove checks anchored, so this shouldn't move anything it shouldn't move.

Not sure how performant this is- this definitely feels like it could be handled better, at lest.


katsukai:
Please merge this, people been using the chest exploit way too much and this fixes this.

Cyprex:
It's already testmerged

Cyprex:
Haven't heard of issues with this, merging.

PR #1090, opened 2024-10-10
Removes random auto spitting

Heheniy:
## About The Pull Request
Random auto spitting of item you are holding in mouth removed
It is fucking annoying when you WANT to hold something in your mouth but game just goes nuh-uh, spit

### DIFF
--- a/code/modules/mob/living/carbon/human/human_movement.dm
+++ b/code/modules/mob/living/carbon/human/human_movement.dm
@@ -94,10 +94,6 @@
 					update_inv_shoes()
 				//End bloody footprints
 				S.step_action()
-		if(mouth)
-			if(mouth.spitoutmouth && prob(5))
-				visible_message(span_warning("[src] spits out [mouth]."))
-				dropItemToGround(mouth, silent = FALSE)
 		if(held_items.len)
 			for(var/obj/item/I in held_items)
 				if(I.minstr)


Hmeister-real:
> ## About The Pull Request
> Random auto spitting of item you are holding in mouth removed It is fucking annoying when you WANT to hold something in your mouth but game just goes nuh-uh, spit
> 

It's there so you don't get a free slot to hold stuff, and you can't just carry a knife in your mouth 24/7

Hmeister-real:
Also rather than removing it, give items that you feel shouldn't drop out of your mouth the same function as the cloth gag, and give reasoning for each. Would make the PR seem less like cope

Squird5:
> Also rather than removing it, give items that you feel shouldn't drop out of your mouth the same function as the cloth gag, and give reasoning for each. Would make the PR seem less like cope

They already do. If you have an item in your mouth, you can not talk. That's how the cloth gag works to begin with. If you have a fiber bundle in your mouth, you can't cast spells because your mouth is obstructed. The pipe and cigs I think are the exception to that.

Hmeister-real:
> > Also rather than removing it, give items that you feel shouldn't drop out of your mouth the same function as the cloth gag, and give reasoning for each. Would make the PR seem less like cope
> 
> They already do. If you have an item in your mouth, you can not talk. That's how the cloth gag works to begin with. If you have a fiber bundle in your mouth, you can't cast spells because your mouth is obstructed. The pipe and cigs I think are the exception to that.

i meant as in, don't drop out of your mouth. I know everything gags you aside from smokes, but i meant that the creator of this PR should make certain items stay inside the mouth with reason rather than completely removing the function

Useroth:
It's intentional so that it's not a free slot, like mentioned in one of the comments above.
However, if you want, you could give it a minimal time before you start rolling for autospit, ie. 30-60 sec or so, so that you can use it in a pinch during work but not to carry stuff around reliably.

Koteshkin:
How do you remove a roach then? It seems to dont have a sprite in mouth and you cant really get it out unless spit automaticly.

PR #1092, opened 2024-10-10
+1 farming for servant

wow50000:
## About The Pull Request
gives servants farming 1

## Why It's Good For The Game
servantcucks cant even go to the soilsons, honestly I dont even agree with this change just that someone else asked for it
image


Cyprex:
Nah, they shouldn't be farming

PR #1091, opened 2024-10-10
No more legendary pilgrims

wow50000:
## About The Pull Request
changes 	
        maximum_possible_slots = 1
        and
	pickprob = 0.0001
as per maintainer request 
image

## Why It's Good For The Game
"legendary" my ass

PR #1095, opened 2024-10-11
Revert "No more legendary pilgrims (#1091)"

Heheniy:
This reverts commit 4a8f7a451c5e735055f734f00aa0dda1fc438616.

## About The Pull Request
That pr is just a bullshit


radarseas:
honestly, I'd keep the 1 slot per-round part of it, since they are supposed to be legendary and if anything up the pick prob to like 10%? 20%? so roundstarters will be more likely to get it than latejoiners.
(if anything most of them will die to stunlock + instant crit from rous anyway)

(minus smith, keep smith at 5% because that shit is insanely strong.)

We don't have enough pilgrims as is, it'll add some more victims and towner-esc roles around maybe if they existed more often with the incentive of a legendary class with the stats to boot?

Cyprex:
Sure. For what it's worth I'd consider a PR reducing their odds, or making them max 1 per round!

PR #1100, opened 2024-10-11
cheap swamp weed zigs and expensive love potions to merchant stock

Elli-Skala:
## About The Pull Request

adds both swampweed zigs (8 mammon) as well as love potions (600, may be to low) to the merchant luxury tab. 
## Why It's Good For The Game

the merchant has every other drug from the purity vender outside of swamp weed and this has always disappointed me. the lowest swamp weed from purity can be is 12 ( compared to normal zigs lows of 5 and 3 from the merchant) also allows for some merchant selling normal zigs as swamp weed zigs

the love potion package gives a alternative to getting a item thats other wise locked behind playing a mage (shameful) ~or a hour of dungeoneering~ at a price that hopefully should only mostly be seen by  towners!.

### DIFF
--- a/code/modules/cargo/packsrogue/luxury.dm
+++ b/code/modules/cargo/packsrogue/luxury.dm
@@ -10,6 +10,11 @@
 	cost = 3
 	contains = list(/obj/item/clothing/mask/cigarette/rollie/nicotine)
 
+/datum/supply_pack/rogue/luxury/coolersigs
+	name = "Swamp Weed Zig"
+	cost = 8
+	contains = list(/obj/item/clothing/mask/cigarette/rollie/cannabis)
+
 /datum/supply_pack/rogue/luxury/perfume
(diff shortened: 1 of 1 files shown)


Keazra:
More lovebombing, yes.

Cyprex:
Not sure buyable love potions are a good idea, but Gyran & seem to think so.

Heheniy:
there is literally swamp weed zig at the vendor outside merchant

Elli-Skala:
> there is literally swamp weed zig at the vendor outside merchant

Correct, but swamp weed was the only thing the merchant didn't have from that vender, his whole stick is that he can buy it cheaper and still get a cut

Keazra:
Maybe love potions should be buyable for an expensive price from the seedy vendor at the south part of town, instead of through the merchant?

PR #1102, opened 2024-10-11
Limb Artery Severity Fix

katsukai:
## About The Pull Request
Halves the lethality of limb arteries, giving you double the time to fix this wound before it becomes fatal, if you start suturing the wound it should reduce bleeding even more too.

## Why It's Good For The Game
Makes combat last bit longer, neck and heart arteries are still super lethal just made the easier to hit arteries less lethal for the sake of not dying so fast in-game.


MickDick1:
This should be reverted if we ever get pressure on bleeding wounds working. But for now, sure.

Useroth:
Don't call it a "fix" if it's a rebalance. @katsukai

PR #1103, opened 2024-10-11
Feint effectiveness buffed.

katsukai:
## About The Pull Request
How currently Feint works. If you successfully feint someone they cannot defend for 5 seconds and you need to wait 10 seconds to try to Feint again.

While feinting without feint style selected the immobility stun was 0.5 seconds. Too short for any normal person.
With Feint style enabled immobility stun was 1.2 seconds. Good but maybe not for everyone who has slow reflexes.
This PR increases both immobility stun times, without the style selected 1.5 seconds, with FEINT style enabled 3 seconds. 

## Why It's Good For The Game
Nobody uses Feint style, because you can use it without having it enabled and other styles like DEFENSE and AIMED offer a lot of benefits without risk. Having more different varieties to play and them being effective helps the game not to be stale. Having the immobility stun be 3 seconds opens a lot of ways to utilize Feint like guaranteeing a satisfying CHOP or a nice SMASH because of their long swing delays or even repositioning behind your opponent.

### DIFF
--- a/code/game/objects/items/rogueweapons/rmb_intents.dm
+++ b/code/game/objects/items/rogueweapons/rmb_intents.dm
@@ -29,7 +29,7 @@
 		perc += (ourskill - theirskill)*15 	//skill is of the essence
 		perc += (user.STAINT - L.STAINT)*10	//but it's also mostly a mindgame
 		perc += (user.STASPD - L.STASPD)*5 	//yet a speedy feint is hard to counter
-		
+
 
 
 	if(L.d_intent == INTENT_DODGE)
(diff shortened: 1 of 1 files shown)


katsukai:
> Imo this is overtuned, 1.5 seconds is a lot and merely doubling it still wouldn't warrant me using feint as an intent. One second without feint selected would be better, 2.5 to 3 seconds with feint enabled.
> 
> Also does feinting really prevent someone parrying for ten seconds? that's huge.

I think you're misreading, you can only feint someone every TEN seconds there is a COOLDOWN and they cannot PARRY for 5 seconds ONLY if you're successful, there is a cooldown on using this attack. Nitpicking on fracture of a second when the intent is completely dead ingame, I dont know it doesnt hard stun you, you can still attack or switch intent to dodge completely negating the Feint. There is lot of counterplay.

Red-Honk:
> > Imo this is overtuned, 1.5 seconds is a lot and merely doubling it still wouldn't warrant me using feint as an intent. One second without feint selected would be better, 2.5 to 3 seconds with feint enabled.
> > Also does feinting really prevent someone parrying for ten seconds? that's huge.
> 
> I think you're misreading, you can only feint someone every TEN seconds there is a COOLDOWN and they cannot PARRY for 5 seconds ONLY if you're successful, there is a cooldown on using this attack. Nitpicking on fracture of a second when the intent is completely dead ingame, I dont know it doesnt hard stun you, you can still attack or switch intent to dodge completely negating the Feint. There is lot of counterplay.
> 
> Depending on feedback I can lower non intent Feint into 1 second or lower. There is just not enough reason to use Feint as your main combat style.

There is no such thing as a cooldown, you can always feint, however, feinting before recovering from this so called "cooldown" gives your character a lower chance to feint successfully, even if they can, the more they wait, the higher the odds.

The goal of this intent is to break the flow of your target, open them to a counter-attack and putting them on the defensive, and in team fights, being stunned is a death sentence if chased by a sizable group, not to mention, being unable to turn for that time allows the others to flank behind and surround their prey and if behind them, they can't strike back, making them easy pickings for the remaining 5 seconds that feint lasts, not to mention that someone else that doesn't have the cooldown can perform it, and you have an utter death sentence.

And finally, just because something is under-utilized, doesn't mean it is bad, considering it costs no stamina to perform.

But if you seriously want to buff it, fine, make the non-feint style 0.7 that should be nearly enough time to strike at them or actually strike if using a swift weapon while they can't move, as for the feint style one, it looks fine as is, but if you insist on buffing it, make it 1.4 so, it is mathematically double the effectiveness.

PR #1104, opened 2024-10-12
Objects Burn Better Fix

Squird5:
## About The Pull Request

Objects were only taking 2% of the fire damage and since fire has been weakened, I've increased that to 15%

## Why It's Good For The Game

Sticks will burn again


Marrone321:
Does this also fix bushes/grass not burning? Just wanted to check in on it, seems like it probably does.

PR #1110, opened 2024-10-13
Make AI a bit less OP

Gyrandola:
Too much crack was given.


Hmeister-real:
aw

katsukai:
Have you tested this? Because if you want it to be 3 seconds you should type 30 because 3 is like one third of a second. Because code is weird.

Cyprex:
It was 5 originally, we lowered it to 1 and now 3

PR #1112, opened 2024-10-13
EXPLOSIONS

Squird5:
## About The Pull Request

Reworks explosions as a prerequisite for reworked Fireballs

## Why It's Good For The Game

Will allow for Fun New Things, like explosion based fireballs that aren't just Hotspot spawners

## Contents

**Explosion VisFX:** You can now add new visual effects to explosions using the `visfx` flag in the `explosion()` proc. Add the new explosion VFX to the list in explosion.dm
**Walls (can) Block Explosions:** Walls will take damage from explosions without the blast dealing damage to the other side, unless it is a soft material like a Tent.
**Better Projectile Explosions:** Projectiles like Fireball no longer explode inside of walls (entirely), but rather backstep to the outside. The wall still takes full damage by a mini-blast with the highest values at 1 tile range.
**Damage based Turf destruction:** Explosions now use the same damage system as weapons when it comes to damaging turfs. Bombs can blow up walls without it being an RNG check for outright deleting them!
**Blast Mining:** Since rock is a turf, it also takes damage. Ores can ~~hopefully~~ be dropped by explosions, but provide ZERO Mining XP.

### DIFF
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -403,7 +403,7 @@
 	take_overall_damage(brute_loss,burn_loss)
 
 	//attempt to dismember bodyparts
-	if(severity >= 2)
+	if(severity <= 2)
 		var/max_limb_loss = rand(0, floor(3/severity))
 		for(var/X in bodyparts)
 			var/obj/item/bodypart/BP = X


RadiantFlash7:
Interestingly, this makes fireball garbage at killing a lot of PVE mobs aside from skeletons, yet Overpowered in PVP.
I'm not sure how to feel about it in truth.

Koteshkin:
Fireball tore off my arm while exploding 2 tiles away from me. Is that.. supposed to happen?

RadiantFlash7:
To follow up what the other guy said, one fireball impact seems to tear off limbs. I don't even want to know what a greater fireball does. This is a bit ah. Much.

zeratino:
fireball does in fact tear off limbs

why the fuck am i getting instantly delimbed when im in full plate

code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -349,63 +349,65 @@
...
+				damage_clothes(max(brute_loss - bomb_armor, 0), BRUTE, "bomb")
+//				if (!istype(ears, /obj/item/clothing/ears/earmuffs))
+//					adjustEarDamage(15,60)
 
 	take_overall_damage(brute_loss,burn_loss)
 
 	//attempt to dismember bodyparts
-	if(severity <= 2)
-		var/max_limb_loss = round(4/severity) //so you don't lose four limbs at severity 3.
+	if(severity >= 2)
```
dwasint, line 406: this is incorrect this means light and moderate delimb not heavy and moderate

dwasint, line 406:
```suggestion
	if(severity <= 2)
```

Gyrandola (approved):

PR #1113, opened 2024-10-13
Removes the pilgrim migrant wave

Hmeister-real:
## About The Pull Request

 The pilgrim wave is nothing but filler for the other waves, completely useless and is just bloat.

## Why It's Good For The Game

migrant waves less annoying to wait for without it


Cyprex:
What migrant waves would you consider interesting?

Hmeister-real:
> What migrant waves would you consider interesting?

the ones that are unique and don't infinitely respawn outside of the migrant system (pilgrims/towners) ((note, towners are literally just better pilgrims half of the time))

PR #1115, opened 2024-10-13
Forgeable Belts and Chains

Nakalon:
### 1 - This PR makes Steel, Gold and Silver belts forgeable as well as Chains.

A single ingot of each will be necessary to create them.

image
image

(Gold and Silver belts aren't pictured because I couldn't figure out how to spawn Gold and Silver ingots at that time.)

### 3 - Changed the description of Chains to reflect their smelt result.

So now they ARE made of Iron and that's what you need to use to create them!

## Why It's Good For The Game

These items are unobtainable outside of spawning with them as certain roles (mainly belts).
When I discovered the Steel belt existed I was surprised to find that not only could I not craft it, but no one else could either.

I believe the more crafting options players have the more interesting it is to engage with crafting and roleplay.
I'm also, personally, not a fan of items that are nigh unobtainable.

## Changelog

- Steel, Gold and Silver Belts are now forgeable and Smeltable
- Chains are now forgeable out of Iron
- Changed the description of Chain to "A heavy iron chain." to reflect its smelting result

### DIFF
--- a/code/game/objects/items/rogueitems/ropechainbola.dm
+++ b/code/game/objects/items/rogueitems/ropechainbola.dm
@@ -160,7 +160,7 @@
 
 /obj/item/rope/chain
 	name = "chain"
-	desc = "A heavy steel chain."
+	desc = "A heavy iron chain."
 	icon = 'icons/roguetown/items/misc.dmi'
 	icon_state = "chain"
 	slot_flags = ITEM_SLOT_HIP|ITEM_SLOT_WRISTS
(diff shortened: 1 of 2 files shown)


Nakalon:
Probably adding a smelt result for the belts for consistency sake.

Nakalon:
> If chains are going to be made out of steel, I would remove their ability to be smelted. Enough people already use them as free materials to smelt already.

I wasn't aware of this, I can just change the description instead and make them be made out of Iron, how does that sound?

PR #1120, opened 2024-10-14
Crit craft mistake removal

Heheniy:
## About The Pull Request
Removes crit mistake from crafting which is just annoying

### DIFF
--- a/code/datums/components/crafting/crafting.dm
+++ b/code/datums/components/crafting/crafting.dm
@@ -230,7 +230,6 @@
 						return ", missing tool."
 
 					var/prob2craft = 25
-//					var/prob2fail = 1
 					if(R.craftdiff)
 						prob2craft -= (25*R.craftdiff)
 					if(R.skillcraft)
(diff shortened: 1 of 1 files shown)


Nakalon:
This will trivialise crafting if you do not have the required knowledge.

And it is the only thing stopping a person with 0 skill from making a full set of armour.

Instead of destroying the ingot they will now just have to spend time hammering it long enough for the weapon or armour to eventually pop into existence.

ThatReformedEngineer:
> This will trivialise crafting if you do not have the required knowledge.
> 
> And it is the only thing stopping a person with 0 skill from making a full set of armour.
> 
> Instead of destroying the ingot they will now just have to spend time hammering it long enough for the weapon or armour to eventually pop into existence.

Pretty sure all that this will do is make crafting things like a leather cloak with zero skincrafting more tolerable in that it won't sudden crit fail multiple times in a row. Not actually mess with how the smithing system works with spoiling bars if you are unskilled? So the chances of someone making a full set of armor with zero skill isn't an actual problem, because if they already got a hammer, tongs, and forge, then they can just level the skill like normal at that point.

Heheniy:
this has nothing to do with spoiling metal bar while smithing

Nakalon:
> this has nothing to do with spoiling metal bar while smithing

My bad, I saw your other smithing PR and assumed they were related. However you have to admit this PR's description is a bit lackluster.

PR #1122, opened 2024-10-14
Bath house revision, added grille to tavern, deleted library room from tavern and tried to do make tavern's ground floor a bit less lavish

NikolayByondPlayer:
## About The Pull Request

**Bath House**
While playing found out that other z floors of Bath house are not used and the whole place is at odds with Ratwood's lore and thus tried to make it suit it better. The bathhouse is located in the same place but just one z level bellow in the sewers. It has same functionality but takes less space + there place has an emergency exit so it can still be used by antags taking over.
image
image
image

**Tavern**
Removed library room to make a bit less space to worry about, added grilles above so it would be possible look above but not that easy to jump into innkeeper's place. Also added bars to protect innkeeper from people jumping in all the time.
image
image
image

## Why It's Good For The Game

- I've cut down unused space from the bath house and made it better fit lore
- QoL for Barkeepers


NikolayByondPlayer:
Also there is a river now that goes almost to the ship, wanted to incentivise throwing people to it :3
image

Jaaackx (approved):
Looks good to go.

NikolayByondPlayer:
> Looks good to go once conflicts resolved.

You mean check in IconDiffBot?

NikolayByondPlayer:
Removed some bars for easier access into main sewer area 
image
image

PR #1125, opened 2024-10-14
Bladesinger Improvement

RadiantFlash7:
## About The Pull Request
For a rare role, this role is pretty hot garbage. This aims to improve it.
Increases their starting spellpoints by one, and buffs their starting stats.
at present, they have a -2 to Endurance, which is /hilariously/ bad for ANY role expected in melee. And the bladesinger is meant to be a melee role. Hell, even mages don't have a -2 in END.
I've given them a total buff of 6 trait points from their 3. Currently, pally's get 5, and warriors get 7. 
Edit: per feedback increased sword skill to 3, so mages can't just hit a dummy and be on the same level as the bladesinger, and removed axe skill to compensate.

## Why It's Good For The Game
In truth, the bladesinger role is pretty terrible. It's probably the worst role for melee, due to it's endurance, and it's limited spell casting means it can't even do that. I get wanting to have a 'spread thin, so they're not great at either melee or magic', but the stat penalty was a bit too much.
increasing them up to 6, with 2 INT, 2 END, 1 con, and 1 speed feels like a good in between, while still keeping the rare role appropriately effective.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/combat/rare/bladesinger.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/combat/rare/bladesinger.dm
@@ -26,12 +26,11 @@
 	head = /obj/item/clothing/head/roguetown/helmet/leather
 	if(H.mind)
 		H.mind.adjust_skillrank(/datum/skill/combat/maces, 1, TRUE)
-		H.mind.adjust_skillrank(/datum/skill/combat/axes, 2, TRUE)
 		H.mind.adjust_skillrank(/datum/skill/misc/athletics, 2, TRUE)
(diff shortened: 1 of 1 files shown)


RadiantFlash7:
To add to this, I've found out recently, that you can't actually parry while charging a spell.

Heheniy:
i think his sword skill should be improved by 1 too because you can train apprentice yourself, which makes bladesinger worse then mage who trained sword skill himself because he has more magic points and madium armor training is just difference in 2 pieces of armor

RadiantFlash7:
> i think his sword skill should be improved by 1 too because you can train apprentice yourself, which makes bladesinger worse then mage who trained sword skill himself because he has more magic points and madium armor training is just difference in 2 pieces of armor
I /could/ bump it up to 3 so it's journeyman.

PR #1108, opened 2024-10-12
Soup Craft (bounty)

Luctus-Comptus:
food change. soup crafting added. buyer said they would do a wiki or what have you. this is 1/3 of a cooking system rework. groundwork has been layed without influencing existing food craft so all default prior recipes exist. Its heavily modified but I dislike not giving credit to whoever did it. im relatively sure the base code for it is generalpantsu on azure. but im not certain. whoever it is core base code isnt my creation. massive modifications to get it to work here is.

*also as a lil bonus love I made each BITE drink from water-source give half mug worth per swallow roughly 5x what was prior.

PR #1128, opened 2024-10-15
Revert "Soup Craft (bounty)" [Testmerge only]

Cyprex:
Reverts Rotwood-Vale/Ratwood-Keep#1108

Breaks Blacksmithing


Nakalon:
Thank you!

katsukai:
Apparently it also broke cloth dipping into bucket. Making cleaning surfaces harder

PR #1129, opened 2024-10-15
Lowers legendary pilgrim chance

wow50000:
## About The Pull Request
Lowers the chance of rolling legendary pilgrims by a reasonable degree as a temp fix to discourage rolling pilgrim to maximise chances of getting legendary. Adds a max cap of 1 per role.

## Why It's Good For The Game
https://github.com/Rotwood-Vale/Ratwood-Keep/pull/860 over buffed legendary pilgrims as well as increasing their chances to roll. Due to this, a large amount of players have been drawn towards the pilgrim role on top of existing pilgrim players leading to the town being increasingly empty.

Pilgrims by design are meant to be around about the same level as towners, only the fact that they have been on the road running from the great rot lorewise. I can see the idea that legendary pilgrims should exist due to survival of the fittest(the best are more likely to make it out.) but as it stands, legendary pilgrims are significantly stronger than all adventurers and mercs and stand a very good chance of beating one or multiple at once.

Legendary pilgrims are hyper powerful, able to rival most combat classes with stats and skills alone. "while sporting equipment that is either unobtainable or requires 2/3rds of the round grinding for most roles to get. 

As well as having the skills to absolutely drive out any existing towners. Due to this combination, I believe that legendary pilgrims should have their chances to roll lowered as well as a maximum cap on them as a temporary fix until they are either nerfed or better integrated into the game. 

Because as it stands, legendary pilgrims are relatively high chance to roll, and have arguably the best power budget in the game on top of ultimate freedom of alignment or antag status which other powerful roles do not have (King, Bandits, Knights etc.)
TLDR: mary sue class should be very rare.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/pilgrim/rare/Lwoodcut.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/pilgrim/rare/Lwoodcut.dm
@@ -8,7 +8,7 @@
 	allowed_sexes = list(MALE, FEMALE)
 	allowed_races = RACES_ALL_KINDS
 	outfit = /datum/outfit/job/roguetown/adventurer/masterwoodcutter
-
+	maximum_possible_slots = 1
 	pickprob = 5
 	category_tags = list(CTAG_PILGRIM)
 	


katsukai:
Looks good to me

PR #1134, opened 2024-10-16
Food bounty fix

Luctus-Comptus:
im tired and very sick. here. stop reverting things when it takes 2 seconds to fix. I'll cut you.

### DIFF
--- a/code/modules/roguetown/roguejobs/blacksmith/forge.dm
+++ b/code/modules/roguetown/roguejobs/blacksmith/forge.dm
@@ -28,6 +28,15 @@
 			W.forceMove(src)
 			update_icon()
 			return
+	if(istype(W, /obj/item/rogueweapon/tongs) && on)
+		var/obj/item/rogueweapon/tongs/T = W
+		if(T.hingot)
+			var/tyme = world.time
+			T.hott = tyme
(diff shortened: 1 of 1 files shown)


hugmeorelse:
The revert is just a test merge, so people can smith in the meanwhile until a fix is in. :) Nothing personal!

Red-Honk:
I mean, they were able to smith by heating it up in the smelter and pulling out the bar with tongs, sure, it was more tedious, but it worked.

Nakalon:
> I mean, they were able to smith by heating it up in the smelter and pulling out the bar with tongs, sure, it was more tedious, but it worked.

It was horrible. No one wanted to make smith because of it. Yes you WERE able to smith, but in an extremely convoluted way which involved setting up the anvil and water next to the furnace, keep it running, and dropping your tools to manually insert the bars inside when they cooled down.

Defending smiths here because I love it hehe

Cyprex:
Okay I know what happened. So this PR first removes the lines that make smithing work- and then it adds them back in another commit. However, we squash merge PRs- that collapses the two into one commit which doesn't change the forging lines at all. Hence there were no diffs when merging this, and hence this doesn't actually add back the forging lines. Wew.

PR #1078, opened 2024-10-05
lowers adventurer cap to 10

wow50000:
## About The Pull Request
WOE BE UPON ADVENTURER MAINS GO BACK TO THE WANDERER QUEUE

## Why It's Good For The Game
image
someone on discord asked for it

PR #1135, opened 2024-10-16
Revert "lowers adventurer cap to 10"

Useroth:
Reverts Rotwood-Vale/Ratwood-Keep#1078


NikolayByondPlayer:
Even 10 is too many...

PR #1136, opened 2024-10-16
Grenzelhoft tabaxi

Heheniy:
## About The Pull Request
Allows tabaxi be grenzelhoft mercs
image


Hmeister-real:
but they are meant to be from the desert, like the Kajhit, surely

PR #1138, opened 2024-10-16
Changes to Stamina and Fatigue Calculation

### LINKED ISSUE #1063: Stamina and Fatigue are swapped.
Stamina and Fatigue bars are swapped, thus leading people to code targeting the wrong variable. For instance athletics is coded so that levels in it only raise your blue bar and do nothing for your green because they're coded to change rogstam which is actually fatigue as shown below:

See UI names being swapped:
```
/atom/movable/screen/rogfat
	name = "stamina"
	icon_state = "fat100"
	icon = 'icons/mob/rogueheat.dmi'
	screen_loc = rogueui_fat

/atom/movable/screen/rogstam
	name = "fatigue"
	icon_state = "stam100"
	icon = 'icons/mob/rogueheat.dmi'
	screen_loc = rogueui_fat
```

See athletics code:

```
/mob/living/proc/update_rogstam()
	var/athletics_skill = 0
	if(mind)
		athletics_skill = mind.get_skill_level(/datum/skill/misc/athletics)
	maxrogstam = (STAEND + (athletics_skill/2 ) ) * 100
	if(cmode)
		if(!HAS_TRAIT(src, TRAIT_BREADY))
			rogstam_add(-2)
```

Pretty sure athletics should be raising your green bar, not your blue bar. I don't think any physical role actually cares terribly much about the blue bar, it does affect physical roles slightly with swift intent and such but not nearly as much as having extra green would aid. Right now raising athletics helps mages far more than it does non-magic users.

Nakalon:
## About The Pull Request

Hopefully this closes #1063 

While investigating this issue I found out that Fatigue is the Green bar and Stamina is the Blue bar.

Of course, congratulations to me for discovering the obvious, but its normal to want to want to swap those names.
In most games what you spend to attack is stamina. The one that regenerates on itself after some time is stamina.
It would make sense if fatigue was the blue one, which is the one you regain when you go to sleep. So... tiredness!

However that is not the case here, and changing that would be an absurdly complex task, which would probably bring more bugs than solutions.

Looking at the issue, as we see here:

image

The names are indeed swapped, no biggy, we swap them back.

But regarding this...

image

I agree! I think its odd there is a calculation for Stamina and not one for Fatigue, which is the one you actually need to attack!

So I added that calculation to fatigue!
Which takes your endurance, adds your athletics level into a nice bundle and multiplies it. That's your Maximum Fatigue now!

image

However, since I don't play mage classes, I kept stamina as it is! Being affected by Endurance!

## Problems and Tribulations

1 - There's a tiny itty bitty chance the hud does not display this correctly.

However, while testing I didn't notice anything odd and looking at how the hud is updated in code it seems to be ready to do the math by percentage, but you never know!

2 - I went back to Endurance at the request of the comments!

_2 - I'm still on the fence about INTELIGENCE being the determining factor in stamina

I was considering CONDITION actually! I will need a Maints opinion on this!_

## Why It's Good For The Game

This will probably strengthen those with an able body and who put some time into leveling their athletics.
And weaken those who aren't so lucky!
Diversity is the spice of life!

## Changelog

Fix: Swapped Stamina and Fatigue hud names back into place
Code: Removed commented out code in rogfatstam.dm
Add: Max Fatigue is now calculated with Endurance and Athletics in mind

### DIFF
--- a/code/modules/mob/living/carbon/rogfatstam.dm
+++ b/code/modules/mob/living/carbon/rogfatstam.dm
@@ -19,7 +19,7 @@
 	var/athletics_skill = 0
 	if(mind)
 		athletics_skill = mind.get_skill_level(/datum/skill/misc/athletics)
-	maxrogstam = (STAINT + (athletics_skill) / 2) * 100 // STAMINA / BLUE most useful for wizards, affected by INT
+	maxrogstam = (STAEND + (athletics_skill) / 2) * 100 // STAMINA / BLUE
 	if(cmode)
 		if(!HAS_TRAIT(src, TRAIT_BREADY))
 			rogstam_add(-2)


Nakalon:
Please note those screenshots were taken when I had removed the athletics division by two.

The updated values should be 85 and 950 respectively!

john-pawson:
Alright, why would INT ever influence your stamina? The entire point of END *is* stamina and fatigue control (With pain on the side too) why INT suddenly make you suddenly more fit?? This also ends up on dumb people that are supposed to be fit (like barbarians) having low stamina for some reason.

Nakalon:
> Alright, why would INT ever influence your stamina? The entire point of END _is_ stamina and fatigue control (With pain on the side too) why INT suddenly make you suddenly more fit?? This also ends up on dumb people that are supposed to be fit (like barbarians) having low stamina for some reason.

Heres why!

image

Hence this!

image

So you want Endurance on both or Condition on Stamina as I propose here?

Nakalon:
> Endurance doesn't have much purpose, so I vote on that.

Very well! I'll change it back to endurance once I'm back at the PC!

Red-Honk:
Let's see, the funny thing about that, I always called the bluebar fatigue and green one stamina, guess someone messed that up somehow, even when they're side by side, regardless, the reason athletics effects fatigue is simple, you sure will get tuckered out like anyone else, buuuuttthh, the more fatigue bar you have the faster you recover, meaning this, if you have better endurance and athletics then your opponent, you will eventually end up winning in a war of effort, because at some point, they won't be able to chase or fight you without barely being able to scrape some recovery back, that is why it focuses on fatigue instead of stamina, so, it's fair for everyone involved, but the longer it goes on for, the more the odds go in your favor.

Nakalon:
> Let's see, the funny thing about that, I always called the bluebar fatigue and green one stamina, guess someone messed that up somehow, even when they're side by side, regardless, the reason athletics effects fatigue is simple, you sure will get tuckered out like anyone else, buuuuttthh, the more fatigue bar you have the faster you recover, meaning this, if you have better endurance and athletics then your opponent, you will eventually end up winning in a war of effort, because at some point, they won't be able to chase or fight you without barely being able to scrape some recovery back, that is why it focuses on fatigue instead of stamina, so, it's fair for everyone involved, but the longer it goes on for, the more the odds go in your favor.

Yes yes! What im doing is make so it applies to both, stamina and fatigue, blue and green bar!

Red-Honk:
> Yes yes! What im doing is make so it applies to both, stamina and fatigue, blue and green bar!

I am uncertain of how good of an idea that is for 1 balance and 2 encouraging a dull grinding mindset or else others will just simply win. (The only benefit I see from this is garrison leader roles force marching troops into building up their endurance.)

Nakalon:
> > Yes yes! What im doing is make so it applies to both, stamina and fatigue, blue and green bar!
> 
> I am uncertain of how good of an idea that is for 1 balance and 2 encouraging a dull grinding mindset or else others will just simply win. (The only benefit I see from this is garrison leader roles force marching troops into building up their endurance.)

Mind you, theres also a negative effect if you have endurance bellow 10!

But one of the reasons you mentioned is why I had made one Endurance the other Inteligence...

Red-Honk:
> Mind you, theres also a negative effect if you have endurance bellow 10!
> 
> But one of the reasons you mentioned is why I had made one Endurance the other Inteligence...

I think that int. should only effect stamina drain when casting spells, the better the mind, the easier it is to conjure an idea to cast type of deal.

john-pawson:
> > Alright, why would INT ever influence your stamina? The entire point of END _is_ stamina and fatigue control (With pain on the side too) why INT suddenly make you suddenly more fit?? This also ends up on dumb people that are supposed to be fit (like barbarians) having low stamina for some reason.
> 
> Heres why!
>(image)
> Hence this!
>(Image)
> So you want Endurance on both or Condition on Stamina as I propose here?

Well, while neither of the pictures say *why* would INT affect it, other than "someone else suggested it", i appreciate you saying your doubts, and i think END should keep being the main factor in fatigue and stamina, it's the *point of the stat*. 
CON is normally used to see how healthy or bulky you are, your resistance to poisons, how much damage your body can take before dying, the amount of blood you can lose without having an effect, how hard your bones are to break and how hard your skin is to bruise. 
END *is* fatigue and stamina, it's how many days you can keep on awake without sleeping, it's how much you can keep on functioning besides the pain on your body, it's how much you can keep on swinging the sword without the chance to catch a break, it's how much you can keep on running even if your legs are giving.

What i mean is, switching fatigue or stamina to another stat would defeat the purpose of having the END stat. Because END ***is*** fatigue and stamina.

Mind if i ask why was fatigue getting changed on the first place? Don't take it as me trying to argue or anything, i can see that you already changed it back to endurance, just curious.

Nakalon:
> Mind if i ask why was fatigue getting changed on the first place? Don't take it as me trying to argue or anything, i can see that you already changed it back to endurance, just curious.

It was stamina that was being changed! Blue bar, since the author said blue bar is more usefull to mages and of little use to melee warriors I went with that information and thought of making it a int based thing! Thats all! My thought process!

PR #1139, opened 2024-10-16
Pick intent now deals pick damage

john-pawson:
## About The Pull Request

The dagger's pick intent, as it is on the name and the attack in itself, is supposed to deal pick damage. Someone forgot to change the damage type from stab to pick, this fixes it.
Ever wondered why you can't pick with a dagger but can with a stake? Or why pick had the same chance of hitting a crit when it *should* have a much higher chance? This is the reason.

## Why It's Good For The Game

Makes the pick intent properly work now! Higher chance of artery crits like it should, can now be used to pick rocks like the stake, all the things it was supposed to do from the start.

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/knives.dm
+++ b/code/game/objects/items/rogueweapons/melee/knives.dm
@@ -62,6 +62,7 @@
 	name = "icepick stab"
 	icon_state = "inpick"
 	attack_verb = list("picks", "impales")
+	blade_class = BCLASS_PICK
 	hitsound = list('sound/combat/hits/bladed/genstab (1).ogg', 'sound/combat/hits/bladed/genstab (2).ogg', 'sound/combat/hits/bladed/genstab (3).ogg')
 	penfactor = 80
 	clickcd = 14


katsukai:
this could be terribly busted because no armor has critical resist to picks nor there is armor against pick damage type.

Hmeister-real:
> this could be terribly busted because no armor has critical resist to picks nor there is armor against pick damage type.

it was always intended to have pick, and it does. it was just bugged

pilotmans:
Apparently pickaxes don't even use the same pick intent as daggers if the ingame stats are to be believed

john-pawson:
> Apparently pickaxes don't even use the same pick intent as daggers if the ingame stats are to be believed

No, they don't use the same intent. Pickaxes use intent/pick and daggers use intent/thurst/pick.

Cyprex:
I'm merging this- but if it ends up being too strong combat capabilities should be nerfed. This should mine rock like a stake, however, or be renamed/removed- if it says pick it should pick.

PR #1147, opened 2024-10-17
Lets slavers be any race

wow50000:
## About The Pull Request
Lets slavers be any race(except seelies)

## Why It's Good For The Game
I want to play a dark elf slaver


Hmeister-real:
they are racist though

PR #1150, opened 2024-10-17
Subsystem (HumanNPC) Delays

ProbablyCarl:
## About The Pull Request
- - -
This should, in theory, slow down the response times and attack rate of the following humanoid NPCs:
 - Bums
 - Goblins
 - Skeletons

The above are now on four second delays, as opposed to 0.3.
To test it in game, open the MC, find the HumanNPC subsystem, look for 'wait' as a variable and change it to what you think appropriate. I've tested 20/40/60(2, 4, 6 seconds.) Alternatively, just swap it back to 5.
- - -

## Why It's Good For The Game
Do I really need to explain this? Half the community is supposedly screeching about how bad it is.

- - -

### DIFF
--- a/code/controllers/subsystem/rogue/humannpc.dm
+++ b/code/controllers/subsystem/rogue/humannpc.dm
@@ -1,7 +1,7 @@
 
 SUBSYSTEM_DEF(humannpc)
 	name = "humannpc"
-	wait = 2 SECONDS
+	wait = 20
 	flags = SS_KEEP_TIMING
 	priority = 50
 	var/list/processing = list()


john-pawson:
Alright, tested it myself, i have to say that the NPCs do need to be slower from 3 ticks, this makes them *waaaaaaaaaaaaaaay* too slow. 

It takes the skeleton about 8 seconds to simply walk into the player, and attack. And if you move a single tile while the skeleton has it's 4 second cooldown after being near you, it needs ANOTHER 4 seconds to simply walk into you again and attack. All the player needs to do is attack and then move a *little bit* and he's effectively beaten the enemy.

2 Seconds is still pretty slow. Not as ridiculous as 4 secons, but definitely better than 2.

From what i've tested so far 12-14 should be the best range (~Same delay normal attacks have). 
Still fast. Still decently dangerous. But you can run away if you're fast enough and they're not attack-spamming weapon-damaging stamina-eating grab-breaking artery-slitting head-decapping skull-bending spine-slipping constatly-running killing machines.

Nakalon:
Althougtht I havent tested myself I'm goign by Johns feedback and 3 to 40 is a BIGGGGGGGG jump

I'd like double it or triple it at max and work from there!

Good PR nontheless!

john-pawson:
> Have you even tested this. The original value was 5
> 
> 15 makes them fucking retarded

I mean if you make it anything >8 i've found it makes it impossible to get near them without being instantly attacked, and doesn't have much of a difference from 3, not only that, but running away isn't exactly very doable when they're able to hit you the moment you're in range before your character is able to move from one tile to the other.

Also. "Retarded" is completely subjective. If the point of them is to be killing machines that just run up to you and attack you until you die, and you should *never* fight them unless you're equipped with good armor and good weapons, then yeah they're "retarded".
If they're supposed to be something that you're expected to fight or encounter regularly when traversing the bog or going into caves, it's a bit impossible to fight them without getting stam-critted from blocking or outright killed, since if you get hit once and slowed down, your chances of outrunning or losing them are *very* slim. Even more in the bog with the mud.

12-14 still makes them dangerous. Just not as dangerous as before, dince the point of the PR was to nerf them. But this only matters depending on how dangerous you *want* them to be, how you're planning on balancing the whole thing, which i don't know, so i made a suggestion that makes them easier than before, but not a joke like 40 or 20.

hugmeorelse:
15 seems like a good place for the mobs to be at.

hugmeorelse:
@Gyrandola You can't go by the old value at all, because other changes were made to speed the AI up. So far, 15 has proven to be quite fast, i.e. fast enough for a simple mob to break through shield than a player mob can.

john-pawson:
> @Gyrandola You can't go by the old value at all, because other changes were made to speed the AI up. So far, 15 has proven to be quite fast, i.e. fast enough for a simple mob to break through shield than a player mob can.

I would like to add that this may be affected greatly by skeletons having a very high STR stat and dealing significantly more damage, and damage output may have a higher role than the speed of the AI if a shield breaking is what you're using for testing. 

So if the problem is the sheilds breaking fast, STR score might be the main problem rather than the speed of it's AI.
Same goes for beheading and breaking armors.

EvenInDeathIStillServe:
Lmfao why would you do this. It means you can literally just walk away from mobs and they'll always be 2 seconds away from you.

PR #1152, opened 2024-10-18
Devotion for Druids = Life Cleric

LofiOtter:
## About The Pull Request

Dendor Life Clerics are now better than Druids at doing druid things because they can make the funny masks now. Let's at least keep them on par devotion-wise. Acolytes don't have this issue.

Also why in God's name are there CLERIC_T0-4 defs if you goobers are gonna make actual devotion earned key off of the skill. I don't understand code all I know is this confused me deeply and made me sad so I just set magic/holy to be equal to how it's set for life clerics.

## Why It's Good For The Game

Acolytes don't have issues with devotion gain compared to their cleric counterparts. To my understanding, druid devotion gain was nerfed because of kneestingers which they no longer have, so I'd like them to at least have their devotion back.


Elli-Skala:
> Also why in God's name are there CLERIC_T0-4 defs if you goobers are gonna make actual devotion earned key off of the skill.

skill level is the regen rate, what the defs are how many tiers of spells one can know (for example templars so they cant learn every spell just the base ones)

LofiOtter:
For the record, I just tested this in-game and it works. Devotion gain for druids is now 3 - as expected - so I am satisfied.

PR #1151, opened 2024-10-18
Sewing Rework Atomization - Part 1 - Clothing Salvage

Nakalon:
## About The Pull Request

I have planed to make a rework of the Seamstress and sewing to make it more exciting and fun to play!

The main idea is to make it function similar to Smithing.
This PR fits that purpose by being the Smelting equivalent of Sewing!

### - Introducing: Scissors!

image

Scissors and Needles are to the Seamstress as Tongs and Hammers to the Smith!

Seamsters spawn with an Iron Scissor now instead of a hunting knife!

image

Don't worry, you're not missing anything! The Scissors are actually a subtype of hunting knife.
(Also, if you lose it you can also ask the smiths to make another, Iron or Steel!)

While a Seamstress may use a Needle to repair clothing, they may use their newly discovered scissor technology to salvage materials out of clothing items!

image

This is an example.
You will always either get 1 fiber or no fibers at all depending on the item.
(If its got fibers in its crafting recipe you'll get 1 fiber)

You may or may not be able to salvage base materials (such as cloth and leather) out of a piece of clothing depending on your sewing skill!

image

Yes! It works for leather also!

image

You may be thinking about "what about sleeve tearing? I'm gonna print unlimited cloth!"

NO. NO YOU WONT! STOP IT!

### - Sleeve Tearing improvements

I always thought tearing sleeves was very obscure and not very obvious.

So I changed that! Now you get proper information on examine about the state of the clothing piece!

image
image
image

### Attention: The above typo has been solved. my bad kings and queens.

This means, I also accounted for sleeve tearing when salvaging something with scissors!

The math works!

image

Wow I have no idea what this means!

### - Experience!

~~Now you get sewing experience for sleeve tearing, but get more experience for salvaging the proper way!~~

You only get experience for salvaging now! Oups!

## Why It's Good For The Game

More power to the Seamstress!
Gives one more step of depth to the role, thus hopefully enticing people to play towner!
Also included are some slight coding improvements, which are always good!

## Changelog

Add: Added Iron and Steel Scissors which may be used to salvage articles of clothing
Sprite: Added Sprites for the Iron and Steel Scissors
Tweak: Seamsters now spawn with Iron Scissors instead of a hunting knife
Tweak: You can now tell how many sleeves were torn form a piece of clothing by examining
~~Tweak: You now get sewing experience for sleeve tearing~~
Code: Removed some duplicated vars and commented out code when it presented itself
Code: Improved Sleeve tearing code and gave it some early returns
[...]

PR #1154, opened 2024-10-18
Sewing Rework Atomization - Part 2 - Seamster Shop Rework

Nakalon:
## About The Pull Request

In this part 2 of the Sewing rework I'm reworking the Tailor shop where the seamster works!

### I had 3 goals when making this:

1 - Improve lighting as the place was extremely dark before which made it very hard to work in and made it feel abandoned.

2 - Giving it the Smithy treatment (since as mentioned in Part 1 #1151 that is the reference for the rework), making them function similarly by moving the counter outside, adding a Vomitorium, shutters, a Peddler and making it so clients can see the seamster working from outside.

3 - Making it more Home-y by adding food, moving some things around and adding some decorations such as the fireplace and replacing the windows with openable ones (for privacy).

4 - Giving the Seamster just a little bit more starting resources by adding two shirts which may be salvaged by the mechanic introduced in Part 1 #1151 

5 - Making the seamster and their shop more sought after and entice RP, this was achieved by adding a tanning rack (which the seamster usually already builds) to support light armour crafting and expand the items a seamster may offer to their clients, and by lighting the outside of the shop, to capture player attention.

6 - Giving the Seamsters more space with which to do whatever. By moving the counter to the street the Seamsters now have more space to build racks or whatever. Up to them!

Client POV:
image

The inside:
image

Seamster POV:
image

Closed Shop, making it absolutely obvious when the Seamster is and isn't around:
image

~~A would be Thief approaches an open window:~~
Changed these into reinforced since we can't have nice things since the monkeys will break them 😡 
image

StrongDMM image:
image

## Why It's Good For The Game

More power to the Seamstress, yada yada!
Should improve roleplay!
Also is part of my Sewing Rework so it strengthens Part 1 and coming parts!

## Changelog

Add: Reworked the Seamster Shop
Add: Added a Tanning Rack, a Vomitorium, and other goodies to the seamster shop, including crackers and water.
Add: Added more lights to the Seamster Shop, including one on the outside.
Add: 2 White plain shirts will now spawn in the Seamster shop to be salvaged for cloth.

### DIFF
--- a/_maps/map_files/rockhill/rockhill.dmm
+++ b/_maps/map_files/rockhill/rockhill.dmm
@@ -11110,6 +11110,12 @@
 "kIh" = (
 /turf/open/floor/rogue/woodturned,
 /area/rogue/indoors/town/garrison)
+"kIm" = (
+/obj/structure/roguewindow/openclose/reinforced{
+	dir = 8
+	},
+/turf/open/floor/rogue/wood,
+/area/rogue/indoors/town)
 "kIu" = (
 /obj/structure/stairs,
 /turf/open/floor/rogue/woodturned,
@@ -301092,8 +301098,8 @@ evo
 evo
 evo
 evo
-dzT
-dzT
+kIm
+kIm
 evo
 evo
 xna


NikolayByondPlayer:
Nice stuff! 
Also I feel like area with windows is a bit too easy to break into, maybe place a reinforced glass there or make a thief climb something to get behind the shop?

Nakalon:
> Nice stuff! Also I feel like area with windows is a bit too easy to break into, maybe place a reinforced glass there or make a thief climb something to get behind the shop?

Ease to break into if they're open! If they're closed its no less difficult than it was when they were plain windows!
At least thats what I'm think, but I could be wrong!

I will wait on more feedback before changing this! Because I didn't want to affect the level of security the shop had, as I never saw a break in taking place!

Thank you for commenting!

Nakalon:
Updated the windows to be reinforced!!

Nakalon:
> people play seamster still? poor bastards.

The idea is exactly to make it not suck!

PR #1159, opened 2024-10-19
Smoke pack and 2 new purity items

Luctus-Comptus:
test merge. to see if pricing and any other issues arise.

*replaced single zigs in purity with branded zig pack sprite made by the talented @sinnerpen[discord] (I picked snakes because I like them alot ok?)

*said zigs come with some honey and nicotine in them, the honey heal is an RNG proc. high roll 30hp total back and thats... extremely lucky. it degrades into sugar at a 33% chance that degrades into nutrients...at a 33% chance, so simulates appetite suppressant effect. will not even on a high roll be more than a berry.

*2 new chems... will say the easier one first. both chems come witha  resprited pre existing TG syringe, its a coder sprite but it isnt terrible. I gave it a custom soundfile I made. resprited it the best I could and its modled after autoinjectors so you cannot remove the contents or refill it

*NOURISH is the first syringe, extremely simple on the cheaper end, full meal and drink suppliment. got honey in it for a lil heal as well same as a zig but all at once. and water for thirst.

*Snake Bite, this one is a doozy and took me forever to do; it is expensive and it is deadly. do not OD on this or your probably RR'd. single dose is all thats safe, drop more and....dont do a drop more. buff last 60 seconds. you are extremely reselient during this time and boundless energy. it removes the sleepyness debuff from you and banks your sleep points. good for when you cant sleep but want remove the debuff and use the points later. 2 str 2 con -3spd. locks stam loss. resistant to magic, resistant to limb break and dismember, resistant to pain stun. once it ends tho your fatigue and stamina hit the fucking floor. gone, come back down to earth hard. has some custom annoying to accomplish visual effects I put into it while you are stimed on it. oh and ahead of time, boathans that think they can pump themselves full of this shit, its a toxin, not a drug sooooo.... go ahead... try.

thank your local snake oil....snake thing

also its modular because I like doing modular.

image


katsukai:
General impressions, 

Branded zigs, these are cool, can't complain. Maybe bit too cheap.

NOURISH, don't know how to feel about this. Kinda feels outside the setting, my opinion should be priced higher because it seems really cheap compared to Inn's price of a meal and a drink.

Snake Bite, Seems absolutely insane and kinda similar what Moondust does but maybe better? I feel like there should be more risk, maybe even a chance of a heart attack.

Generally feels these luxury items could be more expensive.

PR #1164, opened 2024-10-19
Adds new Seed & Feed special, added Seelie language to Polyglot special.

katsukai:
## About The Pull Request
Adds exclusive special for Soilsons, Towners and Pilgrims. Weight is 100, so the same as Outlaw and Excommunicated.
image
What it does, +1 Wrestling, Stash with bag of 10 seeds, and a sickle.
image
What seeds inside the bag.
image

Also added, Seelie language to Polyglot special.
image

## Why It's Good For The Game
Hoping this special would add more people to play peasant roles, and having specials for less played roles is good.


Nakalon:
Good work!!!

PR #1166, opened 2024-10-19
Goblin Rework(Ish)

ProbablyCarl:
## About The Pull Request
- - -
Player Facing:
 - Goblin roles re-enabled once again, this time with a bit of a facelift. You're still required to have Graggar as a patron.
 - Goblin Guard and Chief are still gender locked.
 - Goblin Chef and Smith are now free to play for both genders.
 - Goblin Rabble added. This is the 'commoner' equivalent for the Goblin encampment. They're not given access to the interior by default, having to wait for it to be opened. This, alongside Smith and Chef, are two introductory roles.
 - Player Goblins should now actually be able to wear their armour. Rejoice?
 - The Goblins have travel tiles, allowing them to go from their fort to the sewers, as was requested.
- - -
Other:
 - Goblin camp deleted from the game files. Didn't make sense and had only served to uglify the file structure even worse than it had already been.
 - Goblin Camp map pass done by @NikolayByondPlayer.
- - -

## Why It's Good For The Game
Objectively, I couldn't tell you one way or another. This was done on request, for which I figured I'd give it a bit of a lookover. It's still really rough, especially with how it had been added in the first place, but it should be better than prior. The issues of fragging can be solved with administrative action.

### DIFF
--- a/_maps/map_files/generic/CentCom.dmm
+++ b/_maps/map_files/generic/CentCom.dmm
@@ -600,7 +600,7 @@
 	},
 /obj/item/natural/feather,
 /turf/open/floor/rogue/hexstone,
-/area/rogue/indoors)
+/area/rogue/indoors/banditcamp)
 "bk" = (
 /obj/item/inhand_tester/big,
 /turf/open/floor/rogue/blocks,
@@ -984,13 +984,15 @@
 /area/rogue/indoors)
 "cw" = (
 /obj/effect/landmark/start/bandit{
-	dir = 4
+	dir = 4;
+	icon_state = "arrow"
 	},
 /obj/structure/chair/wood/rogue{
-	dir = 8
+	dir = 8;
+	icon_state = "chair2"
 	},
 /turf/open/floor/rogue/herringbone,
-/area/rogue/indoors)
+/area/rogue/indoors/banditcamp)
 "cx" = (
 /obj/structure/table/wood{
 	dir = 10;
@@ -1110,7 +1112,7 @@
 "cR" = (
 /obj/structure/roguewindow,
 /turf/open/floor/rogue/dirt/road,
-/area/rogue/indoors)
+/area/rogue/indoors/banditcamp)
 "cS" = (
 /obj/structure/mineral_door/wood/donjon{
 	dir = 4;
@@ -1247,7 +1249,7 @@
 /obj/item/paper,
 /obj/item/paper,
 /turf/open/floor/rogue/hexstone,
-/area/rogue/indoors)
+/area/rogue/indoors/banditcamp)
 "dm" = (
 /turf/open/floor/rogue/blocks/stonered,
 /area/rogue/outdoors/woods)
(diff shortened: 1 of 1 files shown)


hugmeorelse:
> > > > Absolutely against having Goblin Kingdom ever again. Goblins yes, but not in this form.
> > > 
> > > 
> > > Was at the request of another individual. I'm not too familiar with the problems that this brought. Care to elaborate? We could probably fix it easily, like a lot of the antags.
> > 
> > 
> > Off the top of my head:
> > 
> > * Too many "external" roles making the town feel even emptier, when the server is trying to shift towards town centric gameplay and roleplay flow.
> > * Goblin King making global announcements everyone can hear, resulting in announcement battles between goblin king and king.
> > * Goblin base getting regularly raided by the whole server for loot or for carnage. (And it looked nothing like a goblin base)
> > * Goblin players misbehaving a lot, but arguably not specifically goblin kingdom related.
> > 
> > I dont mind playable goblins at all, but I was never a fan of the whole "Goblin Kingdom" thing. They are tribal not feudal, they can survive in the bog, make their own base and specialize themselves in different roles without the need for 5 different jobs and a hardcoded camp in my opinion.
> > Goblin Kingdom is cursed, it didnt survive a TM neither here nor on Blackstone back when it was first coded. We must find a better way.
> 
> Provided I can get a map for it, I'd absolutely be willing to flavor this more towards a tribal sort of setup. I'm not particularly creative when it comes to mapping, so my areas would look worse than the current 'Kingdom'.

We could possibly put them in the sewers and maybe have them be like a sewer criminal gang if we want a more town-centric gameplay. Just requires some tweaking.

Nakalon:
> We could possibly put them in the sewers and maybe have them be like a sewer criminal gang if we want a more town-centric gameplay. Just requires some tweaking.

i think this idea is extremely interesting!

radarseas:
not sure how I feel with 6 roundstart guards and 2 smiths for a heretical-worshipping goblin camp, its good they're defensible but.. I think 2 smiths is waaaay too many for these guys, let alone 6 guards at round-start even if they do lack armor.


Do they have anything to heal fractures and injuries? else these guys are gonna get paralyised and fractured a lot without alchemy or surgery tools and skill, let alone miracles of any kind.

Hmeister-real:
goblins cannot wear goblin armor, and they can wear normal humen armor. This needs to be fixed ASAP

wow50000:
the forced graggar will always result in the goblins being purged like bandits, especially since the church will always use that reason to purge them.

NameChangeStonekeep:
Why the fuck did you give my female goblin sprites antennae on your server and screw them up with these ugly skin tones? i permitted them to be used. And not butchered by others, I'm considering revoking their use of them if you ain't going to respect the sprite work

Nakalon:
Sadly this seems to thing the town more than it already is. 90 player rounds and the town still seems deserted.

This may be a cool feature but it does have some side effects...

Gyrandola (approved):

PR #1175, opened 2024-10-20
[Do not merge] items crafting readability treatment

Red-Honk:
## About The Pull Request

Follows the spirit of this PR: https://github.com/Rotwood-Vale/Ratwood-Keep/pull/1173.

# [Changelog]:
[Baits]: Bag baits no longer need a table to be crafted.

### DIFF
--- a/code/modules/roguetown/roguecrafting/items.dm
+++ b/code/modules/roguetown/roguecrafting/items.dm
@@ -5,7 +5,7 @@
 /* craftdif of 0 = NONE */
 
 /datum/crafting_recipe/roguetown/woodcross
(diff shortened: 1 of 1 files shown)


Nakalon:
> Any update on this?

I would advise against merging these.
It is unnecessary verbose at some points (some recipes include NONE as a crafting requirement), leads to events of "Player sewes a CLOTH (2 FIBERS, NONE)" appearing in chat and does less to actually organise the recipes in a way that is easily navigable than would lead you to believe.

I believe crafting has a future much better than the system it currently has, however I don't think this is a step in the right direction.

The core idea is good but it would have to be handled in a much more complex way and less so in the "raw data input" sort of way it's being done here

PR #1184, opened 2024-10-21
Fix and balances

Luctus-Comptus:
alot in this, my usual mixed fix's and stuff

*werewolves after the PR was reverted that everyone had wanted reverted for a long time... was way too overtuned, now before you have a stroke this is on observation and feedback. infi stam for them was removed and about everything from stats to armor was dialed back about 30% across the board. so see where it is at and needs adjustment. this was to avoid an outright disable till rework, you might be angry about it but they could get the VL treatment... oh and I put an ear fuck timer on the play global sound of there howls, first howl of the night from every werewolf will be heard. then the sounds put on a 10min timer. the actual HOWL talking is still only 1min. but no more sound spam.

*upcoming medical changes coming I did some groundwork on. end goal tighten the gap in healing between doc and miricle spammers with miracles eventually being slightly stronger. on this pass, sewing, cauterizing and bandaging going thru clothes and I made the cauterizer docs get not useless. its self heating now. toggle similar to a stun mace.

*arteries are back up to a crit bleed out being bad. since you can now bandage and sew thru clothing the arterial bleeds was brought back to where they was before. embed chance for weapons on a crit droped from 75% to 40% and disembowling not happening possibly fixed back to...actually working.

*devotion; made less painful to level up. 50% increase to baseline prayer gains to all. break points for tiers lowered to gain the spells faster. pestrans got start spells combo'd to diagnose AND lesser so they dont have to pay for an eternity to get cure rot, its now tier 3.

*misc tweaks; adjusted purity prices for smokes and stuff, adjusted stats by 1 for forge cleric and created a gimped shittier astra fire spell for them to start with to light forges and shit with, gave an axed and 2 in axes to deal with skellies since hammer is a joke, backpack for the smith starter supplies they had that completely filled their satchel.

### DIFF
--- a/code/controllers/subsystem/rogue/devotion.dm
+++ b/code/controllers/subsystem/rogue/devotion.dm
@@ -6,10 +6,10 @@
 #define CLERIC_T4 4
 
 #define CLERIC_REQ_0 0
-#define CLERIC_REQ_1 100
-#define CLERIC_REQ_2 250
-#define CLERIC_REQ_3 500
-#define CLERIC_REQ_4 750
+#define CLERIC_REQ_1 75
+#define CLERIC_REQ_2 150
+#define CLERIC_REQ_3 350
+#define CLERIC_REQ_4 500
 
 // Cleric Holder Datums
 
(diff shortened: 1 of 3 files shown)


katsukai:
does a cloth really stop an artery?

wow50000:
would rather have had a mace

PR #1189, opened 2024-10-21
Map fixes, again

Heheniy:
## About The Pull Request
Generally same but trees almost not touched because people love slop and hate good chages

Replaced bones at south-west bog mine with proper ones, that can be passed
Fixes double closet at hand's room
Fixed torches inside walls where i managed to find them
Added lights to admin test area
Moved shylock near smiths to be between peddlers
Added shylock near merchant shop
Added shylock to steward tower
Made merchant's goldface look less out of place
Standartized stone keep walls on 1st floor
Added light to dark areas in town
Added second key chest to steward and separated non-keep keys into it
Fixed some areas and caves (ie replaced all "indoor" areas in terrorbog with just "bog" because of how "indoor" works (blocking sky even withut roof), highground at eastern town near church in not "roofs" but just town "outdoors", fixed mountain areas like on screenshot below, etc)
image

Expanded deepcaves
image

No more easy access to vault from river
image

Farmhouse door starts unlocked (if there any farmers spawned they can lock it, if not- no need to break in to get tools for farming)
Various small tweaks around basin and terrorbog


Heheniy:
fuck

Heheniy:
i guess i need to remake this again...

Issue #1178, opened 2024-10-20
[BUG] Unexpected Behavior for Farm Animals


2mgmozga:
## Issue

Farm animals in the village are supposed to be moving around. I have noticed over multiple rounds over the last few weeks that they just stand there idly and refuse to eat. In the chance that they get offspring, that offspring does move around normally, but the starting animals still stay in place.

## Round ID: 
Over the course of multiple rounds

## Steps to Reproduce: 
It happens every round. There are no specific steps to reproduce

## Additional context

This makes butcher barely playable as you are just hoping that the animals breed and make unbugged offspring.

PR #1190, opened 2024-10-21
AI tick from 20 to 5

Gyrandola:
I am quite convinced whoever pushed for such a big delay on the NPC subsystem doesnt understand how it works.
With such a low tickrate, NPCs barely react and are incredibly easy to cheese.

Should also resolve https://github.com/Rotwood-Vale/Ratwood-Keep/issues/1178.

For reference:
The original tick was 5
Neon lowered it to 1, making npcs too reactive
Someone else upped it to 20, making them lobotomized.
I am reverting it to the original value.

For those who claim we need to compensate for lag, the correct solution would be removing all the unnecessary micro-dungeons that bloat the system. There are like 50 Skeletons, 30 spiders and 20 goblins at round start. For fucks sake.

Morale of the story please dont touch AI.

### DIFF
--- a/code/controllers/subsystem/rogue/humannpc.dm
+++ b/code/controllers/subsystem/rogue/humannpc.dm
@@ -1,7 +1,7 @@
 
 SUBSYSTEM_DEF(humannpc)
 	name = "humannpc"
-	wait = 20
+	wait = 5
 	flags = SS_KEEP_TIMING
 	priority = 50
 	var/list/processing = list()


Nakalon:
Remember the original proposed value on the PR was 40 😭 😭 😭 😭

hugmeorelse:
As has been stated, AI has been optimized in other ways as well beyond that tick value being changed. Changing it back to 5 will not replicate the same behavior as we had originally.

You had upped it from 1 to 3, but the change was still incredibly brutal without really much of a difference for players. 3 to 5 would not work out as you would think.

Based on feedback from numerous people, 15 seems to be around the sweet spot. We were testing out 20 to see how it felt compared to 15. Feedback for 20 is that it should be a little faster.

hugmeorelse:
> I see it with my own eyes whenever I hop on. NPCs have a 2 seconds delay and they are unable to chase someone. Extremely cheesable.
> 
> These fabled AI optimizations that for some reason makes upping the entire AI subsystem tick to 20 reasonable havent been explained to me yet.
> 
> Also NPCs have very recently been given limited stamina, we can definitely go back to the original value and cut the ignorance.

Read above. I posted the feedback I've received.

Also, the PR that gives NPC staminas did not appear to work at the time for players as was reported, hence why the tick rate was managed. Even at 15, as someone noticed, they can bash down a shield faster than a player can.

hugmeorelse:
> I have done some more testing on the live server and I have to concede spawned/ambush NPCs handle themselves decently. I believe round-start NPCs like the dungeon ones may be strangely affected in some way, as earlier today I observed quite the opposite in terms of reactivity, and the round-start farm animals are also acting weird.
> 
> Will close this until I find out more.

I appreciate you taking the time to investigate further.

PR #1191, opened 2024-10-21
Double bed craft

Heheniy:
## About The Pull Request
Makes possible to craft double beds, both their variants

Changes nicebed craft result
on screenshot below from variant on the left to the one on the right
image

### DIFF
--- a/code/modules/roguetown/roguecrafting/structure.dm
+++ b/code/modules/roguetown/roguecrafting/structure.dm
@@ -448,7 +448,7 @@
 
 /datum/crafting_recipe/roguetown/structure/nicebed
 	name = "nice bed"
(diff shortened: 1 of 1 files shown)


Heheniy:
> there one problem with that. there is a sleep quality buff for each bed type. inn beds are extremely extremely good. faster healing, fster wound recovery

so i made them require higher skill to craft

Heheniy:
> oh hay your right, was thinking of a different bed. you can drop the single back down to 2 and double to 2 as well. no need to bump it so high for a craft.

they look kinda fancier so i dont want them to be spammed and still be somewhat an indicator of wealth

Heheniy:
you are free to make any adjustments

PR #1192, opened 2024-10-21
Removes latejoin queen consort

Heheniy:
## About The Pull Request
because king can get married during round and then latejoining queen starts to cause drama noone wants


katsukai:
I do agree that this makes sense.

Marrone321:
What say you to a timer, such as an hour and a half the Queen Consort role closes up, as opposed to after the round starts?

wow50000:
queen being away to kingsfield was always funny though.

PR #1194, opened 2024-10-21
Changes forge cleric int for con

wow50000:
## About The Pull Request
See title

## Why It's Good For The Game
Forge clerics as of now have the skills for adventuring but also the int to learn how to replace blacksmiths due to having more int, this should keep their smithing skills below blacksmiths as well as boosting their combat durabilities while also keeping it lower than other combat roles.


Elli-Skala:
> no

Could you explain why

PR #1210, opened 2024-10-22
Treasury is now actual bank

Heheniy:
## About The Pull Request
All deposints and withdraws rom treasury now affects treasury itself instead of only bank accounts of people
Fining and giving money via steward console only affects people accounts, not treasury itself
Now treasury IS lord's account (kinda) (amount of money in lord's account always equal to treasury)
When priest coronates a person, a new bank account is being created for them (if they dont have one) and treasury is being assigned to it, while old king looses everything
Kingsfield tax randomised more properly
Hand now receives wealth horde info too

One known issue is that wealth horde can trigger twice roundstart and those money will not properly assign to lord account

### DIFF
--- a/code/controllers/subsystem/rogue/treasury.dm
+++ b/code/controllers/subsystem/rogue/treasury.dm
@@ -103,7 +103,7 @@ SUBSYSTEM_DEF(treasury)
 /datum/controller/subsystem/treasury/proc/update_ruler_money()
 	for(var/mob/living/carbon/human/X in GLOB.human_list)
 		switch(X.job)
-			if("Lord")
+			if("Duke")
 				bank_accounts[X] = SStreasury.treasury_value
 				return
 
--- a/code/modules/jobs/job_types/_job.dm
+++ b/code/modules/jobs/job_types/_job.dm
@@ -191,7 +191,7 @@
 	if(give_bank_account)
 		if(give_bank_account > 1)
 			SStreasury.create_bank_account(H, give_bank_account)
-		else if(H.job == "Lord")
+		else if(H.job == "Duke")
 			SStreasury.create_bank_account(H, SStreasury.treasury_value)
 		else
 			SStreasury.create_bank_account(H)


Nakalon:
> One known issue is that wealth horde can trigger twice roundstart and those money will not properly assign to lord account

Known issue you introduced?
Known issue that you resolved?
Or Known issue that just exists and you aren't changing at all?

Heheniy:
Known issue that kinda existed but was affecting nothing and just now has affect in the form of difference between treasury money and money awailable to lord in amount of 1 wealth horde (around 15 mammon)

It is not cricical but i am not able to catch what causes that

Nakalon:
> Known issue that kinda existed but was affecting nothing and just now has affect in the form of difference between treasury money and money awailable to lord in amount of 1 wealth horde (around 15 mammon)
> 
> It is not cricical but i am not able to catch what causes that

Then I advise you find a way to solve this issue as adding a feature that is built on something that is bugged is sure to cause problems down the line and is, generally, not good practice!

I'll have a look later see if I can help you, if you'd like

code/modules/jobs/job_types/roguetown/nobility/lord.dm
```diff
@@ -28,7 +28,7 @@ GLOBAL_LIST_EMPTY(lord_titles)
 	whitelist_req = FALSE
 	min_pq = 10
 	max_pq = null
-	give_bank_account = 1000
+	give_bank_account = TRUE
```
Nakalon, line 31: Regarding this, currently, for some reason most times this var appears I'm jobs it has an arbitrary value. Since you're fixed that here please do so with all other jobs too since you're already touching this system please!

Nakalon (changes requested):

code/modules/jobs/job_types/roguetown/nobility/lord.dm
Heheniy: look at code/modules/jobs/job_types/_job.dm 188 line

Nakalon: So you're telling me account have their own number and aren't directly tied to the player?
If so setting it to TRUE or FALSE would be entirely the same as TRUE is 1 and that checks for numbers superior to 1. I need to check this again when I'm back home.

Nakalon, line 31: I have to get back to this:
image

What this equates to is supposedly initial deposit.

If the value is FALSE they get no account, if the value is TRUE they get an empty account, if the value is superior to 1 they get an initial deposit.

What I'm asking you is that you standardize this var on other jobs:

image

As they seem to have random numbers.

Even I fucked it up and gave the seamstress 2 for absolutely no reason.

image

What I think happened here is, some people who touched this var, like me didn't understand its purpose fully and just thought it meant "bank account number".

I also ask that you add a commend above the var to describe what it does in _job.dm like:

image

Cheers!

Nakalon (changes requested):

code/modules/jobs/job_types/roguetown/nobility/lord.dm
Nakalon: To pile on this, there, in my view, isnt much sense in givign working classes an initial investment since they already spawn with mammons on their person.

Nakalon:
You didn't change the other jobs like I suggested 😭

wow50000:
honestly how it should have acted all along in my opinion

radarseas:
finally, no more worry about storing my mammon in case the steward decides to fine everyone to cover a redicious purchace, because it won't benefit them doing so.

Heheniy:
> This is likely being reverted due to reports of players being able to withdraw infinite money from the treasury/bank.

give more info

PR #1212, opened 2024-10-22
Added invisible bars for inn 

NikolayByondPlayer:
## About The Pull Request

Created invisible bars with the help from the chat and Eora, thank you lads
When you throw items you can hear iron sound, but that will do for now I think, might create a special sprite and inn table some time later 
image

## Why It's Good For The Game

Make it less prison like while still making it possible not to let retarded behaviour be

### DIFF
--- a/_maps/map_files/rockhill/rockhill.dmm
+++ b/_maps/map_files/rockhill/rockhill.dmm
@@ -6083,7 +6083,7 @@
 	climbable = 0
 	},
 /obj/structure/bars{
-	icon_state = "barsbent";
+	icon_state = "openspace";
 	layer = 2.81
 	},
 /turf/open/floor/rogue/ruinedwood/spiral,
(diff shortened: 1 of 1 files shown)


Nakalon:
Good one!

Nakalon:
> Maybe it's just me, but this seems like a bad idea. A structure thats invisible and blocks passing is pretty eh at best, and downright bad at worst.
> 
> Frankly- I'm not even sure why we put bars on the tavern bar anyway? Someone hops the counter, it's a good reason for a tavern brawl. It's not like theres much that can be robbed, like other places such as the merchant or smithy.

I changed my mind. While I understand people may jump the counter and be asses that creates RP by creating a situation that needs solving!!!

Nakalon:
> Good one!

Edit: I changed my mind :(

PR #1214, opened 2024-10-22
Rope and Chain Restraints Breakout Timer Increased

MickDick1:
## About The Pull Request

Changed rope breakouttime to 10 seconds from 5 seconds.
Changed chain breakouttime to 30 seconds from 10 seconds.
Changed chain slipouttime to 3 minutes from 2 minutes

## Why It's Good For The Game

Neon had a pet peeve for restraints actually taking longer than a few seconds to BREAK for some reason. Simply having 11-14 strength means you can snap rope and chains in 5 and 10 seconds respectively due to his changes, whereas it used to be 5 seconds for rope and 1 minute for chains.

Otherwise, slipouttime is called if you're 10 strength or below, and instant break if you're 15 strength.

Being /slightly/ above average in strength should not mean you can SNAP METAL CHAINS IN TEN SECONDS. Rope was slightly buffed also. Restraints now actually work.

### DIFF
--- a/code/game/objects/items/rogueitems/ropechainbola.dm
+++ b/code/game/objects/items/rogueitems/ropechainbola.dm
@@ -12,7 +12,7 @@
 	w_class = WEIGHT_CLASS_SMALL
 	throw_speed = 1
 	throw_range = 5
-	breakouttime = 5 SECONDS
+	breakouttime = 10 SECONDS
 	slipouttime = 1 MINUTES
 	var/cuffsound = 'sound/blank.ogg'
 	possible_item_intents = list(/datum/intent/tie)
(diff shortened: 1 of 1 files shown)


Nakalon:
Ahhh I see it now! So it's better to have low strength? That's odd!

Nakalon:
Wait holy fuck 3 minutes?!? I hadn't seen that. That's so long.

Cyprex:
Chains were adjusted in, the times were much longer originally
https://github.com/Cyprex/Ratwood-Keep/commit/f1ba0da29f35f225a378f2701c3be7d6a6eca03f

Cyprex:
The whole slipout time being for lower strength makes no sense tbf. I'm not sure this should scale with the users stats at all.

PR #1217, opened 2024-10-22
Ports various fixes from Azure

Cyprex:
## About The Pull Request

Ports these PRs, credit goes to @noelle-lavenza 
https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/573
https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/626

It's worth noting I didn't grab the mote code- we have that too(in another file) but our code is somewhat different.

Note that this *might* end up buffing NPCs in situations where there are a lot of them- Before this PR there is an artificial limit to process only 10 per Subsystem run, now it runs for however long the MC allows it to(The arbitrary restriction was bad.) We'll have to see if this requires adjustments to the SS wait or things like that!

### DIFF
--- a/code/game/objects/items/rogueitems/needle.dm
+++ b/code/game/objects/items/rogueitems/needle.dm
@@ -69,7 +69,7 @@
 
 /obj/item/needle/attack_obj(obj/O, mob/living/user)
 	var/obj/item/I = O
-	if(istype(I) && can_repair)
+	if(can_repair)
 		if(stringamt < 1)
 			to_chat(user, span_warning("The needle has no thread left!"))
 			return


code/game/objects/items/rogueitems/needle.dm
```diff
@@ -69,7 +69,7 @@
 
 /obj/item/needle/attack_obj(obj/O, mob/living/user)
 	var/obj/item/I = O
-	if(can_repair)
+	if(istype(I) && can_repair)
```
Nakalon, line 72: Can you leave this one for me pretty please? I'm already retouching this code and don't want to have to fix a merge conflix even if its super easy!

Reference: #1222

Nakalon: I also just discovered in our code "can_repair" does absolutely nothing wow

Nakalon: AHHH I SEE WHY, The var exists because theres supposed to exist a medic only needle, WHICH MAKES SENSE! I'll probably add that at some point!

Cyprex: Will do!

PR #1226, opened 2024-10-23
Remove fox features from Lupian

Dedwolf:
## About The Pull Request

Title

## Why It's Good For The Game

Play vulp if you want to be a vulp. Playing lupian for stats is absolutely lame. 

Side note, when are we switching to stat packs other servers use?

### DIFF
--- a/code/modules/client/customizer/customizers/organ/ears.dm
+++ b/code/modules/client/customizer/customizers/organ/ears.dm
@@ -28,11 +28,9 @@
 	sprite_accessories = list(
 		/datum/sprite_accessory/ears/bigwolf,
 		/datum/sprite_accessory/ears/bigwolf_inner,
-		/datum/sprite_accessory/ears/fox,
 		/datum/sprite_accessory/ears/husky,
 		/datum/sprite_accessory/ears/otie,
 		/datum/sprite_accessory/ears/lab,
-		/datum/sprite_accessory/ears/vulp,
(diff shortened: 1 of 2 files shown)


Nakalon:
IMO this is a bit of an non issue and ultimately will slightly decrease a players options when creating their character.

I believe the problem that should be looked at here is the claim people play lupian for stats instead of vulp.

I really don't like touching people's characters like that.
Plus I don't know what would happen to characters who have these ears right now. Will they have no ears? Will it pick at random? Will it runtime?

Purpleturtle3:
There is no reason for the wolf race to be able to have fox characteristics while the game also has a fox race. Less player freedom is good sometimes.

Purpleturtle3:
I'm tired of my clearly orange fox with fox ears and tail being referred to as a lupian because, for some reason, lupians can look like foxes too. If you want to play a hybrid, wildkin is right there.

Nakalon:
> If you want to play a hybrid, wildkin is right there.

I forgot this was a thing. Alright, I change my mind.

Kepteyn:
If anything, vulps and lupians should be merged into just lupians. They're virtually the same creachur.

RadiantFlash7:
> you people realize hybrids exist, i play a fucking nightstalker for christ sake. thats snake, wolf, fox and coyote mixed

Isn't that as a wildkin though? Thats the race thats like classic 'furry' catch all? Or am I dumb? xD

PR #1232, opened 2024-10-23
spooning

Luctus-Comptus:
* spoon, platter, roll pin craftable

*added to needed area's


LofiOtter:
I have no idea why after peeping over the code, but spoons are not craftable for some reason. This made me very sad and I keep having to buy the extra spoon from the innkeep 😭

PR #1239, opened 2024-10-23
shakey shakey my tail is snakey AKA; mori gives you tail.

Luctus-Comptus:
there was a wrong that needed to be set right. I've added this on every single codebase I ever worked for. colorable rattler tail.

https://github.com/user-attachments/assets/fde3dd16-8fe8-4e05-ac13-d859e0c76608


Nakalon:
Holy shit thats a meaty tail

PR #1242, opened 2024-10-24
Goblin Bounty Squishing

Squird5:
Squish Goblin heads in the head squisher

## About The Pull Request

https://github.com/user-attachments/assets/33c7f885-3c9d-4d69-a49d-7625df9c9ea2

## Why It's Good For The Game

You can now put Goblin heads in the Head Squisher to collect their bounty, based on the same bounty price the Gizan Blimp buys them at.
This reduces the amount of time the Merchant has to spend fucking around trying to sell someone's goblin heads for all of 3 coins or some such.

**Additionally** I corrected two files that were kicking minor errors in the debugger.

### DIFF
--- a/code/__DEFINES/roguetown.dm
+++ b/code/__DEFINES/roguetown.dm
@@ -273,12 +273,6 @@ GLOBAL_LIST_EMPTY(job_respawn_delays)
 #define TRIUMPH_CAT_MISC "MISC!"
 #define TRIUMPH_CAT_ACTIVE_DATUMS "ACTIVE"
 
-
-#define TRIUMPH_CAT_ROUND_EFX "ROUND-EFX"
-#define TRIUMPH_CAT_CHARACTER "CHARACTER"
-#define TRIUMPH_CAT_MISC "MISC!"
-#define TRIUMPH_CAT_ACTIVE_DATUMS "ACTIVE"
-
 /*	........   Nutrition defines   ................ */
 #define MEAL_FILLING 30
 #define MEAL_GOOD 24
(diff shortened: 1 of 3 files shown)


RadiantFlash7:
I do like more passive ways to earn money that allows adventurers to interact with the economy of yeoman more...

code/__DEFINES/roguetown.dm
```diff
@@ -273,12 +273,6 @@ GLOBAL_LIST_EMPTY(job_respawn_delays)
 #define TRIUMPH_CAT_MISC "MISC!"
 #define TRIUMPH_CAT_ACTIVE_DATUMS "ACTIVE"
 
-
-#define TRIUMPH_CAT_ROUND_EFX "ROUND-EFX"
```
Nakalon, line 277: I assume this was what was causing errors?

Nakalon:
This is great by the way!! very nice!!

SayonaraEcho:
Finally....

code/__DEFINES/roguetown.dm
Squird5: Duplicate defines. Kicks a minor yellow error because it's trying to define the same thing twice. It was just annoying to see every time I compiled.

Squird5:
> skelly heads was also one as well ya could do if you fancy in the future, they got bounty value as well.

I checked into that but didn't see that skeleton heads had a value. Goblin skeleton heads do, but only a value of 2 so I didn't bother trying to test if those are also accepted. If so, cool. If not, oh well.

PR #1244, opened 2024-10-24
Helmets and plated chausses now have armor classes

RadiantFlash7:
## About The Pull Request
This PR Adds armor class to helmets and plated chausses. 
Specifically: Any helmet above 250 integrity, OR having FOV Restrctions (usually caused by having a face covering visor toggle) is medium armor. All 'heavy' helmets, listed under the heavy type of helmet, are now heavy armor.
Plated chausses were also given heavy armor tag as well. I was debating making chain chausses medium, but theres no real 'good' armor that can act as light armor for legs in it's place, like there is for helmets.

## Why It's Good For The Game
It became a standard for many classes without heavy armor wearing to wear 'heavy' pieces on their head and legs, because the only thing that was checked was the chest plate. This corrects that particular bit of interaction by causing the same no sprint effect.

### DIFF
--- a/code/modules/clothing/rogueclothes/hats.dm
+++ b/code/modules/clothing/rogueclothes/hats.dm
@@ -626,6 +626,7 @@
 	emote_environment = 3
 	flags_inv = HIDEEARS|HIDEFACE|HIDEHAIR
 	block2add = FOV_BEHIND
+	armor_class = ARMOR_CLASS_MEDIUM	//breaks the 'scheme' of armor class, because it's a unqiue helm, that can't be remade. Go forth, gatemaster.
 	smeltresult = /obj/item/ingot/steel
 
 /obj/item/clothing/head/roguetown/helmet/heavy/knight


RadiantFlash7:
> I'm in the process of adding brigandine armor for the chest, arms and legs which should fit the role of medium/light armor for the legs and arms. Just waiting on sprites.

I'm not too fussed about arm and legs specifically because theres so few options for armor there. You either get steel bracers or leather ones, or none at all. Same with shoes.
However, people have absolutely been running around with steel plated chausses and visored sallets as rogues, while using dodge builds 💢

john-pawson:
Alright i absolutely agree with the plated chausses, but i don't see why helmets are the ones getting the armor classes?? 
A helmet definitely doesn't hinder movement enough or is heavy enough to have any kind of effect on dodging and running, it also means that bards and monks won't be able to use a helmet that protects the face. They already hinder vision and are one of the most important pieces of armor, why make it even more of a pain to wear now??

Besides making every single heavy helmet need heavy training is definitely WAY too much since almost nobody has heavy training, it also means that you cannot wear any helmet without >220 integrity and that protects the face without all the drawbacks. It'll just mean that people will stop using them and now always go for a mask + normal sallet and NOTHING ELSE.

CloverIsLucky:
I also disagree that helmets need their own armor class. Very few roles have heavy armor trainning, but the chauses makes a bit of sense. Meanwhile plate boots should decrease movement speed like wearing clown shoes from SS13.

Hmeister-real:
helmets are heavy as fuck, they should have an armor class. I love this PR so much <3

RadiantFlash7:
> Alright i absolutely agree with the plated chausses, but i don't see why helmets are the ones getting the armor classes?? A helmet definitely doesn't hinder movement enough or is heavy enough to have any kind of effect on dodging and running, it also means that bards and monks won't be able to use a helmet that protects the face. They already hinder vision and are one of the most important pieces of armor, why make it even more of a pain to wear now??
> 
> Besides making every single heavy helmet need heavy training is definitely WAY too much since almost nobody has heavy training, it also means that you cannot wear any helmet without >220 integrity and that protects the face without all the drawbacks. It'll just mean that people will stop using them and now always go for a mask + normal sallet and NOTHING ELSE.

Heavy helmets are /heavy/. They have over 400 integrity. Thats almost as much as plate armor. You really shouldn't be wearing something like that without training tbh. And at present, most of the roles who wear them, do have heavy armor training. Such as knights, and templar. That said, I think the sole exception to this, may need to be the gate masters helmet. Especially since it's non craftable.

Most helmets remain medium or light, which most people can wear as is. That said, It's preferable for people to have a mask, seperate from the helmet, because masks are at most, 200 integrity. Heavy helmets as mentioned prior, are 400.  Monks and rogues do not need 400 integrity head armor, lets be honest here. 

The drawback to wearing a mask and a helmet, vs a visored helmet is pretty clear you can't open it/close it as easily mid fight, if someone is aiming at your face, so you lack as many options when it comes to vision.

john-pawson:
> Heavy helmets are /heavy/. They have over 400 integrity. Thats almost as much as plate armor. You really shouldn't be wearing something like that without training tbh. And at present, most of the roles who wear them, do have heavy armor training. Such as knights, and templar. That said, I think the sole exception to this, may need to be the gate masters helmet. Especially since it's non craftable.

A heavy helmet is probably as light if not lighter than any piece of chainmaille or plated boots/gloves with at MOST a weight of 8lbs, they're "/heavy/" because they're the most expensive and sacrifice vision, 3/4 if they're visored and 1/2 if they're not, and they should be kept with 400 integrity, if not then fights will end up boiling down to "attack the head until helmet is broken".

> Most helmets remain medium or light, which most people can wear as is. That said, It's preferable for people to have a mask, seperate from the helmet, because masks are at most, 200 integrity. Heavy helmets as mentioned prior, are 400. Monks and rogues do not need 400 integrity head armor, lets be honest here.

Well, same problem here, will end up boiling down to "attack head and break helmet", besides they already are limited to a very easily broken haubergeon or leather armor, there's no need to also fuck them over on the head department too. It'd be like making the bevor a heavy equipment training needed, it's one of the most important pieces of equipment and it protects one of the vulnerable parts in a fight, why make it more of a pain when they already have weak armor on every other single bodypart?

Nakalon:
> ## About The Pull Request
> This PR Adds armor class to helmets and plated chausses. Specifically: Any helmet above 250 integrity, OR having FOV Restrctions (usually caused by having a face covering visor toggle) is medium armor. All 'heavy' helmets, listed under the heavy type of helmet, are now heavy armor. Plated chausses were also given heavy armor tag as well. I was debating making chain chausses medium, but theres no real 'good' armor that can act as light armor for legs in it's place, like there is for helmets.
> 
> It became a standard for many classes without heavy armor wearing to wear 'heavy' pieces on their head and legs, because the only thing that was checked was the chest plate. This corrects that particular bit of interaction by causing the same no sprint effect.

Hello!

I agree with this, I'm a big heavy armor hater!

I'm working on light armor based on leather for the hunter and seamstress which will hopefully fill in the gap!
Cheers!

Nakalon:
> I also disagree that helmets need an armor class. Very few roles have heavy armor trainning, but the chauses makes a bit of sense to need one. About plate boots, these should decrease movement speed like wearing clown shoes from SS13.

I do not understand armor training at all.
To me armor type should be a choice!
IMO:
Light armor should be nimble and better at defending against blunt attacks!
Medium armor should be a bit better but less nimble and not as good at blunt attacks
Heavy armour should make you almost unkillable but SLOW

I think skill gatekeeping content, when it can't be trained, does more harm than good!

PR #1246, opened 2024-10-24
Welcome to the Duchy of Rockhill

Useroth:
Replaces all mentions of the lord/lady with duke/duchess where appropriate.
Also some minor changes to text where I noticed "his majesty" and similar.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/nobility/captain.dm
+++ b/code/modules/jobs/job_types/roguetown/nobility/captain.dm
@@ -11,7 +11,7 @@
 	allowed_sexes = list(MALE)
 	allowed_ages = list(AGE_MIDDLEAGED, AGE_OLD)
 	tutorial = "You served your time gracefully as a soldier, and now you've grown into a role which many men dream to become. \
-				You are the Marshal's chosen, elevated to command His Highness's personal retinue. \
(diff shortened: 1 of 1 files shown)


code/modules/jobs/job_types/roguetown/nobility/lady.dm
```diff
@@ -19,7 +19,7 @@
 	max_pq = null
 
 /datum/job/roguetown/exlady //just used to change the ladys title
-	title = "Lady Dowager"
+	title = "Duchess Dowager"
```
BlueMemesauce, line 22:
```suggestion
	title = "Dowager Duchess"
```

code/modules/jobs/job_types/roguetown/nobility/lady.dm
```diff
@@ -1,5 +1,5 @@
 /datum/job/roguetown/lady
-	title = "Lady Consort"
+	title = "Duchess Consort"
```
BlueMemesauce, line 2: duchess consort doesn't seem to be an actual thing as irl duchess consort are just called duchess, but it should be fine just to make it more obvious to players (thats why i added consort to queen in the first place, because players didnt realize queen doesnt actually have legal power)

Squird5:
Nice. Now the Prince and Princess can be renamed to Lord and Lady properly.

PR #1253, opened 2024-10-24
Buffs bandit speed

wow50000:
## About The Pull Request
Buffs bandit base speed to 2, removes hedge knight speed nerf, may make cleaver bandit op. Removed the 1 or 2 extra sword skill putting hedge knight at 4 or 5, making it 4 at all times, remove the str bonus from hedge knight

## Why It's Good For The Game
"ON THE TOPIC OF BANDITS

Look, it's time I put this to proverbial paper. Bandits are in a sorry state and the random loadouts don't help them in the slightest. The current bandit stats are based around the idea that there would be 15-20 of them. This simply isn't the case on Ratwood and likely never will be expect for on very, very rare occasions.

There are some pressing issues but to dumb it down and make it as reasonable and digestible as possible, I've made a short list.

1) Change of the +1 speed to +2, speed is the single biggest tool in keeping yourself alive as a bandit, there isn't any argument against it. The ability to effectively disengage from conflict is how you thrive.

2) Removal, speed buffing  or ability to opt out of Hedge Knight. Hedgie at the moment is generally considered a burden between the massive -2 speed and the armor taking a forever to take off without a dedicated helper. On top of that you generally never have the ability to get metal gear repaired as a bandit so after a fight or two their armor just gets thrown away or melted down. The weapons they get are a massive fucking let down and don't do anything great, plus being in the bog means you NEED to have a polearm on you at all times or simply die.

3) The ability to pick your class, this is the optional part but would make bandits have a new plethera of flavors. There's an open PR for it and I think it's so-so, but it's a lot better than having nothing."
requested on discord https://discord.com/channels/1240735983276654733/1258555111005880520/1299108588803063809

### DIFF
--- a/code/modules/antagonists/roguetown/villain/bandit.dm
+++ b/code/modules/antagonists/roguetown/villain/bandit.dm
@@ -155,8 +155,7 @@
 			else
 				neck = /obj/item/clothing/neck/roguetown/gorget
 			H.mind.adjust_skillrank(/datum/skill/combat/swords, 1, TRUE) 
-			H.change_stat("strength", 1)
-			H.change_stat("constitution", 1)//more stats due to rare role
+			H.change_stat("constitution", 1)
 			ADD_TRAIT(H, TRAIT_HEAVYARMOR, TRAIT_GENERIC)
 	H.change_stat("strength", 3)
 	H.change_stat("endurance", 2)


DatFrostys:
I'd like to say that I don't see bandits needing any debuffs at all right now.

wow50000:
> I recall bandits being quite fast in-game- was this nerfed at some point?

hedge knight has -1 speed mod and humen male bandits will have +1 speed, this gets stupid when other races like tabaxi are factored in

RadiantFlash7:
I was gonna say, it sounds like most of this is fixed by my pr- whenever it gets finalized or test merged

RadiantFlash7:
> hold up bandits get 3 strength and heavy armor? i'm fine with the extra speed honestly but that was a surprise hahah

Hedge knights get heavy armor training.

PR #1261, opened 2024-10-25
Guillotine revisit 

Heheniy:
## About The Pull Request

Guillotine now dont gives watched dismembering debuff, instead, if execution was succesfull, crowd gets mood buff, and if it failed- debuff, with executor getting failed execution mood debuff too.
Maximum level of guillotine sharpness was reduced for this system to work better (ie executor needs to be caring about blade sharpness and sharp it with stone if needed)

Removed code that let to fix (sharpen) guillotine with plasteel as it is not being used.


Nakalon:
Good! Deleting unnecessary code! We should do this more often! Cheers!

PR #1266, opened 2024-10-25
Makes forlorn hope lupian only

wow50000:
## About The Pull Request
See title

## Why It's Good For The Game
Its a lupian order but for some reason all furries are allowed in but not humens or any of the other races ? seeing literally any furries going into the merc group that is very clearly meant to be lupian exclusive makes me cringe or just let other races like humens be in it because clearly the lupians arent that racist if they are letting tabaxi and anyone with fur in.

Be racist properly, im sick of seeing rat people mercs in the clearly lupian merc group.


RadiantFlash7:
I would say remove tabaxi and wildkin, but keep vulps allowed in it.

Lupians and vulp's speak the same language. I assume they're atleast somewhat close culturally that they would be in the same order?

PR #1268, opened 2024-10-26
Nerfs the warhammer

Dedwolf:
## About The Pull Request

Quick nerf to address some good points raised by people that actually took the time to think and have a discussion, not blurt out idedplsnerf

Makes the warhammer wbalance = 0 - It no longer scales off of strength difference, this should make it take longer to exhaust opponents. 
Added charge drain to the pick intent - Should prevent players charging the attack and just waiting, this will reward properly timing the attack. 

I am curious to see how much of an effect the wbalance change will have on the viability of the warhammer. We'll see. 

I was going to wait longer before making this PR but I reckon these are good changes.

## Why It's Good For The Game

a


Nakalon:
So, why IS it good for the game?

RadiantFlash7:
> > So, why IS it good for the game?
> 
> aaaaa

So how exactly does pick differ from smash now?

PR #1270, opened 2024-10-26
crying orb

Nakalon:
## About The Pull Request

image

Starting from Arcane Skill level 5 you will no longer normally break the scry orb (Court Magus)
You can now bypass the scrying orb's cooldown.
Doing so has a chance of breaking it.
Now a player must focus while using the scrying orb (can't move) for a short time based on arcane skill level.

Making it have a chance, even if very little would cause it to eventually break first time, which wasn't fun.

image

I also added a little sound and a small delay based on your Arcane skill:

https://github.com/user-attachments/assets/922f0327-e693-4109-8e47-6740edf4668a

## Why It's Good For The Game

No one likes RNG that fucks you in the ass like this.

## Changelog

tweak: Starting from level 5 Arcane level you no longer break your scrying orb under normal circumstances
tweak: The scrying orb can now be used while under cooldown, the chance of it breaking applies in this case.
tweak: Scrying now plays a little sound for immersion's sake.
balance: Trying to use a scrying orb is now an interruptible action. The time it takes scales down with Arcane Skill.
code: Removed a check in scrying orb code that existed twice for no reason.

### DIFF
--- a/code/game/objects/items/rogueitems/magic.dm
+++ b/code/game/objects/items/rogueitems/magic.dm
@@ -68,14 +68,28 @@
 				success_chance = 97
 			if(SKILL_LEVEL_LEGENDARY)
 				success_chance = 100
-		if(!prob(success_chance))
-			to_chat(user, span_boldwarning("You focus your thoughts on the orb, but feel a sharp pain!"))
-			visible_message("\The [src] shatters!")
-			user.flash_fullscreen("redflash1")
-			new /obj/item/shard(get_turf(src))
-			playsound(src, "shatter", 70, TRUE)
-			qdel(src)
-			return
+	if(world.time > last_scry + cooldown)
+		switch(arcane_skill)
+			if(SKILL_LEVEL_NONE)
+				success_chance = 50
+			if(SKILL_LEVEL_NOVICE)
+				success_chance = 65
+			if(SKILL_LEVEL_APPRENTICE) //Apprentices have this
+				success_chance = 80
+			if(SKILL_LEVEL_JOURNEYMAN) // Adventurer mages have this
+				success_chance = 90
+			if(SKILL_LEVEL_EXPERT)
+				success_chance = 94
+			if(SKILL_LEVEL_MASTER || SKILL_LEVEL_LEGENDARY) // Magus has this
+				success_chance = 100
+	if(!prob(success_chance))
+		to_chat(user, span_boldwarning("You focus your thoughts on the orb, but feel a sharp pain!"))
+		visible_message("\The [src] shatters!")
+		user.flash_fullscreen("redflash1")
+		new /obj/item/shard(get_turf(src))
+		playsound(src, "shatter", 70, TRUE)
+		qdel(src)
+		return
 	playsound(src, 'sound/magic/whiteflame.ogg', 100, TRUE)
 	for(var/mob/living/carbon/human/HL in GLOB.human_list)
 		if(HL.real_name == input)


NikolayByondPlayer:
Wait, but this was on purpose to stop spamming the orb

DatFrostys:
> Wait, but this was on purpose to stop spamming the orb

I had two separate instances of it breaking first try as court magician, not spamming it at all. If that's what it was for I believe it's not great design

Nakalon:
> Wait, but this was on purpose to stop spamming the orb

It was having the unintended effect of blowing up first try...

I dont have the mental fortitute of searching for a better alternative rn

NikolayByondPlayer:
> > Wait, but this was on purpose to stop spamming the orb
> 
> It was having the unintended effect of blowing up first try...
> 
> I dont have the mental fortitute of searching for a better alternative rn

I see, yeah, sounds weird.
Maybe you can also add some delay to orb use?

Nakalon:
> > > Wait, but this was on purpose to stop spamming the orb
> > 
> > 
> > It was having the unintended effect of blowing up first try...
> > I dont have the mental fortitute of searching for a better alternative rn
> 
> I see, yeah, sounds weird. Maybe you can also add some delay to orb use?

Can do a delay yes, hold on

DatFrostys:
there's already a delay

Nakalon:
> I had an idea.

Tell me, I just added a do after and a little sound

Nakalon:
> > > I had an idea.
> > 
> > 
> > Tell me, I just added a do after and a little sound
> 
> make it so using the orb starts a timer/cooldown, if the orb is reused before the timer is over there is a chance it breaks, regardless of user skill. Perhaps this can stack so the more its used in a short amount of time the more likely it will break.

There is already a cooldown! For the scrying orb its 30 seconds.
Would you rather it be that way instead?

Nakalon:
> > > > > I had an idea.
> > > > 
> > > > 
> > > > Tell me, I just added a do after and a little sound
> > > 
> > > 
> > > make it so using the orb starts a timer/cooldown, if the orb is reused before the timer is over there is a chance it breaks, regardless of user skill. Perhaps this can stack so the more its used in a short amount of time the more likely it will break.
> > 
> > 
> > There is already a cooldown! For the scrying orb its 30 seconds. Would you rather it be that way instead?
> 
> okay. Make it so the cooldown can be violated, but doing so has a high risk of breaking the orb.

So, no chance for breaking normally, but ability to ruin the thing by just violating?

Nakalon:
> Yep. Do you think its a good solution?

Sadly I don't think it would be obvious to the player.
The cooldown forces them not to be able to use, being able to bypass that will make it so players never know if the cooldown is over or not... hum...

Nakalon:
> > > Yep. Do you think its a good solution?
> > 
> > 
> > Sadly I don't think it would be obvious to the player. The cooldown forces them not to be able to use, being able to bypass that will make it so players never know if the cooldown is over or not... hum...
> 
> we can add more or less subtle hints. In the orb description, and when the orb is used.

Very well, I'll do that! Thats what I was doing already!

Nakalon:
> we can add more or less subtle hints. In the orb description, and when the orb is used.

It's done, but now im ocncerned about low level Arcane users being able to use it, maybe I should keep that check in place?

Nakalon:
> I guess, yeah.

Hey Gyra, I made the changes but got a bit confused in the logic, Can you check it to see if its to your liking?

RadiantFlash7:
I will point out other mages have the ability to block scrying- if they take it. Worth considering as well c:

radarseas:
I think its fine as like a quick confirmation glance, but having the scrying orb break roundstart immedately kind of sucks, also its bad enough using it as an independant magos anyway, even without the potental it has.

PR #1279, opened 2024-10-27
Places a single purity vendor in the sewers. Remove all others.

Gyrandola:
## About The Pull Request

I don't know who added drugs vending machines all over the place,  but Rockhill is not Amsterdam.
There is now only a single purity machine, and it's in the sewers, where crackheads live.

image

### DIFF
--- a/_maps/map_files/rockhill/rockhill.dmm
+++ b/_maps/map_files/rockhill/rockhill.dmm
@@ -5700,10 +5700,6 @@
 /obj/item/flint,
 /turf/open/floor/rogue/cobble,
 /area/rogue/indoors)
-"fqC" = (
-/obj/structure/roguemachine/drugmachine,
-/turf/closed/wall/mineral/rogue/stone,
-/area/rogue/outdoors/rtfield)
 "fqH" = (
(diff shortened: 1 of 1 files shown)


Red-Honk:
At least add its own room, that risks someone getting ambushed or something, because this just looks lazy and time wasting in my opinion.

Red-Honk:
At least add its own room, that risks someone getting ambushed or something, because this just looks lazy and time wasting in my opinion, like a crack den or the like, a very messy one, places to crash at, relax, so, it's all low tier stuff, but I doubt anyone will notice after the first or second hit.

NikolayByondPlayer:
I have one near bath house, maybe we can just cut others?

Nakalon:
Good! All seems well with the map files too! (seamless wont be flung into the vampire cave again xD)

Nakalon:
> At least add its own room, that risks someone getting ambushed or something, because this just looks lazy and time wasting in my opinion, like a crack den or the like, a very messy one, places to crash at, relax, so, it's all low tier stuff, but I doubt anyone will notice after the first or second hit.

Quick solution: Don't do drugs!

You junkie!

PR #1281, opened 2024-10-27
Remove Nourish

Gyrandola:
## About The Pull Request

This thing has completely destroyed the food economy, allowing people to buy free food from the drug vending machine in the form of syringes. It's kinda hearthstoneish.

We have soilsons, we have butchers, cooks, hunters, maids, servants. Hungry? Interact with them.
Worst case scenario you can butcher and eat someone's limbs (added in the cult update).


wow50000:
why close ?

PR #1283, opened 2024-10-27
Makes knights not so smart

wow50000:
## About The Pull Request
see title

## Why It's Good For The Game
Horc warrior has +4 str +3 con +3 end and the same speed mods, this means a horc warrior in full plate can serve as a more retarded, similar toughness, better trained knight. This should put knights way more above horc warriors in terms of martial prowess.
also should crack down a little on knights learning skills to replace towners... if that ever happens.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/nobility/knight.dm
+++ b/code/modules/jobs/job_types/roguetown/nobility/knight.dm
@@ -74,9 +74,8 @@
 		H.mind.adjust_skillrank(/datum/skill/misc/riding, 4, TRUE)
 		H.change_stat("strength", 4)
 		H.change_stat("perception", 1)
-		H.change_stat("intelligence", 2)
-		H.change_stat("constitution", 3)
-		H.change_stat("endurance", 2)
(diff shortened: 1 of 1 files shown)


DelineFortune:
horc are extra shunned with hard job lock and supposed to spend 1+ hour making their own gear?
instead of knight that is getting everything for free form start?

radarseas:
we're all forgetting without stats on top that knights are just experts and masters at their weaponry, they have authority over a lot of players and generally a lot, **lot more** than your average shunned nimrodded half-orc.

if a half orc manages to become a noble, fullplated in armor, not die during all of that and get it roughly at a time in the round it matters AND train to become a master or expert at fighting in swords and polearms..

_I think they earned it, let them to have it._

---

A knight's main strength over a horc with literally the same gear, authority, skills is int-based fienting anyway, if they take up a side-hobby on top of serving the king, _let em have it._

That and we have other races which can out-strength a horc as knight tenfold anyway, _like say, lupans._

PR #985, opened 2024-09-14
Minor stat buff for shunned races + stat buff for being inhumen gods worshipper

DelineFortune:
## About The Pull Request

Shunned races are minority now as punching bags. They lost their bad reputation because people left the server / started to play respected races for combat roles

Templars are also allowed to be for shunned races because technically they proved themselves as servants of Ten and they are just tools in church's hands (not as acolytes or priest) 

And maybe an opportunity for heretics to use shunned race templars to defile the church

## Why It's Good For The Game

More content, more action in rounds, more ways to play as antag against bored 100500 knights/templars/grenzels/riders/maas/watchmen that can charge-stun and kill your kobold/tiefling
Also more reasons to be racist to shunned races, fear them, call reinforcements (and involve more players into situation,  not 1 vs 1 fight)
Grim dark experience to inquisitors because everyone can be a heretic for blessing of dark gods and try to kill you. 
Templars are servants of Ten, but they can be members of shunned races and help their kins against you. 

Also gives a positive reason to play unhumen worshipper instead of cons that they have: job lock and a reason to be killed for their faith

PR #1286, opened 2024-10-27
Re-made unhumen worshippers + shunned races pr

DelineFortune:
## About The Pull Request
https://github.com/Rotwood-Vale/Ratwood-Keep/pull/985

## Why It's Good For The Game

Changed to fix conflicts.
Less buff for inhumen.


Nakalon:
It's been said by staff they don't want stats to be tied to faith.

Also the way this PR is worded makes it extremely hard to understand what it is about at a glance.

PR #1291, opened 2024-10-28
Timed Gates

ProbablyCarl:
## About The Pull Request

Simple timed gates to prevent bandits from rushing the town before thirty minutes is up. They take place as a 'fog', for which I don't have sprites. Anyone can pass through, with the exception of bandits.

Accompanying map edits to prevent digging around them.

## Why It's Good For The Game

As with a number of my PRs, it was requested that I add this. No one likes bandits sneaking by the gates round start, although you can certainly argue it's a skill issue.


hugmeorelse:
> would throw it on a TM but not on the TGS. flagged for TM pass.

Already TMed.

hugmeorelse:
Let's leave it on TM for now, @Moribun. Gyran wants to do a poll.

Nakalon:
So an invisible wall... I'm not too fond of this solution... I think there are other ways around it...

katsukai:
What are we? Colonial marines??

Nakalon:
> > So an invisible wall... I'm not too fond of this solution... I think there are other ways around it...
> 
> I was told to leave ways around them still, yeah. You can mine to bypass them on the lower levels, and there's a handful of other ways to get past it. This was just to stop people from rushing the gate immediately and then making their way into town.

No, what I'm saying is, ways around adding an invisible wall.

As in I'm sure we can think of another better system to replace this!

Like, understanding why and how they are rushing the gate imediatly.

What if they were under equipped? Doubt they could pose a threat then, and would first need to find ways to equip themselves.

Or what if instead of a wall there was some sort of monster or entity present in the bog during the first 30 minutes of the game.
A scary one...

You see my point?

Nakalon:
> > > > So an invisible wall... I'm not too fond of this solution... I think there are other ways around it...
> > > 
> > > 
> > > I was told to leave ways around them still, yeah. You can mine to bypass them on the lower levels, and there's a handful of other ways to get past it. This was just to stop people from rushing the gate immediately and then making their way into town.
> > 
> > 
> > No, what I'm saying is, ways around adding an invisible wall.
> > As in I'm sure we can think of another better system to replace this!
> > Like, understanding why and how they are rushing the gate imediatly.
> > What if they were under equipped? Doubt they could pose a threat then, and would first need to find ways to equip themselves.
> > Or what if instead of a wall there was some sort of monster or entity present in the bog during the first 30 minutes of the game. A scary one...
> > You see my point?
> 
> There are absolutely easier ways to do it, and methods that aren't as brute force in nature. DM me on Discord and we'll chat about it further to brainstorm ideas. Easier that way.

My discord is PinkSuzuki, I go by Desdemona / Delphine in the server!
And I don't defend easy ways to do it! I defend harder and better :P

Koteshkin:
Disgusting decision.

Also it prevents saigas (and maybe other mobs) from passing. So you cant ride a saiga inside the gate even if you arent bandit. Need to wait almost forever to pass.

RadiantFlash7:
This could probably use a check to stop garrison members fromleaving as well tbh. The entire keep garrison rushing into the bog to frag bandits 5 minutes into the round is no better then the bandits rushing the bog gate 5 minutes in.

Gyrandola:
> This could probably use a check to stop garrison members fromleaving as well tbh. The entire keep garrison rushing into the bog to frag bandits 5 minutes into the round is no better then the bandits rushing the bog gate 5 minutes in.

this sounds like a good compromise

Gyrandola:
fix conficts and I will full merge it

Gyrandola (approved):

PR #1292, opened 2024-10-28
1210 Revert for TM

ProbablyCarl:
Temporary revert of 1210's stuff for the sake of a TM. To see if the exploit still exists.

At request.


Heheniy:
it was meant to b TEST merge...

PR #1293, opened 2024-10-28
wolf handjob

Luctus-Comptus:
-1 doesn't do an integer overflow here and that makes me angry

so I made a thing.


Heheniy:
defined INFINITY does exist

PR #1294, opened 2024-10-28
Fixed map fixes, seamstress friendly, super ready to merge

Heheniy:
## About The Pull Request

#1189 + fixed guild smith spawn issues

### DIFF
--- a/_maps/map_files/rockhill/rockhill.dmm
+++ b/_maps/map_files/rockhill/rockhill.dmm
@@ -8956,11 +8956,11 @@
 /obj/structure/table/wood{
 	icon_state = "tablewood1"
 	},
-/obj/item/clothing/suit/roguetown/shirt/shortshirt/merc{
+/obj/item/clothing/suit/roguetown/shirt/shortshirt/random{
 	pixel_x = -2;
 	pixel_y = 5
 	},
-/obj/item/clothing/suit/roguetown/shirt/shortshirt/merc{
+/obj/item/clothing/suit/roguetown/shirt/shortshirt/random{
 	pixel_x = 4
 	},
 /turf/open/floor/rogue/wood,


Nakalon:
This is in fact not ready to merge and not seamstress friendly, as it negates changes to the map that were made in the past, that I know of:

Water barrel, 2 clothing articles and the oven in the seamstress shop.

If these are missing I wonder what else is missing.

I also have to wonder why you insist in changing so much of the map in one go, while still offering no explanation on what you're doing and no reasoning as to why it would be benefic to the game...

Heheniy:
everything is in place asade of barrel because why seamster even have it?
image

Heheniy:
everything is in place asadi of barrel because why seamster even have one
image

Nakalon:
> everything is in place asade of barrel because why seamster even have one image

I sincerely recommend you check Mapdiff, then you will see what I mean.

FunnyScreenshot_20241028-182223

I'm red you will see what was changed in your PR.

You will see that the barrel was removed, 2 clothing items were swapped and the oven was replaced by a fireplace.

These were changed I introduced myself and you are now reverting. Knowingly or not.

I know why I put the barrel there. Do you?

Again, I would say this PR fails to mention these changes if ANY change was at all documented or if it had anything on its description.

Which it doesn't. It's empty. So all we have is the Mapdiff to go by. And yours lights up like a god damn Christmas tree.

Heheniy:
i cant see why seamstres should be fully self sufficient on water and food side with the barrel and stove

Nakalon:
> i cant see why seamstres should be fully self sufficient on water and food side with the barrel and stove

I can't see why it matters since this wasn't obviously a choice you made knowingly.

What was then your reasoning behind substituting the 2 cloth shirts with 2 Merc shirts which have both their sleeves ripped round start?

Heheniy:
those were shirts from older version of map, i didnt touch them

Nakalon:
> those were shirts from older version of map, i didnt touch them



> those were shirts from older version of map, i didnt touch them

If you hadn't touched them they wouldn't appear in red.

This is the Mapdiff from the PR where I last touched the Seamstress Workshop:
image

As you can see, I changed the Fireplace into an Oven and I switched those 2 merc shirts for 2 normal shirts because the merc ones had their sleeves ripped.

You off your own volition decided to remove the barrel, and since you were either unable or did not bother to solve merge conflicts properly you decided to keep the other two things as they were and just manually place the seamster spawner.

Nowhere in this PR or the one you link are these changes documented.

What I would like to see is you documenting this in the PR and arguing why removing a barrel, changing two shirts and replacing an oven with a fire place is beneficial to the game.

Along with all these you see in red:
image

(This is just level 3)

image

You also changed Centcom for some reason which is a different map file?

PR #1300, opened 2024-10-28
Better crafting naming

Heheniy:
## About The Pull Request

"I sew a cloth (2 fibers; NONE)!" -> "I sew a cloth!"


Nakalon:
How does this behave with multi result recipes?

PR #1304, opened 2024-10-28
removes chudcode from cook

wow50000:
## About The Pull Request
removes sexism from stat assignment

## Why It's Good For The Game
chudcode bad


code/modules/jobs/job_types/roguetown/peasants/cook.dm
```diff
@@ -43,4 +43,4 @@
 		head = /obj/item/clothing/head/roguetown/cookhat
 		shoes = /obj/item/clothing/shoes/roguetown/simpleshoes
 		neck = /obj/item/storage/belt/rogue/pouch/coins/poor
-		H.change_stat("constitution", 1)
+		H.change_stat("constitution", 2)
```
Nakalon, line 46: Just add this and the above one to the if(H.mind) condition, no need to put this in twice!

Nakalon (changes requested):

PR #1314, opened 2024-10-29
WIP - Big burial changes PR

zoktiik:
## About The Pull Request
Committing incomplete work I did weeks ago that's been idly sitting on my local machine. Needs a lot of testing and perhaps some input.
Changes funeral code to work a lot more simply - the game sets and checks for a single variable *in the mind* as opposed to two variables *in the mob* (one of which, `burial`, being entirely useless), since both the carbon mob and the rogueobserver share the same mind datum.
It is set to be *very aggressive* about kicking you back to the lobby after you get a memorial. Your mind datum is checked when you call your mob's Login() - so either switching to the mob or reconnecting to the game. 

## Why It's Good For The Game
Burials are less scuffed and it's easier for people to return to the game. In fact, they're, uh, *very much encouraged* to keep playing the game, on a different character, if they happen to get their body buried.


Nakalon:
> > When you say it needs testing what do you propose?
> > Do you propose it is test merged? That you will test it at a later date?
> 
> I'll try to perform local testing first when I get around to finishing it. Pull request draft status should imply that I don't intend for it to be testmerged yet.
> 
> And, note to self: make respawn aggressiveness into a config option (default: not aggressive)

Good!!! Thanks!

Gyrandola:
life check?

Gyrandola:
4af168f0-43bb-413b-9875-f91e5694b4cf_text

Nakalon:
> do you understand that prodding someone with "live check" on a 3-day-old draft PR with no conflicts, that they're not even being paid for, implies that you want to remind them to finish their PR sooner rather than later

Nuh huh, dont read it like that, Gyran is just very excited for this im sure!

Nakalon:
> I can understand you wanting to defuse a situation, but this isn't how you go about it. I would like to have Gyran respond, if any response is to be made, as I believe them to be more than capable of doing so. You can speak for yourself.

Well you don't have to be mean about it.
I shared may honest thoughts

PR #1315, opened 2024-10-29
Thugs, Confessors & Mercenary Guild

ProbablyCarl:
## About The Pull Request
Moves the mercenaries into the town, round start, and assumedly on the barge, so they can actually operate out of their building. Gives them keys, too, for said building.

Adds confessors back in, removes their objective list and instead puts them in the same boat as Inquisitors, with it being an optional mail in sort of deal for confessions. They have two slots.

Introduces Grabbers under a new name, as 'Thugs', with a bit of a different flavor. They're effectively the same thing, all things considered. Essentially just Merchant's Guild mercenaries. They have two slots.

## Why It's Good For The Game
Mercenaries are supposed to be part of a larger guild. At least Grenzelhoft mercenaries, but they're never portrayed as such accurately. This gives them, the Desert Riders and general sellswords a base to operate out of, utilizing a building that had been intended for them.

Confessors are needed to push forward additional attempts at excising heresy from the Keep and general towners. They're the Inquisitor's lap dogs.

Grabbers, now known as Thugs, are pre-round hired mercenaries for the Shopkeep's and Merchant's to work with, so they've no longer the worry of a mercenary turning around and going 'haha this my shop now'. It's effectively the same as it was prior, role and all, just with a new name and flavor text.

### DIFF
--- a/_maps/map_files/rockhill/rockhill.dmm
+++ b/_maps/map_files/rockhill/rockhill.dmm
@@ -332,12 +332,6 @@
 	},
 /turf/open/floor/rogue/grass,
 /area/rogue/outdoors/rtfield)
-"aoY" = (
-/obj/structure/bed/rogue/inn/hay,
-/turf/open/floor/rogue/ruinedwood{
-	icon_state = "weird1"
-	},
-/area/rogue/indoors/town/shop)
 "apc" = (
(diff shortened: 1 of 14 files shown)


Nakalon:
This is kinda sexy

PR #1319, opened 2024-10-29
Deep caves expand

Heheniy:
## About The Pull Request
Expandes deep caves and removes some forgotten cave area in the middle of deep caves
image


zakanater19:
is the forgotten cave area middle left Z1? cause that's supposed to be a way up into the underground lake area, it's why i left it in

Nakalon:
> is the forgotten cave area middle left Z1? cause that's supposed to be a way up into the underground lake area, it's why i left it in 

I'm not at all surprised the author failed to understand why something in the map was the way it was...

Heheniy:
here are areas of z1 on z2
image
there is no template or logic that that leftover area follows so i see no reason to keep it

zakanater19:
this is the gap thing i was talking about, you can see the dark green tiles are air leading down/up into the z1 cavern

Screenshot 2024-10-29 174435

Heheniy:
cant see it anywhere else, thats the reason
image

zakanater19:
The others are not guaranteed to be clear, but that area is since it's the only one past the bog, but yeah should be big enough not to be closed off

Heheniy:
it is 30% for a turf to be a wall, none of the holls can be fully walled 
image

Nakalon:
This would be much easier if the "Why it's good for the game" wasn't empty and the only argument for this change wasn't "But everything should be the same everywhere".

PR #1324, opened 2024-10-29
Removes negative luck from tiefling

wow50000:
## About The Pull Request
see title

## Why It's Good For The Game
Tiefling players are exceptionally fragile wiht -2 con and -1 luck, this may or may not solve it Tieflings also make the worst miners due to it as well as well as not being able to land crits as easily.


SayonaraEcho:
I'm tiefling main and i like this pr. Fortune is a crucial thing in combat actually because you're hitting people with othen CRITICAL MISS or rare critical strike than others.

Nakalon:
> would let a tiefling rail me. 10/10

Mori I play a Tiefling...

PR #1325, opened 2024-10-29
Recruit positions fix reopen

Heheniy:
## About The Pull Request
Fixed not being able to rectuit (as priest, captain, etc jobs) youngfolk
Removed veteran from mercenary list and moved into guard because HE IS guard
image
image


RadiantFlash7:
Technically, the veterans a hired mercenary at round start, as weird as it sounds.

Still- I kind of like finally being able to be paid when the rest of the garrison is being paid, so I can't bitch much about this xD

katsukai:
Would be funny if the Veteran just spawns in a farm shed and is just jobless.

PR #1328, opened 2024-10-29
Nerfs Church Wrestling/Unarmed

babamandaa:
## About The Pull Request

I think it's a little bit crazy that acolytes are better at wrestling, skill-wise, than knights, men at arms, and bandits. I will state that, for the record, the church does NOT need nerfs, but that's much more of an issue with the church's role in the game and not their actual base stats.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/church/monk.dm
+++ b/code/modules/jobs/job_types/roguetown/church/monk.dm
@@ -71,8 +71,8 @@
 			shoes = /obj/item/clothing/shoes/roguetown/sandals
 			armor = /obj/item/clothing/suit/roguetown/shirt/robe/astrata
 	if(H.mind)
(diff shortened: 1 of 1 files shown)


DelineFortune:
Wrestling is only one way for church people to protect themself against 100500 mercs, guards, adventurers.

Its always low number of guards and templars.

Whats point of this PR? To make you loot everything for free and do anything you want? They have no access to armor/weapon even to protect themself

babamandaa:
> I disagree entirely with this on hearth I axed all the PR's related to cqc nerfs to church and will continue that tradition here.

If I tried to buff Holy's scaling to make the priest and acolytes much more capable of defending themselves without breaking the adventurer classes, would wrestling nerfs be considered?

Zmowlcher:
The church has high skill but low stats. If you lose a combat encounter as a knight against the church, i have no words.

PR #1329, opened 2024-10-29
Ruralization - Two Cottages

Nakalon:
## About The Pull Request

Two lovely cottages fit for couples have been built outside Rockhill!

image

### Here, we have the Farmer Cottage:

image

As you can see, fair space for lovemaking and the cooking of homemade meals.

image

Outside you may find a small patch of land to be plowed, a hoe has been stocked inside the houses wardrobe.

### Here, we have the Woodman Cottage:

image

While it is small, you have space in the back for all your lumbering needs!

image

However, the house includes a second floor, where the bedroom is.

image

And for all you lovebirds, we have included a balcony fit for a romantic dinner...

image

We hope you enjoy the rural side of Rockhill your future children will thank you for it!

## Why It's Good For The Game

There are two reasons for doing this:
1 - If everything goes well, space will be required in the city to accommodate for more interesting shops and attractions.
Creating some houses outside of the city allows for houses within the city to be repurposed later without losing anything in the process.
2 - There was a debate in favor of Deurbanization in discord's feedback channel.
(Here's Wow malding on the topic because its funny)
image
These houses I created accommodate players who may want to RP a low class medieval family life. Something the houses inside the city are just not inviting people to do.
They are modeled after the Farmer and Woodcutter Towner roles.

If this is well accepted I plan to expand and with time build a little village there. But in parts, to simulate the construction of it as if it was taking time like in real life.
I already added the well there to be the "neighborhoods" source of water to ensure the peasants there can subside.

This also hopefully brings some dynamic and light to a part of the map that was left unused. (Darkness, no trees, it was like it was WAITING to have something built on it!)

### DIFF
--- a/_maps/map_files/rockhill/rockhill.dmm
+++ b/_maps/map_files/rockhill/rockhill.dmm
@@ -157,6 +157,13 @@
 /obj/item/clothing/neck/roguetown/psicross,
 /turf/open/floor/rogue/cobblerock,
 /area/rogue/outdoors/town)
+"aei" = (
+/obj/structure/mineral_door/wood/deadbolt{
+	dir = 4;
+	lockid = "farm"
+	},
+/turf/open/floor/rogue/woodturned,
+/area/rogue/indoors/shelter/town)
 "aev" = (
 /turf/open/floor/rogue/dirt,
 /area/rogue/indoors/shelter/mountains)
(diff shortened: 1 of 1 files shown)


Nakalon:
> honestly, the town is far, far too packed full of people, I think looser housing placement to the outskirts to make kidnapping viable is a good thing, for both antags and non-antags alike.
> 
> at the moment grabbing someone in town is either flawless or fighting 20 people at once with no inbetween.
> 
> we need more PRs like this!

Thanks!! That means a lot to me!! Me and @NikolayByondPlayer are actually thinking about this together!

PR #1333, opened 2024-10-30
Lantern for Miner

ProbablyCarl:
## About The Pull Request

At the request of a few people, the miner now has a lantern. It makes sense, given they're meant to be underground a majority of the time.

## Why It's Good For The Game

Lanterns should've already existed for Miners. It doesn't really make sense that they'd be swinging a pick and holding a torch at the same time.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/church/templar.dm
+++ b/code/modules/jobs/job_types/roguetown/church/templar.dm
@@ -48,7 +48,6 @@
 			cloak = /obj/item/clothing/cloak/tabard/crusader/pestra
 		if("Malum")
 			wrists = /obj/item/clothing/neck/roguetown/psicross/malum
-			H.mind.adjust_skillrank(/datum/skill/craft/blacksmithing, 2, TRUE)
 		if("Ravox")
 			wrists = /obj/item/clothing/neck/roguetown/psicross/ravox
 	armor = /obj/item/clothing/suit/roguetown/armor/chainmail/hauberk


Nakalon:
I assume the pick is a type of polearm?

Also very wholesome PR!! Nice!

Elli-Skala:
They start with flint already, they should be able to make standing fires from just coal and stone as they mine lighting up the mine as they go, a lamp is just QOL they can pay for

Clocks-scp:
> They start with flint already, they should be able to make standing fires from just coal and stone as they mine lighting up the mine as they go, a lamp is just QOL they can pay for

And on rounds when they CAN pay for that convenience, they also need a steel pick. The request is mostly framed around being able to successfully perform the role in a smooth and not mind numbing way on rounds when there are limited smithing opportunities. There will always be a thing they need from a smith, the steel pick. But if they started with a lamptern instead of a torch, they can more reliably manage their inventory and haul a cart while mining with the iron pick at round start, even underground. As it stands, the torch is a finicky pain in the ass and having a lamptern would mean that miners could go spend their first day using their iron pick to get ore, instead of standing around for both a steel pick and lamptern and coming in with, usually, nothing. This isn't game making or game breaking, it just would make the opening minutes of game more productive.

So. How it helps the game? It helps miners start out swinging and bring in ore with their iron picks. As it stands, the need for a steel pick AND a lamptern means that you just wait for both or get a Mason's key and get both.

I don't think that the guy who's underground all the time starting with a light source would detract from things, especially with the miner's existing dependence on smiths both for steel picks and repairs and also for someone to deliver ore to. The Malum gaming bond is going to stay strong even if the smith doesn't have to spend one extra ore setting them up.

radarseas:
towners already suck anyway, yeah they could use this QOL honestly.

I mean, ranger adventurers already get a lamptern, as does the hunter. so... its not like this hurts balance, if anything it fixes balance.

Squird5:
I would suggest moving the hunting knife into the belt or backpack and not into the mouth

code/modules/jobs/job_types/roguetown/adventurer/types/pilgrim/miner.dm
```diff
@@ -11,11 +11,12 @@
 /datum/outfit/job/roguetown/adventurer/miner/pre_equip(mob/living/carbon/human/H)
 	..()
 	head = /obj/item/clothing/head/roguetown/armingcap
+	mouth = /obj/item/rogueweapon/huntingknife
```
Nakalon, line 14: CARL, THEM HUNTING KNIFE IS IN THE MOUTH CARL!!

Nakalon (changes requested):

Cyprex:
No from me, lanterns should be a rare upgrade. These aren't cheap(Or that's the idea, anyways!)

PR #1334, opened 2024-10-30
Makes templars the knights equivalent of a paladin

wow50000:
## About The Pull Request
see https://github.com/Rotwood-Vale/Ratwood-Keep/pull/1356


Nakalon:
Why's this motherfucker getting sewing? 😡

DelineFortune:
Paladins are worse 
You need to spend a lot of time to get gear
Have no title
Worse stats 
Can get skills 2 on training dummy/ask preist to get 3

radarseas:
this PR is honestly, fine.

yeah I'm aware they're a lot more skilled now but adventurers should not out-stat a...
_3 job limited role that has obligations unlike them and is **supposed to be the less shitty version**_.

I'm aware adventurers are set back but to be honest I think templar need some buffs, they literally are incompetent in every other weapon class to the point of being virtually unable to use it.

Iffy on sewing but its _not like these guys can make anything better than what they already have with it_. They literally start with a gambeson.

I'd remove any of the skills at novice though. these guys shouldn't know **swimming** or **bows** and **crossbows** or **sneaking** when they've been in a church their entire lives.

PR #1335, opened 2024-10-30
Confessor Correction

ProbablyCarl:
## About The Pull Request

Locks Confessor to Psydon. Does a few other things.

## Why It's Good For The Game

WHY HAD I NOT DONE THIS BEFORE? FUCK FUCK FUCK!!!!!!!!!


Hmeister-real:
i would make it change you to Psydon rather than only let you play as a Psydonian like the inquisitor did, because that way it's easier to pick and less confusing for new players to join as

PR #1338, opened 2024-10-30
Removes churn spell

Heheniy:
## About The Pull Request

Removes churn spell from game (left it commented out if someone one day would decide to rebalance all spells (haha)).

## Why It's Good For The Game
This spell is one of the most cancerous mechanics of the game.
No matter how well prepared you are, how good is your equipment or starts are, how much effort did you put into everything you achieved - if you are undead- even naked, limbless, the most retarded necra templar/acolyte/whatever can nuke you with a single click of a button, that has 0 charge time, CD only of a few seconds and costs 20 devotion (literally nothing).

https://cdn.discordapp.com/attachments/1240810707877101598/1300957762368438303/unknown_2024.10.29-18.57.mp4?ex=67236395&is=67221215&hm=809631433ed0e192c0aa50c163026d0cda135ca65c49f2116a714601e2b542b7&

https://cdn.discordapp.com/attachments/1240810707877101598/1300957928605356032/unknown_2024.10.29-18.57_1.mp4?ex=672363bc&is=6722123c&hm=e64330141020afb7fc07fb761c3c568dad70ef244245628bd223580487f436fa&

### DIFF
--- a/code/controllers/subsystem/rogue/devotion.dm
+++ b/code/controllers/subsystem/rogue/devotion.dm
@@ -121,7 +121,7 @@
 	if(!H || !H.mind || !patron)
 		return
 
-	var/list/spelllist = list(/obj/effect/proc_holder/spell/targeted/churn)
+	var/list/spelllist = list()
 	if(islist(patron.t0))
 		spelllist += patron.t0
 	else if (patron.t0)
--- a/code/datums/gods/patrons/divine_pantheon.dm
+++ b/code/datums/gods/patrons/divine_pantheon.dm
@@ -75,7 +75,6 @@
 	worshippers = "The Dead, Mourners, Gravekeepers"
 	mob_traits = list(TRAIT_SOUL_EXAMINE)
 	t1 = /obj/effect/proc_holder/spell/targeted/burialrite
-	t2 = /obj/effect/proc_holder/spell/targeted/churn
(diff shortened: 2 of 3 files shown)


Nakalon:
Don't comment out code by the love of god, just remove it if you're removing it.

If you actually want to balance it, then do so and don't remove it.

Xaytan:
Counterpoint: The undermaiden rebukes.

Nakalon:
Gyran is working on the Lich still and has said this can't be removed.

So I'm pulling out one of the nuke keys and voting for a no go. 🔑

Unless you're inclined to find a different approach?

SayonaraEcho:
> while removing it wouldn't be a bad idea, I'm sure we can do better like making it only target a single person, or something like azure did, else this removes the whole "putting the dead to rest" gimmic necrans are supposed to have that seperates them from any magical random.
> 
> I will however say every paladin getting a "full powered version" of the same spell doesn't help neither, if anything it should be something exclusive to cleric/morticians as compensation for being magically inferior to every single other patron. Especally that every second adventurer is a paladin these days.
> 
> for example, this PR: https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/709 does a good job at keeping it without decimating the use of it.

Yeah, how about to port from Azure Peak remake of churn undead? I think it's better than undead explosion with one button.

PR #1340, opened 2024-10-30
Harder Smithing for the Unskilled

Nakalon:
## About The Pull Request

This was made as a result of various discussions of this topic on discord.

image
***Funny thing by Pom***

This makes i much harder to get into smithing if you have 0 blacksmithing skill.

If you have 0 smithing skill you will have a 50% to not make any progress while smithing.
You then have a 25% chance to ruin the bar.

Old values were 25% and 12% respectively.

This makes it more costly to learn smithing. While not making it impossible.

## Why It's Good For The Game

There are 4 smithing classes. Guildsmith, Smith Apprentice, Towner Blacksmith, and Forge Cleric.
If players want to smith, they should choose one of these roles.
Choosing a role should be a compromise, you're getting certain skills in exchange for not getting others.
There isn't a main character, and everyone is necessary for the town to work.
This will also hopefully reduce smithery break-ins.
Goodfaith players will not suffer form this change.

## Changelog

balance: Decreased the smithing success chance for characters who have no smithing skill.

### DIFF
--- a/code/modules/roguetown/roguejobs/blacksmith/anvil_recipes/_anvil_recipe.dm
+++ b/code/modules/roguetown/roguejobs/blacksmith/anvil_recipes/_anvil_recipe.dm
@@ -31,7 +31,7 @@
 		moveup += round((user.mind.get_skill_level(appro_skill) * 6) * (breakthrough == 1 ? 1.5 : 1))
 		moveup -= 3 * craftdiff
 		if(!user.mind.get_skill_level(appro_skill))
-			proab = 23
+			proab = 50
 	if(prob(proab))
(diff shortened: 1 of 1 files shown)


Gyrandola:
fair.

Nakalon:
> fair.

Asked the opinion of smith mains, they are for this change!

Cyprex:
How does this affect repairs? They're a good way to pick up smithing.

Nakalon:
> How does this affect repairs? They're a good way to pick up smithing.

Shouldn't affect repairs at all since this only changes difficulty on the "Advance recipe" proc!

PR #1348, opened 2024-10-31
Fixes seraphina painting icon

john-pawson:
## About The Pull Request

The seraphina painting, when taken from a wall, lacked an icon due to an accidental uppercase letter, this fixes it, that's it.

## Why It's Good For The Game

Items having icons are good, right? At least i think they are.


Nakalon:
I'd normaly ask for testing evidence, but I trust this to be the case because I've seen it happen before!

john-pawson:
> I'd normaly ask for testing evidence, but I trust this to be the case because I've seen it happen before!

I've tested this, if that's what you're concerned about. I can send a screenshot of it working properly if needed.

PR #1354, opened 2024-10-31
Arming Sword and Helmet Fix

Nakalon:
## About The Pull Request

Fixes the Arming sword being invisible:

image

Fixes the Kettle helmet appearing very odd when worn:

image

This helmet is also smeltable now.

## Why It's Good For The Game

Bug fix is good!

The smeltable change is because all helmets can be smelted, except this one, which didn't make much sense!

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/swords.dm
+++ b/code/game/objects/items/rogueweapons/melee/swords.dm
@@ -84,7 +84,7 @@
 	slot_flags = ITEM_SLOT_HIP
 	name = "arming sword"
 	desc = "A short arming sword, designed as a knightly sidearm. Best used with a shield or out of desperation."
-	icon_state = "swordshort"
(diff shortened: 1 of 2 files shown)


Squird5:
What exactly was wrong with the kettlehelm? Had a two tile offset at some point?
Also I suggest you look at the "smelting fix" again

Nakalon:
> What exactly was wrong with the kettlehelm? Had a two tile offset at some point? Also I suggest you look at the "smelting fix" again

Kettle helmet is a subtype of helmet

image
image

Which means it inherits whatever vars are in helmet unless otherwise stated, which is what was happening with "smelting result" being ash.

As for what was wrong, it was the worn Icon state. It was appearing on the ground basically.
I could show you it but it takes me like 15 minutes to start up the local server, which I had to do twice for this, for the Helmet and for the Sword.

PR #1356, opened 2024-10-31
Makes templar a slightly better paladin

wow50000:
## About The Pull Request
tunes down templar stats, ports the paladin skillblock 1:1

## Why It's Good For The Game
Templars should be more appealing than paladins because you have to put up with the priest, right now they get worse overall skill and useless stats that waste power budget, this makes them slightly better paladins with more int and one more str as well as roundstart gear.
accidentally did the changes on main for https://github.com/Rotwood-Vale/Ratwood-Keep/pull/1334 so re uploading it


DelineFortune:
Keep per +2 its important for find sneaking persons around and theives in church
Also helps them to use bows and aim with swords better
Its kind of nerf otherwise

PR #1358, opened 2024-10-31
Bank Revert

Gyrandola:
## About The Pull Request

An overcomplication of an already delicate system that created a lot of bugs and weird economic phenomena.


Nakalon:
What about #1210?
I think this is only one part of the system which was introduced in that PR.

Gyrandola:
> What about #1210? I think this is only one part of the system which was introduced in that PR.

that has also been reverted, see second commit. I know its weird the diff is minimal.

Nakalon:
> > What about #1210? I think this is only one part of the system which was introduced in that PR.
> 
> that has also been reverted, see second commit. I know its weird the diff is minimal. You can check manually.

Yessss fantastic!
This was causing a lot of trouble!

Heheniy:
> lot of bugs and weird economic phenomena.

can you elaborate?

PR #1364, opened 2024-11-01
Runelocks

ProbablyCarl:
## About The Pull Request
Just pulls https://github.com/Tree445/Hearthstone/pull/218 and modifies it for our use, alongside fixing it up some. Runed firearms, as opposed to matchlocks. More fantasy feeling, IMO. 

The idea is to give three shots via the ammo pouch to the Lord Inquisitor, with an unloaded runelock pistol roundstart. They're reusable projectiles, same as arrows and bolts, but are guaranteed(mostly) to embed and may not fall out until later, rendering the weapon useless in the meantime. Totally pointless to keep and haul around if you lose the projectiles, too, since you can't get more. Additionally, they provide an identical effect to werewolves that the silver dagger does, should it hit them while transformed.

You get one shot, a lengthy reload, and not much to counter someone rushing you beyond hitting them with a golf ball sized projectile. Try not to miss.

For this, since the Inquisitor spawns with some additional equipment, they now spawn with a backpack, and their bank account was updated due to their lack of a pouch now.
## Why It's Good For The Game
It's not an arquebus, and it certainly isn't the club's special musket, which I'd have preferred over this, but it's something at the least. Variety to ranged combat would be good, even if this is so limited in number.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/church/puritan.dm
+++ b/code/modules/jobs/job_types/roguetown/church/puritan.dm
@@ -15,7 +15,7 @@
 
 	outfit = /datum/outfit/job/roguetown/puritan
 	display_order = JDO_PURITAN
-	give_bank_account = 36
+	give_bank_account = 200
 	min_pq = 5
 	max_pq = null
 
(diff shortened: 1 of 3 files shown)


katsukai:
50 damage 200 pen and people feared crossbows.

Nakalon:
> > 50 damage 200 pen and people feared crossbows.
> 
> Can't be obtained, so I didn't think to change it.

Can you? Because if you're adding them it's because you want them in the game.
If they aren't ready to be put in the game right out of the box then there is no sense in adding them! And I love this idea honestly... More stuff for my woodcutter fantasy!!!!

Gyrandola:
> > > 50 damage 200 pen and people feared crossbows.
> > 
> > 
> > Can't be obtained, so I didn't think to change it.
> 
> Can you? Because if you're adding them it's because you want them in the game. If they aren't ready to be put in the game right out of the box then there is no sense in adding them! And I love this idea honestly... More stuff for my woodcutter fantasy!!!!

Yes I agree, dont add unused content.
I suggest giving it to the inquisitor with 3 shots as we talked about.

Nakalon:
> > > > 50 damage 200 pen and people feared crossbows.
> > > 
> > > 
> > > Can't be obtained, so I didn't think to change it.
> > 
> > 
> > Can you? Because if you're adding them it's because you want them in the game. If they aren't ready to be put in the game right out of the box then there is no sense in adding them! And I love this idea honestly... More stuff for my woodcutter fantasy!!!!
> 
> Yes I agree, dont add unused content. I suggest giving it to the inquisitor with 3 shots as we talked about.

yes if you do so, revise those numbers!

code/modules/jobs/job_types/roguetown/church/puritan.dm
```diff
@@ -15,7 +15,7 @@
 
 	outfit = /datum/outfit/job/roguetown/puritan
 	display_order = JDO_PURITAN
-	give_bank_account = 36
+	give_bank_account = 200
```
Nakalon, line 18: Please make this obvious in PR body so people dont complain when they're coin pouch gets snatched!

Nakalon:
Do you want this not to have craft able ammo? Could be cool to add. If you let me I'll make this craftable in the future too (will make it not so easy)

PR #1392, opened 2024-11-02
New Cultist restrictions

RadiantFlash7:
## About The Pull Request
blacklists roles that can't take zizo as a patron, from being cultists round start. 
Adds the following roles to the blacklist:
	"Bandit",
	"Knight",
	"Retinue Captain",
	"Gatemaster",
	"Bog Master",
	"Inquisitor",
	"Confessor",
	"Acolyte"
Notably, removes merchant from blacklist role. The merchant absolutely can and should be able to be a cultist.

## Why It's Good For The Game
Roles that can't BE zizoites, because they are thoroughly vetted, should not be starting as zizo cultists. I mean come on now. Having the guard captain and knights be roundstart cultists and offering the royal family 40 minutes into the round is pain.


Gyrandola:
Real

PR #1395, opened 2024-11-03
removed cant sleep from all but chest armor

Luctus-Comptus:
told some people when I finished the med rework I would do this. I disagree with the armor sleep shit entirely its an extra un-needed step. but I get its application for plate and heavy chest shit. this removes the flag from all but chest armor. gloves, hats etc. IF you are wearing plate shit odds are you ahve a chest piece for it since this is most important. it makes it where only removing that is needed basically. this trims the bullshit without removing the principle behind it.


Nakalon:
I generally agree with this but wouldn't remove it from helmets

Gyrandola:
Yeah, keep it for helmets honestly imagine sleeping lying down with a bucket helmet

Nakalon:
> left off of crown. I hate kings but a king would absolutely sleep with their crown regardless of discomfort....

sure makes sense!

PR #1375, opened 2024-11-01
Now the announcements will use the proper titles

TheFenix323:
## About The Pull Request
Made the throne now store the title of the person using it's commands.
announcemet
Outlaw
Decree
Taxes
Scom decree

## Why It's Good For The Game
It fixes #1251

PR #1399, opened 2024-11-03
Changes how the title of the Lord is stored for announcements

TheFenix323:
## About The Pull Request
Now it's all stored in a global var that is already inizialiced with "Duke" in case any of the announcements are called outside of the Throne.
This var will now only change when the priest coronates a new ruler.

Duchess
Duke

## Why It's Good For The Game
Fixes an issue introduced with #1375, that caused the outlaw special trait to make an announcement with a wrong title.

image


Nakalon:
speedmerging cuz fix

PR #1403, opened 2024-11-03
Werewolf no longer delete things they are holding.

TheFenix323:
## About The Pull Request
Makes Werewolfs drop the things they are holding when they summon their claws.
(Also fixed the proc having wrong indentation).

Holding the sword.
Holding something

Now you drop the object in the ground.
sword droped

## Why It's Good For The Game
They had the potential of deleting objects with how it worked before.


Nakalon (approved):
everything looks pretty and fine, you're pretty and fine, thanks for this <3

PR #1404, opened 2024-11-03
Clinic Map Adjustments

ProbablyCarl:
## About The Pull Request
Significantly reduces the number of items the clinic receives roundstart, alongside making it less boxy. Cleans up the surrounding area, removes floating floors / walls and tries to make use of the space as best as possible on the interior. Not much more to say.

Guards no longer have a _warhammer_ and instead get cudgels. Reduces their swordsmanship skill entirely to null, since arming swords no longer spawn for them and they don't get a short sword by default.

## Why It's Good For The Game
It's subjective, ultimately, but the balance was kind of out of whack.


Nakalon:
you got conflicts yoooooooo

PR #1405, opened 2024-11-03
Limits green lights to the keep, replaces the city ones with lantern posts

Gyrandola:
## About The Pull Request

Title. We dont have an electrical grid system thing yet and it was a bit of a culture shock.

here is the lantern post for the city:
Registrazione 2024-11-03 140433


john-pawson:
Hey, i was messing around with the lamp posts a while back, might want to make it so you can't take the lamptern off to prevent people stealing the free lampterns hanging from a post immidiately.
You can put a torch on the post and it'll work exactly the same as a lamp though, so you could just let people steal the lamptern and put a torch in.

JohnnyJohnJohnson:
> Hey, i was messing around with the lamp posts a while back, might want to make it so you can't take the lamptern off to prevent people stealing the free lampterns hanging from a post immidiately. You can put a torch on the post and it'll work exactly the same as a lamp though, so you could just let people steal the lamptern and put a torch in.

The torch on a lamptern post won't stay lit, it goes out on it's own like a torch on the ground. They should work like sconces if that's going to be possible.

PR #1413, opened 2024-11-03
racial harmonize 

Luctus-Comptus:
again didnt want this to go to waste. this removes all racial negatives and gives a single point bonus into a stat. so people can not feel gutted playing a fucking reskin. and im certain some "well actually" assholes will argue what the one point should be. shut it. the entire point is race shouldnt give a massive negative or buff. unless seelie. because small. their shits weird so I left it alone.

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/furry/akula.dm
+++ b/code/modules/mob/living/carbon/human/species_types/furry/akula.dm
@@ -38,10 +38,10 @@
 	specstats = list(
 		"strength" = 0,
 		"perception" = 0,
-		"intelligence" = -1,
+		"intelligence" = 0,
 		"constitution" = 0,
 		"endurance" = 1,
-		"speed" = 1,
+		"speed" = 0,
 		"fortune" = 0
 		)
 	enflamed_icon = "widefire"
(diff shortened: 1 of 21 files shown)


wow50000:
Nah I like my racist stats

DelineFortune:
Azure peak were right

Gremillion:
i play almost only elves and i love the stat difference. it would feel really weird if wimpy skinny elves were almost on the same level as horcs or even humens for example. ratwood players cant be trusted to roleplay "this race is known as weak" without having the mechanics force it. something like this would end up creating even more lore-ignoring/snowflake characters.

wow50000:
> > Nah I like my racist stats being extreme
> 
> and I like closing your pr's

honestly the PR's you close have been a great choice to close and I love when you give feedback on why you closed them

babamandaa:
respectfully, if players feel gutted playing a reskin then maybe what we need is a balance update instead of just making all of their stats the same? races are a big part of player agency - an orc player will want to be strong but dumb, a dwarf player will want to be tough but slow, an elf player will want to be weak but smart. if some races (apparently vulpkanin) aren't fitting what players want them to be, it might be a good idea to update them to be more in-line with what vulp players want them to be :3

removing that is something that literally nobody has asked for, and there should definitely be a more elegant solution than just burning it all down. also sending hugs because you seem stressed recently ;w; ppl are worried about you.

PR #1414, opened 2024-11-03
agnostic (not really)

Luctus-Comptus:
neutral faith, can be taken for either undecided of who to follow while not being faithless themselves. cleric and acolyte spells are pulled from various of the 10 to give it a mix. its more the "I worship the ten" role


Heheniy:
dont give them churn

also i think it is posible to make list(s) of spells and then randomly give them spells out of it

Gyrandola:
the faithless shouldnt be able to cast miracles at all

code/datums/gods/patrons/agnostic.dm
```diff
@@ -0,0 +1,19 @@
+/datum/patron/lost
+	name = "The Gods"
+	domain = "The Ones Provided"
+	desc = "Although gods obviously exist in this world, none have specifically earned your unconditional devotion. You are more grounded in a broad spectrum. This is not to say you don't believe in them or even dislike them. You have just yet to be convinced by one. You even speak with whichever one will listen."
+	worshippers = "the neutral, the inquisitive, the indecisive"
+	associated_faith = /datum/faith/lost
+	preference_accessible = TRUE
+	mob_traits = list(TRAIT_NOMOOD)
+	undead_hater = FALSE
+	t1 = /obj/effect/proc_holder/spell/targeted/blesscrop /////////////////// this is ment to simulate as a cleric/acolyte asking each for their favor. only thing I could think of a lil from all with praying for anyone to revive someone as the last.
```
Nakalon, line 10: I agree with Gyran that these should have no miracles at all. Faithlessness or Agnostics should be reserved to those outside of the church. Doesn't make much sense for them to be in a faith role if they're not deeply connected with faith itself.

Nakalon (changes requested):

PR #1415, opened 2024-11-04
Foresworn Buffs / Identity Clarification

babamandaa:
## About The Pull Request

This PR reworks Foresworn - before, they were a slightly faster but ultimately worse version of Brigand; now, they're skilled defensive powerhouses that will make any boggard shit their pants!

## Why It's Good For The Game

I genuinely have only ever seen Foresworn played once, and the player admitted afterwards they picked it by accident. While it's interesting from an RP perspective, old foresworn is just kind of boring mechanically - 2 less strength and one less wrestling in exchange for a bit of speed and different starting weapons. This PR lets the Foresworn punch just as high as their strong and/or ranged counterparts by making them something entirely different instead of a weird in-between of Knave and Brigand.

It could be argued that giving them expert weapons is a step too far; however, since the flavor of Foresworn is "Bandit Sellsword/Mercenary" I think it makes sense for them to have skills on par with them. If that seems shaky, maybe instead buff their athletics to 4 and return their skills to journeyman! Although I don't think it's a problem - a Knave with 12 speed (already a pretty low roll, usually they have 14-15) has 180% defense with dodge, so giving sword/shield foresworn 170% defense with parry will be fine.

TL;DR: Current bandit classes are Strong, Fast, and Not-As-Strong-Or-Fast. This PR changes that to Strong, Fast, and Tough. My intention isn't to buff bandits as a whole, it's just to make this class of bandit more appealing.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/antag/foresworn.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/antag/foresworn.dm
@@ -25,7 +25,7 @@
 	H.mind.adjust_skillrank(/datum/skill/misc/climbing, 3, TRUE)
 	H.mind.adjust_skillrank(/datum/skill/misc/sewing, 1, TRUE)
 	H.mind.adjust_skillrank(/datum/skill/misc/treatment, 1, TRUE)
-	H.mind.adjust_skillrank(/datum/skill/misc/athletics, 3, TRUE)
+	H.mind.adjust_skillrank(/datum/skill/misc/athletics, 4, TRUE)
 	belt = /obj/item/storage/belt/rogue/leather
 	pants = /obj/item/clothing/under/roguetown/trou/leather
 	shirt = /obj/item/clothing/suit/roguetown/armor/gambeson


katsukai:
*sigh I been arguing with Brigands being 4 STR for awhile with the PR owner, knew it would cause problems with other choices being less viable. Foresworn being the tank choice is alright but asking for Mary Sue buffs like bumping weapon skills to Expert is too much my opinion. There is enough powercreep.

code/modules/jobs/job_types/roguetown/adventurer/types/antag/foresworn.dm
```diff
@@ -48,9 +47,11 @@
 		if("Spear") //Deserter watchman. Maybe should be shield and spear? plenty of shields at the bandit hideout though.
 			r_hand = /obj/item/rogueweapon/spear
 			head = /obj/item/clothing/head/roguetown/helmet/skullcap
+			H.mind.adjust_skillrank(/datum/skill/combat/polearms, 1, TRUE) //Wielded defense goes from
```
katsukai, line 50:
```suggestion
			
```

katsukai: Expert skills too much my opinion.

code/modules/jobs/job_types/roguetown/adventurer/types/antag/foresworn.dm
```diff
@@ -48,9 +47,11 @@
 		if("Spear") //Deserter watchman. Maybe should be shield and spear? plenty of shields at the bandit hideout though.
 			r_hand = /obj/item/rogueweapon/spear
 			head = /obj/item/clothing/head/roguetown/helmet/skullcap
+			H.mind.adjust_skillrank(/datum/skill/combat/polearms, 1, TRUE) //Wielded defense goes from 
 		if("Sword & Shield") //Mercenary on the wrong side of the law
-			backl= /obj/item/rogueweapon/shield/wood
+			backl= /obj/item/rogueweapon/shield/buckler //Before, wielding the sword with two hands would give you 130% defense and sword + shield would give 150%; Now, wielding the sword with two hands gives 150% and sword + shield gives 170%.
 			beltr = /obj/item/rogueweapon/sword/iron //iron sword like literally every adventurer gets
 			head = /obj/item/clothing/head/roguetown/helmet/leather/volfhelm
+			H.mind.adjust_skillrank(/datum/skill/combat/swords, 1, TRUE)
```
katsukai, line 55:
```suggestion
```
Same here

code/modules/jobs/job_types/roguetown/adventurer/types/antag/foresworn.dm
```diff
@@ -34,11 +34,10 @@
...
 	neck = /obj/item/clothing/neck/roguetown/gorget
-	H.change_stat("strength", 2) //less buffs than brigand but less int debuff also somewhat faster
+	H.change_stat("strength", 2) //Removing the speed, buffing CON and INT to double down on Foresworn being a "Bandit Tank" (Brigands are hard-hitting and great grapplers, Knaves are fast/ranged, Foresworn are sturdy and skilled)
 	H.change_stat("endurance", 2)
-	H.change_stat("constitution", 2)
-	H.change_stat("intelligence", -2)
-	H.change_stat("speed", 2)
+	H.change_stat("constitution", 3)
+	H.change_stat("intelligence", -1)
 	ADD_TRAIT(H, TRAIT_MEDIUMARMOR, TRAIT_GENERIC)
```
katsukai, line 41:
```suggestion
	ADD_TRAIT(H, TRAIT_MEDIUMARMOR, TRAIT_GENERIC)
	ADD_TRAIT(H, TRAIT_CRITICAL_RESISTANCE, TRAIT_GENERIC)
	ADD_TRAIT(H, TRAIT_STEELHEARTED, TRAIT_GENERIC)
```
Make it tankier and not so emotional.

babamandaa:
i like all of your suggestions! gonna commit all of them :3

katsukai:
Looks good, you have my blessings.

code/modules/jobs/job_types/roguetown/adventurer/types/antag/foresworn.dm
```diff
@@ -25,7 +25,7 @@
 	H.mind.adjust_skillrank(/datum/skill/misc/climbing, 3, TRUE)
 	H.mind.adjust_skillrank(/datum/skill/misc/sewing, 1, TRUE)
 	H.mind.adjust_skillrank(/datum/skill/misc/treatment, 1, TRUE)
-	H.mind.adjust_skillrank(/datum/skill/misc/athletics, 3, TRUE)
+	H.mind.adjust_skillrank(/datum/skill/misc/athletics, 4, TRUE)
```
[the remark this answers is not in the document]
atomicplan666: Remember, athletics is the BLUE bar.

RadiantFlash7:
Funnily enough,  the actual test merge of Bandits had most of these changes to begin with. They proved to be too much. We actually had more foresworn then brigands or knaves during the test merge, though a lot of them went to using the spear as a goto because the spear was 'the meta'. That said, some people said they prefered the speed bonus from foresworn and a more balanced, generalist stat layout over the straight strength of brigand.

The general idea of bandits is that they are worse geared, with better stats then bogguards, putting them on an even fight- and then outdoing bog guards as they get better gear- either from the idol or a smith they abduct. 

Bucklers are actually quite a bit better then wooden shields at parrying, and bumping up their skill to expert puts them significantly above bogguards, which is very much not ideal at all. At that point, the role isn't a tank role, but just capable of flat out winning 1 on 1s. 

It's worth noting that endurance directly affects your stamina bar, and while knaves do have significant dodge ability- especially if they go for daggers- they have any endurance increase in stats. Dodges are /costly/ stamina wise, and you're likely to exhaust a knave if they try to straight up fight using dodge only. You have high defense, and probably won't get hit- for a limited time.

Foresworn on the other hand would become a powerhouse with more endurance and parry/chance to hit buffs.

While I don't /mind/ them becoming 'tanky-' This PR doesn't particularly do that very well. Con by itself purely affects total health of the limb, and in no way affects critical hits to my knowledge. And most people die from crits, be it arteries, broken bones, or otherwise. Meaning Con is pretty much only useful for the health of a limb until it's in a non-critted, crippled state. For that matter, crit resistance trait is helpful, but doesn't actually do much for crit resistance, ironically. It'll stop groin crits, knockout blows to the head, neck dislocations, and if I recall correctly, prevents crits on ears.
--------------
In summary, The expert skill in swords and the buckler should probably go. The con change is... okay? I'd argue it's really not worth losing the speed bonus that foresworn get. Athletics increase is fine I suppose.

code/modules/jobs/job_types/roguetown/adventurer/types/antag/foresworn.dm
```diff
@@ -1,4 +1,4 @@
-/datum/advclass/foresworn //Strength class, starts with axe or flails and medium armor training
+/datum/advclass/foresworn //Tanky class, starts with good con/end and spears or sword/shield
```
RadiantFlash7, line 1: solid whoops on my part. That comments was copy pasted from brigand. Foresworn was designed more of a generalist, then a strength class.

code/modules/jobs/job_types/roguetown/adventurer/types/antag/foresworn.dm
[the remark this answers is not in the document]
RadiantFlash7: > Remember, athletics is the BLUE bar.

Incorrect. Athletics affects the green bar at a minimum. Refer to:
https://github.com/Rotwood-Vale/Ratwood-Keep/blob/6d2dd6b65e554ac8fd59a72e36a04968743b6698/code/modules/mob/living/carbon/rogfatstam.dm#L5

babamandaa:
the expert skills got taken back - now, they're back to 130/150 defense :3

code/modules/jobs/job_types/roguetown/adventurer/types/antag/foresworn.dm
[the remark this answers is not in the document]
Nakalon: > Isn't that impactful, my opinion.

I know it isn't! But I'd rather the tank class not be athletic. But that's also an issue with how athletics actually work...

Nakalon:
After careful deliberation and getting some feedback I have ultimately decided against this.

The core idea is nice enough, but as it was pointed out, unfortunately this will not achieve the desired effect.
Balance is a very risky thing to mess with, but it was proven by the extended duration the new bandits have been in testing that the values they have now are in a better spot than they were during testing (which somewhat resemble some of the proposed values in this PR).
It's very important for us Maintainers to ward of bloat and possible Power/Feature creep, which sadly means we sometimes have to close PRs like this, which aren't bad, but also don't exactly achieve what they set out to!

I'm sorry!

babamandaa:
i actually appreciate this sentiment! thanks for giving it consideration anyways :3

PR #1416, opened 2024-11-04
Disables player deadites and removes rotmen. Fixes cure rot.

Gyrandola:
## About The Pull Request

title.

## Why It's Good For The Game

cure rot actually works now. player deadites are once again disabled because theres basically zero rp you can do as a zombie.

PR #1418, opened 2024-11-04
deadite removal || if you need this gyran its here...

Luctus-Comptus:
this solves your deadite problem while leaving rotting in as a mechanic.... it just removes players transforming to zombies 4min into rotting. strips every drop of zombie and antag datums from the process.....

I dont fucking know why I bother. probably because this is such a big problem even if im leaving I dont want people still fucked over by it.  #1416 

its open to maintainers just do what you need with it.

### DIFF
--- a/code/datums/components/rotting.dm
+++ b/code/datums/components/rotting.dm
@@ -36,25 +36,13 @@
 /datum/component/rot/corpse/process()
 	..()
 	var/mob/living/carbon/C = parent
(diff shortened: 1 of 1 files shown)


Gyrandola:
Appreciate the help, but dead players transforming into deadites _is_ an intended mechanic. the problem was cure rot not working. I'll be closing this, thanks though.

PR #1423, opened 2024-11-04
Allow admins to toggle aghost invisibility

Gyrandola:
## About The Pull Request

- Ports https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/759

proof of testing:
image

## Why It's Good For The Game

Not every admin wants their character to disappear when they aghost.
This also prevents the annoying bug where the character simply disappears into the void and can no longer be gotten back into.


Nakalon (approved):
big bless

PR #1424, opened 2024-11-04
Re adds milk cheese

wow50000:
## About The Pull Request
Re adds milk cheese

## Why It's Good For The Game
woe be upon balance guys (also its funny)


Nakalon:
I disagree with this.

Having it be breatmilk is leagues funnier.

Nakalon:
After internal discussion we have decided against this.
Either because it was a fair balance change at the time...
Or because we're biased for breast milk and like to have jugs of it at home to drink for our own enjoyment.

Sorry!

PR #1427, opened 2024-11-04
Privacy in the Keep

Gyrandola:
## About The Pull Request

Removes royal chambers access from captain, knights and hand.
Only the royal family has access to their own private chambers (heirs and duke/duchess).

## Why It's Good For The Game

Quite a few people complained about having no privacy in the keep and constantly being interrupted by people opening doors without permission.


K4rlox:
maybe physician should have access to the royal chamber considering the physician room is directly connected to the king's bedroom?

Nakalon (approved):
yeah!

PR #1431, opened 2024-11-04
Removes Tolerant trait and heavily limits Steelhearted trait

Gyrandola:
## About The Pull Request

- Removes the tolerant trait altogether. 
- Removes the Steelhearted trait from all classes except antagonists.

## Why It's Good For The Game

The way Tolerant is implemented frankly does not make sense to me, its only use is overriding the xenophobic trait, but why should someone even have both at the same time, and furthermore, why should entire jobs automatically be tolerant?

Steelhearted removes stress gain from killing, dismembering and gib-related stuff. An unhealthy amount of classes have this trait in the name of "desensitization", yet what it truly does is allow almost every combat class to get away with murder with no repercussions on the character's psyche. Removing it will passively discourage fragging and make mood matter more.


Nakalon:
Did you handle the one shot thing that was mentioned?

Gyrandola:
> Did you handle the one shot thing that was mentioned?

No and I dont intend to. It's a pretty brainless change

Nakalon:
> > Did you handle the one shot thing that was mentioned?
> 
> No and I dont intend to. It's a pretty brainless change.
> Im not gonna arbitrarily give church and keep roles crit resistance for literally no reason.

I agree! Awesome!

code/modules/jobs/job_types/roguetown/courtier/jester.dm
```diff
@@ -60,7 +60,5 @@
 	ADD_TRAIT(H, TRAIT_LEAPER, TRAIT_GENERIC)
 	ADD_TRAIT(H, TRAIT_NUTCRACKER, TRAIT_GENERIC)
 	if(prob(50))
```
Nakalon, line 62: Dunno if this should just be 100 since its no longer an either or

Nakalon (approved):
Seems alright! Thanks for deleting commented out code!

PR #1435, opened 2024-11-05
Clinic Removal

ProbablyCarl:
## About The Pull Request
To be clear, there's probably vestigial elements in the code still, but I 'unno about that. A quick search got everything.

In any case, this returns the map's old layout before the Clinic, for the most part.
Removes the Clinic Physician. Gives their in game material to the Court Physician.
Adjusts the Town Physician's building to keep up.

## Why It's Good For The Game
The Clinic took emphasis away from the Court Physician, alongside the Town's Doctors, before throwing them into a role that was overall redundant. I _**PERSONALLY**_ feel like it was added for the sake of it, not that there's anything wrong with that, though with no consideration to anything else. Especially as the individual who made the PR specifically called out Court Physician in a negative light, pushing the changes more into a spite direction. I've had a number of folks approach me for this PR in particular, after I proposed a revert on the Clinic mapping and roles.

### DIFF
--- a/_maps/map_files/rockhill/rockhill.dmm
+++ b/_maps/map_files/rockhill/rockhill.dmm
@@ -178,6 +178,7 @@
 	dir = 8
 	},
 /obj/item/reagent_containers/glass/mortar,
+/obj/item/reagent_containers/glass/alembic,
 /turf/open/floor/rogue/ruinedwood{
 	dir = 1;
 	icon_state = "vertw"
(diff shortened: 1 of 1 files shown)


Gyrandola (changes requested):
Give the town doctor a tiny small clinic  / laboratory in the center of town, maybe adjacent to the church. Nothing huge like the one we are scrapping, but lets at least keep it as a concept. The clinic doctor can indeed go, I suppose, since its redundant.

Nakalon:
I agree with Gyran, also make sure the guards are dead and gone plis plis plis

Gyrandola (approved):

PR #1443, opened 2024-11-05
Removal of stockpile machine from farm

Heheniy:
## About The Pull Request
See title

## Why It's Good For The Game
Decision to place stockpile right in the farm is a VERY bad idea as farmers now can earn hundrents of mammons without even leaving their farm fort


Gyrandola:
I agree, but I want to see what other maintainers think

PR #1455, opened 2024-11-05
Guildhall Adjustments 2.0

ProbablyCarl:
## About The Pull Request
Removes the garden. Makes use of the area previously utilized for it.

## Why It's Good For The Game
Objectively, it's bad to take work from the soilsons. They're already starved as is, like the people they refuse to feed!!!!!!


Nakalon (approved):
looks good!

PR #1458, opened 2024-11-05
Now old Court magicans can scry

### LINKED ISSUE #1402: [BUG] Legendary arcanists can't scry
## Describe the bug

Old court mages, starting with legendary Arcane skill, cannot use the scrying orb. It shatters in-hand on 100% of attempts regardless of target or timing. Mages with master or higher aren't supposed to be able to shatter the orb at all on the first attempt, and legendary skill is supposed to completely prevent shattering in 100% of attempts regardless of timing.

## Round ID:

Zard1136-00.47.09

## Steps to Reproduce:

1. Be old
2. Be court magician
3. Attempt to scry someone
4. Fail to scry; only cry

## Additional context

This doesn't apply to mages with skill less than legendary, it seems, who can still scry as expected.

TheFenix323:
## About The Pull Request
Fixes the if switch using an "or" instead of a "to" to include level 5 and 6 of arcana.

magican
Scrying

## Why It's Good For The Game
fixes #1402

### DIFF
--- a/code/game/objects/items/rogueitems/magic.dm
+++ b/code/game/objects/items/rogueitems/magic.dm
@@ -80,7 +80,7 @@
 				success_chance = 90
 			if(SKILL_LEVEL_EXPERT)
 				success_chance = 94
-			if(SKILL_LEVEL_MASTER || SKILL_LEVEL_LEGENDARY) // Magus has this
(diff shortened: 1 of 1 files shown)


Nakalon:
I fucked this 😄 thanks!

Nakalon (approved):
fixes my mistake, im not totally right now learning "to" is something that exists

PR #1463, opened 2024-11-06
Merc guild double table fix

Heheniy:
## About The Pull Request
Fixes multiple table on one tile at mercenary guild
Sligtly moves wall candles and bandit note to make the room look better

## Why It's Good For The Game
Why wouldn't it?


Heheniy:
mapdiff shows non-existing candle?
image
image

Gyrandola:
image
why

Heheniy:
that candle is on the right side of the wall

PR #1472, opened 2024-11-07
Make both soilson and soilbride get keys

TheFenix323:
## About The Pull Request
Moves the repeated things be before the if, including the key which the soilbride lacked.

image

## Why It's Good For The Game
I see no reason for a Soilbride to not have keys.


TheFenix323:
Welp github desktop did a thing

PR #1485, opened 2024-11-07
No more bm'able spells

wow50000:
## About The Pull Request
Removes purge, due to it being an instant artery
there are some chems that need to be purged to be removed, I couldnt find those chems but they should be called something like "purifying elixir" 
probably missed some other things but this is a slop job to fix something absolutely horrendous to the game

## Why It's Good For The Game
Instant artery spell with no counterplay is coal, thanks mori

PR #1487, opened 2024-11-07
Purge reagents now needs a bed

john-pawson:
## About The Pull Request

Tired of doctors being one of the most lethal roles to fight against since they can instantly open your artery? Well no more, now they need you to be nearby or lying on a bed to do such thing.

- Purge now needs a bed to be within two tiles of the person casting it (Only did it this way since it's the same as other spells use, but with shorter range)
No bed
Captura de pantalla 2024-11-07 191435
Captura de pantalla 2024-11-07 191445

Yes bed
Captura de pantalla 2024-11-07 192928
Captura de pantalla 2024-11-07 192940

## Why It's Good For The Game

I think it's a better way of fixing it rather than https://github.com/Rotwood-Vale/Ratwood-Keep/pull/1485 being a, as menitoned by the creator itself, slop fix.
Also prevents doctors from immidiately opening your artery and killing you.

### DIFF
--- a/code/modules/spells/roguetown/acolyte/sawbones.dm
+++ b/code/modules/spells/roguetown/acolyte/sawbones.dm
@@ -248,6 +248,17 @@
 		return TRUE
 	return FALSE
 
+/obj/effect/proc_holder/spell/targeted/purge/cast_check(skipcharge = 0,mob/user = usr)
+	if(!..())
+		return FALSE
+	var/found = null
+	for(var/obj/structure/bed/rogue/S in oview(2, user))
+		found = S
+	if(!found)
+		to_chat(user, span_warning("I need to lay them on a bed"))
+		return FALSE
+	return TRUE
+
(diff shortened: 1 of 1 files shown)


Nakalon (approved):
I like this change but would rather it require you to be layed down. This makes it so you can't really use it in the field!

john-pawson:
> I like this change but would rather it require you to be layed down. This makes it so you can't really use it in the field!

Well then there's the problem that you can push someone into a wall and cut open their artery again, besides if you're planning on working on the field you should probably bring a sleepcloth or make an improvised bed either way

Nakalon:
> > I like this change but would rather it require you to be layed down. This makes it so you can't really use it in the field!
> 
> Well then there's the problem that you can push someone into a wall and cut open their artery again, besides if you're planning on working on the field you should probably bring a sleepcloth or make an improvised bed either way

good enough for me! Thanks for this!

PR #1496, opened 2024-11-08
[TM First] Feast & Famine: Nutrition changes

Gyrandola:
## About The Pull Request

- Adds healthpotion overdose
- ports https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/779

> Nutrition is a bit of a squandered gameplay mechanic in this codebase.
> 
> Not a whole lot of stuff interacts with it like you'd expect, so I've gone ahead and taken the liberty of putting in some very basic mechanical interactions that make hunger deplete slightly/significantly faster depending on your character's stats and level of physical activity.
> 
> To help explain this, max nutrition (full) is 1000 nutrition.
> 
> Anything that causes you to gain fatigue also causes you to lose nutrition/hunger. A 10 fatigue strike causes you to lose 10 nutrition, adjusted by the following values below:
> 
> - If you have any toxin damage above 20% of your max health, 10% of that damage is added on to the base nutrition you lose.
> - If you have less than 30% fatigue remaining, the amount of nutrition consumed is doubled. (You're tired, working over time!)
> - If your CON is below 10, you take an extra 10% increase to the amount of nutrition lost for every point below 10. 7 END equals a 30% increase.
> - If your CON is above 10, the amount of nutrition you lose is reduced by 5% per point.
> - If your STR is above 10, the amount of nutrition you consume is increased by 7.5% per point. Added by popular request. Strengthmaxxers beware.
> - If you've recently eaten (are well fed), the total amount of nutrition you lose is reduced by half. (The well fed nutrition threshold is 700, so this essentially gives you an extra 600 effective nutrition to lose if you keep yourself well fed.)
> - If you're still digesting food, the amount of nutrition you lose is reduced by 10%. Rest and digest!
> - In addition, red healing potions now also cause a considerable loss of nutrition (they're causing you to regenerate rapidly, basically). Dated reds cause the most loss, while fresh red have half as much loss. Miracles and other forms of healing are unchanged.

## Why It's Good For The Game

> Food is very much an afterthought at the moment. Most people don't interact with the thought of consuming food more than a few times a round as it stands, even if they're doing highly physical activities.
> 
> The intent behind these changes is to give people more realistic opportunities to do things like adventurer campfire cooking moments. Carrying cooking kits with you should become a lot more valuable. A frypan is definitely way more of a nice thing to have now, and END has a bit more value for helping you get a lot more out of the food you do eat.
> 
> It is my hope that this'll get people interacting with food producers and cooking more, and make it a bit more of a thing for people to consider in a round.
>

### DIFF
--- a/code/modules/roguetown/roguejobs/alchemist/reagents.dm
+++ b/code/modules/roguetown/roguejobs/alchemist/reagents.dm
@@ -1,20 +1,3 @@
-
-/datum/reagent/medicine/healthpot/overdose_start(mob/living/M)
-	if(ishuman(M))
-		var/mob/living/carbon/human/H = M
-		if(!istype(H.dna.species, /datum/species/werewolf))
-			H.playsound_local(H, 'sound/misc/heroin_rush.ogg', 100, FALSE)
-			H.visible_message(span_warning("Blood runs from [H]'s nose."))
-	. = 1
-
-/datum/reagent/medicine/healthpot/overdose_process(mob/living/M)
-	if(ishuman(M))
-		var/mob/living/carbon/human/H = M
-		if(!istype(H.dna.species, /datum/species/werewolf))
-			M.adjustToxLoss(2, 0)
-	..()
-	. = 1
-
 /datum/reagent/medicine/healthpot
 	name = "Health Potion"
 	description = "Gradually regenerates all types of damage."


Heheniy:
bandits should get much more food at their camp with these changes

code/modules/roguetown/roguejobs/alchemist/reagents.dm
```diff
@@ -19,11 +19,28 @@
 		//some peeps dislike the church, this allows an alternative thats not a doctor or sleep. 
 		M.heal_wounds(2) //at a motabalism of .5 U a tick this translates to 80WHP healing with 20 U Most wounds are unsewn 15-100. This is powerful on single wounds but rapidly weakens at multi wounds.
 		M.update_damage_overlays()
```
Elli-Skala, line 21: > overdose_threshold = 0
Won't this mean potions have a overdose effect but no one can ever actually od on them?

wow50000:
I feel like con shouldnt have an effect on nutrition loss but giving str scaling with nutrition loss makes sense

CloverIsLucky:
> I feel like con shouldnt have an effect on nutrition loss but giving str scaling with nutrition loss makes sense

I agree with WoW, don't make constitution affect nutrition stat.

Gyrandola:
I gave it some time, this has been reverted on Azure as it seems to be way too imbalanced. Forget it

PR #1499, opened 2024-11-08
Runtime fix for fog walls + show fuzzy time remaining to bandits

zoktiik:
## About The Pull Request
*With apologies to Carl.*

- Fixes the insane amounts of runtimes caused by fogwalls trying to access the `mind`s of `mind`-less mobs *everywhere*

And, because I always get carried away:
- Prevents the fog wall from spamming your chat every 0.5s you walk into it (now: 40-second cooldown. Uses a timer, a global list, and a proc to remove the mob from that list.)
- Makes the warning message less unnecessarily direct and IC-ly weird. And more informative (vaguely) as to how long before the barriers go down! 
- Examine will also give you a vague description of remaining time.

Future suggestions: perhaps make a single global timer that gets added on round start, and asynchronously delete all the walls when it triggers, instead of having 346 different 30-minute timers. Likely not going to be a huge performance diff, but it would be cleaner.
Having the timers start on round start, rather than when initialized, would also make the walls *actually* last 30 minutes, instead of 30m minus however long round setup took. CBA to figure that out now because I've spent enough time on this sideproject.

## Why It's Good For The Game
Screenshot from 2024-11-07 16-06-38
Screenshot from 2024-11-07 17-15-47

### DIFF
--- a/code/game/objects/structures/timed_blockers.dm
+++ b/code/game/objects/structures/timed_blockers.dm
@@ -7,15 +7,20 @@
 	anchored = TRUE
 	opacity = FALSE//from TRUE
 	resistance_flags = INDESTRUCTIBLE | LAVA_PROOF | FIRE_PROOF | UNACIDABLE | ACID_PROOF
+	var/walltimer // tracks this wall's specific timer
 
-	var/dissipate_time = 30 MINUTES
-
+#define FOG_DISSIPATE_TIME_MAX 30 MINUTES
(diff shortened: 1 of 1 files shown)


Nakalon:
I love this but its too late to proper review (can't think straight).

Please, if you may, please please please think of a solution outside of "invisible wall" to replace the fog!

I really hate the idea and I believe you're the dragonborn or whatever, destined to take this on!

Gyrandola (approved):

Gyrandola:
Your code is very clean.

PR #1503, opened 2024-11-08
YUUUGE burial changes PR!! Now mind-based and connection-agnostic!

zoktiik:
## About The Pull Request

Here's my hastily-written, incomplete changelog. I've only tested the force-respawn logic a bit, but it's pretty brutally simple, and doesn't quite matter since it's not the default. 
Since this touches configs (harmlessly) here's a ratwood-config companion PR: Rotwood-Vale/Ratwood-config/pull/23

- **Funeralization now happens on the MIND DATUM level instead of the living mob.** Uses a single `funeral` variable within the mind datum. This will very reliably tie your buried mob to its ghosts and spirits, making sure that you are unaffected by a previous life getting buried.
- Living mobs can be funeralized and have it affect them, their ghosts, and their underworld spirits.
- Underworld spirits track a reference to their previous mind when they Journey, whether using the verb or the "AFTER LIFE" button. Adminghosts becoming spirits will have no previous mind reference.
- When buried, underworld spirits will be teleported back to their spawn after a 5s delay, and their toll will be immediately paid. It will also reverse any effects (paralysis, being summoned) from Speak with Dead.
- Speaking of Speak with Dead, it got a pretty big rework (more refactor). It no longer uses a 2-minute `sleep()`, opting to use timers instead. You can no longer get spirits without a client, or who are funeralized, or who are also already summoned. The proc now also boasts significantly more stable and consistent behavior (and faster casting, by using the new `spirit_list` as mentioned below). Abyssal language granting will never remove the language if you already have it.
- Revival will remove your funeral status and tell you as much. <sub>Except admin revives.</sub>
- Cure Rot will also remove your funeral status **if you're alive at the end of it** (i.e. you were a zombie.) 

- Shovel intents are swapped. SCOOP intent is first, STRIKE is second.
- Coffins can now be buried, and gravemarkers/Burial-Rites can be used on a coffin, or agrave with a coffin in it, to funeralize it.
- **Funeral Rites now only funeralizes the tile directly in front of you. It checks for a coffin, then a grave if there is no coffin or it can't be funeralized.**
- Gravemarker construction only funeralizes the first grave (who would have thought?)
- **Only the first container within a grave or an unburied coffin is checked.** All others are skipped. This does not, however, apply to corpses. Mass graves ahoy!

- New global list: `spirit_list`. Avoids having to iterate through the whole carbon list just for roguespirits.
- Using `mob.get_spirit()` is a lot less expensive now thanks to the new list. It also now has a check to only get the spirit if it shares a corresponding mind datum (from the same respawn).

- Because the poll was so indecisive, there is a new Ratwood config option: `FORCE_RESPAWN_ON_FUNERAL`. You (as in the one with config or VV access) can toggle whether players are aggressively forced back to lobby when you funeralize them! This is less useful immediately now, more useful in case any reworks, tests or events are done. Default is OFF. Behavior for not forcing is well-fleshed-out, especially with regards to underworld spirits.
- A new proc, `funeral_login()` called on `/client/New()`, handles both login notifications for funeralized corpses (no force respawn) or immediately sending funeralized minds back to lobby (if toggled on).
- In both configurations, **burials will succeed regardless of whether the player is connected or not.**
- If you get returned to the lobby directly, whether by using the Journey/AFTER LIFE procs while funeralized, or you get force-respawned (connected or not), the game will output a short message to chat AFTER the MOTD if you got respawned from a recent burial. 

## Why It's Good For The Game
Burial now has an actual place in the game, with much more stable and consistent behavior overall.

### DIFF
--- a/code/modules/mob/living/living_defines.dm
+++ b/code/modules/mob/living/living_defines.dm
@@ -171,8 +171,6 @@
 	var/list/mob_descriptors
 	var/list/custom_descriptors
 
-	var/funeral = FALSE // Whether the body has received rites or not.
-
 	/**This variable updated in mob_movement.dm primarily. Mainly a shitcode measure for existing shitcode because this is SHITCODE!
 	 * All it does is track when a mob is sneaking so we don't have to constantly reset alpha values as this fucks with how things are intended to be.
 	 * if you really need to cancel someone who is sneaking, call update_sneak_invis(TRUE).*/


Gyrandola:
@Useroth

Gyrandola (approved):

PR #1510, opened 2024-11-09
Stun Mace Return

ProbablyCarl:
Makes stunmaces drain stamina and fatigue. When a target is at no stamina, it acts like it used to, electrocuting and stunning the target. The mace does 6 damage when active, 25 when off.
Also gives the guards gorgets. Because they used to have neck protection.
## Why It's Good For The Game
The mechanic was removed temporarily pending rework. I'd argue this is technically a rework.

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/special.dm
+++ b/code/game/objects/items/rogueweapons/melee/special.dm
@@ -147,8 +147,8 @@
 			target.electrocute_act(5, src)
 			charge -= 6
 		else//TODO: Check target.STACON!!!!!!!!!!
-			target.rogstam_add(-25)
-			target.rogfat_add(15)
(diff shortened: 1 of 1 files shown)


Nakalon:
I'mma wait on Gyrans aproval since they're the one this is being done for 😳

Gyrandola (dismissed):

hugmeorelse:
They should get chain coifs instead of gorgets.

ToothyDeer:
we did it guys. we got ss13 sec in roguetown

Heheniy:
4 hits to win any fight and 7 to leave target completely without stamina

no

hugmeorelse:
Un-TMed.

Gyrandola (approved):

PR #1515, opened 2024-11-09
Lightning bolt Nerf/adjustment

RadiantFlash7:
## About The Pull Request
Lightning bolt no longer hard stuns. Instead, it causes a second long paralysis, similar to repulse. 

## Why It's Good For The Game
This should be much less terrible feeling for people on the recieving end, while still being rather useful. A ranged knockdown that allows a mage a bit of breathing room to cast another spell, that isn't a hitscan fight ender.


RadiantFlash7:
Ooh. This is the wrong branch

PR #1529, opened 2024-11-10
Makes improvised surgery tools easier to craft

babamandaa:
Currently, you can make a stone table with one stone, and a stone cross (that is _taller than a person_ with two stones. So... why the hell does an improvised, hand-held saw take three stones to craft? This PR changes the total amount of ingredients to make a functional surgery kit (Two sets of forceps, a speculum, and a bonesaw, not counting a scalpel which can be any knife) from 9 (!!!) stone, 4 fibers, a _small log_ and 5 sticks to 1 stone, 7 sticks, and 4 fibers, something much more manageable since stones can't be stacked like sticks and fibers can.


Nakalon (approved):
I agree, I don't think these were well thought out.

PR #1541, opened 2024-11-11
Watch nerf

Heheniy:
## About The Pull Request
maces pick(3,4) -> pick(3,3,4)
polearms, knives, whips, bows 3 -> 2 (not their main weapons but still could be trained to use them)
climbing 3 -> 1 (why it was so high?)
sneaking 3 -> 0 (why they had it?)
riding 2 -> 0 (too fancy and expensive for common watch, also no saigas in town)

## Why It's Good For The Game
Town watch, who are already getting trained but still not full-on soldier, have too high skills, some even better then MAA, when their job is to uphold law in the town, while last ones are real soldiers.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/garrison/citywatch.dm
+++ b/code/modules/jobs/job_types/roguetown/garrison/citywatch.dm
@@ -52,6 +52,7 @@
 	guard.mind.adjust_skillrank(/datum/skill/combat/whipsflails, 2, TRUE)
 	guard.mind.adjust_skillrank(/datum/skill/combat/knives, 2, TRUE)
 	guard.mind.adjust_skillrank(/datum/skill/combat/bows, 2, TRUE)
+	guard.mind.adjust_skillrank(/datum/skill/combat/crossbows, 2, TRUE)
 	guard.mind.adjust_skillrank(/datum/skill/misc/climbing, 1, TRUE)
 	guard.mind.adjust_skillrank(/datum/skill/misc/athletics, 3, TRUE)
 	guard.mind.adjust_skillrank(/datum/skill/misc/swimming, 2, TRUE)


Nakalon:
I tend to agree with this but would need balance review form my peers

Heheniy:
heavy armor includes medium

Bobinater:
This has got to be the worst take I've seen in a long time...

Maces are supposed to be a watchman's /signature/ weapon, and regularly only rolled at 3, to further cuck them would be heinous.
Polearms are what ANY fighting force historically have been trained with even if they were things like the night watch or anything of the sort, if you don't have it you're going to get shit stomped because polearms are (as they should be) the meta. Where this really gets me is you want to gimp literally every combat skill they have, knife fighting is because guess what, they carry a knife and fight people carrying such so it isn't even a surprise to start. But you forget yourself and the fact that when rolling in combat, if you have zero weapon skill on what you're being ATTACKED with, it makes you eat shit. That's why EVERY combat role has them up. If you want to talk bloat we can talk vanguard because they have neigh OG bandit statblock skill wise.

Where this is the single worst take I've seen is wanting to remove climbing. You do realize Watchmen are supposed to be the ones you send to chase people down, club them and drag them back, right? You do realize that bandits, vanguards, etc all get climbing 3 because they need it just the same and if they can't climb up the same thing that a bandit does, bandits can just sit on rooftops all day and can't be touched by the force meant to fight them in the town.

Uweinee:
The climbing change feels like it's taking it a bit far? It's already annoying to chase bandits when they're fast and high on moondust-- now we want to make it impossible to climb after them? Huge thumbs down.

Nakalon:
I just remembered bandits are on the stronger side of things. Idk if this will be good.

Heheniy:
watch literally have whole town and keep on their side against bandits, comparing dry skills is stupid

Gyrandola:
Nobody seems to want this change.

PR #1543, opened 2024-11-11
Specials Not-Perma / Special Buff Removal

ProbablyCarl:
## About The Pull Request
Specials are no longer permanent on a per round basis, as opposed to per character. See argument below. Removes the secret buff to night vision. I don't like that this was touched at all, but it's whatever. Just removes the undocumented perception increase. Not the trait inclusion.

## Why It's Good For The Game
image


Gyrandola:
leak...

PR #1544, opened 2024-11-11
Watchmen Armoury Access

ProbablyCarl:
## About The Pull Request
Removes a copy of the Sheriff's keychain for some reason and gives armoury access to the Watchmen.

## Why It's Good For The Game
They bash down the door anyways, and there's rarely a Sheriff. While the former isn't reason to give them keys, I don't believe this would be too big a change. It just gives them access to an area they otherwise couldn't make use of in a round normally, alongside letting them make use of the skills they have.

### DIFF
--- a/code/game/objects/items/rogueitems/keyrings.dm
+++ b/code/game/objects/items/rogueitems/keyrings.dm
@@ -127,9 +127,6 @@
 	else
 		desc = ""
 
-/obj/item/keyring/sheriff
-	keys = list(/obj/item/key/town_dungeon, /obj/item/key/town_barracks, /obj/item/key/town_armory)
-
 /obj/item/keyring/marshal
 	keys = list(/obj/item/key/town_dungeon, /obj/item/key/town_barracks, /obj/item/key/councillor_rooms, /obj/item/key/keep_barracks, /obj/item/key/keep_dungeon, /obj/item/key/manor)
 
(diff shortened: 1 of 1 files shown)


Nakalon:
Isn't the sherif supposed to be like warden

Nakalon:
> > Isn't the sherif supposed to be like warden
> 
> Sheriff is the leadership of the Citywatch. So, kind of. Warden before Warden existed.

So shouldnt we keep access to the armory limited and punish players who bust the doors down?

Bobinater:
> Contributor

Punish them how? Punish them why?

You realize the reason they bust it down is because otherwise it gets sat on the entire round and they themselves are shit outta luck for example without a shield. You could say some shit like 'go to the smith' but I'm not waiting 30 minutes in line for a fucking piece of round wood.

radarseas:
honestly, yeah.


Its not like the watch's armory even has anything significant enough to really change much about the watch, if anything its just more restraints to encourage them to not kill people and shields to encouarge them to stick with their maces instead of using a polearm and decapping everyone.

the most, dangerous thing in there is like, _one, singular_ crossbow (needed to easily deal with rogues and those _with infinite stamina like werewolves_) and two steel daggers, _which they start with_.

Gyrandola (approved):

PR #1545, opened 2024-11-11
Avatar of Dendor

ProbablyCarl:
## About The Pull Request
Redoes the volf pit. Lowers it a Z-Level and replaces the volves with a singular Verewolf that's been cracked out. A lesson from Kingsfield. To get out of the pit now, you need the gate open to climb out, alongside needing a high enough climbing skill. Someone could probably throw down a chair, but I 'unno if that'd be enough for someone without climbing skill in general. Get tossed out I guess.

## Why It's Good For The Game
I found this funny. That's all. More interesting that three volves.


NikolayByondPlayer:
Won't it be too hard to get out?

PR #1548, opened 2024-11-11
Allowed Rogue patron changes

JohnnyJohnJohnson:
The PR allows Rogues to have any patron instead of Xylix.
## About The Pull Request

The change allows Rogues to select any of the Divine Ten as their patron God instead of only being able to select Xylix or inhumen Gods.

## Why It's Good For The Game

It feels unnecessarily restrictive to RP if the only god Rogues can mechanically worship is Xylix. It makes sense that Xylix would be the most common deity worshipped by Rogues, but I can see reasons why Rogues would worship other deities while still taking up their trade. For example, Swashbuckler is a variant of Rogue and a Swashbuckling Zorro-styled vigilante would have Ravox as their patron since a vigilante is about fighting injustice. It allows for more creative freedom with backstories and I can't think of a downside to allowing Rogues to be freeform with their patron.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/combat/rogue.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/combat/rogue.dm
@@ -10,7 +10,6 @@
 	category_tags = list(CTAG_ADVENTURER)
 
 /datum/outfit/job/roguetown/adventurer/rogue
-	allowed_patrons = list(/datum/patron/divine/astrata, /datum/patron/divine/noc, /datum/patron/divine/dendor, /datum/patron/divine/abyssor, /datum/patron/divine/ravox, /datum/patron/divine/necra, /datum/patron/divine/xylix, /datum/patron/divine/pestra, /datum/patron/divine/malum, /datum/patron/divine/eora, /datum/patron/inhumen/graggar, /datum/patron/inhumen/baotha, /datum/patron/inhumen/matthios)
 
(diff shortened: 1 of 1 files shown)


code/modules/jobs/job_types/roguetown/adventurer/types/combat/rogue.dm
```diff
@@ -10,7 +10,7 @@
 	category_tags = list(CTAG_ADVENTURER)
 
 /datum/outfit/job/roguetown/adventurer/rogue
-	allowed_patrons = list(/datum/patron/divine/xylix, /datum/patron/inhumen/graggar, /datum/patron/inhumen/baotha, /datum/patron/inhumen/matthios)
+	allowed_patrons = list(/datum/patron/divine/astrata, /datum/patron/divine/noc, /datum/patron/divine/dendor, /datum/patron/divine/abyssor, /datum/patron/divine/ravox, /datum/patron/divine/necra, /datum/patron/divine/xylix, /datum/patron/divine/pestra, /datum/patron/divine/malum, /datum/patron/divine/eora, /datum/patron/inhumen/graggar, /datum/patron/inhumen/baotha, /datum/patron/inhumen/matthios)
```
Gyrandola, line 13:
```suggestion

```
By default all patrons are allowed, no need to specify it.

JohnnyJohnJohnson: > By default all patrons are allowed, no need to specify it.

Ah, sorry then. I've literally never worked on BYOND code before. I hope it's good now, and that you actually like the idea.

PR #1550, opened 2024-11-12
Bandit migration

Rayforx:
## About The Pull Request
Adds back bandits being able to respawn with waves in a reduced manner.
I also ended up talking with the person who had removed it instead of reducing it, and he does not know why it was removed instead of being a reduced migration.

## Why It's Good For The Game
Currently, bandits tend to die very quickly, it can be 40 minutes into the round or an hour. But having some way that a few players (2 instead of 4) can come back to make some action for the bog gate, or to replenish their lines. Can do a lot of good. And this is even more with the respawn changes, and the future bandit rework.

### DIFF
--- a/code/datums/migrants/migrant_wave.dm
+++ b/code/datums/migrants/migrant_wave.dm
@@ -90,15 +90,6 @@
 
 /datum/migrant_wave/bandit_down_two
 	name = "Bandit Raid"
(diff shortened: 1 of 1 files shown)


Gyrandola:
Im merging this since the bandits are now facing a much stronger gate.

Gyrandola (approved):

PR #1554, opened 2024-11-12
Citywatch Pass 2.0

ProbablyCarl:
## About The Pull Request
Removes the crafting recipe for their armour. Adds two extra helmets and armours inside their armoury. Also includes a singular replacement stun mace. Folks argued that the armour, with its smash protection and integrated neck protection was too strong to be crafted. This rectifies that without removing a way for the Watch to actually get replacement suits. For the most part. Considering their armour breaks nonstop.
image

## Why It's Good For The Game
Their armour frequently breaks from fights. This allows them to get replacements, without allowing a blacksmith to arm the entire server in really good plating.


Nakalon:
No commented out code!!!!!!

Gyrandola:
dont comment, remove. we got history anyway.

Gyrandola (approved):

PR #1552, opened 2024-11-12
Goldface starts Locked

Nakalon:
## About The Pull Request

Now the Goldface starts locked instead of... unlocked?
Why the fuck did it need a key if it started unlocked! What?!

Also, the lock imprinter now admin logs, because a certain admin was trolled 😈 

image

## Why It's Good For The Game

Less Mercenaries breaking into merchant.
More admin logs. Less trolling.

PR #1555, opened 2024-11-12
Merchant key heat death

Nakalon:
## About The Pull Request

- Removed the 2 Shop keys and 1 Merchant key from the Merchant shop.

- Locked the door to that fucking thing I don't remember the name of that takes items and turns them into gold

This goes along with #1552

"Buh buh buh Desdemona, if theres no merchant how will we make use of this?!!??!! 😭 😭 😭 😭 😭 😭 😭 😭 "

YOU DON'T.
Do NOT. You do NOT. You DON'TN'T.

Shophand spwans with a shop key and can get a Merchant key from the Goldface if theres no merchant.
Merchant spawns with both keys.

If there are no Merchant or Shophand you don't make use of their services. Simple as.

## Why It's Good For The Game

The answer to "Why play the role if anyone can just break in and do it?"
Players should not be able to accomplish anything and everything by themselves.

### DIFF
--- a/_maps/map_files/rockhill/rockhill.dmm
+++ b/_maps/map_files/rockhill/rockhill.dmm
@@ -4284,7 +4284,6 @@
 "dDk" = (
 /obj/structure/closet/crate/roguecloset,
 /obj/item/reagent_containers/powder/spice,
-/obj/item/key/merchant,
 /turf/open/floor/carpet/stellar,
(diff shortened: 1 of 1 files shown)


Gyrandola:
when the store is closed i just break in

Gyrandola (approved):

PR #1558, opened 2024-11-12
Physician's Apprentice

ProbablyCarl:
## About The Pull Request
Another five second Carl addition, made at the request of a few folks.
We don't have sprites, so I'll let this sit for a bit, but it's fairly simple.

Adds an apprentice role for the court's Physician, permitting newer players to learn the medical system in a much more relaxed manner. Rather than being tossed into the deep end.

Spawns right behind him.
image

## Why It's Good For The Game
Players currently have no dedicated role to calmly learn the nuances of our medical system, for better or worse.

Do we want this? I 'unno. Just putting it here in case we do.


Nakalon:
I don't think role bloat is good. No reason why new players shouldn't play a role. I think already think smith apprentice is silly.

PR #1561, opened 2024-11-12
Remove bandit on-examine

Gyrandola:
## About The Pull Request

Title. (also removes a duplicated outlaw examine).

## Why It's Good For The Game

Since bandits only have one point of access. This might make it easier for them to hide their identity and encourage vanguards to check more closely who's passing through.

### DIFF
--- a/code/modules/mob/living/carbon/human/examine.dm
+++ b/code/modules/mob/living/carbon/human/examine.dm
@@ -128,13 +128,8 @@
 			. += "<span class='userdanger'>OUTLAW!</span>"
 		if(mind && mind.special_role)
 		else
-			if(mind && mind.special_role == "Bandit")
-				. += "<span class='userdanger'>BANDIT!</span>"
 			if(mind && mind.special_role == "Vampire Lord")
 				. += "<span class='userdanger'>A MONSTER!</span>"
-
-		if(name in GLOB.outlawed_players)
-			. += span_userdanger("OUTLAW!")
 
 
 		var/commie_text


RadiantFlash7:
It's an interesting idea, But. Now we have no way to tell who a bandit is at all? Theres no list anywhere or anything like that. Hell, Removing the outlaw brand as well means people can't know who an outlaw is. Ever.

Gyrandola:
> It's an interesting idea, But. Now we have no way to tell who a bandit is at all? Theres no list anywhere or anything like that. Hell, Removing the outlaw brand as well means people can't know who an outlaw is. Ever.

you can tell if someones a bandit because they act like it, instead of because they have a giant sign above their head.
You're mistaken on the outlaw brand, I removed a duplicate, scroll up in the code and you'll see the other one.

RadiantFlash7:
> > It's an interesting idea, But. Now we have no way to tell who a bandit is at all? Theres no list anywhere or anything like that. Hell, Removing the outlaw brand as well means people can't know who an outlaw is. Ever.
> 
> you can tell if someones a bandit because they act like it, instead of because they have a giant sign above their head. You're mistaken on the outlaw brand, I removed a duplicate, scroll up in the code and you'll see the other one.

I suppose thats fair. God knows everyones been executed for being a bandit alone in the past.

Elli-Skala:
This means only late join bog guards/bog guard on round with no bandits who set a nick name can check who is and isn't a bandit with the heart menu due to the menu showing you the bandit's job

HissyNays:
Then you need to make bandit posters state all names of current bandits. Otherwise it's an extremely huge buff to bandits that as a side effect removes necessity to act like a bandit. I can see no good will come from this.

Acwald:
Theres barely any antagonist left, and at most times bandits just get dogged on by the whole town gathering up on five people camping on their hideout. Besides,  bandits barely do something worth noticing. I think a good alternative would be bandit posters giving descriptions of bandits looks instead of names, so guards will have to match the descriptions and ask questions, etc. Just giving their name the same as the bandit label

awawaawawaawa:
> Theres barely any antagonist left, and at most times bandits just get dogged on by the whole town gathering up on five people camping on their hideout. Besides, bandits barely do something worth noticing. I think a good alternative would be bandit posters giving descriptions of bandits looks instead of names, so guards will have to match the descriptions and ask questions, etc. Just giving their name the same as the bandit label

That's a really solid idea.
Get some papers please action up in here. Fit the description. Check the posters.

2PlainGhoul:
Gyran Is has become quite clear that you DO NOT play the game

atomicplan666:
> > It's an interesting idea, But. Now we have no way to tell who a bandit is at all? Theres no list anywhere or anything like that. Hell, Removing the outlaw brand as well means people can't know who an outlaw is. Ever.
> 
> you can tell if someones a bandit because they act like it, instead of because they have a giant sign above their head. You're mistaken on the outlaw brand, I removed a duplicate, scroll up in the code and you'll see the other one.

so based

RyszardParys:
> I suppose thats fair. God knows everyones been executed for being a bandit alone in the past.

That's because bandits are supposed to be "Known highwaymen who are wanted for their past crimes." You were executed/put into a dungeon for these past crimes, it makes sense.

KarimXZ:
Would make the inquisitor and confessors a fair bit more useful too.

Nakalon (approved):
This thread is full of people with a lot of ideas but not the knowledge to see them come to fruition. Until such a time as those ideas become real this will have to suffice. Without this change, as there are currently no other ways to get inside the town but to pass the gate, bandit players just can't interact with the city at all.

This will be a band-aid until a better system can be worked on.

wow50000:
can we remove rebel huds ? maybe even the on examine too

ToothyDeer:
Despite this being merged I can still see BANDIT!!!! on people

PR #1568, opened 2024-11-13
Leg Sprite Fixes & Halfkin Colourable Sacks & Twistable Groin

EvenInDeathIStillServe:
Makes side leg sprites actually appear properly instead of looking like you're standing on one leg.
image

And makes it so halfkin can colour their testicles if they so wish.
image

Also you can twist groins if you grab them.


Gyrandola:
twisting groin stamcrit meta is back.

Gyrandola (approved):

PR #1574, opened 2024-11-13
More Artificer Misc Changes

Nakalon:
## About The Pull Request

- Repathed Cogs, Added Unstable Wooden Cog, renamed Advanced wooden cog to Reliable Wooden Cog

- Special cogs can now be used to stabilize or unstabilize a contraption

image
image
image

- Simplified Cart upgrade code

- Legendary Artificers can now explode themselves again
Now that they can stabilize the contraptions they don't need this free pass

- Solved Essence typo

## Why It's Good For The Game

While in game I noticed Artificer players never sold contraptions because of how they'd only serve them due to skill check.
This will hopefully solve that problem.

### DIFF
--- a/code/game/objects/structures/fluff.dm
+++ b/code/game/objects/structures/fluff.dm
@@ -545,7 +545,7 @@
 	var/broke = FALSE
 	var/datum/looping_sound/clockloop/soundloop
 	drag_slowdown = 3
-	metalizer_result = /obj/item/roguegear
+	metalizer_result = /obj/item/roguegear/bronze
 
 /obj/structure/fluff/clock/Initialize()
 	soundloop = new(list(src), FALSE)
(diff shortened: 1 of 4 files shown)


Nakalon:
I FORGOT SOMETHING HERE I THINK DONT MERGE

Nakalon:
Fixed, safe to merge now

ThatReformedEngineer:
Okay my only question is why would an artificer want to make their devices more unstable? Unless there is some mechanic that involves purposefully causing a mishap beyond inflicting danger.

Nakalon:
> Okay my only question is why would an artificer want to make their devices more unstable? Unless there is some mechanic that involves purposefully causing a mishap beyond inflicting danger.

Depends, evil intent probably. The misfiring of the contraptions also has its uses. Purposefully causing them to detonate would at some point be useful (specially if under duress)

ThatReformedEngineer:
> Depends, evil intent probably. The misfiring of the contraptions also has its uses. Purposefully causing them to detonate would at some point be useful (specially if under duress)

Yeah fair enough, it'd make portable smelter (bronze pipebombs) more reliably effective in causing a misfire detonation.

Gyrandola (approved):

PR #1576, opened 2024-11-13
Cutting Stump is no longer Ass

Nakalon:
## About The Pull Request

Now chopping a stump works the same way as chopping a lot:

image

## Why It's Good For The Game

As of now cutting a stump only delays the player for no real gain, which lead to a lot of stumps everywhere.
Players would rather have to push the cart with them and climb on top of the stump to store the logs in their cart and do this for every tree than to actually remove the stump.
Since players liked this woodcutting mechanic, I decided to expand it to stumps to solve the problems highlighted above.

### DIFF
--- a/code/game/objects/structures/roguetown/rogueflora.dm
+++ b/code/game/objects/structures/roguetown/rogueflora.dm
@@ -177,7 +177,7 @@
 /obj/structure/flora/roguetree/stump/attacked_by(obj/item/I, mob/living/user)
 	if(user.used_intent.blade_class == BCLASS_CHOP && lumber_amount)
 		var/skill_level = user.mind.get_skill_level(/datum/skill/labor/lumberjacking)
-		var/lumber_time = (40 - (skill_level * 5))
+		var/lumber_time = (120 - (skill_level * 15))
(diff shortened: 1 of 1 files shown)


2PlainGhoul:
Yeah, I never had this issue before -- this seems like a non-problem

Nakalon:
> Yeah, I never had this issue before -- this seems like a non-problem

oh... kay?

Hmeister-real:
yeah this isn't really an issue, stumps are just part of cutting a tree

Gyrandola:
Stumps are supposed to be not worth removing. Theyre stumps.

Nakalon:
> Stumps are supposed to be not worth removing. Theyre stumps.

Stumps cause you to have to climb on top of them to get the logs into the cart, every single time, and if you decide to remove them you get 1 stick and one small log, now you have to fish for that stick because you don't want it occupying space in your cart.

This takes away the stick element, makes it so you can just put everything in your cart at once and doesn't give you much more than it already did, except for the small chance of the small log it would otherwise spawn become an essence of lumber instead

Gyrandola (approved):

PR #1584, opened 2024-11-14
Tent cost increase

Heheniy:
## About The Pull Request
Increases resource cost of tent wall/door

## Why It's Good For The Game
Considering you neer only 1 lvl of crafting to make this and no climbing skill to climb, it is currently too cheap and often is being used for different cheases


Nakalon:
heres an idea: add skill level needed to climb?

awawaawawaawa:
> heres an idea: add skill level needed to climb?

You need a low skill level in climbing to climb tentwalls anyway. Its not zero like he's implying.

MickDick1:
Still. Climbing a tent wall shouldn't be possible in the first place.

Make it legendary. Problem solved.

Gyrandola:
Honestly good PRs because how the fuck do you climb a tent, let alone a medieval one

PR #1588, opened 2024-11-14
Re-arranges order of Heartfelt wave

RadiantFlash7:
## About The Pull Request
Re-arranges order of party members lost for the heartfelt wave.

## Why It's Good For The Game
The entire reason for my last PR allowing both men and woman as lord and lady of heartfelt, is that getting someone to PICK the lady of heartfelt is like pulling fucking teeth. Even if we have 5 people rolling for the heartfelt wave, No picks the lady, and then we're left with losing over half the roles from the wave because of it. 

Hopefully this adjustment will make heartfelt waves more likely to succeed in the lobby.


Nakalon:
I love seeing alternate solutions to problems!

Gyrandola (approved):

PR #1594, opened 2024-11-14
Fix for Smelter Bug

Nakalon:
## About The Pull Request

My previous PR caused a bug that basicly made steel impossible to make.

While that was extremely amusing to me and caused a very wicked event where people were scrapping for steel everywhere, its not what I intended.

This fixes it.

Made 6 steel this time:
image

## Why It's Good For The Game

I fucked.

### DIFF
--- a/code/modules/roguetown/roguejobs/blacksmith/smelter.dm
+++ b/code/modules/roguetown/roguejobs/blacksmith/smelter.dm
@@ -151,6 +151,7 @@
 								var/obj/item/R = new I.smeltresult(src)
 								ore += R
 								qdel(I)
+					maxore = initial(maxore)
 					playsound(src,'sound/misc/smelter_fin.ogg', 100, FALSE)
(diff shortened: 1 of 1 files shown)


ThatReformedEngineer:
You did indeed fuck. My guess is that whatever process you used to remove the extra item in the great furnace for smelting steel doesn't actually do it's job fully, and is leaving like some phantom item in it that the furnace thinks still remains.

Or maybe I am wrong, who knows.

Nakalon:
> You did indeed fuck. My guess is that whatever process you used to remove the extra item in the great furnace for smelting steel doesn't actually do it's job fully, and is leaving like some phantom item in it that the furnace thinks still remains.
> 
> Or maybe I am wrong, who knows.

What was happening was, in order to reduce the number of items it outputs I had to reduce the maximum amount of items the funance could take mid smelting. However, I did forget to revert the maximum amount of items back to its initial value, so it was stuck at 3!

PR #1595, opened 2024-11-15
Prevent lich from raising goblins

Gyrandola:
## About The Pull Request

Title. Does not address the root cause of goblin skeletons apparently being immortal but I don't want to look into it right now.


Nakalon (approved):
okay pookey

PR #1597, opened 2024-11-15
General bug fixes

john-pawson:
## About The Pull Request
This fixes some bugs i've found, since i didn't want to make 3 separate PRs for just bugfixing, they're all here on one.
Changelog:
- Heat treatment furnaces no longer crash the game when destroyed without an item inside! (Feel free to check it yourself, just spawn in a furnace, and then destroy it or delete it with admin commands)
Captura de pantalla 2024-11-15 001852
- Steel daggers now give steel when smelted instead of iron!
Captura de pantalla 2024-11-15 001923
Captura de pantalla 2024-11-15 001943
- Cooking skills and magic skills are now properly used with sleep exp instead of regular exp! (This made it so people didn't need to sleep and spend points to win the skills)
- Magic skills now cost the same amount to level as combat skills, considering that they're mainly used for combat. Added since now they actually use the sleep experience system.

## Why It's Good For The Game

Bug fixes! Also prevents magic users from spamming fireballs into trees until they get legendary magic skill. Same for cooking, but without the fireballs and the trees.

### DIFF
--- a/code/datums/skills/magic.dm
+++ b/code/datums/skills/magic.dm
@@ -2,6 +2,8 @@
 	abstract_type = /datum/skill/magic
 	name = "Magic"
 	desc = ""
+	dream_cost_base = 2
+	dream_cost_per_level = 1
 	randomable_dream_xp = FALSE
 
 /datum/skill/magic/holy


Gyrandola (approved):

code/modules/roguetown/roguejobs/blacksmith/treatingfurnace.dm
```diff
@@ -47,7 +47,8 @@
 		icon_state = "treatment0"
 
 /obj/structure/treating_furnace/Destroy()
-	inserted.forceMove(drop_location())
```
Cyprex, line 50: wtf, why would this have hard crashed the game- this should have just runtimed.

PR #1599, opened 2024-11-15
Map-Rockhill - Church revamp

NikolayByondPlayer:
## About The Pull Request

Revamped church to be smaller but still be functional and also have some new social places:

The whole church is separated into two areas: Public area and private church area. Public area is mostly on second and first floor. Church area takes half of first and second floor, also third and -1 floor.

1) Second floor now has a small healer's wing and library entrance. Altar area now has also a confessional room and two smaller altars. 
image

2) First floor now has public area for library and reading places, toilet. Area for church has meeting area, kitchen, storage room and a place for morticians. 
image

3) -1 floor has place for Inquisiton, cells and a secret entrance that can be locked-unlocked.
image

4) Third floor has 3 rooms for templars, room for priest, room for churchlings, room for acolytes
image

## Why It's Good For The Game

Can make place smaller, while still saving functionality.

### DIFF
--- a/_maps/map_files/rockhill/rockhill.dmm
+++ b/_maps/map_files/rockhill/rockhill.dmm
@@ -672,6 +672,12 @@
 	},
 /turf/open/floor/carpet/red,
 /area/rogue/indoors/town/church)
+"ayW" = (
+/obj/effect/decal/cobbleedge{
+	dir = 10
+	},
+/turf/open/floor/rogue/tile,
+/area/rogue/indoors/town/church)
 "azf" = (
 /obj/machinery/light/rogue/wallfire/candle/l,
 /turf/open/floor/rogue/hexstone,
@@ -998,7 +1004,7 @@
 "aLi" = (
 /obj/structure/fluff/sellsign{
 	desc = "We shall take care of them";
-	name = "Leave your wounded"
+	name = "Leave your wounded here"
 	},
 /turf/closed/wall/mineral/rogue/decostone,
 /area/rogue/indoors/town/church)
@@ -1401,9 +1407,12 @@
 /turf/open/floor/rogue/herringbone,
 /area/rogue/indoors/town/church/chapel)
 "bcq" = (
-/obj/structure/fluff/walldeco/psybanner,
-/turf/closed/wall/mineral/rogue/decostone/mossy/blue,
-/area/rogue/indoors/town/church/chapel)
+/obj/structure/fluff/railing/border{
+	dir = 6;
+	pixel_x = 4
+	},
+/turf/open/floor/rogue/blocks,
+/area/rogue/outdoors/town)
 "bcu" = (
 /obj/structure/feedinghole,
 /turf/open/floor/rogue/blocks/newstone/alt,
@@ -1814,6 +1823,13 @@
 	dir = 1
 	},
 /area/rogue/indoors/town/church/chapel)
+"brC" = (
+/obj/effect/decal/cobbleedge{
+	dir = 9;
+	pixel_x = 13
+	},
+/turf/open/floor/rogue/tile,
+/area/rogue/indoors/town/church)
 "brD" = (
 /obj/structure/mineral_door/wood{
 	locked = 1;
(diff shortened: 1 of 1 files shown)


NikolayByondPlayer:
Some screenshots:
image
image
image
image

AukariBunn:
This map makes some good points and has some good ideas, however, it has much more cons than pros.

Lets begin with one trouble common for the whole map:
The church looks too disolate. Are we speaking about the very same organization that can afford windows made of silver and gold? Even in the poorest societies churches tend to be shinning diamonds in the pit of shit and piss. Just look at the old european catholic churches that used to be in cities where people thrown their waste out of windows, or at the modern orthodox churches in Russia: golden domes and lots of decorations surrounded by buildings that havent been renovated for the last 50 years.
Its been said that the map is much smaller while also saving functionality. The functionality is indeed saved, but the efficiency is doomed. Now the church is a simulator of constant running from a place to place to do the most simple things. This is really bad layout.

The entrance
image
This might've been a fine idea, however, as its still a part of the streets in that variation, making the doors to the infirmary and stairs senseless. They arent a part of the church, they're outside the building. It breaks the building integrity. Imagine as if you lived in a flat and your contractor suddenly decided to place the entrance into your bathroom from the staircase.

The infirmary
image
The infirmary is too remote from the primary gathering place, which prevents the church from controlling the place. A lot of players will most likely die without treatment there because the acolytes wouldnt even know there's anybody there and inspecting it every 5 minutes would eventually annoy everyone. Also its appartness from the kitchen removes one of its main purpose: being a place of rest for the most shunned, which cannonically makes the church less significant.

image
The stairs are too hard to disassociate them from the rest of the room at the first glance, thats really bad for design. Changing the turf at the staircase from wood to stone would already make it much better.

image
No crosses are too remote from the altar. It could work, but the problem is that there are also just 3 tiles to approach each cross, while right now its 8. Not to mention that the people would stuck between the benches so eventually they wont be used at all and everyone would prefer just to stand or to sit at the front one. The rest wont just exist in players' minds.

image
The acolytes bedroom is too remote from the entrance. its not quite used because its too remote even now, congrats on deleting it from the game in practice. Its only good for ERP now, people wont walk 36 tiles and unlock 3 doors just to sleep as they rarely do it now while having to spend like 2 times less time to get there.

image
Churches really have balconies and they certainly dont do them like that. Also nobody places windows between bedrooms and halls.

image
The priest's room is too huge. Not to mention a window facing LITERAL ROCK. I do understand what decorative purpose it serves, but its not a wall decoration. Its still a window.

image
Two chandelier for one zone is too much while the other side doesnt have lighting at all. In reality they would place either two by the left and the right side, or just one right in the middle, but not two by the upper and lower sides.

image
The passages are too small, it would be hard for even two players to walk there without troubles.

image
The kitchen is too remote, terrible for game design. Reducing space is NOT good if you're making its impossible for players to actually use the place. You do realize we're doing it for the people to play and not for display? You must always think of how it will be used during the game before making anything.

image
This place is quite alright and will serve some roleplaying purpose,

image
Lots of meaningless stocks and the passage between the cells is too small. In certain cases it wont be enough even for one person.
Whats this even? And what are those bars doing there?
image

image
I have lots of questions about this tunnel. Its just a random cave without anything, it doesnt fit the rest of the building.

NikolayByondPlayer:
> AukariBunn

Hello! Thanks for feedback - gonna address some things: like turf on the stairs and probably chandeliers

About other feedback:
- Since Rockhill is a bit of shithole I wanted to make church be less grand and make it look like it has seen some disregards.
- We will probably add a door from church if it won't be used much or maybe move it - need to see it in game
- Sleeping places are made for function mostly, don't want people to hang out there for long
- Kitchen was made to serve church mostly
- LD part can override LA in terms of realism

NikolayByondPlayer:
Did some stuff suggested by Miss PT and Blutz
image
image

NikolayByondPlayer:
Changed statues to crosses, did small deko in some parts of church:

image
image

NikolayByondPlayer:
Fixed open space that leads to cells from confessional
image

Gyrandola (approved):

PR #1601, opened 2024-11-16
Adds a new tail and 2 new body markings, adds a tail back to lupian, tweaked vulpkian stats

Dedwolf:
## About The Pull Request

Adds the australian shepherd tail from another codebase (bubber)
Adds two new body markings made by myself, the wolf chest markings and the fox chest markings
Adds the fox2 tail back to lupians
Removes the eevee tail feom lupians
Tweaked vulpkian stats

## Why It's Good For The Game

More customization options is good.
Foxes being simply weak with better perception doesn't match their folk lore expectations, i've tweaked the stats so that they can better fit the tricky fox archetype

Default
image
New
image-1

Default
image
New
image-1


RadiantFlash7:
I'm not a fan of the vulp stat changes. They're the only ones who have good PER out of all the races. Leave vulps with 2 PER, IMO.

RadiantFlash7:
> > I'm not a fan of the vulp stat changes. They're the only ones who have good PER out of all the races. Leave vulps with 2 PER, IMO. If you want to adjust it otherwise, fiddle with the other stats.
> 
> so true, yeah no worries check out the new stats

works better!

code/modules/client/customizer/customizers/organ/tail.dm
```diff
@@ -27,7 +27,8 @@
 		/datum/sprite_accessory/tail/otie,
 		/datum/sprite_accessory/tail/shepherd,
 		/datum/sprite_accessory/tail/wolf,
-		/datum/sprite_accessory/tail/eevee,
```
Gyrandola, line 30: You did not mention removing this

Gyrandola (approved):

PR #1603, opened 2024-11-16
Mimic & Poltergeist

ProbablyCarl:
## About The Pull Request
Adds both of the creatures mentioned in the title.

Functionality:

Mimic pretends it's a normal chest, up until the point someone comes close to it. When activated, it'll unfold and reveal itself. Upon death, it'll drop whatever was placed into it via mapping, or whatever is on its tile when spawned. There exists a single mimic(by chance) on the map, and it's in the wizard's tower, replacing the coffin that had spice in it, as a 50/50 between a mimic or a chest spawn. Waiting on map stuff to be done before actually mapping these in proper.
image
image

Poltergeist is something I pulled near 1:1 off of what I did years ago for Stalker 13. It's supposed to be a floating skull, like a D&D demilich, but for now we're using a temporary sprite. It'll appear like it did in Stalker 13 for the moment, as a ball of electricity.
The unfinished sprite is in the code, but, of course, is unused and marked as 'new' for now.
For its functionality, it floats around, makes weird noises and throws whatever it can see at players. Like a violent haunting, generally causing a mess. It's near unhittable, though with incredibly low health. Use Churn Undead to get rid of it, or swing when it becomes temporarily visible for a chance to do something.
There also exists an event for a single time per round, during an omen as a 10% chance, that it'll spawn a handful of them around the map.
image
image

## Why It's Good For The Game
I thought these were neat enough to add. That's about it. Mimics serve as a way to punish people rushing loot spawns, for better or worse. Poltergeists are unique fluff mobs that can spook people.

### DIFF
--- a/roguetown.dme
+++ b/roguetown.dme
@@ -2107,6 +2107,7 @@
 #include "code\modules\events\rogue\goblininvasion.dm"
 #include "code\modules\events\rogue\haunts.dm"
 #include "code\modules\events\rogue\lightsout.dm"
+#include "code\modules\events\rogue\poltergeist.dm"
 #include "code\modules\events\rogue\skellyinvade.dm"
 #include "code\modules\events\rogue\skellysiege.dm"
 #include "code\modules\events\rogue\vines.dm"


Gyrandola:
Id like to see a clip of each mob

Nakalon:
Absolutely I like new mobs but we have to see test proof

Gyrandola (approved):

PR #1604, opened 2024-11-16
Roundend mugshot credits

Gyrandola:
## About The Pull Request

- Ports roundend credits from https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/357.

image


Hmeister-real:
The text needs a background because sometimes (like with dark purple or yellow) they are really hard to read. But past that I really like it

Crowbar764:
Maybe make it so people with no mugshot get skipped.

Nakalon:
Gyran we should really use this technology for the bandits!!!

Nakalon (approved):
it just works

PR #1606, opened 2024-11-16
Bandits at it again - Hordemaster moved in

Rayforx:
## About The Pull Request
Ports bandit hordemaster from [ BANDIT DLC PT 2 - PRAYING FOR NO MAP CONFLICTS EDITION #659 ](https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/659)
image

image

And a few sacks were added in
image

The balance was slightly changed, so the bandits have to work for their triumph instead of the original PR starting with 150 > was changed to 10

The PR will most likely be completely whack in terms of balance, but as this was mostly a straight port. Then I choose not to touch much of it so that I can hear feedback in a TM.

## Why It's Good For The Game
Remove the awful items being dropped that are boring and annoying to work with and let the bandits choose from a list of items for what they want to buy, from a general list and class level.

### DIFF
--- a/code/modules/antagonists/roguetown/villain/bandit.dm
+++ b/code/modules/antagonists/roguetown/villain/bandit.dm
@@ -11,7 +11,7 @@
 		"I WILL NOT LIVE IN YOUR WALLS!",
 		"I WILL NOT FOLLOW YOUR RULES!",
 	)
-	var/favor = 25
+	var/favor = 50
 	var/totaldonated = 0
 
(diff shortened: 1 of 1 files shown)


Koteshkin:
That dragon from a map looks SO DAMN AWESOME.

Rayforx:
> Good system, but the shared points and buying from other people's open menus should be fixed ASAP.

I'm not one hundred percent sure how the bug of opening other people's menus are being done.

But I was going to increase the price of all the goods, so it's triple of what it was before.

Gyrandola (approved):

PR #1614, opened 2024-11-17
Fixes the bronze spear having no icon while wielded

john-pawson:
## About The Pull Request

Title! Bronze spear has no sprite while wielded, someone should probably make one, but that isn't me, as i am no spriter.
I already tested it but i forgot to send a screenshot. It works, trust me bro.

## Why It's Good For The Game
I think it's good to be able to see the spear while you're wielding it. Bug fix!


Nakalon (approved):
Ah this makes sense

PR #1615, opened 2024-11-18
Rogue nerf

wow50000:
## About The Pull Request
Nerfs rogues, rogues have been too hyper tuned compared to every other class
List of rogue changes:
more speed
speed buffs
dodge expert
more skills

removes the bonus speed removes some end removes bonus skills

## Why It's Good For The Game
heres some footage of me playing humen male warrior against a tabaxi rogue

https://github.com/user-attachments/assets/a5279dc7-c22c-4fac-8348-f4092de81d9b

notice the stupidly high chance of them bypassing parries, as well as stupidly low chances to hit them as well as taking forever for them to be stammed out, rogues can also disengage at no cost of stamina as well as grapple pretty good
even with the best neck protection in the game rogues can just bypass it and have a very good chance of bypassing the kite shield and the best neck protection in the game

with roundstart gear, no class has any chance of beating a robust rogue main. Off recording we did about 6-7 fights without full steel gear and using the spawn shield and the rogue beat me every single time.

"just grab them"
they can dodge and stab me
"just swift intent stab them"
they can move too fast to hit and do that back
"just parry"
they bypass parries and armour
"just charge them"
charges are easy to sidestep at such high speeds

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/combat/rogue.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/combat/rogue.dm
@@ -51,6 +51,7 @@
 			beltr = /obj/item/rogueweapon/huntingknife/idagger
 			beltl = /obj/item/rogueweapon/huntingknife/idagger
 			H.change_stat("strength", -1)
+			H.change_stat("constitution", -1)
 			H.change_stat("perception", 2)
 			H.change_stat("speed", 3)
 			H.change_stat("endurance", -1)
@@ -86,6 +87,7 @@
 			beltr = /obj/item/rogueweapon/sword/iron
 			beltl = /obj/item/storage/belt/rogue/pouch/coins/poor	
 			H.change_stat("endurance", -1)
+			H.change_stat("constitution", -1)
 			H.change_stat("speed", 2)
 			H.cmode_music = 'sound/music/combat_duelist.ogg'
 	ADD_TRAIT(H, TRAIT_MEDIUMARMOR, TRAIT_GENERIC)


Gyrandola:
expert swords on an adventurer class is quite crazy

Gyrandola (approved):

john-pawson:
Well i agreed with the first commit there's no reason for swashbucklers to have expert swords for some reason, and the rogues having the best athletics didn't make sense (Probably added since dodging costs double as much stamina as parrying)

But you over did it on the second and third commits.

Swashbucklers: -1 END, -1 CON and +2 SPD? Really? You made the fighting or dueling class that doesn't even have dodging as an option SPD their only good stat? They can't even dodge that much with the shitty END, and you've effectively made them *the* weakest class ever for about no fucking reason.

Thieves: -1 CON -1 END and -1 STR for 2 PER and 3 SPD. It's just.. the exact same thing. Not only that but getting a single grab on any of these two means an instant win since you've entirely removed their wrestling skill. I agree that it should've been lowered to something like apprentice since they're not exactly a wrestling class, but this just makes them WEAKER THAN FUCKING MAGES ON COMBAT.

Besides, the armor penetration is a problem of the dagger, and it's surprisingly easy to avoid since it has a full 1.2 second delay to the attack giving you plenty of time to move away.

Hmeister-real:
this is good. but hoooolllly shiiit that video makes you look so unrobust good god

atomicplan666:
rogues are dogshit now and watchmen are OP. very nice.

john-pawson:
> Swashbucklers are supposed to be duelists, why wouldn't they have expert swordsmanship? Sword skill and being light on their feet are kind of the only things they have going for them.

Balance reasons since adventurers are supposed to have at MAX journeyman (Exceptions are of course for like monks with wrestling and rangers with bows), but their high SPD with swift weapons should be plenty to have an advantage on dueling against other people.

PR #1636, opened 2024-11-22
The Three Musketeers

Slorbin:
## About The Pull Request

Count them. Three! With additional special text if all three can't roll at once.

They get pretty good musket skill. They only spawn as a migrant wave, and only one can spawn! They get far less other skills overall than other combat classes, then.

oh also swashbucklers get their rapier back bc they were balanced specifically around having the rapier.

## Why It's Good For The Game

more guns on funny guys

### DIFF
--- a/code/datums/migrants/migrant_waves/musketeer roles.dm
+++ b/code/datums/migrants/migrant_waves/musketeer roles.dm
@@ -1,6 +1,6 @@
 /datum/migrant_role/musketeer/musketeer
 	name = "Musketeer"
-	greet_text = "A jolly adventurer armed with a musket and a rapier, seeking to save the girl and the day! Don't lose sight of your two companions, you're inseperable!"
+	greet_text = "A merry musketman seeking to save the girl and the day! Don't lose sight of your two companions, you're inseperable!"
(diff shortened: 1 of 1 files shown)


DeerJesus:
also, fun little fact, an arquebus and a musket are the same thing. synonyms

code/datums/migrants/migrant_waves/musketeer roles.dm
```diff
@@ -0,0 +1,44 @@
+/datum/migrant_role/musketeer/musketeer
+	name = "Musketeer"
+	greet_text = "A jolly adventurer armed with a musket and a rapier, seeking to save the girl and the day! Don't lose sight of your two companions, you're inseperable!"
```
Gyrandola, line 3: Please something more fitting than "jolly adventurer"

DeerJesus: Like what? They're jolly adventurers

Frienderanian:
I would just make it so that you need three to spawn. I get what you're going for- but it isn't hard to get three people. Heartfelts spawn with three.

Gyrandola (approved):

PR #1638, opened 2024-11-22
Zizoid Beast Invulnerability | End Message Touchup

ProbablyCarl:
## About The Pull Request
We're not meant to be able to kill the Zozoid ascension creature, but for some reason, it's a big ol' joke. This makes it immune to damage, as had been intended with its health set so damn high. Unless I'm mistaken? I'm pretty tired as I write this. Ascension is meant to be a big deal, in any case. This PR is probably overkill, but whatever.

As an aside, it also touches up the end round message to be more clear. IMO, at least.

## Why It's Good For The Game
Cult ascension was more of an inconvenience, with the creature killed easily. The mood debuff was a bigger issue, by comparison. This should rectify that by throwing it well into the "OH GOD OH FUCK" realm of being broken, as was the intent originally, unless I'm mistaken.

The message was also not clear. This fluffs it up a bit and straight up states that it's the end in fifteen minutes, not just 'fifteen minutes remain'.

### DIFF
--- a/code/datums/gods/patrons/curses.dm
+++ b/code/datums/gods/patrons/curses.dm
@@ -12,10 +12,10 @@
 	var/trait
 
 /datum/curse/proc/on_life()
-	return 
+	return
 
 /datum/curse/proc/on_death()
-	return 
+	return
 
 /datum/curse/proc/on_gain(mob/living/carbon/human/owner, silent)
 	ADD_TRAIT(owner, trait, TRAIT_CURSE)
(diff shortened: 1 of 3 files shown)


Xaytan:
IMO, giving it some regeneration would be a better idea. Still technically possible to kill it, but allows it to retreat and heal if it's in danger.

Gyrandola:
> IMO, giving it some regeneration would be a better idea. Still technically possible to kill it, but allows it to retreat and heal if it's in danger.

The round ends after the ascension anyhow (or it should at least)

Gyrandola (approved):

HissyNays:
"Unless I'm mistaken?"
You are mistaken, it's meant to be /near/ unkillable and it was, having a chance to kill it is fun. Now no one will fight it ever and run, and to run is quite easy.
Fun draining PR

PR #1639, opened 2024-11-22
Druid Buff | Scomm Ring Icon

ProbablyCarl:
## About The Pull Request
Provides druids a hefty buff.

Adds them being friendly with the following:
 - Spiders
 - Rats
 - Volves
 - Moles

They're also provided a walking stick by default, with randomized skill in polearms between 1 or 2.

Not related to this PR's main focus, but snuck in, is scomm rings getting a temporary sprite by way of using the scommstone icon. It doesn't have its two special states set, however, and is one icon only. It's not proper at all, but it's something to look at that makes SOME sense, as opposed to it being invisible.

## Why It's Good For The Game
Druids are kind of a forgotten role. I 'unno. This was at the request of a few people, with suggestions from players.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/church/druid.dm
+++ b/code/modules/jobs/job_types/roguetown/church/druid.dm
@@ -26,7 +26,6 @@
 /datum/outfit/job/roguetown/druid/pre_equip(mob/living/carbon/human/H)
 	..()
 	H.faction += list("rats", "spiders", "wolfs")
-	H.ambushable = FALSE
 	belt = /obj/item/storage/belt/rogue/leather/rope
 	beltl = /obj/item/storage/belt/rogue/pouch/coins/poor
 	beltr = /obj/item/flashlight/flare/torch/lantern
@@ -59,7 +58,3 @@
 	var/datum/devotion/C = new /datum/devotion(H, H.patron)
 	C.grant_spells(H)
 	H.verbs += list(/mob/living/carbon/human/proc/devotionreport, /mob/living/carbon/human/proc/clericpray)
-
-	if(!H.has_language(/datum/language/beast))
-		H.grant_language(/datum/language/beast)
-		to_chat(H, span_info("I can speak the primal language with ,b before my speech."))


awawaawawaawa:
> ### 2) Why should this feature be locked to druids? Why not make it modular by setting it as a trait?
> 
> This change would make it much easier to work with for gamemasters, and also paves the way for druids who break tenets of dendor (i.e by abusing their immunity to slay mobs that can't fight back) to have their trait removed. Furthermore, it would allow for the mechanic to be tutorialized naturally, as players tend to check their traits on spawn, they would easily see some immersive message like:
> 
> _"**Dendor's Watch** - I am a companion of Dendor, patron of the wild. Woodland creatures pose no threat to me."_
> 
> Once you saw the text, you would immediately understand you are safe from beasts. Being able to roleplay around the boons your patron has granted you is essential, so this seems like the greatest benefit to the average player if you set it as a trait. (You could achieve a similar thing by simply printing some text in chat when you spawn as druid, but it would often be skipped over as most players do not check that text for changes.)
> 
> Another benefit is that in future a mechanic could even be implemented so that upon violating Dendor's will (butchering a mob for an example that's easy to code) you lost the trait, to promote druidic roleplay where nature is revered rather than exploited. This would not be in the scope of the current PR, but by making the feature a trait you would pave the way for anyone else to add it in future!


A lot of druids like to hunt, I among them. 'Dendor provides' being a common saying among Dendorites as a whole. Hunting is a part of that ethos. I really don't like the idea of Druids being penalized for roaming wild, and hunting prey. Not a fan of that particular bit.

Elli-Skala:
I imagine moving druids away from fist fighting and wrestling (like every other church caster) and more to pole arms expertise could be interesting. We have the druidic trickery as a crafting option if we wanted to open it more via allowing druids to craft higher teir dendor blessed wooden staffs

Gyrandola (changes requested):
Druids being friendly with wolves etc is OK I suppose, however:

Druids being not ambushable makes them almost unable to hunt (they aren't necessarily hippies), and also makes them not ambushable by goblins, which most definitely are not creatures of Dendor.

It might be a bit aids to address the aforementioned request, so feel free to atomize the rest of the changes because they are good.

Gyrandola (approved):

PR #1646, opened 2024-11-23
Nerfs clerics

john-pawson:
## About The Pull Request

Currently forge clerics are a little bit of a jack of all trades, having the ability to smith while still being able to do combat very easily, this nerfs the forge clerics making them less of a war cleric with smithing, and more of a smith with combat capabilities.
Also removes expert maces from war clerics since adventurers aren't supposed to have more than journeyman on weapon skills.(Excluding things like unarmed and wrestling for monks, or archery and crossbows for rangers, as those are way less valuable than things like swords or polearms)

By the way, the endurance was changed to constitution since forge clerics already get a spell that's literally a way of regaining stamina without having to rest.

## Why It's Good For The Game

Less people going around into the bog and being able to fight efficiently *and* give themselves plate armor through smithing, while not interacting with anyone else on the round.
Also makes clerics not have expert for no reason at all.


john-pawson:
> One comment I can make, is my suggestion to go the paladin route. Give cleric expert maces if the character is old.

I don't think clerics were ever supposed to be this proficient at weapons either way, since they're more of a holy class than a battle class (like the paladin) it'd probably make sense to give them a higher holy skill when old.

PR #1649, opened 2024-11-23
Nerf's poison arrows

RadiantFlash7:
## About The Pull Request
This PR halves Toxin damage from Poison arrows. Please note: Poison arrows would /still/ be viable after doing so. You would only need to land 2 to crit someone, at which point they're either screwed and in critical condition on the floor, or shortly will be. 
Further more, at 50 toxins that 1 arrow delivers, you're still probably going to make the target puke their guts out, stunning them that way.

## Why It's Good For The Game
I once more, come to you regarding poison arrows. I can not stress this enough, This isn't a salt PR.
This is me making the PR /after/ I trounced the entire bandit troup in a 1v6 as a vanguard archer.
Similarly, a few rounds later I did the exact same thing as a bandit archer with poison arrows in a 1v4 against keep garrison members
This is /genuinely/ broken. And I feel disgusted for using it. But theres also no reason /not/ to use it. It's an 'I win' tactic that take 20 minutes of prep at most.

The current ways of countering poison is- Magically having a leach on you at all times, and using it on yourself, having a physician/sawbones on standby that can purge toxins, and I believe the expensive as shit antitoxin medicine Mori added. Unfortunately, The most common method (leeches) won't actually save you. A single leech will pull you out of crit after recieving a single poisoned arrow, but it'll die and fall off while your still coughing and puking up blood every few seconds, pretty much still full of toxins. At this point, I doubt a regular leech will actually save someone if they're hit with 2 poison arrows.

### DIFF
--- a/code/game/objects/items/rogueweapons/ranged/ammo.dm
+++ b/code/game/objects/items/rogueweapons/ranged/ammo.dm
@@ -134,7 +134,7 @@
 	if(istype(target, /mob/living/simple_animal)) //On-hit for carbon mobs has been moved to projectile act in living_defense.dm, to ensure poison is not applied if armor prevents damage.
 		var/mob/living/simple_animal/M = target
 		M.show_message(span_danger("You feel an intense burning sensation spreading swiftly from the puncture!")) //In case a player is in control of the mob.
-		addtimer(CALLBACK(M, TYPE_PROC_REF(/mob/living, adjustToxLoss), 100), 10 SECONDS)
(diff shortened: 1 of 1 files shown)


ThatReformedEngineer:
There is the secret fourth option of mixing charcoal by crushing up some ash and salt, then pouring the mixture into a bottle or some kind of container when it comes to dealing with poison arrows. But no one ever bothers with it because you're more liable to die of a stabbing than poison.

Gyrandola (approved):

atomicplan666:
Good PR, a singular arrow shouldn't kill, as said before it's bad sauce

PR #1650, opened 2024-11-23
Hordemaster Foresworn Bevor Fix

BlackjackIsTired:
## About The Pull Request

Fixes Hordemaster Bevor buying for Foresworn and adds the Hound Mask as an additional purchasable option.

## Why It's Good For The Game

Fixes the issue where Foresworn would get scammed for hundreds of favor because a single typo was made in the code(Classic).

As an example of what the issue was for those unaware, Foresworn were able to purchase a Bevor for on avg a 100 favor. As a result of the typo in the spawn code, you would spend the favor, it would be removed from your balance and you would get nothing. (Then the Hordemaster would go "Thank you~", ruining your day because that greedy fat fucker just stole your shit.)

Also added the Steel Hound Mask as a purchasable option for everyone because style is important. Same cost as the regular Steel Mask.

### DIFF
--- a/code/modules/cargo/packsrogue/armor.dm
+++ b/code/modules/cargo/packsrogue/armor.dm
@@ -41,7 +41,7 @@
 	cost = 50
 	contains = list(
 					/obj/item/clothing/mask/rogue/facemask/steel/hound,
-					/obj/item/clothing/mask/rogue/facemask/steel/hound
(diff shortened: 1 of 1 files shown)


Gyrandola (approved):

BlackjackIsTired:
I forgot to push something, but lets pretend I didn't and it works perfectly.

PR #1655, opened 2024-11-24
Round End Timer

EvenInDeathIStillServe:
when the vote for round end happens it tells u when the round will end. amazing. spectacular. wow !

image


HissyNays:
Honestly, it's a good little detail, it allows me to know exactly when to shoot duke in the face before round ends.

PR #1664, opened 2024-11-25
Keep That Thang Loaded

ProbablyCarl:
## About The Pull Request
Gives the Priest a special specifically for spawning with dual runelocks. "Runic Faith".
Unloaded, of course.
Alongside firearms skill rather high.
Also moves some of their items around to fit it, so nothing needs to be deleted. They'll need to mug the Inquisitor for pistol shot or go to the Artificer.
## Why It's Good For The Game
Is this good? NO. It's funny. That's why I propose it.


Coolgat3:
GEM PR, just imagine her descending from second floor onto the altar then starting to blast filthy matthiosians

ChocoboLord:
This will go so hard with the clergy combat track

Gyrandola:
Halve the weight and reduce the skill level to 4 or 5 please.

Gyrandola (approved):

PR #1666, opened 2024-11-25
Fixes the mote lantern and magemote.

Kevak:
## About The Pull Request

This fixes the sprite null for the more lantern and magemote. I'd like to request further caution be given to these massive cleanup pr's because there are almost certainly dozens of obj sprites that have been orphaned by these PR's just like this one has. Please make bug reports about nullsprite objects you see so they can be fixed! The fix is using the replacement sprite that cleanup pr replaced the mote with, honestly I like how it looks better than the old one! 

## Why It's Good For The Game

Fixing a nullsprite bug is always good for the game.

## Proof Of Testing

https://github.com/user-attachments/assets/91b7279c-ca11-4182-911e-1cd0f6398176

### DIFF
--- a/code/modules/mining/lavaland/necropolis_chests.dm
+++ b/code/modules/mining/lavaland/necropolis_chests.dm
@@ -156,8 +156,8 @@
 /obj/effect/wisp
 	name = "friendly wisp"
 	desc = "Happy to light your way."
-	icon = 'icons/obj/lighting.dmi'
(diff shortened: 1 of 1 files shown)


Kevak:
Yeah, that's what happened in this case

Gyrandola:
Thanks a lot

Gyrandola (approved):

PR #1667, opened 2024-11-25
Goblin Touchups | Goblin Shaman

ProbablyCarl:
## About The Pull Request
New:
 - Goblin Shaman introduced. Acts as a healer for Goblins. Locked into having four spells by default. Prestidigitation, Fetch, Lesser Heal and Diagnose. They're a mixture of clergy and physicians, for what it's worth, even spawning with the surgical kits.
 - All Goblin roles are given Orcish, if they don't already possess the language, as a fallback for if staff wants to make non-goblins a Goblin role. As dumb as that might sound.
 - Goblin Chief is now unlocked to Half Orcs by default.
 - Goblin 'Boltslinger' introduced as a Guard class. Just a crossbowmen of low skill, on par with the other classes.
 - Goblins are now in the same faction as NPC goblins, making them allied.

Changes:
 - You're not required to be a Graggarite pre-spawn to play as Goblins. Instead, it's forced after spawn onto your character, changing whatever religion you had.
 - Goblin travel tiles on the interior can be used by anyone, allowing escape. The outside tiles are still locked to Goblin trait characters exclusively.
 - Goblins are given 'goblin eyes', which, in theory, should allow night vision, but they're fucky and don't work well at all in this purpose. If at all.
 - Mineral drops around the Goblin's camp has been significantly improved, with smatterings of medium and high yield rock all over the place.
 - Goblin Smithery outfitted by default with basic equipment. Two furnaces, an anvil, a forge and a barrel of water.
 - Stone around the encampment modified to hug the walls, so there's no dead space for mining.
 - Rabble moved into the encampment by default.
 - Significantly ups the stats of the Chief, so he can hold his own in a fight.

## Why It's Good For The Game
Goblins need a lot of QoL stuff. I'm not too happy with how they are currently, and this rectifies a number of issues, while also introducing some new content.


NikolayByondPlayer:
Can you maybe add one small ringing bell to entrance? So waiting gobbos can ring it

Also do we need a gate if only goblins can pass?

Faire-Fuzzy:
I like these changes, not being able to do anything when your captured called the Fog quickly as it gets dull fast sometimes. If their entrance could be moved to somewhere just out of town too that would be great I think, Where they are their a little repetitive and a nuisance no can deal with short of walling up their space then their trapped and just as bored.

Gyrandola (approved):

CrushKidClyde:
I have three suggestions for Gobbos as a dedicated Warg enthusiast.

1.) Gobbo should have a few paths to enter and exit from. 
     Using existing dilapidated buildings of the old expanded areas of the town now abandoned to the rot the Gobbos could have entry of their own into the Bog and a few hidden paths for egress. 

2.) I promote a new Roll for Gobbos that is kind of silly but I think has its merits: the Warg.
    Using Volf and tuning it to only speak Goblin but understand Common and Goblin. This would go with some Tolkien logic but would lend to the mechanics reinforcing gobbo focused teamwork.
    Changing the Volf to be Rideable by Gobbos would not lend to a fast cavalry but it would create a lethal threat. Currently Gobbos Ride the noble Saiga, theme could be due for the removal of the horse for Warg mounts.
    To clarify the idea would to have Volf be left alone and a duplicate mob be created with a rename and a couple changed, Rideable, & understand common & Gobbo / and speak only Gobbo.  
   
3.) The Magical loot Gobbos have maybe there is a broken/stolen teleporter mirror that allows for the Gobbo Portals? Like we see admins spawn at events to spawn in gobbo hordes?  Could be used for alot of shenanigans.

PR #1671, opened 2024-11-25
Cure rot Fix+sawbone treatment to 4

RadiantFlash7:
## About The Pull Request
Fixes not being able to cure rot as a pestran or being unable to debride it as a physician on both player deadites, and rotted corpses.
Also reduces sawbones knife skill to 3 and gives them level 4 treatment (So they are able to revive)

## Why It's Good For The Game
Cure rots been broken for like a week due to rotted players no longer getting the zombie antag status. I removed a check that was causing cure rot to fail. Simple as.
Also apparently sawbones can't revive people without level 4 treatment, which is half the point of the role. So.... yeah. That needed fixing.


Gyrandola:
Real?

RadiantFlash7:
> Real?

Real.

Gyrandola (approved):

PR #1672, opened 2024-11-25
Cleanup 4

Gyrandola:
## About The Pull Request

- Ports https://github.com/Vanderlin-Tales-Of-Wine/Vanderlin/pull/16.
- Ports https://github.com/Vanderlin-Tales-Of-Wine/Vanderlin/pull/17.


Kevak:
Why are you deleting defines and helpers?? Never delete those, they have so many critical dependencies tied to them. Have you been checking how many runtimes these cleanup PR's have been throwing? Just because it compiles does not mean it works.

Kevak:
Additionally nuking the power system wholecloth will make it harder for coders and mappers to do anything with the planned proper citymap's powernet.

Gyrandola:
Ill get back to this later

dwasint:
> Why are you deleting defines and helpers?? Never delete those, they have so many critical dependencies tied to them. Have you been checking how many runtimes these cleanup PR's have been throwing? Just because it compiles does not mean it works.

All keeping unused defines and helpers does is lead future coders into unknowingly breaking stuff because they thought adding a helper that doesn't work worked.


> Additionally nuking the power system wholecloth will make it harder for coders and mappers to do anything with the planned proper citymap's powernet.

Unless you are planning on doing large scale power grids I would recommend redoing it, the powernet code is bad as is, and if its not being heavily used can be simplified alot

Nakalon (approved):
What a champ

PR #1675, opened 2024-11-26
Removal of loveless & Pacifism Specials

ProbablyCarl:
## About The Pull Request
Pacifism was cancerous and genuinely made certain elements of the game, even outside of combat, actually unplayable. Like, genuinely unable to interact with stuff. It's better restricted to pacifism bans, as with Loveless, which is to follow.

Loveless doesn't make sense as a special, because it locks you out of the following:
 - Receiving / Giving Hugs (Causes a 'freak out', atop that.)
 - Receiving / Giving Kisses (Causes a 'freak out', atop that.)
 - Literally anything involving mechanical ERP.

## Why It's Good For The Game

Pacifism doesn't even need to be argued, does it?

As for Loveless, it was intended as a curse, not a a player special. Rolling this can ruin some folks' rounds, for better or worse. As funny as it is. It also makes certain roles requiring mechanics completely impossible to play, such as Bath Wench.

### DIFF
--- a/code/datums/special_traits/traits/traits.dm
+++ b/code/datums/special_traits/traits/traits.dm
@@ -425,7 +425,6 @@
 /datum/special_trait/unlucky/on_apply(mob/living/carbon/human/character, silent)
 	character.STALUC -= rand(1, 10)
 
-
 /datum/special_trait/jesterphobia
 	name = "Jesterphobic"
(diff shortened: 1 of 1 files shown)


NikolayByondPlayer:
But shouldn't specials be also dangerous?
So they would give you great benefits or disadvantages. So if you roll you know you have risk of having something that is not too great.

Gyrandola (approved):

PR #1678, opened 2024-11-26
Gives Noc A Patron Boon - Nerfed a solid bit. Gib feedback in PR pls.

Kevak:
## About The Pull Request

This gives Noc a patron boon. Specifically he provides darkvision equivalent to the spell. Additionally, this makes the darkvision spell provide slightly better darkvision to those whom already have darkvision or noc's boon. It feels thematic for a _moon god_ of _knowledge_ to enable its followers and knowledgeseekers to see in dark places. After all- **Noc Sees All.** Note: This does not touch werewolf or undead nightvision, those are beyond Noc's purview. Werewolves are Dendorite- Noc simply reveals the beast for all to see. Whereas undead are Zizo's creatures.

I intend to give all the patrons (excluding atheism) whom do not presently possess patron boons, small patron boons/blessings like this or the other various patrons whom currently have effects. I have a few ideas but I'd love suggestions.

## Why It's Good For The Game

Patron gods should all give a little buff. It's weird that a handful do but most don't. I intend to amend that.

## Proof Of Testing

https://github.com/user-attachments/assets/aa63f979-966a-4c9a-98ca-d67d4dde1f32

### DIFF
--- a/code/__DEFINES/lighting.dm
+++ b/code/__DEFINES/lighting.dm
@@ -80,10 +80,11 @@
 #define LIGHT_RANGE_FIRE		3 //How many tiles standard fires glow.
 
 #define LIGHTING_PLANE_ALPHA_VISIBLE 255
+#define LIGHTING_PLANE_ALPHA_NOC 245
 #define LIGHTING_PLANE_ALPHA_LESSER_NV_TRAIT 243
 #define LIGHTING_PLANE_ALPHA_NV_TRAIT 235
 #define LIGHTING_PLANE_ALPHA_DARKVISION 220
-#define LIGHTING_PLANE_ALPHA_DARKVISION_BETTER 195
+#define LIGHTING_PLANE_ALPHA_DARKVISION_BETTER 200
 #define LIGHTING_PLANE_ALPHA_MOSTLY_VISIBLE 192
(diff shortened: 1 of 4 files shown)


RadiantFlash7:
I would recommend keeping this PR open, and giving all the patrons stuff in the same PR, rather than one at a time. Better coding practice; and it's something that makes sense to put in all ages once rather then piecemeal

Kevak:
> I would recommend keeping this PR open, and giving all the patrons stuff in the same PR, rather than one at a time. Better coding practice; and it's something that makes sense to put in all at once rather then piecemeal

That requires having ideas for all of them. But yes it would be better practice! Though admittedly I would prefer to do it on a per-deity basis for better feedback.

john-pawson:
Alright i'll have to agree it seems a little bit too powerful, you should probably limit it to 2-3 tiles of range for vision or a *very* slight buff, don't want people to start picking patrons **ONLY** because of the advantages.

Kevak:
Mm good point. I’ll retune the efficacy of the darkvision once I roll out of bed. Methinks bumping the alpha up like.. 15-20 some points should make it much harder to see while still providing something tangible and useful if you squint.

Kevak:
Think I'm vibing with tweaking the alpha up to 245 for the nocboon. Just baaarely enough to see and navigate with but you gotta squint a bit. The feedback has been appreciated!

Screenshot 2024-11-27 000016

Issue #1512, opened 2024-11-09
[REQUEST] blindfolds


wow50000:
## Is your feature request related to a problem? Please describe. 
nah

## Describe the solution you'd like
putting cloth in the mask slot acts as a blindfold

## Describe alternatives you've considered
nah

Eriripng:
working on this, but consider yourself safe now, my child

PR #1681, opened 2024-11-26
Add blindfold to the game

polvohime:
## About The Pull Request

Per request of [this](https://github.com/Rotwood-Vale/Ratwood-Keep/issues/1512) issue I took the liberty to add to the server the item "blindfold" which can be used to, as you may guess, blind someone when wearing it. I am not very well-versed into DM, but I found this easy enough to add and I liked the idea too.

## Why It's Good For The Game

This PR adds a new item - and taking into consideration the events/scenes that happen in the game -which I believe will enhance player experience. The blindfold will make a players screen go entirely black until it is removed, and it can easily be tied up together with one piece of cloth from the craft menu.

## Evidences
Screenshot_2
Screenshot_3
4a51702c60fb4390a787e0016bcc7d75

#### Note
It would be cool to have a proper sprite to it, though it escapes my .......... skills to actually sprite it. If anyone is willing to make _at least_ an UI icon for it, I think that would be very neat.

### DIFF
--- a/code/modules/mob/living/carbon/human/update_icons.dm
+++ b/code/modules/mob/living/carbon/human/update_icons.dm
@@ -901,18 +901,17 @@ There are several things that need to be remembered:
 				client.screen += r_store
 			update_observer_view(r_store)*/
 
-
 /mob/living/carbon/human/update_inv_wear_mask()
 	..()
 	update_body_parts(TRUE)
 
 	if(wear_mask)
-		if(!(ITEM_SLOT_MASK in check_obscured_slots()))
+		if(!(SLOT_WEAR_MASK in check_obscured_slots()))
 			overlays_standing[MASK_LAYER] = wear_mask.build_worn_icon(default_layer = MASK_LAYER, default_icon_file = 'icons/roguetown/clothing/onmob/masks.dmi')
 		update_hud_wear_mask(wear_mask)
 
-	apply_overlay(MASK_LAYER)
 	var/mutable_appearance/mask_overlay = overlays_standing[MASK_LAYER]
+
 	if(mask_overlay)
 		remove_overlay(MASK_LAYER)
 		if(gender == MALE)


Eriripng:
> what if instead of being its own item you could just equp cloth to the eye slot instead?

i can give that a try too, my idea was that you have to tie a blindfold before, yknow, applying to someone's head - much like you do with a sash or a headband; so i went with that in mind

give me a day or two and i can make it so you just need to put the cloth over the mask slot *salute*

code/modules/clothing/masks/_masks.dm
```diff
@@ -8,6 +8,7 @@
 	var/modifies_speech = FALSE
 	var/mask_adjusted = 0
 	var/adjusted_flags = null
+	var/darkness_view = 2 // default
```
Cyprex, line 11: What is this for? This doesn'T do anything as it is, as far as I can tell!

Eriripng: i *think* masks don't have that variable by default, so while that specific line is irrelevant, it *can* be useful in case we wish to add another item that blinds the player (by setting the darkness_view = 1)

we can also move that to here, which would work too!

image

Gyrandola: No point leaving unused variables for possible future code

Eriripng:
> to make things easier you could use the sack's method for making characters blind, would make things a lot easier if you make it so all you need is cloth in the eye slot. I can explain how if you'd like. https://github.com/Rotwood-Vale/Ratwood-Keep/blob/9d2f6c5f790873dee5dba81759ff5d502d2c2222/code/game/objects/items/rogueitems/bags.dm

this is actually perfect! i will take a look on this and alter the code!

Eriripng:
# Changes ✔
- Removed crafting recipe and blindfold item entirely
- Removed unused variable (var/darkness_view)
- Now you can use cloth in the mask slot (yay!)
- Mob will see only a black screen if wearing a cloth in the mask slot
- Mob will also have a blindfold covering their eyes

## Notes 📝
I am slightly concerned with the logic in `update_inv_wear_mask()` in this [commit](https://github.com/Rotwood-Vale/Ratwood-Keep/pull/1681/commits/203abea6cbe54ba3d06cef09c8b6740b580b975f) in case it messes around with any other type of mask item. I HOPE it doesn't; but let me know if any of you notice anything unusual.

### Considerations 🙈
- Make a dmi specifically for the cloth on eyes? Currently we are using just the one from SS13. It's not that bad, but maybe we could get a cute one specifically for Ratwood?
- Maybe add the possibility to blind someone using silk as well?
- Not sure if we have any black cloth item, but add a black blindfold in case we do?
- Currently if you are blindfolded you can still see who is interacting with you (punching, shoving, grabbing) and maybe make it so they can't see it at all? I don't know how to do this, so if this idea is interesting, I would be willing to try!

## Evidences
9484a3bbb618964d949b05358f950ef9

Gyrandola (approved):

Kevak:
Someone on the discord popped out some sprites. 

[https://discord.com/channels/1240735983276654733/1264293885291794492/1311116547670020196]

Eriripng:
> Lets be honest. Currently we have sack with same functionality, but for bandits it doesnt prevent someone to know the hidden path when they haul a person iin. Did you fix that and will cloth prevent people from learning the way?

I think @Dedwolf had a PR approved to solve this specific issue just a few hours ago. And given that the cloth functions quite similar to the sack, I believe it should also prevent people from finding the secret path!

Eriripng:
> Someone on the discord popped out some sprites.
> 
> [https://discord.com/channels/1240735983276654733/1264293885291794492/1311116547670020196]

Thank you! Here's a few screenshots of it in-game.

Screenshot_2
Screenshot_3
Screenshot_4
Screenshot_5

Thank you Betrocious for your generous donation xx

Gyrandola:
image

Eriripng:
god bless our QAs (the players)

i will take a look in a few hours *salute

Eriripng:
# Changes ✔
- Fixed overlays not being removed after equipping any mask item (hopefully)
- Changed sprite to a black cloth over eyes instead (thank you Betrocious)

## Evidences
Screenshot_2

###### I can't record with gyazo but I tested this time removing both the cloth and the mask over the eyes. So HOPEFULLY it's all working now...

PR #1684, opened 2024-11-26
Templar Helmet FOV

Squird5:
## About The Pull Request

I was asked to "fix" the Templar helmets, because wearing them is fucking awful and you can't see anything. Worse, if someone is at your diagonal, you can never parry them, because they are never actually 'seen' mechanically.

* Templar Helmets are now a middle ground of Visored Sallet and Knight's Helmet

<sub>I was going to try and increase the field of view from 90° to 120°, but the way the view cone is currently built, it only works in 45° increments and I really, really do not feel like trying to build a new vision system right now. I made some new view cone overlays, but I am not up for trying to code and test a whole new formula for finding who is in your line of sight, especially with no idea how performant it would be compared to the current system.</sub>

## Why It's Good For The Game
Templar can actually see what they're doing again and defend against people attacking from their corners.

### DIFF
--- a/code/modules/clothing/rogueclothes/hats/templar.dm
+++ b/code/modules/clothing/rogueclothes/hats/templar.dm
@@ -1,13 +1,18 @@
+/obj/item/clothing/head/roguetown/helmet/heavy/templar
+	emote_environment = 3
+	adjustable = CAN_CADJUST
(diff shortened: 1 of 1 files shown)


Gyrandola:
Alright you win.

for now.

PR #1685, opened 2024-11-27
The Great PQ'ing

Vizoplez:
## About The Pull Request
Keep jobs are meant to be played by people with some game-sense due to how important they are to the game.

Why the hell are PQ requirements, therefor, 8 on most Keep jobs?

You can go an Astrata Acolyte, revive 8 people in one round, and then immediately be able to play Knight with no idea how else to play the server.

## Why It's Good For The Game

Raising Knight, Priest, and Duke PQ significantly will give staff more incentive to PQ nuke instead of ban these players.

If they're bad at a  job, they should be removed from it for a long time; not a slap on the wrist.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/bandit.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/bandit.dm
@@ -17,7 +17,7 @@
 	display_order = JDO_BANDIT
 	show_in_credits = FALSE
 	announce_latejoin = FALSE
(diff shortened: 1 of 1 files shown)


Rayforx:
I'm not playing anymore, if this goes through.
Plain and simple.

Rayforx:
You expect people to play 35 x 3 hours for a knight role.
That is 35 days for a normal person, I am heavily against this PR.

RadiantFlash7:
To be frank, this seems like a major non-issue. In all my rounds I have almost never encountered a new player struggling in a role like this more then... twice? I don't really think PQ needs to be raised.

PR #1686, opened 2024-11-27
Goblin Camp Exits

ProbablyCarl:
## About The Pull Request
Provides two additional exits to the Goblin camp. One under the mines, and one next to the fishing hut. Images to follow.
image

image
image

## Why It's Good For The Game
They needed additional exits, due to it being too easy to camp one of them at a time. Might be overkill with where I placed them, but we'll see.


Elli-Skala:
The poor lone fisher

PR #1705, opened 2024-11-28
Assassination and Coup de Grace

Squird5:
Breaking off parts of the Combat Overhaul that need not be part of the current edition

## About The Pull Request

* **Adds Assassination and Coup de Grace attacks**
  * Both attacks bypass all armor except chain mail as they are intended to represent wedging one's weapon between the sections of armor and stabbing the soft human flesh inside.
  * Neither action can dismember, but both can cause artery Critical Hits.
  * Landing a hit past the armor is scaled by skill level.

--- 
* **Assassination** is an attack from behind, requiring an Aggressive Grab to initiate
  * Daggers, Combat Knives and Elven Daggers can all be used to Assassinate

* **Coup de Grace** is an attack from on top, requiring you to share their tile and have an Aggressive Grab
  * Daggers, Hunting Knives, Combat Knives and Elven Daggers as well as Basic Swords, Arming Swords, Short Swords and Messers can all be used to Coup de Grace

## Why It's Good For The Game

Fixes an exploit related to Daggers and Pick Intent and gives a better alternative to the "wedging the blade under their armor"

## Videos

https://github.com/user-attachments/assets/ecf42156-71bd-469b-912b-e38c2a66fbcf

https://github.com/user-attachments/assets/f0e1a9a3-5440-47c2-8c8f-3ed94d0508b2

### DIFF
--- a/code/modules/mob/living/carbon/human/species.dm
+++ b/code/modules/mob/living/carbon/human/species.dm
@@ -1601,15 +1601,15 @@ GLOBAL_LIST_EMPTY(roundstart_races)
 	var/Iforce = get_complex_damage(I, user) //to avoid runtimes on the forcesay checks at the bottom. Some items might delete themselves if you drop them. (stunning yourself, ninja swords)
 
 	var/blade_class = user.used_intent?.blade_class
-	if(get_dir(user, H) == H.dir && H.pulledby == user && user.grab_state >= GRAB_AGGRESSIVE)	//Check for Assassination
+	if(get_dir(user, H) == H.dir && H.pulledby == user)							//Check for Assassination
 		if(I.can_assin && user.used_intent.ican_assin)
-			if(prob(user?.mind?.get_skill_level(I.associated_skill) * 15))				//Skill check, 15-95%
+			if(prob(user?.mind?.get_skill_level(I.associated_skill) * 15))		//Skill check, 15-95%
 				blade_class = BCLASS_ASSASSIN
 				pen = 100
 
-	if(!get_dist(user, H) && H.pulledby == user && user.grab_state >= GRAB_AGGRESSIVE)			//Check for Coup de Grace
+	if(!get_dist(user, H) && H.pulledby == user)								//Check for Coup de Grace
 		if(I.can_cdg && user.used_intent.ican_cdg)
-			if(prob(user?.mind?.get_skill_level(I.associated_skill) * 15))				//Skill check, 15-95%
+			if(prob(user?.mind?.get_skill_level(I.associated_skill) * 15))		//Skill check, 15-95%
 				blade_class = BCLASS_ASSASSIN
 				pen = 100
 


Inanichts:
pretty cool

notafurryy:
I think this could be changed to a normal grab rather than an agro one, there is no way rogues or any class with knives/daggers can get an agro grab on other combat classes.

RyszardParys:
So pick intent is ass now?

RadiantFlash7:
> So pick intent is ass now?

It's actually much better. It's got 70 AP still, and does 1.5 damage

RadiantFlash7:
> > So pick intent is ass now?
> 
> It's actually much better. It's got 70 AP still, and does 1.5 damage
> This PR quietly buffed it

Though in hindsight. That means it does about 12 damage a stab now, with 10 taken off from 10 armor. Maybe not as much of a buff as I thought for any plate gear

HissyNays:
Wow, this looks good.

Coolgat3:
I have a huge problem with the assasination part of this PR.
It requires you to attack someone from behind, while they are aggro grabbed.
The only way that is happening as any dagger wielding role is if you attack without them knowing you are doing so.
Which is against the rules... How are we supposed to utilize this..?

notafurryy:
> I have a huge problem with the assasination part of this PR. It requires you to attack someone from behind, while they are aggro grabbed. The only way that is happening as any dagger wielding role is if you attack without them knowing you are doing so. Which is against the rules... How are we supposed to utilize this..?

They changed that, look at commits

notafurryy:
> How will rogues/dagger users be able to reasonably utilize these against (as pictured) a knight? What would the demo look like if it was an actual rogue against an actual knight attempting to resist being grabbed, or trying to grapple in return?
> 
> Rogues (very often) having less than 10 strength, plus low (novice or apprentice) wrestling skill means that aggressive grabbing is a considerable barrier to this. Additionally, being locked in a grab as a rogue means that you're sacrificing your dodging, which is the primary means used to avoid being hit and killed.
> 
> If this doesn't pass through chainmail, then you're further limited in what parts of the body can be picked through. If someone so much as has a hauberk, you aren't going to be able to do this on their chest, arms or legs. I (think) this is to prevent heart crits using pick, which is reasonable, but it seems only restrictive instead.
> 
> If your opponent is worth using these intents against because of their heavy armor, it's also very likely they're going to have superior wrestling skill and strength that will prevent you from being able to use this. Interesting counterplay this does not make, in my opinion.
> 
> Pick intent was never an exploit or bug in the code, it was a deliberate choice before this server was made to give daggers some usage against heavily-armored opponents. It only recently became a larger point of contention because chest crits were fixed, resulting in heart crits being possible with the pick. Being stabbed in the heart through some of the best available armor is obviously bad, that much I don't disagree on.

You should look at the code first, the pick intent still has enough ap to go through other armor, it just can't go through full plate anymore, also rogues shouldn't be able to 1v1 the knights that are supposed to be elite soldiers and if they want to do that they should use trickery, traps and alike

Gyrandola (approved):

PR #1713, opened 2024-11-28
Spellsowrd merc update! And slight balance tweaks

notafurryy:
## About The Pull Request

Adds a new _enchanted_  sword, currently only available to the spellblade merc

enchanted sword3

Small rebalance to the class:
-lower sneaking
-lowers the ammount of slots
-allows more species while still being region thematic

Adds LORE by albo3

## Why It's Good For The Game

Allows a bit more freedom when choosing the merc class also adds an unique piece of equipment that could also be used elsewhere. Also, balance tweaks from player feedback.


RadiantFlash7:
Oh god, a burn weapon? You do know that makes it impossible to kill people with it, right?

Like you can't even die from being permanently engulfed in flames.

notafurryy:
> Oh god, a burn weapon? You do know that makes it impossible to kill people with it, right?
> 
> Like you can't even die from being permanently engulfed in flames.

The weapon doesn't set you on fire, it deals burn damage instead of brute, and you can still get crits with the cut and stab intent, you can't cut off limbs though.

RadiantFlash7:
I threw my thoughts into dev general on discord, but I'll throw them here as well.
The damage factor for cut should be removed because it's worse then every other sword on cut, and cut is what most people will be using when an opponent is moving too much to 'freeze'.
The defense is probably too low, to be frank. Spellswords are meant to be dodge mains, but sabers are designed around being able to be used one-handed. You can't constantly stay on dodge intent because you'll get tired out, and if a one handed weapon's defense is so low... You're never gonna parry anything without a shield; Which encourages them to buy a buckler.

And finally- The burn damage is a bit of an issue. I'd reccomend looking at stun maces to be able to toggle burn damage on and off, because if a weapon is unable to kill deadites or skeletons, thats a /major/ problem. Those are some of the most common enemies to ambush you, and if you run into one as a spellsword with this weapon, You're SOL.

notafurryy:
> I threw my thoughts into dev general on discord, but I'll throw them here as well. The damage factor for cut should be removed because it's worse then every other sword on cut, and cut is what most people will be using when an opponent is moving too much to 'freeze'. The defense is probably too low, to be frank. Spellswords are meant to be dodge mains, but sabers are designed around being able to be used one-handed. You can't constantly stay on dodge intent because you'll get tired out, and if a one handed weapon's defense is so low... You're never gonna parry anything without a shield; Which encourages them to buy a buckler.
> 
> And finally- The burn damage is a bit of an issue. I'd reccomend looking at stun maces to be able to toggle burn damage on and off, because if a weapon is unable to kill deadites or skeletons, thats a /major/ problem. Those are some of the most common enemies to ambush you, and if you run into one as a spellsword with this weapon, You're SOL.

I will be making most of those changes

Kevak:
Should enable wildkin. If halfkin are enabled, wildkin should be too.

Gyrandola (approved):

PR #1721, opened 2024-11-29
Maniac Buff

ProbablyCarl:
## About The Pull Request
As well, it gives them dark vision, so they don't have to rely on torches / lamps and can stalk in the dark.

Returns their strength, endurance and constitution increases, though to 16, not 20.

## Why It's Good For The Game
These are changes based on active testing on the server.
I took a hammer far too strong to Maniac, prior to the aforementioned active testing, so we've taken to buffing it incrementally on the server, and this is what we settled on being ok-ish.

### DIFF
--- a/code/modules/antagonists/roguetown/villain/maniac/maniac.dm
+++ b/code/modules/antagonists/roguetown/villain/maniac/maniac.dm
@@ -22,7 +22,6 @@
 		TRAIT_STEELHEARTED,
 		TRAIT_NOMOOD,
 		TRAIT_SCHIZO_AMBIENCE,
-		TRAIT_NOROGSTAM,
 		TRAIT_DARKVISION,
 	)
 	/// Traits that only get applied in the final sequence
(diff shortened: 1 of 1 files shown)


RadiantFlash7:
This is a wild buff change, good god. Maniac gets legendary knifes and master wrestling as it is. Granting them 18 end and 18 str, and there is pretty much no chance anyone will kill them aside from mobbing them with 3 people. 

As it is right now, all the maniac player has to do is grab someone and their toast. My assumption is the person who wrecked a maniac was quite literally the one role who can combat grabs, a monk?

Even without grab combat, legendary knife is a damn good skill, ontop of whatever stats a person has from their job they rolled maniac as.

radarseas:
I think this is for the greater good. /however/ I'd be careful with infinite stamina, maybe swap it for crit resistance and stun resistance instead so they can escape a lucky chest bleed's stun, drink a red, suture it up and maybe survive?

remember adventurer rogues/rangers _do exist still_, when you get a rogue/ranger that can tackle a knight and expert-dodge at 90% with no stamina counter and has expert knives and wrestling, **there's a serious problem.**

RadiantFlash7:
> > Personally Im against buffing maniac for now, but we'll see what the others think.
> 
> 
> 
> We had an Adventurer shit on the Maniac earlier. It was some of the funniest shit imaginable.

Adding to this- Apparently it was an heiress maniac. Against a ranger. The heiress gets a lot of stat penaltirs from my understanding, so I'm not surprised.
Frankly, raising the maniac str con and end to 15 might be a good idea? But 18 is massively excessice

john-pawson:
> _I hate that dodging has no drawbacks other than being elevated where you might potentally trip and fall from turning on defensive mode, when you lack stamina as an issue._

Little reminder that you can't dodge on armors you're not proficient in and it takes double the amount of stamina. So yes, it does have drawbacks. Just to let you know.

Gyrandola:
Infinite stamina is a no-go.
Lower the STATS to 16 from 18 and we have reached a compromise.

Gyrandola (approved):

PR #1729, opened 2024-11-29
Fem Tabaxi SPD nerf

Savotta:
## About The Pull Request

Femtabaxi now have -1 STR, 1 INT, -2 CON, -1 END, 2 SPD and 2 PER (so i essentially just dumbed down their speed from 3 and gave them an extra perception point. yay). Males remain unchanged with 0 STR, 1 PER, 0 INT, -1 CON, -1 END and 2 SPD.

## Why It's Good For The Game

idk people just really didnt like the fact that femtabaxi had no malum for a whopping +3 spd i think

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/furry/tabaxi.dm
+++ b/code/modules/mob/living/carbon/human/species_types/furry/tabaxi.dm
@@ -37,7 +37,7 @@
 		OFFSET_SHIRT_F = list(0,0), OFFSET_ARMOR_F = list(0,0), OFFSET_UNDIES_F = list(0,0)
 		)
 	specstats = list(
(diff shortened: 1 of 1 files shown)


Elli-Skala:
I belive the stats in the pr body and actual changed are different

Savotta:
> I belive the stats in the pr body and actual changed are different

They inherit the -1 STR, 1 INT, -2 CON from femhuman code, all this PR does is remove the +1 to their already racial 2 SPD and give them an extra perception point to make up for the CON malum.

RadiantFlash7:
This is a probably not a bad idea. +3 to any stat, for any race is a bit much.

Gyrandola (approved):

PR #1731, opened 2024-11-29
Removes pinkballs from the purity vendor

wow50000:
## About The Pull Request
See title

## Why It's Good For The Game
Removes pinkballs from the drug machine due to it being an easier to carry health potion that really does not fit the theme of the game, trash addition. A mapper can remove it from the map before this trash gets properly removed.


RadiantFlash7:
In truth, these probably should be removed universally, not just from the vendor, but from- every source. Patch kits, the bandit vendor, etc. As nice as these things are, they're fairly major power creep.

Kevak:
Yeah these need to be removed fullstop. Being able to carry around 10 red potions in one inventory slot is absurd. While you're at it, snakebite should be removed too. P much all the mori meds are kinda absurd.

Gyrandola:
i will be waiting for your next removals

Gyrandola (approved):

PR #1734, opened 2024-11-30
Speedballing Nerf

Dedwolf:
## About The Pull Request

Drug abuse has consequences.

Combining ozium and moondust, speedballing, is now far less ideal for combat, you'll get confused, dizzy, jittery and you'll gasp. It won't be enough to kill you but it should make dosing up on both drugs to counter them less viable. 

Combining moondust and moondust purest now makes you overdose. Pure and simple. You can't take two hits of regular moondust, why would taking both dusts NOT give you an overdose? This is likely an oversight.

## Why It's Good For The Game

I always hate answering this question -- You tell me. 

Sorry for video quality i am australian

https://github.com/user-attachments/assets/fca71f7b-7d9c-4f91-b88b-dfe29bfa89e0

### DIFF
--- a/code/modules/reagents/reagent_containers/powderspice.dm
+++ b/code/modules/reagents/reagent_containers/powderspice.dm
@@ -264,7 +264,10 @@
 
 /datum/reagent/moondust
 	name = "Moondust"
-
+	description = ""
+	color = "#bfc3b5"
+	overdose_threshold = 16
+	metabolization_rate = 0.2
 
 /datum/reagent/moondust/overdose_process(mob/living/M)
 	M.adjustToxLoss(0.25*REM, 0)


Gyrandola:
Could you remove the white flashing? I would rather not go to jail for giving someone an epilectic attack.

Inanichts:
Cant see shit LOL

notafurryy:
Overdosing does not seem to work currently? Overdoses for normal moondust do not work at all currently and for uncut moondust you only start dying after the third pile, could maybe use this pr as an opportunity to fix it too.

Kevak:
Note that bathoa’s patron boon negates overdose effects. Make sure not to break that.

Also yeah get rid of that flashing effect or at least slow the animation way down.

Coolgat3:
> Could you remove the white flashing? I would rather not go to jail for giving someone an epilectic attack.

That's pure moondust high start, was like that since we started

RyszardParys:
Nice fucking 10p video you got there, good sire.

PR #1739, opened 2024-12-01
Exploit/game abuse removal

RadiantFlash7:
## About The Pull Request
Three main things, I'll add more as people bring them up to me as long as this PR is open:

1. Spitfire spell no longer adds 'stacks' of nonexistant fire, that trigger when someone is set on fire,
2. Steelmasks in hoardmaster as no longer remotely likely to be cheaper then the sell price of steel even on low rolls of prices.
3. Bandits can no longer use shylocks. Stop trying to act like a towner and sell to a stockpile for mammons.

## Why It's Good For The Game
Various things are about that really should be fixed, or had unintended consequences. This is just a few off the top of my head that got bragged/shared on the discord recently.

### DIFF
--- a/code/modules/roguetown/roguemachine/ATM.dm
+++ b/code/modules/roguetown/roguemachine/ATM.dm
@@ -11,6 +11,9 @@
 	if(!ishuman(user))
 		return
 	var/mob/living/carbon/human/H = user
+	if(H.mind?.special_role == "Bandit")
+		to_chat(H, "<span class='warning'>The idea of giving hard won coin to nobles repulses me!</span>")
+		return
 
 	if(HAS_TRAIT(user, TRAIT_MATTHIOS_CURSE))
 		to_chat(H, "<span class='warning'>The idea repulses me!</span>")
--- a/code/modules/spells/roguetown/wizard.dm
+++ b/code/modules/spells/roguetown/wizard.dm
@@ -267,10 +267,10 @@
 /obj/projectile/magic/aoe/rogue2
 	name = "spitfire"
 	icon_state = "fireball"
-	exp_heavy = 0
-	exp_light = 0
-	exp_flash = 1
-	exp_fire = 0
+	var/exp_heavy = 0
+	var/exp_light = 0
+	var/exp_flash = 1
+	var/exp_fire = 0
 	damage = 15	//no armor really has burn protection. So assuming all three connect, 45 burn damage- average damage of fireball with firestacks nerfed. Thats a big 'if' however. Notably, won't cause wounds,
 	damage_type = BURN
 	homing = TRUE


NikolayByondPlayer:
Ahahah someone used shylocks as bandits? :DD
We can make so it eats their arms when they try to us it :D

Hmeister-real:
> Ahahah someone used shylocks as bandits? :DD We can make so it eats their arms when they try to us it :D

would be funnier if it artery crits their finger

Gyrandola:
you changed the magic type path to something thats potentially confusing

Crowbar764:
@RadiantFlash7
About the Shylock nerf.

That is removing yet another thing bandits could be doing instead of fragging and thus maybe a step in the wrong direction?

But more importantly, creating a Shylock account is the /only/ way for Bandits to steal the treasury which is, you know, a big objective for them.

If you are going to keep this change, consider adding a new feature on the steward machine that lets you directly take mammons out. I am sure the Steward players would thank you for this too.

RadiantFlash7:
> you changed the magic type path to something thats potentially confusing

I could always change its name to something a bit clearer. I was trying to get rid of the fireball parent since it inherently had firestacks being added

radarseas:
I think bandits being unable to use shylocks is fine. _it should have been that way from the start_, using one puts you below the master rod's control and is against your sacred matthiosian oath of _**FREEDOM!!**_, if anything the message prompted should be changes from `"the idea of giving nobles my hard earned coin repulses me"` to something to reflect reflect that oath.

honestly big antags like vlord shouldn't be able to use it either while you're at it, as funny as the idea of a vlord going off to farm/mine to get 2000+ mammon is.

zeratino:
fuck my stupid bandit life

code/modules/roguetown/roguemachine/ATM.dm
```diff
@@ -11,6 +11,9 @@
 	if(!ishuman(user))
 		return
 	var/mob/living/carbon/human/H = user
+	if(H.mind?.special_role == "Bandit")
+		to_chat(H, "<span class='warning'>The idea of giving hard won coin to nobles repulses me!</span>")
+		return
 
 	if(HAS_TRAIT(user, TRAIT_MATTHIOS_CURSE))
 		to_chat(H, "<span class='warning'>The idea repulses me!</span>")
```
Heheniy, line 19: just make it a single check for matthios patron i suggest

Rayforx:
I do have to agree that it's dumb that bandits are using the shylock to become a towner, but it is also the only way that the bandits can steal from the steward's tower, which is one of their "main" objectives. It would be better to make a check for the stockpile to prevent bandits from being able to deposit than to take the entirety of one of the main ways that bandits rob the keep away.

There is also a certain amount of risk that comes from bandits having a shylock (stun rod goes brr), and it does technically make them "open" for being scry'ed since the steward knows their name. So, it becomes a bit of a risk/reward situation.

Gyrandola (approved):

PR #1745, opened 2024-12-01
Nerf Lich and VLs

ghost:
## About The Pull Request
According to Gyran, No antags should have the trait inf rog stams and the no pain traits therefore Vampire Lords and lichs got those traits removed.  
They also mentioned that minor antags are not supposed to be powerful, so lich had several of their traits removed(norogstam, no pain, no sleep no limbdisable and shock immunity) to ensure they're on par as the others as a minor antag.

## Why It's Good For The Game
No, it isn't good, but since everyone want to nerf antags to the ground 🤷


Gyrandola:
Pathetic

PR #1749, opened 2024-12-01
Slight Buff to Maniac, Remove blue colour to peoples marked by the wonders

ghost:
## About The Pull Request
A Slight Buff to Maniac which is removing the knocked down and freaking out when hit by a hallucination, only for them though.
Remove the blue screen when you are marked by the wonders, instead you get cursed with hallucinations.
I'll try to find a way to remove the mood effect AND the hallucination effects for marked peoples when the maniac die when i am not sleep deprived.

## Why It's Good For The Game
No comment, tried it in a private server, it work for both things.

### DIFF
--- a/code/modules/antagonists/roguetown/villain/maniac/maniac.dm
+++ b/code/modules/antagonists/roguetown/villain/maniac/maniac.dm
@@ -23,7 +23,6 @@
 		TRAIT_NOMOOD,
 		TRAIT_SCHIZO_AMBIENCE,
 		TRAIT_DARKVISION,
-		TRAIT_NOSLEEP, //Buddy, You're trying to wake up.
 	)
 	/// Traits that only get applied in the final sequence
 	var/static/list/final_traits = list(


Gyrandola:
Maniac was just buffed, he doesnt need more buffs right now. 

Giving visions to those who see the wonders is cool.

radarseas:
nosleep is actually a pretty heavy-handed nerf as they can't heal off burns and other injuries through without a mechanic to circumvent the disadvantages, unlike vampires without miracles.

it leaves them pretty royally fucked in quite a few majority of circumstances where losing 1 speed is barely an issue. There's a reason they had so many ungodly powerful traits that let them absolutely slaughter people while fatally injured tenfold, because they couldn't really \heal\ to compensate under a majority of circumstances.

Gyrandola:
> > > losing knocked down is good, else people can swift-intent neck due to never missing on downed people, _also it lets them, somewhat blend in if they fuck up, dodging a hallucination instead of being instantly meta'd by freaking out and falling over._ While still putting the pressure on in a fight.
> > > nosleep is actually a pretty heavy-handed nerf as they can't heal off burns and other injuries through without a mechanic to circumvent the disadvantages, unlike vampires without miracles.
> > > it leaves them pretty royally fucked in quite a few majority of circumstances where losing 1 speed in comparison is barely an issue. There's a reason they had so many ungodly powerful traits that let them absolutely slaughter people while fatally injured tenfold, because they couldn't really \heal\ to compensate under a majority of circumstances.
> > 
> > 
> > Maniacs can still dodge the hallucinations. So, when people start being affected by the totems, it will not be easy to identify whos the maniac.
> 
> the issue is that this hallucination can proc In a fight, and being hit with it would be basically a death sentence,, this atleast give them a chance if they actually get affected by one of the hallucination

Right, i fear this will be easy to metagamed as "He fell, so he is not a maniac", "he didnt fall, so he is a maniac".
Let's see how it goes though.

Gyrandola (approved):

PR #1750, opened 2024-12-01
Removes vampire crit weakness

wow50000:
## About The Pull Request
See title.

## Why It's Good For The Game
Old Vlord was a glass cannon which would work on a frag heavy server as don't get hit and frag everyone, however in this setting I believe vampire should be a better rounded antag which can't be insta killed nor insta kill people, especially since the sleep spell was reworked to not insta sleep people. There is still the issue of inf stam but crit weakness makes it too easy to kill vamp lords accidentally.

someone else should tone down vampires in another way such as skills or stats to reflect this.


Gyrandola:
Definition of a salt pr

PR #1640, opened 2024-11-22
Allows druid to use beasttongue

Vizoplez:
## About The Pull Request
The Verevolf zodomites wouldn't shut up about it.

I don't care about this PR at all and wont' be responding to balancing.

If they want it merged, they need to argue for it.

## Why It's Good For The Game
Easier to fuck verevolfs for druids ig

PR #1751, opened 2024-12-01
Eoran acolytes can perform marriages.

Vladmar11:
Allows eoran acolytes to marry people. Removes marriage-ability druids.

Why druids are not supposed to perform marriage: 

1. Dendor is the god of Nature and Earth, not Love or Family and such things. Dendor is the mad man who actually care about his beasts and nature itself. Druids are his followers, they need to worship him  - protect animals, nature, live in unity with it.

2. Someone was telling that druids can perform marriages because there are no eorans. So, we added them already.

3. Druids already had some buffs like https://github.com/Rotwood-Vale/Ratwood-Keep/pull/1639 (wooden stuff and polearm skill 1-2, friendly animals spiders, rats, volves, moles), https://github.com/Rotwood-Vale/Ratwood-Keep/pull/1640 (they can speak with werevolves, beasts).

I think this is too much for them to gain marriage ability. This is the special option for eorans.

image

### DIFF
--- a/code/game/objects/structures/fluff.dm
+++ b/code/game/objects/structures/fluff.dm
@@ -1038,7 +1038,7 @@
 
 /obj/structure/fluff/psycross/attackby(obj/item/W, mob/living/carbon/human/user, params)
 	if(user.mind)
-		if((user.mind.assigned_role == "Priest") || (user.mind.assigned_role == "Acolyte") && (user.patron.type == /datum/patron/divine/eora))
+		if((user.mind.assigned_role == "Priest") || ((user.mind.assigned_role == "Acolyte") && (user.patron.type == /datum/patron/divine/eora)))
 			if(istype(W, /obj/item/reagent_containers/food/snacks/grown/apple))
 				var/marriage
 				var/obj/item/reagent_containers/food/snacks/grown/apple/A = W


Gyrandola:
Eora acolytes performing marriage rites makes a lot more sense than druids.

Lord-Preston:
I think we should keep druid marriages for a few reasons. Practically speaking, there is almost always one druid and they’re typically free. This is not the case for the priest, and I am not sure the availability of Eoran acolytes since they’re new. Druids also travel around more, so they can marry people out in the Terrorbog who otherwise would not head into town. That is why I removed the location requirements for marriage. When I play druid I travel around helping Terrorbog settlers usually, but acolytes are usually expected to stay around the church/town.

As for lore reasons, yes Dendor is the God of Nature. In nature animals can be very ritualistic (of course not to the extent of humans), and some animals mate for life like wolves and many birds. Given the broad number of races one can play in-game, I think it makes sense that Dendorites would have some sort of marriage ritual for followers of Dendor who want to have one mate for life. I imagine Druidic circles would be the primary religious orders tending to religious needs in remote locations around the world, including marriage.

Actually, I think a unique flavour text depending on who performed the wedding would work best, like calling Druidic marriages lifebonds.

code/game/objects/structures/fluff.dm
```diff
@@ -1036,9 +1036,9 @@
 	max_integrity = 80
 	chance2hear = 10
 
-/obj/structure/fluff/psycross/attackby(obj/item/W, mob/user, params)
+/obj/structure/fluff/psycross/attackby(obj/item/W, mob/living/carbon/human/user, params)
 	if(user.mind)
-		if(user.mind.assigned_role == "Priest" || user.mind.assigned_role == "Druid")
+		if((user.mind.assigned_role == "Priest") || (user.mind.assigned_role == "Acolyte") && (user.patron.type == /datum/patron/divine/eora))
```
Heheniy, line 1041:
```suggestion
		if((user.mind.assigned_role == "Priest") || ((user.mind.assigned_role == "Acolyte") && (user.patron.type == /datum/patron/divine/eora)))
```
for better reading

Gyrandola (approved):

PR #1761, opened 2024-12-02
New Lightning

Gyrandola:
## About The Pull Request

Ports:
- https://github.com/Vanderlin-Tales-Of-Wine/Vanderlin/pull/12.

This might require some tweaking before we achieve an atmosphere that satisfies most people.


Gyrandola:
Giving this a try. Leave your feedback so stuff can be tweaked accordingly, please dont be an ape about it.

Gyrandola:
I feel iffy about this. Our lightning systems are already very similar as, I believe, ours was reworked some time in the past by azarak. I like ours better.

PR #1762, opened 2024-12-03
Removes RnG from some roles

wow50000:
## About The Pull Request
Removes random stats from some roles, secretly buffs blacksmith, nerfs fisher knife skill(may be an issue)

## Why It's Good For The Game
RnG either makes a class feel underpowered or overpowered depending on round, this either means you should kill yourself if you get a bad roll or be more powerful than you should.

Smiths were stealth buffed because forge clerics exist and have miracles


Coolgat3:
I dunno I think the feeling of rolling good stats is awesome

Gyrandola (approved):

PR #1769, opened 2024-12-03
Remove Sleeve System

Gyrandola:
## About The Pull Request

A convoluted, sprite layering system that exists just to lets players tear the sleeves off only certain dresses. The real issue? Since every clothing sprite requires its sleeves to be separated in another icon file, It exponentially increases the TIME & COST of creating each new clothing piece. I don't think it's worth it at all.
Sleeve tearing is sacrificed for now but it can be reimplemented in a way thats much less aids.

- This additionally fixes all current missing-sleeves sprites.

### DIFF
--- a/code/modules/clothing/rogueclothes/new/armor.dm
+++ b/code/modules/clothing/rogueclothes/new/armor.dm
@@ -4,7 +4,6 @@
 	desc = "A long leather coat made with quality materials for experienced hunters or noble explorers. The leather still offers some protection."
 	icon = 'icons/roguetown/clothing/newclothes/dripobj.dmi'
 	mob_overlay_icon = 'icons/roguetown/clothing/newclothes/onmob/onmobdrip.dmi'
-	sleeved = 'icons/roguetown/clothing/newclothes/onmob/sleeves_armor.dmi'
 	icon_state = "leathercoat"
 	blocksound = SOFTHIT
 	body_parts_covered = CHEST|GROIN|VITALS|ARMS|LEGS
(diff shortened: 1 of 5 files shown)


radarseas:
> why?

outfits render for the outfit itself, and each sleave as seperate textures. which is extrodinarily inefficient for making new ones and left several broken from a side-view where there was no sleaves rendering, its most noticable with noble-dresses or formal silks, or the vampire lord's regal silks without the cape.

Coolgat3:
but why remove the quickest way to get cloth in an emergency situation.. i'd say leave it in just don't change sprite v.v

DeerJesus:
leave the mechanic of sleeve tearing in just remove the different sprites

TheFenix323:
> leave the mechanic of sleeve tearing in just remove the different sprites

The sleeve tearing system is still there

Gyrandola:
I'm psychologically unable to bring this to completion for the time being, because I need to manually stitch back together all the sleeves to their corresponding main sprites.

dwasint:
> > why?
> 
> outfits render for the outfit itself, and each sleave as seperate textures. which is extrodinarily **inefficient** for making new ones.
> 
> this left several broken from a side-view where there was no sleaves rendering, its most noticable with noble-dresses or formal silks, or the vampire lord's regal silks without the cape. where you can see the sleaves from the front but at a side-view, it doesn't cover your arms.
> 
> it also means for every new outfit you need to make the outfit _**AND**_ the _**SLEAVES**_ over the arms seperately.
> 
> TLDR: a small sacrifice for great progress

honestly if the reason is the workload I can make a script that splits icons into generic sleeves. or we just forgo that and use masks

NPC1314:
Some sort of masking system would probably be the best solution, or just leave it as it is. While annoying it does actually work as designed.

Issue #534, opened 2024-07-22
[BUG] When a mask is taken off, and put into the belt slot, it still blocks the mouth


Hmeister-real:
## Describe the bug

It shouldnt, but if you put it on your belt it still blocks the mouth

## Steps to Reproduce:

Take mask from face onto belt, still cant drink, sorcery

Squird5:
Also makes you bald

![ratwood_helm bug 1](https://github.com/user-attachments/assets/84e11d62-ab0a-448a-b364-98930de81bc8)
![ratwood_helm bug 2](https://github.com/user-attachments/assets/9a1c6f4e-4a02-42d3-9350-30440ce7eeae)

Hmeister-real:
This is still a big issue, with masks and other armors that can go in the belt slots (you cannot sleep with armor in the belt slot either now so it's more of an issue than before)

PR #1780, opened 2024-12-03
Helmets no longer hide hair if they are in your belt.

TheFenix323:
## About The Pull Request
Makes the proc that hides hair to check if the item is in the belt slot, doing this it won't remove the hair overlay when is not in the head slot.

https://github.com/user-attachments/assets/b831bdb3-13c7-4124-902d-91ae1d83b51d

## Why It's Good For The Game
It fixes a bug.

### DIFF
--- a/code/__HELPERS/ratwood.dm
+++ b/code/__HELPERS/ratwood.dm
@@ -79,6 +79,8 @@
 		return TRUE
 	var/list/worn_items = get_all_worn_items(human)
 	for(var/obj/item/item as anything in worn_items)
+		if(item.worn_in == SLOT_BELT_L || item.worn_in == SLOT_BELT_R)
+			continue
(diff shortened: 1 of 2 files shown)


Hmeister-real:
make it fix this too
https://github.com/Rotwood-Vale/Ratwood-Keep/issues/534

It got falsely marked as completed by Mori and never got opened again

TheFenix323:
> make it fix this too #534
> 
> It got falsely marked as completed by Mori and never got opened again

Seems it fixed that too

https://github.com/user-attachments/assets/9750bfd1-7d89-49f8-8fd3-4508e85ad99b

TheFenix323:
nevermind is still an issue

Gyrandola:
one fix at a time

Issue #1766, opened 2024-12-03
[REQUEST] Make Rotting/Deadite turn SLOWER.


Hyro-Fox:
## Is your feature request related to a problem? Please describe. 
Deadites/Rotting Starts the moment you die, It a ticking time before you actually turn into one, but here's the problem. 2-3 fucking minutes is enough for you to get rotten, any Astratans can't help you.
Deadites? Sure, if a Physician or Pestran managed to cure rot, they get revived right away.
Rot? No. Physicians can't Unrot, Pestrans only manage to cleanse off the Rot but that's about it, Physicians used to do BOTH but they are useless but to revive the heart if they don't turn before... side note: If they die and take 1 minute, it is another minute praying to get lucky + skill enough to resuscitate.

## Describe the solution you'd like
Make Rotting / Deadites turn slower, 25% slower to spare the headache. Make Physicians and Astratans happier, it should spare some Pestran some of the headache.

## Describe alternatives you've considered
Either Make Deadites unable to bite faces or just can't fight at all if they get skullcracked or jaw broken, they SHOULD not be able to fight or bite if they are tetraplagic.

## Additional context
No addictional context, just make this less of a massive headache than already is. Dying outside of bog and being brought back just 2 minutes to turn into deadite inside a church? Nah, prolong this one, please.

PR #1781, opened 2024-12-03
Heavily buffs rot timers

wow50000:
## About The Pull Request
changes rot from 10 mins to 45, skeletonise from 25 to 90, dust from 35 to 135

## Why It's Good For The Game
Requested in #1766 makes acolytes jobs easier and gives people more time to haul bodies

### DIFF
--- a/code/datums/components/rotting.dm
+++ b/code/datums/components/rotting.dm
@@ -62,13 +62,13 @@
 	for(var/obj/item/bodypart/B in C.bodyparts)
 		if(!B.skeletonized && B.is_organic_limb())
 			if(!B.rotted)
-				if(amount > 10 MINUTES)
+				if(amount > 45 MINUTES)
 					B.rotted = TRUE
 					findonerotten = TRUE
 					shouldupdate = TRUE
 					C.change_stat("constitution", -8, "rottenlimbs")
 			else
-				if(amount > 25 MINUTES)
+				if(amount > 90 MINUTES)
 					if(!is_zombie)
 						B.skeletonize()
 						if(C.dna && C.dna.species)
@@ -77,7 +77,7 @@
 						shouldupdate = TRUE
 				else
 					findonerotten = TRUE
-		if(amount > 35 MINUTES)
+		if(amount > 135 MINUTES)
 			if(!is_zombie)
 				if(B.skeletonized)
 					dustme = TRUE


Gyrandola:
Rot must remain central to the setting. Upgrading the time it takes for people to zombify to 45 whole minutes would negate its existence.

wow50000:
rot is an unfun mechanic imo but thats fair, the setting requires it

wow50000:
> Considering how rot is supposed to be one of the main threats lore-wise i don't think cuadrupling the rot timer is exactly good since it kind of almost removes it entirely. Not only that but it also makes death, an already pretty easily curable thing, now even more of a non-problem.
> 
> Why not instead make it so there's ways to _slow down_ the rot? Like modifying the winding sheets so they slow down rotting or just adding a drug that delays rotting slightly, instead of making it take so long that it just isn't going to show up anymore if it takes more than half an hour to even appear?

people treat death like nothing because they have 10 backup characters lined up or they want to die IC, not because of rot timers.

john-pawson:
> people treat death like nothing because they have 10 backup characters lined up or they want to die IC, not because of rot timers.

Well i don't see why we should make the problem worse by making rot timers higher.

Also, i'm pretty sure this doesn't exactly apply to everyone. If you die as a knight mid-round there's a pretty big chance you'll not be able to just go in as a backup character again, same goes with about every single role that doesn't have a lot of slots open (Adventurer, Pilgrim, Mercenary, etc.) and for those that do have a lot open, you all lose your gear most of the time. 
Excluding of course the occasional death to spider in the middle of the bog, in which they'll probably just go metagaming and pick up their stuff back.

Also, i never said people treat death like nothing because of rot timers, if anything it's the only thing preventing every single person from getting revived. (Asides from the deliverate decapitations where they hide the head, in which case they, ridiculosly, can still be revived with the head of some other guy or zombie.)

john-pawson:
> > people treat death like nothing because they have 10 backup characters lined up or they want to die IC, not because of rot timers.
> 
> Well i don't see why we should make the problem worse by making rot timers higher.
> 
> Also, i'm pretty sure this doesn't exactly apply to everyone. If you die as a knight mid-round there's a pretty big chance you'll not be able to just go in as a backup character again, same goes with about every single role that doesn't have a lot of slots open (Adventurer, Pilgrim, Mercenary, etc.) and for those that do have a lot open, you lose all your gear most of the time. Excluding of course the occasional death to spider in the middle of the bog, in which they'll probably just go metagaming and pick up their stuff back.
> 
> Also, i never said people treat death like nothing because of rot timers, if anything it's the only thing preventing every single person from getting revived. (Asides from the deliverate decapitations where they hide the head, in which case they, ridiculosly, can still be revived with the head of some other guy or zombie.)

PR #1774, opened 2024-12-03
Turning no longer cancels running

Gyrandola:
## About The Pull Request

Title.

Ported from
- https://github.com/StoneHedgeSS13/StoneHedge/pull/171

PR #1808, opened 2024-12-05
Revert "Turning no longer cancels running"

Hmeister-real:
Reverts Rotwood-Vale/Ratwood-Keep#1774 because it makes running very finnicky.
The whole reason behind stopping when you turn is so that people cannot be inhumanly fast and accidently crash into walls, when you turn you now almost lose control because of how fast you are goind and slam into shit. It's not a good change, the only thing it benefits is rogues and fae who are super fast already and now don't ever need to slow down untill they get low stam. I say this as a rogue main myself


Crowbar764:
Can we make this an in-game preference option at least. I'm sure it's a matter of personal opinion which is better and we should let people decide themselves

Hmeister-real:
> Can we make this an in-game preference option at least. I'm sure it's a matter of personal opinion which is better and we should let people decide themselves

Going to try and make it a toggle

Gyrandola:
Vote is against it, but if you make it a toggle I'll accept it.

PR #1810, opened 2024-12-05
Fix nerve master

Vladmar11:
Fixed the code and made it possible to open the nerve master with a keyring if there is a steward key in there.

### DIFF
--- a/code/modules/roguetown/roguemachine/steward.dm
+++ b/code/modules/roguetown/roguemachine/steward.dm
@@ -15,8 +15,7 @@
 	max_integrity = 0
 	anchored = TRUE
 	layer = BELOW_OBJ_LAYER
-	locked = FALSE
-	lockid = "steward"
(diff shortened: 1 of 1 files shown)


Gyrandola (approved):

code/modules/roguetown/roguemachine/steward.dm
```diff
@@ -33,15 +32,15 @@
...
+			if(KE.lockid == keycontrol)
 				locked = !locked
 				playsound(loc, 'sound/misc/beep.ogg', 100, FALSE, -1)
 				update_icon()
 				return
-		to_chat(user, span_warning("Wrong key."))
-		return
+			else
+				to_chat(user, span_warning("Wrong key."))
+				return
```
TheFenix323, line 43: Won't this cause the lock to not open if the first key is not the steward key?

PR #1816, opened 2024-12-05
Pantheon mood adjustments

Hmeister-real:
## About The Pull Request

Adjusts negative moodlets to work with the inhumen pantheon. Fixes /issues/1689

## Why It's Good For The Game

Better mood


Xaytan:
Hey, aren't a lot of these backwards? You're having it return false for can_apply if they DON'T have the traits that block them?

Hmeister-real:
> Hey, aren't a lot of these backwards? You're having it return false for can_apply if they DON'T have the traits that block them?

No the code is just funny looking like that

Gyrandola (approved):

PR #1822, opened 2024-12-05
gives swashbucklers lockpicking

Slorbin:
## About The Pull Request

gives swashbucklers lockpick 2

## Why It's Good For The Game

pirates, treasure hunters, mercs and rogues all get it. not sure why the rogue subtype doesnt


john-pawson:
Pirates, trasure hunters, rogues, all people that usually go around looting and need to open locks if it's either for opening stolen chests, or dug up ones. Now why would mercs or swashbucklers have them?
Yes i know that you're not touching mercs or giving them the skill, i'm just saying how it doens't make sense for that role to have lockpicking. We really need less roles that are just good at everything.

Gyrandola:
read above

PR #1826, opened 2024-12-06
Map Adjustments

RadiantFlash7:
## About The Pull Request
This PR makes a few map changes.
First of all the Lavamines is now actually woth mining at:
image
image
Prior to this change, it was entirely randomized, and not worth mining here, compared to other areas.

Secondly: There is no longer a mineable gap on the second level, on either side of the vanguard gate. This was an oversight, according to mappers, and has been corrected.
Thirdly: Some new minor buildings and things have been added. To quickly list them and provide pictures:
A shack has been added within the vanguard gates, to replace the one that has been missing since the major map change. (We all know what that old shack was used for, this one simply re-adds it for the same purpose) 
image
The chapel in the bog now has an astrata statue, and features a under section dedicated to noc: 
image
image

And finally, a little cliffside shack outside the vanguard gate, infested by deadites. 
image

## Why It's Good For The Game
More reasons to go to the bog, Minor map additions,

### DIFF
--- a/_maps/map_files/rockhill/rockhill.dmm
+++ b/_maps/map_files/rockhill/rockhill.dmm
@@ -4456,9 +4456,9 @@
 /turf/open/floor/rogue/carpet,
 /area/rogue/indoors/town/manor)
 "dyf" = (
-/obj/structure/bars,
-/turf/open/floor/rogue/ruinedwood,
-/area/rogue/outdoors/mountains)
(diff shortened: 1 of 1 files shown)


Gyrandola (approved):

Gyrandola:
The supposes secret entrances are also very easy to cheese

PR #1829, opened 2024-12-06
Shunned races can join as barkeeper

SayonaraEcho:
## About The Pull Request

See the title.
image

## Why It's Good For The Game

It's kinda strange that you can't join as barkeeper if you're shunned. There are a lot of roles like: 
Templar, acolyte, merchant, guild smith, artificer, cook. Barkeeper is yomen too. 

More people will join in round as barkeeper.


Hmeister-real:
If merchants can be shunned races, I don't see why barkeeps can't either

Gyrandola (approved):

PR #1831, opened 2024-12-06
More integrity for freezing sabre

Heheniy:
## About The Pull Request
max_integrity 150 -> 200

## Why It's Good For The Game
150 is too low, it can break after fighting 3 skeletons and there is no way to replace it
even daggers have bigger integrity


RadiantFlash7:
The main reason the saber can break is fighting mace wielding skeletons. Maced cause a LOT of integrity damage when parrying. Same with a lot of weapons against maces truly.

Gyrandola:
Mhmmmm.. isn't the entire point of this magic sabre that it's fragile?

PR #1845, opened 2024-12-07
Disarming the Architect

notafurryy:
## About The Pull Request

Removes the blunderbuss and the weapon accessories related to it, gives the architect lockpicking instead, and a backpack!

## Why It's Good For The Game

Removes the remnants left over from the gunsmithing part of the architect making the role less combat focused, unlike previously where it sacrificed storage which is essential for a builder the architect is supposed to be in favor of the blunderbuss.


Gyrandola:
We want to empower towner roles

PR #1854, opened 2024-12-08
Fixes snakebite post-nerf

SpaceMelon123:
## About The Pull Request

Previously, snakebite buffed your stamina via a trait. However that trait is now removed, but the massive stamina nuke that completely destroys your entire green and blue bar when it wears off still remains.

Given how much more useless the drug is now, there is absolutely no need for such a massive downside. The drug already reduces some stats, blinds you, harms you when used, and kills you if used with other drugs. It doesn't need to be any more downsides.

To make it even worse, for some reason it was coded so you get hit by it twice?

Anyway, this PR just removes the stamina / fatigue nuke.

## Why It's Good For The Game

Fixes an oversight from the nerf.

### DIFF
--- a/modular/Smoker/code/Smoker.dm
+++ b/modular/Smoker/code/Smoker.dm
@@ -76,12 +76,6 @@
 	alpha = 255
 	show_when_dead = FALSE
 
-/datum/reagent/drug/snekbt/on_mob_end_metabolize(mob/living/M)
-	M.rogstam_add(-2500)
-	M.rogfat_add(2500) // crash you fucking junkie.
-	..()
-	. = 1
-
 /atom/movable/screen/fullscreen/bittensm/Initialize()
 	. = ..()
 	filters += filter(type="angular_blur",x=5,y=5,size=1)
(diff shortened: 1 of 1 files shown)


Gyrandola:
so, what does snakebite actually still do?

SpaceMelon123:
> so, what does snakebite actually still do?

TRAIT_PROSOPAGNOSIA
^ does nothing

TRAIT_SCHIZO_AMBIENCE
^does nothing

TRAIT_NOLIMBDISABLE
^this is broken and doesn't actually work properly a lot of the time

TRAIT_NOPAINSTUN
^Ozium already does this

TRAIT_ANTIMAGIC
^The only unique thing

The downsides are it:

- Reduces your vision to 3-4 spaces
- Hurts you in a way Red doesn't heal
- Cannot be taken with other drugs like moondust
- Adds stress
- Nukes your entire stamina and fatigue bar (twice).

This PR simply removes the stamina nukes that basically means you are completely fucked when it wears off, which is now unnecessary with how much it got nerfed.

Gyrandola (approved):

PR #1861, opened 2024-12-09
[Port] Some QoL from Azure Peak again (Armour coverage examine, better colours for Rockhill

SayonaraEcho:
## About The Pull Request

1. Added coverage to armor inspect text. https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/1008

image

2. "Adjusts the hexcodes of the heraldry colors, as well as those in the colors.dm of roguetown, to fit better visually. No more piss-yellow carpets in the Keep. Plus they're better on the eyes." - [MortoSasye](https://github.com/MortoSasye) https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/995

image

3. Speeds up bait deploying time to 2 seconds.
"So, apparently using baits probably takes as long as it does because in OG roguecode there was a planned hunting skill that never saw the light of day that presumably would influence the speed of deploying/retrieving bait. This skill was never really implemented, so all we have are very inconveniently lengthy bait usage times. This long usage time doesn't really add anything to the gameplay experience at all and really shouldn't be as long as it is anyway, so let's just speed it up." - [Tenryumi](https://github.com/Tenryumi) https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/968

**_Also, someone deleted/broke an ability to examine armor durability coifs. I'll try fix it soon_**

### DIFF
--- a/code/__DEFINES/colors.dm
+++ b/code/__DEFINES/colors.dm
@@ -61,19 +61,19 @@
 
 
 //roguetown
-#define CLOTHING_RED			"#a32121"
-#define CLOTHING_PURPLE			"#8747b1"
+#define CLOTHING_RED			"#8b2323"
+#define CLOTHING_PURPLE			"#764b8d"
+#define CLOTHING_BLACK			"#2b292e"
 #define CLOTHING_VANGUARD		"#2e2931"
 #define CLOTHING_VANGUARD_DETAIL		"#262529"
-#define CLOTHING_BLACK			"#414143"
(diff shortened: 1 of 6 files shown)


Gyrandola (approved):

Elli-Skala:
> "So, apparently using baits probably takes as long as it does because in OG roguecode there was a planned hunting skill that never saw the light of day that presumably would influence

But we did this

PR #1863, opened 2024-12-09
Vamp undead faction and better telepathy message

Heheniy:
## About The Pull Request
Gives vamp undead faction (because they are undead) so zombies and skeletons wont attack them, which is annoying
image
Also makes telepathy message a bit better
image

## Why It's Good For The Game
it is not


Gyrandola:
I hope the faction thing doesnt break anything

PR #1858, opened 2024-12-09
Gives morticians novice sewing

wow50000:
## About The Pull Request
See title

## Why It's Good For The Game
"Morticians should have Novice Sewing due to the fact they use Wrapping Sheets for ease of transport/respecting the dead. I ask that someone please and thank you PR this. 

As a side thing, I believe they should POSSIBLY also be able to snap heads back onto bodies for the coin thing, buuut yknow, I primarily want the Novice Sewing thingy. "
Requested on discord by a mortician player.
https://discord.com/channels/1240735983276654733/1258555111005880520/1315466884640145480

PR #1867, opened 2024-12-09
More mortician buffs

wow50000:
## About The Pull Request
Gives morticians expert treatment, on par with life cleric

## Why It's Good For The Game
Allows morticians to re sew heads and clean up bodies before burial. Requested in https://github.com/Rotwood-Vale/Ratwood-Keep/pull/1858#issuecomment-2527139706


john-pawson:
Won't this end up on morticians acting as acolytes instead of, well, morticians?
I get the whole re-sew the head but you can simply give them the re-attachment spell if that's the only reason for it.

PR #1875, opened 2024-12-11
This PR Grants: Armpit related interactions based on existing feet ones

LeonxnoeL:
## About The Pull Request

Adds four new interactions involving armpits, based on the existing feet interactions.

Key features:
- These interactions, along with the modified nuzzle armpit one, check for the accessibility of the arms instead of the chest.
- I believe the point above might allow these actions to be performed on some clothed individuals.
- Seelies cannot force anyone on their armpits but they can be interacted with, including now the ability to be nuzzled as well.

## Why It's Good For The Game

With the addition of these interactions, the game's current selection of foreplay actions in the sex system is enriched, with inspiration taken from the existing feet related actions that served as an example of what fits into the immersive experience and overall design of the game.


ophaq:
Zased.

Gyrandola (approved):

PR #1886, opened 2024-12-12
Rework tiefling accent and add the ability to select accents

Vladmar11:
Added the ability to select any accent or remove it.
Changed tiefling accent, helped by SayonaraEcho.

image
image
image


ophaq:
This is awesome. Next we should port over families... Maybe I'll figure it out if no one else does.

Gyrandola:
> This is awesome. Next we should port over vanderlin families... Maybe I'll figure it out if no one else does.

vanderlin has families? wowza

Gyrandola (approved):

PR #1892, opened 2024-12-13
Slight Fisher polearm skill buff and knife nerf

JohnnyJohnJohnson:
## About The Pull Request

I took one point out of knife skills and put it into polearms. That's it.

## Why It's Good For The Game

Some staff didn't like a common Towner/Pilgrim class having expert-level skills with a weapon, but I figured instead of just nerfing it I'd move the point into another weapon that made sense for the class. Fishermen would be good at harpooning larger aquatic beasts, and it lets them be assembled into a spearman role if they join an angry mob or are drafted into a militia. Very few roles in general are good with polearms, but this one makes sense to be.


Heheniy:
it makes 0 sense
fisherman uses fishing rod, not harpoon

JohnnyJohnJohnson:
> it makes 0 sense fisherman uses fishing rod, not harpoon

Funnily enough, several people have told me that using the fishing rod as a weapon uses polearm skills lol

Also you're acting like because there is no mechanic in-game for spearfishing or a harpoon item (yet) that these activities aren't something fishermen would do in this setting.

Gyrandola (approved):

PR #1896, opened 2024-12-13
Architect is back to keep

Heheniy:
## About The Pull Request
Moves architect back into keep and gives him his own little room and workplace

## Why It's Good For The Game
Architect was not designet to be another town role. He is NOT a mason. Artificers are combined old masons and carpenters. Why would we need redundant town role which is literally artificer + a little bit of blacksmith but with a gun and bit fuck-you drill (both ow which shouldnt be roundstart given to any towner (and drill shouldnt be given to towners AT ALL))?

### DIFF
--- a/_maps/map_files/rockhill/rockhill.dmm
+++ b/_maps/map_files/rockhill/rockhill.dmm
@@ -4291,6 +4291,9 @@
 /obj/item/key/walls,
 /obj/item/key/walls,
 /obj/item/key/walls,
+/obj/item/key/architect,
+/obj/item/key/archive,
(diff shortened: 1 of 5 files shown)


Gyrandola:
You knew this is something I would reject since it is my wish that the town gets more content.

If your argument is that the architect has too many skills, which is a fair one, then nerf him.
The drill might be strong but it is also a unique item for a 1 slot towner job.

Heheniy:
more redundant jobs for town != more content

Gyrandola:
You should have removed mason long ago then

PR #1902, opened 2024-12-14
Blacksmiths can make Locks + Keys again

SunriseOverYourHead:
## About The Pull Request

I think that the Artificer is a fantastic role for obtaining unique, late-game, high-end recipes. He is also a role that facilitates obtaining certain items through easier, lower-difficulty (Either in material or skill-level) recipes. Locks and keys are, however, not high-end items. 

By making them dependent on the Artificer, not only it raises the difficulty to obtain them **once** since you now require a **great forge** to make Bronze, but **twice** by making them dependent on an alloy of copper and tin which means twice the amount of mining to get the minimum amount of ore to make bronze, without factoring RNG. With how easy doors are broken, and with how trivial they are to lockpick, locks and keys should be plentyful and easy for any skilled artisan, such as a smith, to make.

**This PR increases the amount of keys and locks an Artificer can make, and also gives Smiths once again the ability to make locks and keys from iron.**

## Why It's Good For The Game

It's important to allow builder and non-town roles easy access to locks and keys. This allows players to both build and secure forts, houses and towers, and also rewards players who are willing to take the time to build elaborate defensive schemes. If you can not get easy access to locks and keys, especially in roles outside of the keep, what point is there in building a fort if people can just walk in? Therefore, having locks and keys be more accessible is a net-benefit in terms of building and fort-making. **They should not be gatekept behind a GREATFORGE recipe.**

Frankly, it makes sense that the artificer would be a better role 'lore-wise' to make locks and keys. But the reality is that there is very seldom an artificer playing AND willing to make locks and keys. This issue becomes even worse once you consider that lock and key production is essentially only possible by a role of the keep. While you could argue that making a great forge for bronze is doable by mining and getting a riddle of steel, this should not be the route that players take to obtain a measly set of locks and keys.

The reality is that the Smith is a role much better suited for the production of those simple metallic trinkets. It is not entirely out of the question that a smith would know how to make rudimentary locks and keys, especially medieval ones.

**Overall, this is a good addition to the game because it allows carpenters, masons, pilgrims, towners and bandits to build secure houses and rooms without needing a greatforge, giving incentive to build.**

## An Even Better Alternative:

I think that an even better alternative would be to allow for tin and bronze to be ground together in a special crafting recipe to generate 'Bronze Dust'. This PR fixes the issue of locks and keys being completely unobtainable on rounds without an anrtificer, but cogs and other such items that can be used to make gates and levers are still impossible to access for anyone except the keep. (And let's be honest, if only the artificer can make those things due to the skill + greatforge combo, we'll only see engineering recipes being used once in a blue moon.)

Alternatively, we could switch to copper and/or tin gears, which are metals whose only current purpose is making bronze. But either way this is outside of the scope of this PR, which focuses on locks and keys as a glaring issue of the current recipe scheme.

### DIFF
--- a/code/modules/roguetown/roguejobs/artificer/artificer recipes/artificer_recipes.dm
+++ b/code/modules/roguetown/roguejobs/artificer/artificer recipes/artificer_recipes.dm
@@ -109,14 +109,14 @@
 	skill_level = 1
 
 /datum/artificer_recipe/bronze/locks
-	name = "Lock"
-	created_item = /obj/item/customlock
+	name = "Locks 5x"
+	created_item = list(/obj/item/customlock, /obj/item/customlock, /obj/item/customlock, /obj/item/customlock, /obj/item/customlock)
 	hammers_per_item = 5
 	skill_level = 1
 
 /datum/artificer_recipe/bronze/keys
-	name = "Keys 2x"
-	created_item = list(/obj/item/key_custom_blank, /obj/item/key_custom_blank)
+	name = "Keys 5x"
+	created_item = list(/obj/item/key_custom_blank, /obj/item/key_custom_blank, /obj/item/key_custom_blank, /obj/item/key_custom_blank, /obj/item/key_custom_blank)
 	hammers_per_item = 5
 	skill_level = 1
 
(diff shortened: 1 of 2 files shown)


Gyrandola:
You should make the iron recipe give 2 locks / keys and the bronze 5, so its not useless

SunriseOverYourHead:
> You should make the iron recipe give 2 locks / keys and the bronze 5, so its not useless

Done. Commits squashed.

Gyrandola (approved):

wow50000:
Could you take a look at making arrows smithed again ?

PR #1906, opened 2024-12-14
Dumb nobles getting captured

Rayforx:
## About The Pull Request
Adds a special trait that makes Heirs and Hands captured at round start.

Regarding gameplay, A noble gets stripped with only their keys and pouch and put in the bandit camp. Only the Heir, Hand and Courtier can be this special so it only has a HUGE gameplay negative to the loop.

Video of the current code

https://github.com/user-attachments/assets/5d586f6f-dde1-4452-8ddd-f521eb868aed

## Why It's Good For The Game
It's pretty funny. And it could open up for some roleplay at roundstart.

### DIFF
--- a/code/datums/special_traits/traits/traits.dm
+++ b/code/datums/special_traits/traits/traits.dm
@@ -446,9 +446,9 @@
 
 /datum/special_trait/captured_noble
 	name = "Captured Noble"
-	req_text = "Be a heir or hand."
-	greet_text = span_boldwarning("You have been captured by foul bandits!")
-	allowed_jobs = list(/datum/job/roguetown/prince, /datum/job/roguetown/hand)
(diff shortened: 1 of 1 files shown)


NikolayByondPlayer:
Good shit, also is map dif broken?

Heheniy:
but what if there is no bandits

Rayforx:
> but what if there is no bandits

Then, they must wait for bandits to spawn. A minimum of four bandits are always able to join.

Gyrandola:
Shouldnt the keep be aware someone has been kidnapped / is missing?
Else they will assume nobody rolled the role in the first place

Rayforx:
It was a funny idea, but we probably have to scratch it for now.

PR #1924, opened 2024-12-15
Buffs Barkeep & Cook Jobs

ophaq:
## About The Pull Request

Gives the barkeep a satchel and cloth for cleaning, and the cook gets a few extra item and skills when starting out to help them with their roles. 

## Why It's Good For The Game

The barkeep, who makes a lot of money, should have a satchel to store things if they need to.

Additionally, the cook is now competitive with the cheesemaker towner class (which is a glorified cook), but without the starting cheese and less skills thank cheesemaker has still. The main skill changes being fist fighting and wrestling to act as a bouncer in the kitchen if needed. The cook also gets a little bit of farming like the cheesemaker has, and also gets a satchel with a spare flint and cloth in it (for cheese and cleaning like the barkeep). The cook also has a hunting knife on their belt now too, which will help in making signs if they want to make a menu board, or just assist in cutting things if they lose their cleavers. Although the butcher knife still functions similarly to the hunting knife, there are things it cannot do, and good cooks always have a variety of knives at their disposal.

If any of these changes are too controversial, let me know.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/peasants/cook.dm
+++ b/code/modules/jobs/job_types/roguetown/peasants/cook.dm
@@ -24,8 +24,6 @@
 		H.mind.adjust_skillrank(/datum/skill/craft/crafting, 1, TRUE)
 		H.mind.adjust_skillrank(/datum/skill/misc/sewing, 1, TRUE)
 		H.mind.adjust_skillrank(/datum/skill/craft/hunting, 2, TRUE)
-		H.mind.adjust_skillrank(/datum/skill/combat/wrestling, 3, TRUE)
-		H.mind.adjust_skillrank(/datum/skill/combat/unarmed, 2, TRUE)
 		H.mind.adjust_skillrank(/datum/skill/misc/athletics, 2, TRUE)
 		H.mind.adjust_skillrank(/datum/skill/labor/farming, 1, TRUE) 
 		if(H.age == AGE_OLD)


Gyrandola:
why does every class need to be a fucking wrestler?
Enough powercreep

ophaq:
> why does every class need to be a fucking wrestler? Enough powercreep

Is every class a wrestler? Tbh I didn't notice. IIRC barkeep has it and traditionally cook always was bouncer in the kitchen. I can remove it from barkeep, or both even.

Gyrandola:
> > why does every class need to be a fucking wrestler? Enough powercreep
> 
> Is every class a wrestler? Tbh I didn't notice. IIRC barkeep has it and traditionally cook always was bouncer in the kitchen. I can remove it from the cook, maybe even the barkeep too if that's something that would be more balanced.

I can see the bartender having it, but the cook? A labor job should not justify wrestling skills or every construction worker would be a martial artist

> Also if there are a lot of classes getting to be a wrestler, would it be better if we just stopped doing that and start stripping most of the classes of wrestling so that people who do wrestle as a profession are more valuable?

Yeah.

ophaq:
> > > why does every class need to be a fucking wrestler? Enough powercreep
> > 
> > 
> > Is every class a wrestler? Tbh I didn't notice. IIRC barkeep has it and traditionally cook always was bouncer in the kitchen. I can remove it from the cook, maybe even the barkeep too if that's something that would be more balanced.
> 
> I can see the bartender having it, but the cook? A labor job should not justify wrestling skills or every construction worker would be a martial artist
> 
> > Also if there are a lot of classes getting to be a wrestler, would it be better if we just stopped doing that and start stripping most of the classes of wrestling so that people who do wrestle as a profession are more valuable?
> 
> Yeah.

I'll take the wrestling/fighting off the cook then. Maybe I will strip a lot of classes with the kung fu later on.

Gyrandola (approved):

PR #1937, opened 2024-12-16
Disables latejoin goblins

Heheniy:
## About The Pull Request
Disables latejoin goblins

## Why It's Good For The Game
Goblins have no their own latejoin spawnpoint, making them spawn at harbor on latejoin, which should not occure.
Some people even starts to thing they are "town goblins" (no, they arent)


ophaq:
Wait, we can finally allow late join goblins? We gotta keep this.

Koteshkin:
Why disallow them to latejoin instead of just adding them a spawn point within the camp?

Gyrandola:
What carl says + what EvenInDeath says

PR #1938, opened 2024-12-16
Silver Cutlery WW+Vampire interaction

SunriseOverYourHead:
## About The Pull Request

Silver cutlery can no longer be picked up by V.lords and transformed werewolves, and applies typical debuffs and effects when held, bashed with, or equipped.

Being bashed by those non-combat silver items will only incur a small amount of burn damage and a minimal amount of fire stacks, unlike silver daggers, which apply a proper stun.

I have also noticed that the elvish daggers and sabers, despite being made with silver, are smelted into steel and do not apply silver effects into mobs. I will fix that in a later PR.

## Why It's Good For The Game

Upholds silver item lore against WWs and Vampires. I figure that the funny-attack-effects function may need tweaking in the future so people don't go hunting vampires with silver spoons.

### DIFF
--- a/code/game/objects/items/rogueitems/cup.dm
+++ b/code/game/objects/items/rogueitems/cup.dm
@@ -66,8 +66,6 @@
 		var/mob/living/carbon/human/H = target
 		if(H.dna && H.dna.species)
 			if(istype(H.dna.species, /datum/species/werewolf))
-				H.Paralyze(10)
-				H.Stun(10)
 				H.adjustFireLoss(10)
 				H.fire_act(1,10)
 				to_chat(H, span_userdanger("I'm hit with my BANE!"))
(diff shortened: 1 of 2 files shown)


ophaq:
Finally, I can stab a werevolf with a fork to kill it.

radarseas:
I'd make it not stun when attacking with. also the vampire lord's manor needs a rework so it doesn't spawn with a goblet.


_then again the idea the goblet has been there for centuries because the lord can't physically move it is funny._

Gyrandola:
Like the idea but this is really bad for balance.

radarseas:
> Still, that isn't really a justification not to change it.

goblets are few and far between, in comparison to smaller items like knives and forks, that's the issue. If a vampire PICKs one up that's 1000% on them and they deserve it, but anything else shouldn't do anything used as a weapon or thrown, unless its as large as a platter, in which case setting on fire at most?

NikolayByondPlayer:
Goblets are rare though, seems fun

SunriseOverYourHead:
> Like the idea but this is really bad for balance.

On-hit effect changed and description updated.

It no longer stuns but applies a small amount of burn damage and fire stacks. Is this agreeable? Let me know.

Gyrandola (approved):

JohnnyJohnJohnson:
I'm gonna stick a fork in the vampire lord

PR #1944, opened 2024-12-16
Scholar cant do alchemy fix + bigger teaching range

AukariBunn:
Recently during the playtests i found out the scholar cant brew potions since they get apprentice treatment (alchemy require journeyman treatment). So i increased the skill level by one. Also added cooking.

Code_rmIvyKh1td

Some people asked me to increase the teaching range from 1 to 2 so they could use the tutoring ability while sitting at a table with their student.

Code_QFVfehRGiG

Also made some minor changes to the map:
Added a separate chest for the spare archive keys
StrongDMM_7GQabJTQID
Moved the window a bit to make it more logical
StrongDMM_eSA7LYGxiL
Also added a hunting knife into the kitchen. Adding a whole cleaver would be too much, while having to make lots of stone knives every round to chop some meat for red isnt an option either. The iron dagger which the scholar gets when spawns cannot chop.

### DIFF
--- a/_maps/map_files/rockhill/rockhill.dmm
+++ b/_maps/map_files/rockhill/rockhill.dmm
@@ -1711,9 +1711,6 @@
 /obj/item/clothing/shoes/roguetown/sandals,
 /obj/item/clothing/under/roguetown/tights/stockings/silk/random,
 /obj/item/clothing/under/roguetown/tights/stockings/silk,
-/obj/item/key/archive,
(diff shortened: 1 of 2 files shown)


AukariBunn:
This is not an urgent matter
If you find any errors during the game report them to me immideately, i will be working on another update after this PR, so the next fixes if needed will take a plenty of time.

Gyrandola (approved):

PR #1947, opened 2024-12-17
Clerk Keys

ProbablyCarl:
## About The Pull Request
Removes the vestigial Clerk keyring and provides them the Steward's keyring instead. The only difference is that it has a Steward key in it, but is otherwise identical. This is because the chests with the spare that spawns in roundstart is now locked, and the Clerk is shafted into being permanently stuck without a Steward or other manner of intervention.

This means, if this PR goes through, that the following keys exist going forward.
 - The Hand's copy.
 - The Steward's copy.
 - The Clerk's copy.(New.)
 - The Spare copy.(Previously taken by the Clerk anyway.)

## Why It's Good For The Game
Clerk being stuck in the tower is cringe. They already took the spare anyway, atop that.


Gyrandola (approved):

Tastyfish:
Thank you for the quick fix!

PR #1948, opened 2024-12-17
Fixes Elvish Weapons Not Having Silver Effects

SunriseOverYourHead:
## About The Pull Request

Elvish weapons are made using silver. In spite of this, the Elvish Saber and the Elvish Dagger were BOTH smelted into steel AND failed to apply the silver effect that weapons made out of silver apply on werewolves and vampires.

This PR fixes this and allows elvish sabers and elvish daggers to apply the silver debuff to those antagonists.

## Why It's Good For The Game

This PR improves item consistency, as those two weapons made out of silver will now properly be smelted into silver.
Moreover, this also makes them powerful against Werewolves and Vampires, as silver weapons should be.

I believe that not having those effects was an oversight when elf weapons were added. Let me know if this is intentional, though.

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/knives.dm
+++ b/code/game/objects/items/rogueweapons/melee/knives.dm
@@ -365,8 +365,115 @@
 	force = 19
 	icon_state = "elfdagger"
 	item_state = "elfdag"
+	smeltresult = /obj/item/ingot/silver
 	can_cdg = TRUE
 	can_assin = TRUE
+	var/last_used = 0
+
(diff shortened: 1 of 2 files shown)


john-pawson:
I remember a PR being before that fixed this exact problem but i can't find it anywhere, so i don't know if someone intentionally removed it after that.

Either way, i think this is a good change. Why wouldn't silver weapons have the properties of silver?
Nice PR

Gyrandola (approved):

PR #1950, opened 2024-12-17
Restores Crit Protection on Adjustable Armor Pieces

Xaytan:
## About The Pull Request

ResetAdjust was causing any adjustable armor pieces to lose their crit protection values when being picked up, put down, equipped, spawned in an equip slot, or really interacted with in any way beyond examining them.

This was causing runtimes when checking protection values of armor, and causing a number of armor items to not provide their intended crit protection. This list includes:

Coif
Chain Coif
Steel Chain Coif
Knight's Helmet
Visored Sallet
Templar Helmets with adjustable visors
Any other armor items that are adjustable, these are just the list I found while troubleshooting.

Currently no adjusts actually change the prevent_crits list of any armor. If somebody adds an armor piece that does that, well, we can make it their problem to implement it.

## Why It's Good For The Game

It's a bugfix. Less runtimes, more things doing what they are intended to do. Also, knight's helmets will now properly protect against maces. As they are intended to do.

image
image

image
image

### DIFF
--- a/code/modules/clothing/suits/toggles.dm
+++ b/code/modules/clothing/suits/toggles.dm
@@ -45,7 +45,6 @@
 	flags_cover = initial(flags_cover)
 	block2add = initial(block2add)
 	body_parts_covered = initial(body_parts_covered)
-	prevent_crits = initial(prevent_crits)
 
 /obj/item/clothing/equipped(mob/user, slot)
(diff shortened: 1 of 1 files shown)


Xaytan:
Oops.

Gyrandola (approved):

PR #1951, opened 2024-12-17
Smash Protection Proliferation

Xaytan:
## About The Pull Request

Adds smash crit protection to the padded gambeson, arming jacket and padded cap.

## Why It's Good For The Game

Gambesons are currently overlooked pretty heavily, as are padded caps.

This change creates a pretty strong potential rivalry between the padded gambeson and the hauberk, and gives padded caps a niche as a way to prevent smashing weapon crits (Skullcracks) while using light or medium helmets.

This allows for the trade off of face and eye protection in exchange for more resistance to being crit by blunt weapons like maces while wearing medium helmets that don't provide smash protection of their own, or trading the otherwise objectively better hauberk for the safety of knowing your bones won't get snapped like twigs by a guy with a mace.

Meanwhile, a gambeson does nothing to protect from assassination and coup de grace crits while a hauberk does, and the gambeson is far less protective overall.

Mace supremacy is pretty heavy right now, and there are clear meta picks for all armor slots. This adds new viable choices and some give and take to armor selection. This will hopefully both make these more popular options, and perhaps lead to diversification of weapon choices since blunt instruments will no longer be a one-size-frags-all pick.

Padded caps, due to their incredibly low durability, will still be basically useless on their own. They will only really help when combined with a real helmet. Thanks to the way durability damage propagates, the padding will save you from smash crits until your real helmet breaks then be promptly obliterated by the first couple of blows that get past it.

image
image

### DIFF
--- a/code/modules/clothing/rogueclothes/armor.dm
+++ b/code/modules/clothing/rogueclothes/armor.dm
@@ -352,7 +352,7 @@
 	name = "padded gambeson"
 	desc = "A gambeson with additional padding layers, hardened to make it more durable. It still cannot compare to proper armor."
 	icon_state = "gambesonp"
-	prevent_crits = list(BCLASS_CUT, BCLASS_BLUNT, BCLASS_TWIST)
+	prevent_crits = list(BCLASS_CUT, BCLASS_BLUNT, BCLASS_TWIST, BCLASS_SMASH)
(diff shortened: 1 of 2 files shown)


Xaytan:
Also, you can still absolutely beat the shit out of somebody with a padded gambeson + full plate, you'll pain crit them from all the damage going through even though they aren't dying and don't have any broken bones.

Crowbar764:
Two smash intent nerf PRs at the same time?

Really?

Gyrandola (approved):

PR #1952, opened 2024-12-17
Re-adjusts Tin and Copper ore price on Stockpile.

SunriseOverYourHead:
## About The Pull Request

Playing for a few rounds I realized that I probably made Copper too pricey. Copper rings are not that expensive, and Artificer players were having trouble getting the mammon to get copper for Bronze.

This PR makes both Tin and Copper a bit more expensive than iron, still giving smiths a net-positive when used to make valuable items, but not so expensive that other roles can not buy them at all.

## Why It's Good For The Game

Makes the balance of copper in the economy a bit more saner now that I've had some experience seeing how its being used in play. This also makes miners not get instantly rich when hitting veins of copper and selling to the stockpile.

Meanwhile, this also makes roles that require copper like Architect and Artificer be able to get them more easily. Copper's price was massively overinflated for what it offered. Tin's was a bit too low with the new Pewter ware.

Hopefully this will be the last price edit that I need to do for a while.

### DIFF
--- a/code/modules/roguetown/roguestock/stockpile.dm
+++ b/code/modules/roguetown/roguestock/stockpile.dm
@@ -62,10 +62,10 @@
 	desc = "Chunks of copper used for smithing and alloying."
 	item_type = /obj/item/rogueore/copper
 	held_items = list(6, 0)
-	payout_price = 20
-	withdraw_price = 30
-	transport_fee = 5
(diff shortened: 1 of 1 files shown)


Gyrandola (approved):

Elli-Skala:
You forgot copper bars.

PR #1957, opened 2024-12-17
Adding buyable shops

NikolayByondPlayer:
## About The Pull Request
1)Added buyable shop, made so smith shop on the right would be bought with coins.
Also nuked stairs to the flat above, want to test  shops first then we can add it back for town smiths depending on how people use place following weeks.
image
image

2) Removed locks from graveyard since it rarely used

## Why It's Good For The Game

Want to test if people would actually use these shops, price for shops is only 5 let's leave it at that it can be adjusted later. Main thing is to test idea. 
If not I will choose dif role for this road. Right now I want to make it a shop street.

### DIFF
--- a/_maps/map_files/rockhill/rockhill.dmm
+++ b/_maps/map_files/rockhill/rockhill.dmm
@@ -2007,8 +2007,10 @@
 /turf/open/floor/rogue/concrete,
 /area/rogue/under/town/basement)
 "blx" = (
-/obj/structure/chair/stool/rogue,
-/turf/open/floor/rogue/blocks/stonered,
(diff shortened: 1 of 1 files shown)


wow50000:
Not a fan of the houses being buyable, assuming anyone can buy it. It removes another reason to go towner smith.

NikolayByondPlayer:
> Not a fan of the houses being buyable, assuming anyone can buy it. It removes another reason to go towner smith.

I will probably add additional town smithery after I move garrison anyways

Gyrandola (approved):

PR #1973, opened 2024-12-18
Pickaxe for back

SayonaraEcho:
## About The Pull Request

This change allows the pickaxe to be carried on the back.

image

## Why It's Good For The Game

Well, you can carry an axe on back or belt but you can't carry a pickaxe like that.

QoL for miners


EvenInDeathIStillServe:
Did you check how this looks ingame? Back slots use experimental icon generation and I'd like to confirm whether this actually looks _good_.

PR #1976, opened 2024-12-18
Rip speedsters

Purpleturtle3:
## About The Pull Request

TLDR: Now you only get speed maluses if below a certain speed; if you are above 11, you will get less movespeed. You now need 17 speed to reach the cap of movement speed.

## Why It's Good For The Game

Speed is super op stat, we all know it, the difference between having 9 and 11 was incredibly noticeable.

### DIFF
--- a/code/modules/mob/living/living_movement.dm
+++ b/code/modules/mob/living/living_movement.dm
@@ -44,24 +44,32 @@
 	return ..()
 
 /mob/living/proc/update_move_intent_slowdown()
-	var/mod = 0
-	switch(m_intent)
-		if(MOVE_INTENT_WALK)
-			mod = CONFIG_GET(number/movedelay/walk_delay)
-		if(MOVE_INTENT_RUN)
-			mod = CONFIG_GET(number/movedelay/run_delay)
-		if(MOVE_INTENT_SNEAK)
-			mod = 6
+    var/mod = 0
+    switch(m_intent)
+        if(MOVE_INTENT_WALK)
+            mod = CONFIG_GET(number/movedelay/walk_delay)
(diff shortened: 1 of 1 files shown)


Xaytan:
9 vs 11 speed _should_ be noticable. That's a two point difference, and the middle aged humen with -1 should be clearly walking slower than the young half elf he whips the hell out of on strength and endurance.

Hmeister-real:
> Seriously if you have 13-14 speed and you somehow get caught you're straight up unrobust.

Yeah my bad I tried to RP and got grabbed from behind, that means I'm not a gamer. This is nothing but a shitty cope PR, lmao.

Purpleturtle3:
> > Seriously if you have 13-14 speed and you somehow get caught you're straight up unrobust.
> 
> Yeah my bad I tried to RP and got grabbed from behind, that means I'm not a gamer. This is nothing but a shitty cope PR, lmao.

Just turn on combat mode so you can automatically dodge people you can't see.

notafurryy:
> > > Seriously if you have 13-14 speed and you somehow get caught you're straight up unrobust.
> > 
> > 
> > Yeah my bad I tried to RP and got grabbed from behind, that means I'm not a gamer. This is nothing but a shitty cope PR, lmao.
> 
> Just turn on combat mode so you can automatically dodge people you can't see.

that still tears through your stamina, and most roles have high endurance so odds are they can catch up to you and grab you as you run out of breath

RadiantFlash7:
Ehhh. This is a massive buff to knights veterans and other roles that have giant stat bonuses but speed penalties. Not a fan.

Purpleturtle3:
Unrobust speedsters win this dae...

PR #1977, opened 2024-12-18
Backend Lesser Raise Undead Spell + Part of Prep For Patron Spelllists

Kevak:
## About The Pull Request

This is pretty much entirely backend fiddling. Made a lesser raise undead spell that caps sapient skeletons to current mage-arcana. This means max skeletons of 6 at legendary skill, less at lower skill. Those with it can still raise infinite nonsapient skeletons, but those are waaay less threatening. Only 6 sapients max if a mage with this really grinds for higher arcana.

Additionally, if the lesser necromancer is out of slots for sapient skeletons and tries raising a body with a dude in it, it will send that dude straight to the underworld, skip the wait.

## Why It's Good For The Game

Occultist needs to not have Infinite Skeleton Works but having a necromancer about is still fun. Capping it to arcana skill feels like a good compromise. Also some backend fiddles in prep for patronlocked additions to the spell list.

## Proof Of Compile
I forgot to record testing and this is p much all backend work and I don't want to wait 10 minutes for the server to boot again to grab screencaps of variables switching so here's proof of compile.

Screenshot 2024-12-18 113609

### DIFF
--- a/code/datums/mind.dm
+++ b/code/datums/mind.dm
@@ -48,6 +48,12 @@
 	var/spell_points
 	var/used_spell_points
 
+	//Lesser Necromancy Trackers
+	var/boneboys = 0
+	var/bonemax = 0
+	var/boneboy = FALSE
+	var/bonenecro = null
+
 	var/linglink
 	var/datum/martial_art/martial_art
 	var/static/default_martial_art = new/datum/martial_art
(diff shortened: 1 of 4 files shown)


Kevak:
Pushing this to refugees

Issue #953, opened 2024-09-09
[BUG] Dungeon surgery room


K4rlox:
## Describe the bug

Dungeon surgery room has only 1 forceps, making all kinds of surgeries impossible

Physician does not have access to the room, previously being able to unlock the room, as the room doubles as a surgery room

## Round ID:
Dragon724

## Steps to Reproduce:
be physician
try to unlock dungeon
be able to enter
try to unlock surgery room
cant unlock

enter surgery room
1 forceps, cant do surgery

JagOfTroy:
So from the looks of this, I'm going to assume the bug is still in existence since there hasn't been any verification that this is a closed issue.
Will try to repro to see if Physicians have access(I feel like they do from the last round I played).

JagOfTroy:
![Screenshot 2024-12-19 132117](https://github.com/user-attachments/assets/5fa15a3a-cc96-42bd-a066-078a22d1b64b)
Can confirm, only one set of forceps in the dungeon surgery chamber. Not sure if intentional or an overlook.


![Screenshot 2024-12-19 133519](https://github.com/user-attachments/assets/b37627df-213a-43ae-a9ec-73976a575d26)
Other than that, I was able to walk from the Court Physician starting point, unlock cell doors and open the dungeon surgery room.
The only cell door I couldn't open was the one in the dungeon surgery room.
![Screenshot 2024-12-19 133809](https://github.com/user-attachments/assets/705a4577-a80d-494d-aee4-44a5dd761201)

JagOfTroy:
This an be closed now via #1985

PR #1985, opened 2024-12-19
Fix for Court Physician Dungeon Surgery

JagOfTroy:
## About The Pull Request
 
The Court Physician originally was not able to gain full access to the patients waiting for him in the dungeon surgery room, neither did he have the proper tools in order to perform surgery. With a small map edit, I included additional forceps toward the back of the room(hard to see with lack of lighting but it's there) and also changed the Cell Door to a Holding Door which allows the Physician to now also have access to the private cell room within dungeon surgery like he should have.

The only other solution here would be to give the Court Physician the key:'keep_dungeon' in addition to 'keep_barracks' which he already has but that may be slightly too advantageous and keeps the Court Physician limited this way to his areas and not just open any cell, willy-nilly.

## Why It's Good For The Game

Fix for issue #953 

It gives full functionality back to the Court Physician if they choose(and likely will) to use the dungeon surgery room.

Screenshot 2024-12-19 135624

### DIFF
--- a/_maps/map_files/rockhill/rockhill.dmm
+++ b/_maps/map_files/rockhill/rockhill.dmm
@@ -3756,6 +3756,14 @@
 /obj/structure/mineral_door/wood,
 /turf/open/floor/rogue/cobble,
 /area/rogue/indoors/town/tavern)
+"ctJ" = (
+/obj/structure/table/wood{
+	dir = 10;
+	icon_state = "tablewood2"
+	},
+/obj/item/rogueweapon/surgery/hemostat,
+/turf/open/floor/rogue/blocks,
+/area/rogue/under/town/basement)
 "cuk" = (
 /turf/open/floor/rogue/concrete,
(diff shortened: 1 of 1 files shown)


Gyrandola (approved):

K4rlox:
i tought i already fixed this issue

JagOfTroy:
> i tought i already fixed this issue in #974

For the most part, yes but not fully. While reproducing, everything seemed to work fine but the issue with the Cell Door is you need a dungeon key and the court physician currently only has the barracks key. His barracks key cannot open the cell door inside of the dungeon surgery so had to replace the cell door with holding door since it responds to the barracks key OR the only other choice would be to give the court physician the dungeon key as well which seemed unbalanced if something were to happen to the physician.

Oh.. and I just added a pair of forceps in the dungeon surgery room itself, there was still only one in there(sake of convenience and keep up pace of game.)

PR #1987, opened 2024-12-19
Removes windows from the street smith shop

wow50000:
## About The Pull Request
Removes the side windows that only existed to cause pain, re adds the funny pillar.
image
Tiles in blue are the removed windows

## Why It's Good For The Game
The funny pillar splits the shop in a nice way while the stone murder holes make the interior unrealistically exposed even when the shutters are closed.

### DIFF
--- a/_maps/map_files/rockhill/rockhill.dmm
+++ b/_maps/map_files/rockhill/rockhill.dmm
@@ -336773,10 +336773,10 @@ wbg
 wbg
 wVI
 uln
-ehT
+uln
 uMc
 lRU
-ehT
+uln
 uln
 uln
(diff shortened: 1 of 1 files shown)


NikolayByondPlayer:
On the other hand it is hard to tell if building is vacant or not right not, not sure getting rid of windows is a good idea - since they help identify if thing is occupied or not

Gyrandola (approved):

wow50000:
> On the other hand it is hard to tell if building is vacant or not right not, not sure getting rid of windows is a good idea - since they help identify if thing is occupied or not

The shutters.

PR #1994, opened 2024-12-20
Gives picklocking to thief special

Heheniy:
## About The Pull Request
see name

## Why It's Good For The Game
Picklocking was added to the game, but was forgotten to be added to the thief special while definetly being a theivery thing


notafurryy:
The special could maybe also give you a lockpick ring or just a lockpick like other specials do, also probably could bump up lockpicking to 5, it's not that strong.

Gyrandola (approved):

PR #1999, opened 2024-12-20
Un-nerfs moondust

Dedwolf:
## About The Pull Request

Adds the nextmove modifier back to moondust, meaning it will once again speed up your actions.

Moondust reduces the move cooldown by 15%
Purest moondust reduces the move cooldown by 30%

Pure moondust reducing the move cooldown by 30% is still really, really noticeable and will still have a significant effect on combat. However i don't care because the keep is far too soft and given too many benefits. 

Since moondust can't be mixed with other drugs i reckon this is still well balanced.

### DIFF
--- a/code/datums/status_effects/rogue/roguebuff.dm
+++ b/code/datums/status_effects/rogue/roguebuff.dm
@@ -116,6 +116,9 @@
 	effectedstats = list("speed" = 4, "endurance" = 4, "perception" = -3)
 	duration = 30 SECONDS
 
+/datum/status_effect/buff/moondust/nextmove_modifier()
+	return 0.15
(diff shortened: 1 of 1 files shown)


Elli-Skala:
unrelated but the change making it so you need 5 units of a drug to make it take effect has broken ozium in a few ways, most noteablely smoking it

Kevak:
Can we get something other than that abysmal epileptic screenflash for moondust?

Gyrandola:
I know this is just gonna be used by bandits for fragging better, plus click delay breaks smithing.

Im not sure why drugs have to be combat enhancers in the first place.

PR #2002, opened 2024-12-20
Goblin Race Stats

notafurryy:
## About The Pull Request

Gives goblins racial stats, making them fast resiliant gremlins rather than blank slates being mostly carried by the role.

## Why It's Good For The Game

Goblins had no racial stat modifiers making for silly stat comparisons to other roles, male goblin guards could get up to 15 str if they had a lucky roll making them have better stats than actual knights while being a third of the size.

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/roguetown/goblin/goblinp.dm
+++ b/code/modules/mob/living/carbon/human/species_types/roguetown/goblin/goblinp.dm
@@ -48,13 +48,13 @@
 		OFFSET_SHIRT_F = list(0,0), OFFSET_ARMOR_F = list(0,0), OFFSET_UNDIES = list(0,0), OFFSET_UNDIES_F = list(0,0), \
 		)
 	specstats = list(
-		"strength" = 0,
-		"perception" = 0,
-		"intelligence" = 0,
-		"constitution" = 0,
-		"endurance" = 0,
-		"speed" = 0,
-		"fortune" = 0
+		"strength" = -3, 
+		"perception" = 0, 
+		"intelligence" = -2, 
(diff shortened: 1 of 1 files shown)


john-pawson:
Love the change, would just like to point out that all other races get an overall +1, and it seems this one gets an overall +0. Might be intended for all that i know though.

notafurryy:
> Love the change, would just like to point out that all other races get an overall +1, and it seems this one gets an overall +0. Might be intended for all that i know though.

I might actually lower it further, goblins are supposed to be more of a nuisence than an actual threat I believe? And they already start with pretty powerful gear, a smith and a doctor so they already have resources to substitute raw power.

Gyrandola (approved):

john-pawson:
> > Love the change, would just like to point out that all other races get an overall +1, and it seems this one gets an overall +0. Might be intended for all that i know though.
> 
> I might actually lower it further, goblins are supposed to be more of a nuisence than an actual threat I believe? And they already start with pretty powerful gear, a smith and a doctor so they already have resources to substitute raw power.

If you plan to, lower constitution. Such little things shouldn't probably be as hard to break than an horc

Femboyms:
after this nerf I'll be gone for months again before coming back, congrats whoever managed to kill the last and only thing that still funny in the server (and I've seen many will quit after this nerf) nobody liked the update, not only werewolf and vampires got nerfed to oblivion but now goblins which were soft antags, please continue and try to nerf bandits maybe you get your extended game one day bc now antags are irrelevant, slowly becoming an ERP server

Limeon666:
Nah, they are right

Issue #1842, opened 2024-12-07
[BUG] Unexpected Behavior for Bookcases (Can't put books back)


Lord-Preston:
## Describe the bug

You cannot put books back on bookcases. I noticed this two rounds in a row, but I only tried bookcases in the Church.

## Round ID:

![Round ID](https://github.com/user-attachments/assets/26bd788d-62b8-46a0-96fd-fba36b0e446f)

## Steps to Reproduce:

1 - Take a book from a bookcase
2 - Try to put it back
3 - It won't go back
## Additional context

I tried several bookcases in the Church and none worked.
![Bookcase Issue](https://github.com/user-attachments/assets/1bac734f-cb26-4b2a-9243-9b65688146ac)

JagOfTroy:
Will attempt to reproduce to make sure it is still an issue. Then go from there!

JagOfTroy:
![Screenshot 2024-12-21 143033](https://github.com/user-attachments/assets/b6c4f5c8-f9c2-47c6-a867-0eb0383cf282)
Yep, still a bug. I can remove books from bookcases but unless there is a secret way to put them back that I don't know of yet, then this is confirmed. Will start looking into a fix.

JagOfTroy:
To put it simply after reviewing the lib_items.dm, I don't think a function was ever written to actually put books back. That's why you can't return the book to the bookcase. There's probably a good reason for this that I don't know but will still keep poking at this bug to see if I can fix it.

PR #2013, opened 2024-12-21
Bookcases now work properly - You can put books back on the book shelf.

JagOfTroy:
## About The Pull Request

The origin of the bug is that the player could remove book cases from the book case but not put them back. After reviewing the code for the bookcases, I determined that there was never a method written to put the books back. Now, I have included a method whereby if you touch the bookcase with the book in hand, it will return that book to the bookcase.

## Why It's Good For The Game

Fixes bug issue #1842 

This gives back functionality to the bookcase that it should have, instead of refusing the user from putting books back. It even returns the book to the list so that it can be grabbed by another player later on if they wish.

Below are my screenshot tests for proof of it working.

Screenshot 2024-12-21 150258
Screenshot 2024-12-21 150306
Screenshot 2024-12-21 150316
Screenshot 2024-12-21 150326
Screenshot 2024-12-21 150337

### DIFF
--- a/code/modules/library/lib_items.dm
+++ b/code/modules/library/lib_items.dm
@@ -69,6 +69,12 @@
 				choice.forceMove(drop_location())
 			update_icon()
 
+/obj/structure/bookcase/attackby(obj/item/I, mob/user, params)
+	if(!istype(I, /obj/item/book))
+		return ..()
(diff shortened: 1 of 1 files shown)


Gyrandola (approved):

Gyrandola:
bless

PR #2023, opened 2024-12-22
Bandit Smith | Hoardmaster Nerfening

ProbablyCarl:
## About The Pull Request
Provides the bandits with a dedicated smith, in the role of the Forgehand. Atop that, it significantly changes around the Hoardmaster by removing almost all class related objects, beyond the most basic required for their function. The rest can be smithed, as is the purpose of the Forgehand. The 'stuff' and 'things' categories have been completely untouched. Only class specific sections were touched, and this was, primarily, if not entirely, weapons and armour of a variety higher than the most basic for each respective class.

For example, the Knave can no longer purchase elven daggers, sabres or poisoned arrows, but he can still purchase iron daggers, swords and full quivers of bolts / arrows. Special exceptions for classes are made, like the Sawbones, who can purchase bombs and killers ice still. Or the Mage, for example, who can buy scrying orbs and null magic rings.

Finally, it provides a basic smithery roundstart. The Forgehand, through the Hoardmaster, can purchase a riddle of steel for a proper forge, should he wish. (Why wouldn't he?)
image

Unrelated to the above, but it shifts around bandit packs into their own file path so they're not just laying about, and cleans up the innards just a tad of the files.

## Why It's Good For The Game
More of a subjective change, for the Hoardmaster's contents, this time. I don't like bandits being able to just poof stuff into existence, especially with how easy it is to gain favor. IMO, at least. Favor still ABSOLUTELY has a use, but it's not nearly as round shaping as prior, with that instead falling to a player based role, the Forgehand.


ShiningM00N:
I think these are very interesting and definitely good changes. I think the only one's I'd say would be too overpowered is Scrying orb for Bandits (How come gobbos get a worse one?) and also null magic for bandits, considering the latter is both expensive at the merchant and exclusive to two roles. Otherwise, cool stuff.

ShiningM00N:
> They already had it. This doesn't add any new items, beyond what the Forehand gets, to the Categories.

This is what happens when you stop playing bandit for weeks :sob: Thanks for clarifying.

Rayforx:
> I think these are very interesting and definitely good changes. I think the only one's I'd say would be too overpowered is Scrying orb for Bandits (How come gobbos get a worse one?) and also null magic for bandits, considering the latter is both expensive at the merchant and exclusive to two roles. Otherwise, cool stuff.

The null magic ring only works for 30 seconds with a 10 min cooldown, and only the mage can buy it (someone is working on a PR that makes you not able to buy from each other tab) both items, are incredibly expensive, and rarely bought since they can't be afforded

PR #2030, opened 2024-12-23
Canopies - You Forgot To Wash Them In Six Years And Now They Stink And Are Grimdark(tm) Update

Kepteyn:
## About The Pull Request

Makes the shop canopies grimmer and darker to better fit into the environment. It also adds a blue canopy variant, which faces the side!

## Why It's Good For The Game

Useroth was not happy with the light and cheery canopies. By soaking them in a mix of dirt, shit and piss, they now look just as shitty as the rest of the town instead of sticking out as something from a Nobledark Fantasy™. A side canopy now allows for a canopy to be added to the sides of shops, rather than just vertically. They layer above players correctly too, and add an extra variety of colour to the mix. More variety!

BEFORE:
b1
AFTER:
a1

BEFORE:
b2
AFTER:
a2

BEFORE:
b3
AFTER:
a3

BEFORE:
b4
AFTER:
a4

### DIFF
--- a/_maps/map_files/rockhill/rockhill.dmm
+++ b/_maps/map_files/rockhill/rockhill.dmm
@@ -2611,6 +2611,18 @@
 /obj/structure/globe,
 /turf/open/floor/rogue/cobble/mossy,
 /area/rogue/indoors/shelter/town/harbor)
+"bDQ" = (
+/obj/structure/table/wood,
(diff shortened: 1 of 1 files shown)


Useroth (approved):
:+1:

NikolayByondPlayer (approved):

PR #2033, opened 2024-12-24
Adjusts Fish Mince In-Line with Steak Amount

ophaq:
## About The Pull Request

This shrimply makes it so you get two mince instead of one from fish now.

## Why It's Good For The Game

Currently with the new fish loot drop table, fish are more scarce (as they should be) and take a lot more work to get. But currently, you cannot make a simple sausage out of a single fish, like you can with a single meat steak slab. This changes that so fish can be up to snuff with a steak, and fisherman's hard work is marginally slightly better for providing meat. They still won't outclass a hunter though. Normally when you filet a fish you get two filets anyway, so mince should be feasible too. Not sure why this would be a bad change in anyway but if it is do let me know.

### DIFF
--- a/code/modules/roguetown/roguejobs/fisher/fish.dm
+++ b/code/modules/roguetown/roguejobs/fisher/fish.dm
@@ -10,7 +10,7 @@
 	max_integrity = 50
 	sellprice = 10
 	dropshrink = 0.6
-	slices_num = 1
+	slices_num = 2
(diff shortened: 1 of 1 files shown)


Gyrandola:
We already have enough food around. Scarcity is good.

thraes:
this would be a good change ONLY if blue potions get the red potion treatment

PR #2040, opened 2024-12-24
Re-enter Corpse Verb

EvenInDeathIStillServe:
adds a re-enter corpse verb to the spirit tab.

doesn't work if  your corpse is a zombini

### DIFF
--- a/code/modules/mob/dead/observer/observer.dm
+++ b/code/modules/mob/dead/observer/observer.dm
@@ -493,13 +493,17 @@ This is the proc mobs get to turn into a ghost. Forked from ghostize due to comp
 	if(mind.current.key && copytext(mind.current.key,1,2)!="@")	//makes sure we don't accidentally kick any clients
 		to_chat(usr, span_warning("Another consciousness is in my body... It is resisting me."))
 		return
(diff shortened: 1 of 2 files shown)


hugmeorelse:
Paid for this, as this has always been a major headache for both players and staff alike. It frustrates players, and it takes up a lot of staff time to handle. There are other bugs that poses the same problem for both players and staff alike that need to get prioritized and fixed.

code/modules/mob/dead/observer/observer.dm
```diff
@@ -493,13 +493,17 @@ This is the proc mobs get to turn into a ghost. Forked from ghostize due to comp
 	if(mind.current.key && copytext(mind.current.key,1,2)!="@")	//makes sure we don't accidentally kick any clients
 		to_chat(usr, span_warning("Another consciousness is in my body... It is resisting me."))
 		return
+	if("undead" in mind.current.faction)
```
Heheniy, line 496: but what if player had undead faction even before death, like occultist (refugee mage subclass)

Gyrandola (approved):

PR #2047, opened 2024-12-25
Nerve Master Withdrawing

Crowbar764:
## About The Pull Request

Adds a simple button to the Nerve Master to let anyone withdraw money directly from the treasury.

image

## Why It's Good For The Game

When bandits stopped being able to create Shylock accounts, they stopped being able to rob the treasury themselves. This simple feature allows them to do this optional objective once again.

This also stops the pointless task of stewards needing coins in hand and paying into their own accounts, and then running to a Shylock to withdraw the money that way, which happens a lot.


Gyrandola:
You should require a bank account to be able to retrieve money from the nerve master

Gyrandola (approved):

PR #2054, opened 2024-12-26
Occultist / Zizoid Mage Rebalance

ProbablyCarl:
## About The Pull Request
Skeletons are a broken mess, both unfair and unfun to fight against.
I absolutely adore occultist in concept and practice, but their minions break the subclass entirely and thrust it into the realm of genuinely cancerous in nature.
As a result, this PR decouples old Occultist from the magos refugee altogether and makes it an outfit for admin intervention exclusively.

Now, occultist in its state for this PR, admin exclusive outfit aside, has no such abilities to raise the dead. It can still heal them, and possess the strengthen undead stun spell, but it can't raise corpses anymore.

'Raise Undead', 'Command Undead' and 'Revoke Unlife' are no longer purchasable for Zizoid mages.

'Strengthen Undead' cost raised from 2 to 4. It was far too powerful utility and combat wise to be at TWO.
'Eyebite' is now purchasable for Zizoid mages with a cost of 3, because it was missing. This makes 'Ray of Sickness', 'Eyebite' and 'Strengthen Undead' the spells added to the list of possible to purchase for a Zizoid mage.

We'll leave the hoards of unkillable creatures to Lich, outside of admin curated events / intervention.

## Why It's Good For The Game
A myriad of player and administrative complaints have prompted me to push this. As someone who plays both skeleton AND fights them regularly, I'd like to think I've a good idea of both sides, and this is the best solution at the moment.
Barring a full skeleton rework, at least, but the Ten knows I refuse to touch that with a ten foot pole.

### DIFF
--- a/code/modules/spells/roguetown/necromancer.dm
+++ b/code/modules/spells/roguetown/necromancer.dm
@@ -15,7 +15,7 @@
 	antimagic_allowed = TRUE
 	charge_max = 15 SECONDS
 	miracle = FALSE
-	cost = 2
+	cost = 4
 	chargedrain = 1
 	xp_gain = TRUE
 
@@ -95,7 +95,7 @@
 
 /obj/effect/proc_holder/spell/invoked/raise_undead_lesser
 	name = "Lesser Raise Undead"
-	cost = 6	
+	cost = 6
 	desc = "Reanimate a corpse as a skeleton. You can maintain two sapient skeletons, the rest will be mindless. The body must have all limbs and its head."
 	clothes_req = FALSE
 	range = 7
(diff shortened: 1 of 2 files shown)


Kregle:
Good. Liches were made major antags so we wouldn't have skeleton hordes teaming up with rebels, vampires, etc etc. Occultists aren't even antags in the code. Keeping it around as an admin load out is more than appropriate.

Xaytan:
Remember that the ability to create skeletons was also given to any spellcasting character with Zizo as their patron too.

Should probably take that away from them. It never should have been done in the first place.

Gyrandola:
Remove their ability to raise the undead so they can still be played as a zizo mage of sorts.

Gyrandola (approved):

ophaq:
Psydon wins this one Zizo!

Kevak:
Do we have specific reports of skeletons breaking anywhere? It is irritating attempting to do bugfixes without actual bug occurrence instances being reported reliably. Hard to replicate and narrow down the issue. I’d also like to remind you that occultist lacked the ability to raise a swarm of any significant degree. It was capped to two skeletons per.

PR #2058, opened 2024-12-26
Vampyre Lord nerfs & Bugfix

Vizoplez:
## About The Pull Request
Caps VLord max blood at 65,000.
Fixes VLord being able to get negative vitae
Fixes VLord sun interaction ignoring disguise (adminbus)
Nerfs VLord statgain early on, buffs at end. Total still ends up being less than what it originally gave you.

## Why It's Good For The Game
Vlord is dummy strong at the moment and can roll on lowpop, but also has some really annoying bugs like the Vitae issue.

image
 This is their stats lowrolled post change at max level.


Gyrandola (approved):

john-pawson:
This nerfs overall vampire stats, but only raises the amount of stats the *thralls* gain at the end. So you've effectively lowered the amount of stats the VLord gets by two, and raised the amount the thralls get when the vlord ascends by one.

PR #2060, opened 2024-12-27
Violent npcs now can be male or female

Purpleturtle3:
## About The Pull Request

Makes it so bums, skeletons, and deadite npcs can be male or female. Also the male to female str diff is only 1 just like it happens with player characters.

## Why It's Good For The Game

I just thought it was weird that the bums, skeletons, and all deadites were males. Now the zombies who can bite you to death can be girls too #girlpower.


Heheniy:
muh male supremacy...

Gyrandola (approved):

PR #2064, opened 2024-12-27
Makes Village Elder Slightly Better

ophaq:
Made VE a master in polearms instead of expert. Gave them a little bit of sewing, hunting, and farming and one more in athletics.

## Why It's Good For The Game

VE having just expert in polearms after living so long is kind of lacking. They have journeyman in a lot of other melee skills but that is pretty bad and may as well be lowered to 1 or 2 but that's debatable. Main point of the VE is to rely on their good polearm skill and train young recruits in times of need or to protect the farm anyway similar to the veteran but just more geared toward the lower masses / soilson area.

Also gave them a little bit of skill in hunting/farming/sewing because it would make sense for them to have some experience in these after living for so long and helping with occasional harvests or animal tending. Raised athletics to 2 as well, which isn't a big increase I think unless I'm wrong.

The VE used to be the woodsman but now he's more of an underplayed role because it requires the player to be old and not many people visit to see him to train polearms anyway. Maybe now he can be slightly more useful on the farm if the VE isn't fucking off to go fishing or complaining about 'youth these days' like old people do. He is also mostly useless due in combat due to his stats anyway since he is elderly.

He can also stand as a guard on the farm or be the venerable trainer for those wishing to seek decent skills in polearms as the actual veteran in the duke's place is too busy running around or training soldiers.

**This is also a discussion I'm bringing up about what to do with VE skills as well. Let me know what could be changed while this is up so this is more balanced.**


Gyrandola:
I dont approve raising a minor role's combat skills to 5. It's insane.
Either way, the Village Elder will soon be reworked into the Major, who's not going to be a combat role.

ophaq:
> I dont approve raising a minor role's combat skills to 5. It's insane. Either way, the Village Elder will soon be reworked into the Mayor, who's not going to be a combat role.

Oh we're bringing back mayor? Okay maybe it is better to not touch VE then for the time being...

PR #2065, opened 2024-12-27
More Fisherman Changes

Vizoplez:
## About The Pull Request
This time, it's a slightly larger re-buff to make the loot table less frustrating but still random enough.

More chance of rings again, less chances of coin pouches; less chances of useless junk like clothing, shards, etc

## Why It's Good For The Game
People thought the nerfs were too harsh, I'm inclined to agree but not for the same reasons. Removing the double-stacked chances of rings on worms seemingly almost entirely fixed the rareness (or lack thereof) issue and this ended up being unusually harsh due to the high pick weights (relatively) of other items.

At some point this'll be reworked entirely into a full fishing system instead of just a gacha table, but for now, this should be fine.


Gyrandola (approved):

Windryn:
Could we please have a greater chance at fishing yielding fish?
It feels frustrating currently, maybe it me but more than half of what I pull isn't actually fish even at Master or Legend

PR #2073, opened 2024-12-29
Returns Necromancy to Necromancers

DiscordWizard:
## About The Pull Request

This pull request returns the less raise undead spell to the Occultist refugee. This change comes with a nerf to the raised skeleton, as well as restrictions, and a small buff to the ray of sickness spell. It also gives skeletons non-combat skills so they can act in other roles besides combat, hopefully pushing the class away from just a frag role.

## Why It's Good For The Game

Occultist right now, with raise dead disabled, is basically useless. It has no spell points to learn spells, one of its spells is useless because it has no more undead to raise, and ray of sickness as far as I've tested and been told by those who use it is pretty underwhelming for a 3 point spell.

This PR returns the ability to raise skeletons to the Occultist while also making it impossible to raise bums, as those were the mobs that players had the most issues with once skelliefied. Some outright lost crit weakness and gained crit resistance, making them essentially immortal. It also returns the ability to revoke undeath as suggested, so to stop shitter skeletons on their tracks or un-do minions that are no longer useful.

The PR also gives all raised skeletons crit weakness, meaning that critical strikes will cripple or outright instantly kill any skeletons hit by them.

I've tested it in my own local server, everything I raised was able to be easy killed and beheaded. Bums, even rotten and deadified, are no longer able to be raised.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/refugees/types/magos.dm
+++ b/code/modules/jobs/job_types/roguetown/refugees/types/magos.dm
@@ -98,6 +98,7 @@
 			H.mind.AddSpell(new /obj/effect/proc_holder/spell/invoked/projectile/sickness)
 			H.mind.AddSpell(new /obj/effect/proc_holder/spell/invoked/eyebite)
 			H.mind.AddSpell(new /obj/effect/proc_holder/spell/invoked/raise_undead_lesser)
+			H.mind.AddSpell(new /obj/effect/proc_holder/spell/invoked/revoke_unlife)
 			head = /obj/item/clothing/head/roguetown/necromhood
 			pants = /obj/item/clothing/under/roguetown/trou/leather
 			shoes = /obj/item/clothing/shoes/roguetown/simpleshoes


ophaq:
We need more skeleton slavery for sure...

Gyrandola:
TMing this and hoping necromancers wont cause too much trouble, considering the new map changes.

DiscordWizard:
> You NEED to give them the release undead spell. That's integral, so they can get rid of shitter skeletons. Aside from that, maybe communicate, too. Check the spells that Zizoids can buy, atop that, though I think I gutted it.

Yeah zizoid spells were gutted, I can give them the release undead spell if you think it's needed.

DiscordWizard:
There, if there's any reason for the spell to not be added, as I see it's on the list of spells the adminspawned necromancer can get, do tell and I'll remove it. It's a single line change

radarseas:
I'd give them command the undead too, although its only two skeletons, its a nice bit of QOL for "hey we're moving to a smith in the lower-town" without having to run half of the terrorbog to find out your summon died to a maneater.

maybe up their now nerfed-cap from two to three, sapients? they die a lot easier now _so people can probably be trusted?_

Gyrandola:
> maybe up their now nerfed-cap from two to three, sapients? they die a lot easier now _so people can probably be trusted?_

Let's not get ahead of ourselves. Let's see how it goes first.

Gyrandola (approved):

PR #2078, opened 2024-12-30
Makes ecclesiast accessible to shunned races

john-pawson:
## About The Pull Request

So far, every single refugee role is allowed to all races, but ecclesiast, that for some reason has even more race restrictions than the local church of the ten, which lets templars and acolytes of any races be a part of it, other than the priest, being the only role at the church that needs to be at least tolerated.

## Why It's Good For The Game

More ecclesiasts around meaning more diversity on the refugees, and still getting no respect or honor whatsoever for being a foreigner and a shunned race too.


Gyrandola:
yeah that was already included in the pyramid

Gyrandola (approved):

PR #2081, opened 2024-12-30
Deathsquad Garbage

ProbablyCarl:
## About The Pull Request
Adds a deathsquad outfit utilised for admin garbage. I wanted to make this more in depth, but this should help with staff sending the equivalent of no-skilled-lepers against nuclear armageddon.

They've an indicator for 'My ally.', when inspecting other DS, alongside a 'unique' music track for combat. Two classes.

## Why It's Good For The Game
This isn't a player facing change, without staff intervention.
It just speeds up the process of having deathsquad set up.


Gyrandola:
Except we are a low admin intervention server. This is pure adminbus

PR #2083, opened 2024-12-31
gives inquisitor custom armor & gives confessor a rag mask

zeratino:
## About The Pull Request

inquisitor gets a batman-esque coat
- covers all limbs as it's lined with plates & goes to his feet
- weakest to pierce damage
- keeps the original cloak due to user request
- no head protection

image

## Why It's Good For The Game

makes the inquisition pleasing to look at 
makes townspeople take the inquisitor seriously/not want to frag him

still can get his ass beat by people if he isn't careful though

very much open to feedback


zeratino:
please excuse the large amount of commits, this is my first pr in like 4 years

notafurryy:
I am not sure they should get reskinned full plate.. as armor... That durability looks a bit high, something along the lines of half plate with medium armor training would be better I think

zeratino:
> I am not sure they should get reskinned full plate.. as armor... That durability looks a bit high, something along the lines of half plate with medium armor training would be better I think

image
attached above is the durability of half plate w/ protection classes

image
attached above is the durability of both a regular steel cuirass & the inquisitor coat, with attached armor classes

notafurryy:
Oh! Fair then, might still want to make it medium rather than heavy, especially considering the roundstart ammount of money the inquisitor gets, the church already has templars that almost always get fullplate

Gyrandola:
VERY cool change but there's no need to powercreep it, I agree with Yawet

zeratino:
> VERY cool change but there's no need to powercreep it, I agree with Yawet

Got it. I thought the current durability & protection areas were fine since the idea was to make it so that it would be much more prone to the threat of breaking, as it has much worse durability than even half plate. 

The idea was so that it gives the Inquisitor some basic protection against a towner/whatever who decides to throw his life away instead of coming peacefully - but still crumple against anything from antags/an organized attack since that’s not what he’s made for.

Do you guys think that giving it scale-mail protection would be fine? I kinda want to keep the protection areas in but it doesn’t matter too much.

zeratino:
> Smail is considered mid-to-high armor, while I wouldn't mind it, i'd still prefer if you kept its durability somewhat low (compared to the original). Like I said, main concern is that with this change, an inquis could just throw away their coat for _actual_ plate armor (similar to why we dislike the Warlock PR for somewhat the same reason).
> 
> We'll see how it does on a testmerge.

Tight. Let me just get all that stuff done, then.

Gyrandola:
Solve conflicts and its ready to be merged.

zeratino:
im not really sure how to do this, ill have ask around and get to it later

Gyrandola:
> image
> @Gyrandola im being told to make a new pr that adds the dmi to fix the conflict
> do you know if this is the right thing to do

I can probably fix this for you, give me a bit

zeratino:
thank you

Gyrandola (approved):

PR #2103, opened 2025-01-02
Priest/Acolyte titles and old priests

notafurryy:
## About The Pull Request

Forces priests to be middle aged/old and gives them a title before the name, Acolytes also get titles!
image

## Why It's Good For The Game

Gives the priests/acolytes a little more recognition when they have their hoods off, also with the changes gyran made it makes sense that priests are older now, having to climb the clerical ladder and prove themselves rather than being chosen.. Making them old.. Like all priests...


Gyrandola:
Nice

Gyrandola (approved):

PR #2108, opened 2025-01-02
Allows shunned races up to play magicians apprentice

Dilligentdog:
To me it seems kinda stupid that that race lock was tolerated up with a exception for just small folk
`RACES_TOLERATED_UP_PLUS_SMOLFOLK`

Magic is something that anyone can be gifted from my knowledge, so why can't a dark elf be a apprentice under a Magos, or a Tiefling. 

These are the only shunned races that weren't allowed in due to the code allowing "smolfolk". So kobolds and verminfolk were allowed to be apprentices

image


Gyrandola:
no.

PR #2116, opened 2025-01-03
Makes citywatch armor more inline with MAA armor

notafurryy:
## About The Pull Request

Removes heavy armor training and removes protection regions/classes that do not fit the armor, watchmen helmets no longer get smash protection (It was the only medium helmet that had it) and removes arm protection from the torso armor. Also removes the heavy armor trait!
## Why It's Good For The Game

Citywatch has an insane level of powercreep and this feels like too small of a change but it's a needed one I feel. The citywatch had better gear than the keeps retinue which feels unrealistic? The conscript bobbies being better equipped than lifelong soldiers of the duke feels wrong, also they have no reason to have heavy armor training, again being civilians with a tad bit more authority over the peasants.

### DIFF
--- a/code/modules/clothing/rogueclothes/temp/citywatch.dm
+++ b/code/modules/clothing/rogueclothes/temp/citywatch.dm
@@ -8,11 +8,11 @@
 	item_state = "citywatch"
 	blocksound = PLATEHIT
 	do_sound_chain = TRUE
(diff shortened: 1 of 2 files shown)


Gyrandola:
Fair

Gyrandola (approved):

PR #2118, opened 2025-01-03
Gooner pr the goonening

wow50000:
## About The Pull Request
Adds good mood buffs from sex (Test on main pls.)

## Why It's Good For The Game
Realism.

### DIFF
--- a/code/datums/stress/positive_events.dm
+++ b/code/datums/stress/positive_events.dm
@@ -10,12 +10,12 @@
 
 /datum/stressevent/cumok
 	timer = 60 MINUTES
-	stressadd = -5
+	stressadd = -3
 	desc = span_green("I came.")
 
 /datum/stressevent/cumgood
 	timer = 60 MINUTES
-	stressadd = -10
+	stressadd = -5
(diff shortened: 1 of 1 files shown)


githubuser4141:
Aren't blue balls supposed to be negative? Why'd you give it a mood bonus?

Gyrandola:
The mood buffs are too big, and the blueballs change makes absolutely 0 sense.

CalamaBanana:
> The mood buffs are too big, and the blueballs change makes absolutely 0 sense.

You never see the negative moodlet except during sex, as that is (to 99%) the only time you have arousal. Is weird as hell to get stressed out from blue balls while you are balls deep.

wow50000:
> The mood buffs are too big, and the blueballs change makes absolutely 0 sense.

I've never had sex, this is how I imagine it feels.

wow50000:
> Aren't blue balls supposed to be negative? Why'd you give it a mood bonus?

Too lazy to change its name, its the buff you get mid sex now.

Gyrandola (approved):

PR #2119, opened 2025-01-03
Goblin racial stat buff

wow50000:
## About The Pull Request
Makes goblins more accurately depicted as hateful creatures instead of unnaturally resilient creatures. Changes their net balance of stats to be +1 instead of 0 to keep in line with the other races.

## Why It's Good For The Game
-3 str means that they cannot deal damage with melee which is what most of their roles consist of. 

This new statblock reflects that goblins are the creation of the god that values strength and domination, being somewhat fast melee damagers that have lower health rather than a weak high con pushover.


notafurryy:
Goblins have no reason to be strong though, they are just smooth kobolds, there is no way that someone that's 3 feet is as strong as your average person, except dwarves, they get a pass

notafurryy:
Also the +1 to str means that "guard" goblins can get 16 str which is just awful, if you want them to give them more str you should consider rebalancing the goblin roles

PR #2121, opened 2025-01-03
Snow effects

Gyrandola:
## About The Pull Request

New snow turf effects from Vanderlin.

image

### DIFF
--- a/code/__HELPERS/time.dm
+++ b/code/__HELPERS/time.dm
@@ -68,7 +68,7 @@ GLOBAL_VAR_INIT(dayspassed, FALSE)
 					foundnd = TRUE
 				if(!foundnd)
 					SSParticleWeather?.run_weather(pick(/datum/particle_weather/rain_gentle, /datum/particle_weather/rain_storm))
-			else if(GLOB.forecast == "rain")
+			else if(GLOB.forecast == "snow")
 				var/foundnd
 				if(SSParticleWeather?.runningWeather?.target_trait == PARTICLEWEATHER_SNOW)
 					foundnd = TRUE


zeratino:
posting some bugs here
image

weird borders happen when shoveled, updates when touched

image
sometimes disembodies people when they lay in it
(i think it happens when they're dragged out of the snow)

image
same with structures

misc stuff (not blatant bugs but nice to have):
water doesn't freeze & turn into ice tiles when it snows

code/datums/particle_weathers/datum_types/snow_storm.dm
```diff
@@ -73,3 +101,341 @@
...
+	START_PROCESSING(SSslowobj, src)
+
+	update_corners(TRUE)
+	update_overlays()
+
+	update_visuals_effects(first = TRUE)
+
+/obj/structure/snow/Destroy(force)
+	update_visuals_effects(src, FALSE)
+	STOP_PROCESSING(SSslowobj, src)
```
dwasint, line 160:
```suggestion
        for(var/atom/movable/movable in get_turf(src))
		if(movable.get_filter("mob_moving_effect_mask"))
			animate(movable.get_filter("mob_moving_effect_mask"), y = -32, time = 0)
			if(ismob(movable))
				movable:update_vision_cone()
			for(var/mob/living/carbon/human/human in view(movable, 7))
				human.update_vision_cone()
	STOP_PROCESSING(SSslowobj, src)
```

dwasint: this fixes people keeping the effect if they are in it while deleting it

Gyrandola: Thank you very much

Gyrandola:
will come back to this later tonight.

ophaq:
I hate snow. I hate snow. I hate snow. I hate snow.

Rayforx:
This fucks with combat, 100% of the time if you are fleeing, then you will always, be unable to run away. Yeah, no. Big yikes.

EvenInDeathIStillServe (approved):

radarseas:
this is needs a rework over the slowdowns and area blacklisting for the terrorbogs plus a timelock into the round for it, it makes it impossible for antagonists like **bandits**, **revs** or **maniac** that rely on stamina to run away from threats because everyone moves equally as slow, even sprinting with a staff without combat mode.

I've spent nearly half of a round for five treks through the terrorbogs because of how potent the slowdown is only to die to three equally slow roles because I cannot outpace them, despite superior stamina stats to outpace them in sprinting.

PR #2130, opened 2025-01-04
The ruler gets the money

Rayforx:
## About The Pull Request
This gives the queen the 1000 mammon when they are the ruler.

I did a quick test, and it should be working 👍 

https://github.com/user-attachments/assets/e4080589-6b91-4177-8ff8-6797eae506d4

## Why It's Good For The Game
It fixes the problem of the ruler not having the 1000 mammon when their is only a dutchess, and their is no duke.


Gyrandola:
Solved it this way https://github.com/Rotwood-Vale/Ratwood-Keep/pull/2132.
This will also prevent round start bad omens

PR #2132, opened 2025-01-04
Round wont start without a Duke/Duchess

Gyrandola:
## About The Pull Request

Prevents round from starting if nobody has readied up as a Duke/Duchess. This is ignored if FASTLOAD is defined or if an admin overrides it with the Start Now button.

## Why It's Good For The Game

Prevents round start bad omens and generally makes the round better by always having a Duke.


radarseas:
cult needs a ruler to sac to win anyway, so this is good. Especally without a priest.

EvenInDeathIStillServe (approved):

PR #2133, opened 2025-01-04
Gives the ability to buy 3 gold + silver instead of 10

Rayforx:
## About The Pull Request

image

## Why It's Good For The Game
Buying 10 gold and silver is hugely expensive, and something that is rarely affordable. Being able to only have to buy 3 will be more manageable for the steward.


wow50000:
Gold was specifically increased so that buying in bulk would be hard.

Gyrandola:
I prefer to keep it like this, we have miners after all.

PR #2137, opened 2025-01-05
Removes the ability to use triumphs to resist confession.

Spaghetti-bit:
## About The Pull Request

Removes the prompt for people to pay one triumph to resist confession. Replacing it with a percentage chance to resist. Antags get an additional 25% to resist.

Formula is...

INT + CON + 10 (+25 if antag)

## Why It's Good For The Game

A mechanic that derived entirely around getting confessions with RP and mechanics leading up to it should not be thwarted by something can be accrued across dozens if not hundreds rounds. One triumph isn't even a ripple in the proverbial pond.

No flaw? Makes sense, 3 triumphs.
Mood boost? Weird but eh... alright?

Resisting Confession? No, why? This doesn't make sense this is a mechanic that  requires another player (one who is irreplaceable) and has to even find the confessed(to be) to interrogate in the first place. Using something that is accrued over multiple rounds.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/church/puritan.dm
+++ b/code/modules/jobs/job_types/roguetown/church/puritan.dm
@@ -134,7 +134,7 @@
 
 /mob/living/carbon/human/proc/confession_time(confession_type = "antag")
 	var/timerid = addtimer(CALLBACK(src, PROC_REF(confess_sins)), 6 SECONDS, TIMER_STOPPABLE)
-	var/responsey = alert(src, "Resist torture? (1 TRI)", "TORTURE", "Yes","No")
+	var/responsey = alert(src, "Resist torture?", "TORTURE", "Yes","No")
 	if(!responsey)
 		responsey = "No"
 	if(SStimer.timer_id_dict[timerid])


Gyrandola:
I agree that using triumph is stupid because everyone has accumulated over 1000. But you shouldnt completely remove resisting.
Maybe make it some kind of stat check? The higher your stats the slower you break.

radarseas:
I'd do the same for cult conversion too, because its dumb that's a thing when you can literally just skele jaunt them instantly for basically the same results anyway.

Purpleturtle3:
Make it so people can still resist god confession but not antag status. Why? Because cultists have no reason whatsoever to ever confess, same as rebels. Meanwhile, the inquisitor could theoretically start grabbing random people around to see what god they follow and act as if any resistance is a confession by itself.

NikolayByondPlayer:
Deleting whole triumph system so far would be even better - it is just not balanced and overpowered

Spaghetti-bit:
> I agree that using triumph is stupid because everyone has accumulated over 1000. But you shouldnt completely remove resisting. Maybe make it some kind of stat check? The higher your stats the slower you break.

Done.

john-pawson:
END would make more sense than CON, considering that's the stat used for handling the amount of pain your character can have until you get pain crit.
Other than that, love the change. At least now getting a confession will actually be possible, instead of the outcome being decided only by the one being interrogated, since he's the one to decide if they want to just spend the entire round resisting.

Oh and, looking at the code:

`if(!resist || (resist_chance && prob(resist_chance)))`

Shouldn't it be `!prob(resist_chance)` since we're checking if they fail the resist check?

Spaghetti-bit:
> END would make more sense than CON, considering that's the stat used for handling the amount of pain your character can have until you get pain crit. Other than that, love the change. At least now getting a confession will actually be possible, instead of the outcome being decided only by the one being interrogated, since he's the one to decide if they want to just spend the entire round resisting.
> 
> Oh and, looking at the code:
> 
> `if(!resist || (resist_chance && prob(resist_chance)))`
> 
> Shouldn't it be `!prob(resist_chance)` since we're checking if they fail the resist check?

Changed from constitution to endurance.

As for the code suggestion; no. It's an order of operations, it's quicker to just check a boolean than a proc for checking chance. That's all <3

Gyrandola (approved):

wow50000:
shit's on backwards 
{31DC5606-F5A4-42AC-8834-0952D7BC8E8D}
{46949F49-B6AE-45E1-93D2-1EEC28D93E56}
shame on all o yew for not testin dis shit

PR #2141, opened 2025-01-05
Makes Butcher Slightly More Playable Now

ophaq:
## About The Pull Request
**Please give me constructive feedback on this and anything else you might think would help make this job and possibly even the soilson's job better.**

Gives the butcher some more versatility to help on the farm and the skills needed to assist with their job.

## Why It's Good For The Game
Currently butcher sucks to play. You are either substituting for the soilson and farming yourself or fucking off doing whatever to go sell animals or something (but often times soilson isn't there so you are still farming to feed your cattle). This makes them somewhat more desirable to play as they can have at least a bit more versatility in what they can offer from the farm. The main reason why I don't even want to play butcher is due to being able to do the same job as soilson. Now it is slightly more in line with them so they aren't struggling as much. Hell, soilson still is objectively better to play but this would definitely make butcher just a smidge more appealing in my opinion.

I wouldn't want to go as far as removing this role and just making soilsons a mix between the two but that's on the table if people want that to help with role bloat. If this happened I would also suggest making the round start soilsons a yeoman since they have their own land they're borrowing and now that the VE is gone.

Cooking (1 -> 3) - Literally says they're a savant in sausage making in the description. Correct me if I'm wrong but having faster sausage making skills with better cooking makes more sense (also I am pretty sure sausage making currently doesn't give cooking exp).

Sewing (0 -> 2) - Currently some leather working recipes require this and being able to craft some basic things from the leather they sell gives them more options. Unless someone wants to move every single leather-based clothing/etc recipe to require hunting skill instead so hunters and butchers are more reliable in making leather goods. Hunters also get this level of skill so it makes them up to par. I kind of want to give them 3 but that might be too much.

Carpentry (0 -> 1) - I put this in there because making extra tables and basic items seems like something they could do and the soilson gets this as well anyway. Also you need carpentry 1 for building extra animal pens.

Treatment (0 -> 1) - Butchers work with knives all the time. I think they would know how to stitch their hand cuts and whatnot when making a mistake tanning or butchering. Also, soilson already has this.

Tested locally, feels better to play.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/peasants/butcher.dm
+++ b/code/modules/jobs/job_types/roguetown/peasants/butcher.dm
@@ -23,12 +23,15 @@
 		H.mind.adjust_skillrank(/datum/skill/combat/whipsflails, 2, TRUE)
 		H.mind.adjust_skillrank(/datum/skill/combat/wrestling, 3, TRUE)
 		H.mind.adjust_skillrank(/datum/skill/combat/unarmed, 2, TRUE)
-		H.mind.adjust_skillrank(/datum/skill/craft/cooking, 1, TRUE)
+		H.mind.adjust_skillrank(/datum/skill/craft/cooking, 3, TRUE)
 		H.mind.adjust_skillrank(/datum/skill/craft/hunting, 4, TRUE)
 		H.mind.adjust_skillrank(/datum/skill/craft/crafting, 2, TRUE)
 		H.mind.adjust_skillrank(/datum/skill/misc/athletics, 4, TRUE)
 		H.mind.adjust_skillrank(/datum/skill/misc/riding, 3, TRUE)
 		H.mind.adjust_skillrank(/datum/skill/misc/reading, 1, TRUE)
+		H.mind.adjust_skillrank(/datum/skill/misc/sewing, 2, TRUE)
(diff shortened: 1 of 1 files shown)


Gyrandola (approved):

Casteroki:
Actually as someone who has worked with animals growing up. You are constantly having to patch up and tend to your animals medically. Half the time they are finding a way to die and you have to keep their silly ass alive. A lot of that is applicable to human care too. Seeing butchers getting some treatment is great!

githubuser4141:
> Actually as someone who has worked with animals growing up. You are constantly having to patch up and tend to your animals medically. Half the time they are finding a way to die and you have to keep their silly ass alive. A lot of that is applicable to human care too. Seeing butchers getting some treatment is great!

But this isn't necessarily applicable to Butcher if they're a dedicated role and might have no prior farming experience. If they're subbing well substitutes are normally worse than the actual thing. So it could go either way: the Butcher has little/no experience in running the farm long-term OR the butcher is themselves a ex-soilson that decided to spec into butchering for whatever reason.

ophaq:
Honestly butcher could use 1 more treatment, a single level in farming and 1 more level in sewing. Maybe that's just me who plays them often.

PR #2149, opened 2025-01-06
nerfs wooden platform cost, makes carpentry and masonry easier 

Elli-Skala:
## About The Pull Request

removes fiber costs for wooden platforms and stone

## Why It's Good For The Game

the requirement for fibers on wood and stone made it so that almost every time someone built would build twig platforms first instead of bothering to gather the fibers. atop of that it just makes more sense to need fibers to tie twigs together

as a second twig platforms are so accessible that often times people will make long twig bridges to get past walls instead of getting a proper carpenter or mason 

# Changes
raised stone platform cost to 2 stone but felt like it wasnt enough, so additionally raised the skill level to stone and wooden platforms to level 2 (same as walls)

### DIFF
--- a/code/modules/roguetown/roguecrafting/turfs.dm
+++ b/code/modules/roguetown/roguecrafting/turfs.dm
@@ -24,7 +24,7 @@
 	skillcraft = /datum/skill/craft/carpentry
 	verbage_simple = "construct"
 	verbage = "constructs"
-	skill_level = 1
+	skill_level = 2
 
 /datum/crafting_recipe/roguetown/turfs/woodplatform/TurfCheck(mob/user, turf/T)
 	if(isclosedturf(T))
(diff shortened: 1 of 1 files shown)


notafurryy:
I don't think that's how you use the word nerf

Hmeister-real:
keep the twig platform change and revert the removal from the other two

radarseas:
yeah I like this, it gives a GENUINE reasoning to use these over spamming sticks.

However please, please up the cost of stone platforms by one singular stone otherwise anyone with a spade can skybridge wherever the fuck they please with nothing more than a sack because they stepped into a mine and happened across 40+ stones just laying there.

Gyrandola (changes requested):
What the other contributors say

Gyrandola (approved):

PR #2151, opened 2025-01-06
Beard growth and shaving

Vladmar11:
Now shaving works as it should.
Port from https://github.com/sylphynford/Penumbra/pull/89
image
image
image


RyszardParys:
Can i set it so my beard wont grow? Let's say i want to keep a specific beard at all times.

ophaq:
I hope you aren't going to force this on all races with a beard or whatever cause there's a lot of characters it wouldn't make sense to have naturally growing beards with. Seems like a weird mechanic.

ophaq:
> I can see us having some exceptions like elves for example. It's worth noting this doesn't make you grow a full beard but only stubble.

What about people like wild kin which are a mixed bag? Such as insect people or even fluvians? I think there should be a toggle option or a detect beard feature so if the person has a beard then they grow more and it can be shaved. There are people who don't genetically grow facial hair too (and sometimes femboys). I think it would be unfair to make anyone that's a male suddenly have to grow stubble or a beard over time.

wow50000:
This only applies to men right ?

Ember16:
this is dumb imo. If people want to opt into it, fine, but forcing it on everyone is pointless micromanaging that is really shitty for people playing rolls that cant easily access a knife and now have to scramble every round to find one just so they dont have to have a giant beard.

Gyrandola:
Time for a town barber

2PlainGhoul:
this Is kinda dumb -- what If you just don't want any facial hair? what If you play a race like Wild-Kin, Fluvian -- Axian -- It just seems like a weird thing to impose on people

Gyrandola (changes requested):
Remove beard growth from Elves and other races that would realistically not grow a beard (wildkin, fluvians, axians, etcetera.)

radarseas:
I'd make any form of deadite (lich, vampyre etc) immune to beard growth too. the idea of a vampyre lord getting a beard when they don't feasably have a knife as an undead ancient abomination is redicious (and inconvenient as you can't do anything about it).

PR #2152, opened 2025-01-06
Melee Combat Accuracy REWORK

BlackjackIsTired:
## About The Pull Request

Reworked Combat Accuracy around penalties that apply to which zones you're trying to hit as well as a cumulative penalty system that penalizes picking a sublimb to get what amounts to a free reroll. 

Example: Before when you targeted the top of the skull with an attack, upon a miss it would make you attack the zone that sublimb is inside. What this means is that you miss your attack on the top of the skull and then get a second roll on the Head, if you fail again, it hits the CHEST. That was one of my concerns and part of the reason I made the previous PR (https://github.com/Rotwood-Vale/Ratwood-Keep/pull/2125). While it DID solve that issue, it wasn't exactly the most immersive or realistic fix. It also included some adjustments to Skill Based Accuracy that I was perhaps too hasty in adding. I took the PR out of the oven too fast so to speak. But as a result of that and as a result of the feedback I've gotten, I made this.

So, to begin with an example on how this new system works. When you make an attack you get a penalty decided based on the difficulty of the part of the target you're aiming for. If you succeed, great! If you fail, that's part of where this new system comes into play. Now, you've aimed for the eyes, you missed, tragic. Now you hit the zone that sublimb is part of. (I mean really, how does someone miss an attack to the right foot and end up hitting the CHEST?). Anyhow, given that the eyes are a part of the Head, you get second a roll against the Head, that's roughly where you were aiming after all. However, the penalties are cumulative, since the penalty for aiming eyes is -50, and the penalty for aiming Head is -20. You now have to overcome a -70 to your roll to hit. Depending on your SKILL LEVEL, this can sometimes be impossible or near impossible. After all, a barely trained swordsman shouldn't be able to strike the eyes of a moving target so easily.

That's generally how the system works with this PR, there's some other changes which I'll document below:

Added a diminishing returns formula for Aimed and Non-Aimed intent(This may need tweaking, it may be too low). 
The base formula for accuracy is still similar to how it was previously, involving Perception as well as the ACC CALC having a x10 multiplier instead of x7, this time however, it will certainly **NOT** increase the likelihood of eye hits across the board.

Weapons now have accuracy bonuses based on length (controllability, ease of use).
Daggers and similar weapons with WLENGTH_SHORT have +10; your swords and other similar, typically one-handed weapons (WLENGTH_NORMAL) have +5. WLENGTH_LONG: your longswords, like bastard swords have no bonus.
WLENGTH_GREAT: your greatswords, spears, and similar weapons have a -5 penalty for accuracy.

Although it’s perhaps weird that spears would be considered less accurate, I might change that in the future.

Lastly, I've made it so that when you miss the neck, it redirects your hit to the CHEST instead of the Head. I think it makes more sense that a missed attack to the neck would hit the chest armor as opposed to the Head, that's however open for discussion.

## Why It's Good For The Game

A greater disparity between skill levels, taking account of your skill, equipment and intent are essential when choosing to target a critically important area on the body. You will truly no longer have to be fully armored to avoid having your throat slit or your eyes stabbed out. Masks are no longer **NECESSARY**, while it can still happen, you'll no longer have to be told by an oracle that despite your full plate, you will get throat slit because there's a very angry maid with a steel dagger coming at you and her accuracy is pinpoint, _she's basically a Rogue with that thing._ Shorter weapons have better accuracy, so jabbing in the eyes with a dagger isn't a bad move, but you still need to be trained in Knife-Fighting to be any good. Aimed intent helps!

### DIFF
--- a/code/modules/mob/living/roguetownprocs.dm
+++ b/code/modules/mob/living/roguetownprocs.dm
@@ -102,7 +102,7 @@
 
 	if(!user.mind)
 		var/npcchance = 60
-		if(prob(npcchance + cumulative_penalty))
+		if(prob(npcchance + cumulative_penalty + acc_zone_penalties[zone]))
 			return zone
 		else
 			cumulative_penalty += (acc_zone_penalties[zone] || 0)


Xaytan:
I think cutting and stabbing weapons should still get an accuracy advantage over blunt ones.

BlackjackIsTired:
> I think cutting and stabbing weapons should still get an accuracy advantage over blunt ones.

It's certainly worth discussion. Could you share your reasons for making such a suggestion? I understand stabbing having an accuracy advantage, it's reasonable and realistic. But from that same standpoint, I'm not so sure why cutting would get a similar buff to accuracy.

john-pawson:
> But from that same standpoint, I'm not so sure why cutting would get a similar buff to accuracy.

Cutting makes sense to have a higher accuracy, it's a thin blade with incredible maneurability, easy to control and move into those precise places like someone's neck under their jaw, or someone's hand while they're fighting, compared to things like a metal ball where your initial inertia dictates most of the accuracy. 
Besides, from a balance standpoint, cutting has absolutely nothing going for it other than slightly decent accuracy, considering how almost any armor is able to stop it, and it doesn't have much going for it compared to it's bigger brother, the stab (Same speed, more accuracy, more penetration, less armors have stab crit resistance, etc.)

john-pawson:
> > compared to things like a metal ball where your initial inertia dictates most of the accuracy.
> 
> That makes a lot of sense, I wasn't thinking about it from a comparative standpoint. I was just thinking about it individually. I'll go ahead implement this in a way that fits with the balancing of the system.
> 
> For clarity, are we just talking about "Stab" and "Cut" intents on weapons or are we talking about **_all_** stabbing and cutting motions. I.E. Pick, Chop, etc.
> 
> @john-pawson @Xaytan

Oh yeah definitely, things like chop or pick should probably stay without a bonus to accuracy, they don't need it and already consdered the heavier/slower versions of cut and stab, so them not having the bonus makes sense.

BlackjackIsTired:
> Oh yeah definitely, things like chop or pick should probably stay without a bonus to accuracy, they don't need it and are already considered the heavier/slower versions of cut and stab, so them not having the bonus makes sense.

This should be a big enough change to improve stabbing and cutting options while reducing the advantage blunt weapons have. It's not too much to be over-tuned, but also not too little to be irrelevant. This will also expand the relevance of precision-like small weapons. For example simply using a Dagger has a +18% Chance to Hit Precision Zones. So dropping your mace and quickly switching to your dagger when your opponent's helmet and/or mask is broken would be a totally valid strategy.

(Perhaps moreso when I attempt to rework lethality in a similar manner along the lines of the new hit chances. It's something I'm planning for now, not at the stage of proper discussion yet, since it's a follow-up to this PR and I want to keep them relatively atomized. (As an addendum, I'm not talking about "lethality" as in instant kills, but when you cut someone's unarmored throat with very low chances, you should be rewarded with a gratifying, brutal noise, sprays of blood and some serious fear of death. Something that when it happens, you should think "OH FUCK-- AHHHHHHHHHH!!!")

Another thing I'm considering adding down the **_line_** is grabbing someone gives both of you accuracy bonuses on each other, so it's a double-edged sword. "Wolfenstein Style", suggestion was given to me by a friend.)

RadiantFlash7:
Generic swords lack a niche currently. I think making them more accurate over other options might be a good idea tbqh.

TonyKarsh:
I think this broke decapitations.

dreamseeker_TPJfR0u1Tr

Gyrandola:
Two things:
Remove the brackets from the code, DM allows brackets but its using them isnt part of the convention
Normal players shouldnt be seeing all that math in the combat messages.

code/modules/mob/mob_helpers.dm
```diff
@@ -25,7 +25,7 @@
 		if(BODY_ZONE_PRECISE_EARS)
 			zone = BODY_ZONE_HEAD
 		if(BODY_ZONE_PRECISE_NECK)
-			zone = BODY_ZONE_HEAD
+			zone = BODY_ZONE_CHEST
```
Gyrandola, line 28: This is probably whats making decaps impossible

BlackjackIsTired:
> Two things: Remove the brackets from the code, 
> DM allows brackets but using them isnt part of the convention Normal players shouldnt be seeing all that math in the combat messages.
> Normal players shouldnt be seeing all that math in the combat messages.

I typically use brackets for better readability, I'll keep in mind to remove them before I PR or break the habit for DM from now on, thanks.
I'd assumed the visible math would be fine since showing attack rolls is a toggle-able option, regardless, I've cleaned up the logs and it gives much less feedback. It just says the chance there was to hit and that's it.

Finally, decapping is fixed.

RyszardParys:
Another accuracy rework boys, I sure do hope that the desert rider actually doesnt have 10% to stab someone in the eye with his scimitar now. :clueless:

john-pawson:
Ok i'll have to comment about this because, it seems that the accuracy penalties are way too high for the low amount that you recieve from skill/PER.

Let's take for example the absolute highest amount of accuracy you can have, a dagger on stab intent, being wielded by a legendary knife-wielder with 17 PER and aimed intent. (Yes, due to how the calculation is, 17 PER gives you higher chance than 20, matter of fact there seems to be a hard cap where anything above 14 gives you 15% chance on AIMED intent.)

This means:
60% chance to hit due to skill
10% chance due to being short
8% chance due to stab intent
15.3% chance due to aimed intent

That's an 93.3% chance to hit the chest. At the absolute highest.
So if we were to aim at the absolute hardest place to hit, an eye, having a -70% chance to hit, you'd have a 23.3% chance to hit it, and then a 53.3% chance to hit the head, barely above 50/50 chance.

The most skilled person that could possibly exist, using the most accurate weapon, can only hit the head half of the time.

I don't think i have to say much but this is WAY too low, you're hitting the eye 1/4 times, and there's still a chance to get parried + an eye hit doesn't garauntee anything, you could stab someone on the eye and just give them a minor cut, or worse, get stopped by the armor.

Of course, most people, even trained fighers, won't be using short weapons, and won't have >15 PER, and won't have legendary skill, so it makes it outright almost impossible to domething like hit an eye, or even a hand, for the most skilled of warriors.

Or for basic warriors, a cutthroat rogue using a dagger would barely have a 28% chance to hit the head, probably the place you should be aiming for most of the time to fight normal enemies on the bog like deadites. It should be at *least* something like >50%, they're trained combatants after all. It shouldn't be that hard. (It'd also mean it would be impossible to do something like hitting their neck, unless they turn off combat mode and their enemy is already on the ground and incapacitated) 

So yeah, i personally think either the skill and the PER bonus should be higher, or the penalties substancially lower. I don't mean that some parts like eyes/neck/hands/feet should be able to be attacked by everyone, but even when masters can't hit some of the precise places, it's kind of too high.

BlackjackIsTired:
> Or for basic warriors, a cutthroat rogue using a dagger would barely have a 28% chance to hit the head, probably the place you should be aiming for most of the time to fight normal enemies on the bog like deadites. It should be at _least_ something like >50%, they're trained combatants after all. It shouldn't be that hard. (It'd also mean it would be impossible to do something like hitting their neck, unless they turn off combat mode and their enemy is already on the ground and incapacitated)

> So yeah, i personally think either the skill and the PER bonus should be higher, or the penalties substancially lower. I don't mean that some parts like eyes/neck/hands/feet should be able to be attacked by everyone, but even when masters can't hit some of the precise places, it's kind of too high.

Thanks for the detailed, helpful feedback. I'll take a look & change up the numbers with this in mind.

> I don't think i have to say much but this is WAY too low, you're hitting the eye 1/4 times, and there's still a chance to get parried + an eye hit doesn't garauntee anything, you could stab someone on the eye and just give them a minor cut, or worse, get stopped by the armor.

This kind of accuracy is what I would have in mind with my lethality rework. It WOULD guarantee something if you were successful. This was a great test for the future, but evidently when you can stab someone in the eye four times and never get a crit or any payoff, it kinda defeats the purpose of aiming for the weak spots at all!

Regardless, since that rework isn't done yet, I'll change the numbers with the current lethality system in mind.

Moving forward, it'd be great to hear if they've made a positive impact. After all, the point of the PR is to discourage things like eye stab spam and not require you to be fully armored, since previously you could just target any unarmored limb and hit nearly all of the time, invalidating your cuirass & hauberk entirely because your ears were unarmored. 

The system will continue to be adjusted based on polite, thought-out, constructive feedback. Thank you.

Xaytan:
I haven't checked, does this remove the abilitiy to guarantee precise hits when your target is on the ground?

That's generally one of the most impactful parts of being prone, and rewards game knowledge when fighting say deadites or skeletons as a less skilled combatant since if you can get them on the ground you can suddenly kill them quite easily.

Xaytan:
Let's also keep in mind what the skill scales actually look like and represent:

0 or 1: Wholly untrained fighter, maybe familiar with an approximate tool. People who work non-physical jobs or noblewomen who aren't expected to fight.

2: Somebody who is either very familiar with the basic tool or has been in a few scraps and picked some things up. Workmen. Hastily trained peasants.

3: People who have received some amount of professional training. Vanguard, watchmen, common bandits, more proper peasant levies.

4: Professional soldiers. Mercenaries, Men at Arms, Templars, the Dungeoneer, watchmen who got lucky and rolled high.

5: Dedicated warriors. Knights, Captain, Sheriff

6: Certified monsters. Vampire Lords, the Veteran

Don't make the mistake of balancing around everyone being at rating 4 or 5. Assume the average person who is supposed to be in a fight is at a 3.

john-pawson:
> Don't make the mistake of balancing around everyone being at rating 5. Assume the average person who is supposed to be in a fight is at a 3. Everything higher than that is a bonus everything lower is a penalty.

This. 100% this. Use Journeyman skill as the standard amount of skill anyone should have, decent chance to hit things, anything below makes it harder and gives you a disadvantage, restraining you on the parts you can aim at making them substancially harder, anything higher is easier and gives you and advantage, opening more options and more consistently hitting limbs and precise zones.

zeratino:
> I haven't checked, does this remove the abilitiy to guarantee precise hits when your target is on the ground?
> 
> That's generally one of the most impactful parts of being prone, and rewards game knowledge when fighting say deadites or skeletons as a less skilled combatant since if you can get them on the ground you can suddenly kill them quite easily.

it does not

code/modules/mob/living/roguetownprocs.dm
```diff
@@ -15,45 +15,112 @@
...
-		chance2hit += (user.STAPER)*2
+		chance2hit += round(max((user.STAPER * 1.75) - (0.05 * (user.STAPER ^ 2)), 0))
+	else
+		chance2hit += round(max((user.STAPER * 1.2) - (0.03 * (user.STAPER ^ 2)), 0))
 	if(istype(user.rmb_intent, /datum/rmb_intent/swift))
 		chance2hit -= 20
 
-	chance2hit = CLAMP(chance2hit, 5, 99)
+	// Clamped C2H so we don't get 115% or -50% rolls.
+	chance2hit = CLAMP(chance2hit, 0, 100)
```
NPC1314, line 84: If I recall correctly clamp between 5 and 99 should yield a result between those numbers.
That would mean this changes it so you again can end up with zero chance to it, or 100% chance to hit.
Would need testing to confirm but yeah, clamp 5 to 95 instead is better if I remember correctly.

BlackjackIsTired:
Added "Wolfenstein Style" Accuracy Bonuses to Grabbing.

Skill Levels made impactful without going wacky. 
Accuracy Penalties rebased around assumed Journeyman skill. 
NPC accuracy more on par with the new Accuracy System.(Be careful, they can get very dangerous)
Added Rerolls that are easier depending on your Skill Level.
PER still adds to your accuracy independently of your intent with no multipliers, so no risk of crazy overpowered bonuses.
Aimed & Swift Intents reworked to either raise your Accuracy the equivalent of one level or lower one level respectively.(Aimed is especially effective for unskilled, Swift should usually only be used if you're grabbing someone or very skilled with your weapon.)
Added feedback flavor text that informs and stylizes your failures.
Added misses with visual, audio & text feedback.
Bonked Smash+Swift Spam without nuking it into oblivion.(It's viable if you're skilled, otherwise you're just flailing).
Added a fuck ton of safeguard checks to prevent runtimes.
Made helpers for the helpers to help out with helping.

There's still some work to do on it, but otherwise it's essential final ver. As always, numbers can be tweaked at any time.

Foxentail:
If you're going to do this, you really should apply the accuracy thing to the mobs too. Its not fun to miss 1000 times meanwhile the mobs hit every. single. time.

Spaghetti-bit (changes requested):
> Bonked Smash+Swift Spam without nuking it into oblivion.(It's viable if you're skilled, otherwise you're just flailing). 

Add a wind up time to smash, that would fix the issue entirely. Similar impale etc.

BlackjackIsTired:
> If you're going to do this, you really should apply the accuracy thing to the mobs too. Its not fun to miss 1000 times meanwhile the mobs hit every. single. time.

Mobs already have a simplified logic applied to them. They don't have skills so they have a placeholder system instead. It just needs feedback for tweaking. Currently it's the same order as user mobs except it's a flat 50% for each roll. Also, if you're missing 1000 times, it's likely you're not adapting very well to the system. There's things you can do to raise your accuracy, such as grappling, aimed intent and targeting easier to hit limbs. A good general guide on how hard something is to hit. The smaller it is on the UI Limb Doll, the harder it will be to hit.

BlackjackIsTired:
> We're going to step away from this for now. Mobs are cancerous to fight still, and don't suffer the problem of accuracy that you do. Simple mobs become terminators at times. Carbon simples are even worse. Expert sword fighting with a sixteen percent chance to hit a skeleton in the neck? AAAAAAAAAAAAAAAAA
> Make it so that mobs abide by a similar system, or miss just as frequently, because they're genuinely unfun to fight.

### _Changelog:_

1. Fixed a bug that caused penalties to improperly apply on rerolls, causing more failures than intended.
2. Fixed a bug that caused NPCs have a 50% chance to not miss after failing every roll.
3. Improved the NPC aiming system to properly incorporate zone penalties.
4. Changed NPC simulated level to Expert : 60%. (They're supposed to be dangerous.)

Spaghetti-bit:
Commonly repeated feedback is that bog mobs feel overtuned. @BlackjackIsTired

BlackjackIsTired:
> Commonly repeated feedback is that bog mobs feel overtuned. @BlackjackIsTired

They have the exact same skill level and penalties as an "Expert". That definitely wouldn't make them overpowered, but it WOULD make them dangerous. Which is the intention. Are the mobs not supposed to be dangerous? Obviously, if you try to fight a spawn of Zizo or Graggar without armor, it SHOULD be dangerous. At least in my opinion. These are the kinds of things that would eventually combined to make the Bog dangerous.

It's not something people can't adapt to if they try. But if you as a Maintainer feel strongly about it, I can drop them down to Journeyman.

(P.S. Thank you for passing the feedback along, I generally don't get feedback given directly to me, unless it's already too late to fix it and it's being UnTM'd. So I appreciate hearing the feedback while I can still make changes to it live.)

BlackjackIsTired:
It's not to my satisfaction. I know I can do better than this.

PR #2155, opened 2025-01-06
Dechudding the last Bastion of toxic masculinity in the Keep, or How I Learned To Stop Caring And Let Girls Play Captain

Savotta:
## About The Pull Request

A one-line code change that allows females to play as Retinue Captain. I think.

## Why It's Good For The Game

Every other town-aligned combat role has allowed females, including Man-at-Arms. I don't see why Captain is still excluded. Plus, I want it. I think.

This was also bound to happen eventually. I think.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/nobility/captain.dm
+++ b/code/modules/jobs/job_types/roguetown/nobility/captain.dm
@@ -8,7 +8,7 @@
 	spawn_positions = 1
 
 	allowed_races = RACES_TOLERATED_UP
-	allowed_sexes = list(MALE)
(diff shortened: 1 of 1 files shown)


Savotta:
> Only chuds are allowed to play Captain. Sorry.

thescronkstate

hugmeorelse:
At this point, there is really no logical reason to not merge this PR. People have shown themselves in polls to be in favor of this, and it also explicitly says in the background text for retinue captain that they are veteran knights that have been promoted into this role. Considering we have dames, there's no reason to reject this PR.

Gyrandola (approved):

PR #2160, opened 2025-01-06
Ranged subclass for Sellsword that ANYBODY can use. 

divifest:
## About The Pull Request

As it says on the tin. Adds a ranged variant of Sellsword, that isn't locked behind a species etc. Basically just a modified fencer, but instead of sword skill it's bow skill. Also -1 wrestling in favor of +1 climbing. Medium armor REMOVED in favor of dodge expert.

## Why It's Good For The Game

Because it adds variety to Sellsword, and offers another avenue rather than generic melee fighter merc. (Also isn't some super special exclusive class only available to xyz)


wow50000:
str and maces is too high imo, otherwise seems decent

divifest:
> str and maces is too high imo, otherwise seems decent

Maces changed to 2, strength down to 1, perception up by one to reflect this suggestion.

Gyrandola:
I believe generic mercenaries are bloat and the last thing I want is to have adventurer 2

PR #2161, opened 2025-01-07
Snow tweaks

Gyrandola:
## About The Pull Request

Reduce snow probability


radarseas:
honestly it could use less movement debuff too, especally with a staff because right now staffs don't seem to help. bandits and the likes really, struggle to get through it without hiting stamcrit from using leaps to outrun crowds.

especally with how long it takes to trek the entire terrorbog.

EvenInDeathIStillServe (approved):

PR #2162, opened 2025-01-07
Fatigue and Stamina renaming and refactor, slight balance tweaks

Spaghetti-bit:
## About The Pull Request

Renames fatigue to stamina (green bar)
Renames stamina to energy (blue bar)

Max Stamina and Max Energy calculation change: (The only balance change that would affect anything)
From `(STAEND + (athletics_skill / 2)) * 10` to `(STAEND + athletics_skill) * 10` 
From `(STAEND + (athletics_skill / 2)) * 100` to `(STAEND + athletics_skill) * 100`

This means that every level of athletics gives `100` energy max (from `50`) and stamina.... `10` from `5`.
Standard max energy is 1000.
standard max stamina is 100.
Endurance is still the king in what determines max energy and stamina.

Slight buff to stamina and energy max.

Stamina recovery remains the same.
Energy recovery increased. MUCH faster when sleeping. Energy (blue) recovers faster overall. Not by much.

Fixes stamina recovery for undead if they are to ever get their stamina boon changed... It was due to them having the no sleep trait. Because that totally made sense 😭 My head hurts.

## Why It's Good For The Game

We have alot of unused stamina code from TG, this is the first pass over.

Energy/Stamina changes were to make athletics a tiny bit better.
Energy recovery changes were intended to be less painful. If you sleep you'll recover all your fatigue almost every guaranteed.

Needs a TM.


code/datums/status_effects/debuffs.dm
```diff
@@ -126,7 +126,7 @@
 			break //Only count the first bedsheet
 		if(health_ratio > 0.8)
 			owner.adjustToxLoss(healing * 0.5, FALSE, TRUE)
-		owner.adjustStaminaLoss(healing)
+		//owner.adjustStaminaLoss(healing)
```
dwasint, line 129: :( just remove the lines if code is unused just remove it so its easier to deal with

Spaghetti-bit: Went through all of it and removed the commented lines.

MarioTheShiba:
this will be good. thank you.

Gyrandola (approved):

PR #2170, opened 2025-01-09
Racelocked Tabaxi noises

Savotta:
## About The Pull Request

Don't look at the sheer number of commits. I'm still learning.

This adds meows, trills and hisses for Tabaxi ONLY. They're species-locked as to avoid having Humen, Half-Kin, various Wild-Kin and what-other species running around acting like kitties. It's also race-locked to a relatively minor playerbase as to understand whether or not the addition of furry emotes to the codebase is viable while maintaining player quality. I would suggest TMing this if it's not denied outright, seeing as I'm somewhat skeptical that people won't just spam it and be retards.

## Why It's Good For The Game

Cats make noise. They're noisy critters. Noise can, in some instances, be good. If this turns into a shitfest where all the kitties do is mrrowmeow and hiss at eachother, it won't be so good. PRing this in the first place is taking a lot of faith,  which probably won't be reciprocated by maintainers. Alas.

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/furry/tabaxi.dm
+++ b/code/modules/mob/living/carbon/human/species_types/furry/tabaxi.dm
@@ -183,20 +183,3 @@
 /datum/species/tabaxi/on_species_loss(mob/living/carbon/C)
 	. = ..()
 	UnregisterSignal(C, COMSIG_MOB_SAY)
-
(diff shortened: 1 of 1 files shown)


Savotta:
these commits make me look like a bumbling retard, i shouldve probably made a new fork. it works doe

Heheniy:
purr where

Savotta:
later

PR #2173, opened 2025-01-09
Brings back the old wise wizard

ChuChu42:
## About The Pull Request

Spawning as a old aged wizard now makes you get blue robes instead like it previously did
Old wizards are supposed to be wise, but for some unknown reason, they were made so if they are old, they become retarded and barely able to cast magic

This pr aims to revert that and let people play "Wise Wizard" again

image

## Why It's Good For The Game

As a wizard, growing old while studying magic only makes you wiser
The old blue robes are iconic, now players get a choice if they want to be the wise wizard court magician or the normal court magician
_**like they were previously able to do so**_


Gyrandola:
Those robes barely fit the new intended setting unfortunately. Iconic for DND perhaps.

PR #2175, opened 2025-01-09
Acid pit cell 

NikolayByondPlayer:
## About The Pull Request

1) Adds a cell above the acid pit for people to get trapped in, if players want to capture someone
image
image
image

## Why It's Good For The Game

Added choice of killing/capturing. If they want to burn they can still pull the second lever


Gyrandola (approved):

Limeon666:
I still think having this is unnecessary, overkill, and makes no sense in the setting, but this is better than instant death lever.

PR #2177, opened 2025-01-09
Racelocked Tabaxi noises without 30 useless commits

Savotta:
## About The Pull Request

Same as the PR I closed due to audio quality concerns. Adds meows, trills, hisses and purrs for Tabaxi only, with two variations for each.

Half-kin aren't getting this. They will be receiving their own sets of audio if I ever decide to PR it. Wild-kin may, possibly, just maybe get this in the future. 

## Why It's Good For The Game

Noise can be good. Spamming noise is NOT so good. Please TM this to see whether or not the Tabaxi playerbase can be trusted with furry noises.

### DIFF
--- a/code/modules/mob/living/carbon/human/voicepacks/genfemale.dm
+++ b/code/modules/mob/living/carbon/human/voicepacks/genfemale.dm
@@ -98,6 +98,6 @@
 			if("hiss")
 				used = list('sound/vo/hiss.ogg','sound/vo/hiss (2).ogg')
 			if("purr")
-				used = list('sound/vo/purr.ogg','sound/vo/purr (2).ogg')
+				used = 'sound/vo/purr.ogg'
 
 	return used


RyszardParys:
Finally, I can meow at the duke!

Savotta:
nvm lol apparently i can fix it just like that heeheehaw

Foxentail:
Half-kin should get this too.

Savotta:
> I have the options as a non-Tabaxi(As God intended.) but I've no way to use them. Why show them in the first place? Properly species lock them!!!!!!!

This has been fixed by the last 3 commits I pushed, properly race-locking it to only be visible to Tabaxi, they just need to be approved. I think. If not, something is broken, but it works just fine on local.

>Half-kin should get this too.

No.

JonathanTheII:
Half-kin SHOULD get this too.

Gyrandola:
> Half-kin SHOULD get this too.

Nah.  We hate felinids.

RadiantFlash7:
> > Half-kin SHOULD get this too.
> 
> Nah. We hate felinids.

So true

Gyrandola (approved):

PR #2179, opened 2025-01-09
Cleanup 9 & Unit tests

Gyrandola:
## About The Pull Request

Pretty much title. Unit Tests grabbed from Vanderlin.
Additionally ports https://github.com/Vanderlin-Tales-Of-Wine/Vanderlin/pull/278.
TMing this just to make sure nothing breaks as always.


Gyrandola:
> Removes skinning results, and seemingly breaks bundles? Also seems to gut standard unit tests, unless I'm mistaken?

AFAIK that skinning is TG stuff and it was based on the TG stack system we dont really use either. 
Bundles should be fixed now, giving it another TM.

PR #2184, opened 2025-01-10
Wood planks are carpentry

Xaytan:
## About The Pull Request

Switches wood plank crafting from lumberjacking to carpentry.

## Why It's Good For The Game

Refining raw materials into building materials should work off the building skill, now that you use wood planks for building.


Xaytan:
Actually, hold my beer.

PR #2197, opened 2025-01-11
Pop locks stockpile autogeneration

Gyrandola:
## About The Pull Request

Locks stockpile resource autogeneration to only fire when server pop is 25 or less.

## Why It's Good For The Game

Hopefully allows for a less artificial economy that relies on player work instead of stimmy checks.

### DIFF
--- a/code/controllers/subsystem/rogue/treasury.dm
+++ b/code/controllers/subsystem/rogue/treasury.dm
@@ -60,9 +60,10 @@ SUBSYSTEM_DEF(treasury)
 						X.demand += rand(5,15)
 					if(X.demand > initial(X.demand))
 						X.demand -= rand(5,15)
-		for(var/datum/roguestock/stockpile/A in stockpile_datums) //Generate some remote resources
-			A.held_items[2] += A.passive_generation
-			A.held_items[2] = min(A.held_items[2],10) //To a maximum of 10
+		if(num_players() <= 25)
+			for(var/datum/roguestock/stockpile/A in stockpile_datums) //Generate some remote resources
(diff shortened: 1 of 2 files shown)


Purpleturtle3:
>Tfw no one plays miner anymore

DelineFortune:
Its not grimdark enough because almost every problem is going to be solved with a code

Every towner even fucking half orc with 6 int can fucking read and sew artery 
Food cannot spoil any more
Auto-stockpile

Id rather to see this system turned off
We have dozen+ of leadership roles that wojak mentallty instead of real nobles
Counsillors? Its rare to see any good of them, people play them for erp and free 4 wrestling

If players have a problem that they have no soilsons / miners / hunters, the duke / a nobleman can and must organise people to work in mines/fields, or force them, or make its kind of punishment (aka grind us 10 gold for kicking a noble and go away)

Elli-Skala:
Do you think making it so the cool down for the stock pile to auto refill be based on the amount of unique keys spawned in multiplied by minutes would be a good balance between super low pop and higher pops?

Gyrandola:
> Do you think making it so the cool down for the stock pile to auto refill be based on the amount of unique keys spawned in multiplied by minutes would be a good balance between super low pop and higher pops?

Elaborate? Give me an example

Elli-Skala:
> > Do you think making it so the cool down for the stock pile to auto refill be based on the amount of unique keys spawned in multiplied by minutes would be a good balance between super low pop and higher pops?
> 
> Elaborate? Give me an example

20 players would mean stock pile refills every 20 minutes, 70 players it refills every 70 minutes making it way less reliable

PR #2200, opened 2025-01-11
Mob attack speed, bow clickability, rain filling buckets

NPC1314:
## About The Pull Request

* Bows, arrows, quiver clickability increased. Just reduced amount of times you try to click them and miss.
Also increases arrow damage by 1, since theyre so weak its comical.

* Most simplemobs spam attacks less fast now. (simplemob AI update messed with their effective swingdelay, these numbers are better)

* Ports this (rain fills open containers)
https://github.com/Vanderlin-Tales-Of-Wine/Vanderlin/pull/265

* Stone pot slightly darker than metal since it used same icon. Fibers a little MORE contrast again, getting there.

* Added the platform fix path typo


NPC1314:
> This is filling containers that are inside. Obviously not an ideal happening.

I´ll disable it for now, probably some flag missing in this repo.

NPC1314:
The fix works in localtest, no more leaky roofs

NPC1314:
So the label is incorrect, not in TM for a while it seems. Closing time?

Spaghetti-bit (changes requested):
Needs to be atomic, keep separate changes to their own PRs. Sprite fixes etc.

Mob attack speed - own PR
Bow clickability - own PR
Rain filling buckets - own PR etc.

Spaghetti-bit:
Please reopen a few PRs to keep things atomic. Thank you for the work you've done though!

PR #2202, opened 2025-01-12
Farming but worse

Xaytan:
## About The Pull Request

Harvesting non-perennial crops like wheat and ruining the produce actually ruins the produce. Get new seeds. Plant again. Wait for it to grow.

Failing to harvest produce now damages perennial plants.

Reduced farming XP granted on unsuccessful harvest.

## Why It's Good For The Game

Farming is now more difficult and expensive to do at low levels. This will reduce output of low skill farmers, and reduce the speed at which non-farmers can learn farming. Also, it's realistic!

image

### DIFF
--- a/code/modules/farming/soil.dm
+++ b/code/modules/farming/soil.dm
@@ -587,7 +587,8 @@
 /obj/structure/soil/proc/ruin_produce()
 	produce_ready = FALSE
 	if(!plant.perennial)
-		uproot()
+		adjust_weeds(-100)
+		plant = null
 	update_icon()
 
 /// Yields produce on its tile if it's ready for harvest


Xaytan:
Don't merge this, I broke something.

Xaytan:
Okay fixed it

hugmeorelse:
I understand what the intent is, however, it does not align with reality. This would make sense if farming was a popular mechanic people enjoyed doing to begin with. However, as indicated by the jobs people take, farming is merely something people do because they have to, not because they enjoy it. Doing this only exacerbates current problems and will cause players to become ever more reliant on auto-stockpile. Harder isn't better.

Instead of making things arbitrarily more difficult, adding further depth to farming would do so much more. It would address the issue you're trying to resolve and make the mechanics more interesting. Doing this, I feel, will do more harm than good.

Overall, I'm very much skeptical about this PR.

radarseas:
the issue this brings is keep has 4000+ mammon to throw at the wall, this doesn't fix maids.

your average mage adventurer/barber doctor interacting with other players to get fishing done and having to substitue for soilhands saying _"fuck off"_ results in them not being able to really grow anything, like westleaf, screwing them out of making any complex drugs.

not great for the player ecomony, unless you give potion brewers or barber docs farming skill which kinda breaks balance because they'll just play medical/mage farmer which is dumb.

Xaytan:
You are very much overestimating the added difficulty.

It is a very small increase. The game is currently plagued with skills that give near full functionality at extremely low skill ratings.

Reworking farming to be more engaging will, hopefully, be a result of porting Neufarming. That is not the intention of this PR. This PR is intended to help address the ongoing problem that doing other people's jobs is really easy.

radarseas:
> Reworking farming to be more engaging will, hopefully, be a result of porting Neufarming. That is not the intention of this PR. This PR is intended to help address the ongoing problem that doing other people's jobs is really easy.

until towner's subclasses are split so you know there's freelance farmers + fishers as a barber doc and don't have to attempt to speedrun level and do everything yourself. Its not really possible.

towner's biggest flaw is that's its _8+ classes that all rely on eachother in one_. Nobody knows how many of everyone else there is and its why we get "whoops all miners, whoops all smiths, whoops all fishers problems."

Its also probably why nobody plays towner, it sounds too useless when in actuality some of the classes like hunter can be rather fun.

PR #2206, opened 2025-01-12
D1 bandit hating

wow50000:
## About The Pull Request
Removes all int bonuses to bandit, sets all classes to a -2 int modifier, removes noble blooded from the hedge knight(Azure slop port)

## Why It's Good For The Game
Int is used in crafting and feints, bandits should not be crafting or learning any skills this easily with the support classes given.

Noble blooded was removed because mattheos hates nobles.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/bandits/types/sawbones.dm
+++ b/code/modules/jobs/job_types/roguetown/bandits/types/sawbones.dm
@@ -38,7 +38,7 @@
 	ADD_TRAIT(H, TRAIT_NOSTINK, TRAIT_GENERIC) // Vital for any surgical role dealing in potentially decaying bodies.
 	H.change_stat("strength", 2)
 	H.change_stat("intelligence", -2)
-	H.change_stat("speed", 1)
+	H.change_stat("speed", 2)
 	H.change_stat("fortune", 1)
 	if(H.age == AGE_OLD)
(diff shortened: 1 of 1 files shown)


RadiantFlash7:
I'm afraid that isn't going to work. You can't give int penalties to mages. Their casting becomes drastically longer, and it makes them useless.

Sawbones was the role intentionally designed as a support role- with only a single one able to be picked- Frankly it should be able to craft and learn things.

Bandit mages got added after, after I was pestered to enable them.

Dilligentdog:
fastest thumbs down i have ever given a PR, this is a slop change especially after bandits have become useless to play till either a gutted hoardmaster is brought back or the old idol is brought back

wow50000:
> fastest thumbs down i have ever given a PR, this is a slop change especially after bandits have become useless to play till either a gutted hoardmaster is brought back or the old idol is brought back

true lmao

radarseas:
Nobles can have Matthios as their patron anyway, _**this makes absolutely no sense.**_ Its also for RP flavor, they don't need the actual benefits, _I'd argue vlord should have noble-blooded too but meh._

-2 int also means ALL of these classes can be fiented insanely easy by paladins and templar whom get a fucking redicious amount of int, _god why would you do this? there's like five throughout 3 hours._

PR #2208, opened 2025-01-13
Bandidos

NPC1314:
## About The Pull Request

Bandits in a bad spot, re-enabling the old statue was mentioned as the alternative.
This is one take.

* More random equipment, knave rogue gets a hood & 1 more lockpicking. Slightly better average gear but not by much.
Less uniform in appearance, harder to identify on clothing besides stuff like helmets and steel masks.
* Re-enables the rewards from Mathios statue oldstyle. Modified loot table, 1st step healing potions.
* Sawbones & Bandit wizard replaced with Matthosian cleric called Zealot.
Spells same as matthosian churchling, got dodge expert but not great speed, shitty armor and a unique 2 handed flail.
Reason is simple, splitting up a tiny team of bandits into combatants and support not had a good outcome IMO.
Cleric meant to stick with the gang, provide some healing and some unusual combat style.

If the Hoard is re-enabled the lineup should stay IMO, if they do get restyled to city-gang the sawbone would fit better than the zealot but this is to make them more fun today, here, now.
With the commented out stuff its easy to switch between Hoard & oldstyle.

Also needs a few map tweaks but I cant solve that right now.


DelineFortune:
>Spells same as matthosian churchling, got dodge expert but not great speed, shitty armor and a unique 2 handed flail.

Rip

Rayforx:
Removing the sawbone that only a few people really play, and bandits do not really use the service off, I'm pretty sure that nobody is going to say anything about it, and having it replaced with some easier-to-use healing skills would be a huge plus.

But removing the mage, is weird and does not make any sense with the PR in question.

DelineFortune:
Make people able to choose between zealot and sawbones 
Sawbones have to carry bandits with crafts/smithy, surgeries and medicine also they have good rp
They also can revive fallen comrades/people 
Two mages are literally 50% of bandits strength 

Whats bandits getting for those?
Copypaste lesser miracle line
Copypaste war flail with rename

radarseas:
I think zeolot would be better as a compliment to these two classes, they're good in their own way, but worse in others. It adds more variety.

sawbones can revive, these guys only do miracles? sawbones is actually rather stronk with their rapid treatment ability and stabilising syringe, that's not considering the cardac massage revival they can do. They are invaluable because of it and their guarrenteed loyalty.

I don't know what to say of the mask randomisation, keep it steel or downgrade to rags and make em get steel masks from the idol, one or the other. Rags will probably lessen earlygame meta since anyone can make one, no refugees start with steel masks to my awareness.

NPC1314:
Eh, honestly without Gyran you guys in trouble, seems not worth the hassle trying to fix your bandits. Best of luck

PR #2209, opened 2025-01-13
Scruffing for Lupians and Tabaxi

Savotta:
## About The Pull Request

Makes it so Lupians and Tabaxi are now scruffable, as per the PR's title.

This also reintroduces neck-twisting. It's functionally the same as twisting one's head, I don't know why neck-twisting was removed in the first place.

To scruff a Lupian/Tabaxi, simply aggrograb their neck and twist. It won't work if the neck is inaccessible due to armour, using the same check as choking. This paralyses them for 2 seconds.

Middle-aged and old Lupians/Tabaxi aren't affected by this, the same way older cats and dogs are probably used to being scruffed.

## Why It's Good For The Game

I don't know if I would call this 'good' for the game, but I've been in the mood to PR race-specific things. It's also pretty funny, despite the fear of it being kind of overpowered. Tabaxi and Lupians now have yet another reason to cover their necks up, and nerfing anthros is all the rage anyway.


Dilligentdog:
will think be turned into a meta move? Possible. Is it funny? very.

100% down for this as a tabaxi main

PR #2210, opened 2025-01-13
More ear hiding

wow50000:
## About The Pull Request
Adds ear hiding to everything that covers ears and is protective, removes ear and nose protection from tricorn because ????

## Why It's Good For The Game
{4A4D1BDD-860D-4B23-817F-22425611B5DA}
ears are soft matter, just wear a helmet that doesn't cover ears.


Spaghetti-bit:
Logically makes sense but the drip loss 😭

Useroth:
No on hiding ears.
You can make the PR separately for the other thing

PR #2218, opened 2025-01-14
The pit

Rayforx:
## About The Pull Request
This adds an alternative to the Acid Pit, a horror show for admins.

Video of the pit
Yes, Falling on the floor hatches will break the fall even when it's activated. If the bastion forgot to turn off the hatches, then it's on them.

https://youtu.be/v2BY9ve5COE

## Why It's Good For The Game
Admins should not deal with tickets from an instant death trap, or a person being griefed with no body left, this, makes it very easy for Bastion to defend, and a lot more fun to deal with.

### DIFF
--- a/_maps/map_files/rockhill/rockhill.dmm
+++ b/_maps/map_files/rockhill/rockhill.dmm
@@ -502,12 +502,6 @@
 	name = "DROP";
 	redstone_id = "lava_bastion"
 	},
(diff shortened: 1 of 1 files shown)


Rayforx:
This will be closed later, as willy will add this to his own PR

Rayforx:
Added into another PR 👍

PR #2220, opened 2025-01-15
Disables Maniac

ProbablyCarl:
## About The Pull Request
Huzzah. We're free from the mockery of this bargain brand Dreamer.

Simply comments it out, in case we want to return it in the future, for whatever reason.
I considered removing the code relating to latespawns as well, but that is already set to return before it does anything, so it's not even something to bother with..

## Why It's Good For The Game
Do I really need to explain this?

### DIFF
--- a/code/game/gamemodes/roguetown/roguetown.dm
+++ b/code/game/gamemodes/roguetown/roguetown.dm
@@ -204,9 +204,11 @@ var/global/list/roguegamemodes = list("Rebellion", "Vampires and Werewolves", "N
 		pick_aspirants()
 		log_game("Minor Antagonist: Aspirant")
 
+/*
 	if(prob(10))
 		pick_maniac()
 		log_game("Minor Antagonist: Maniac")
+*/
 
 	return TRUE
 


Kregle:
Constant issues from the staffie side of things. The community doesn't seem to find any enjoyment from this role. Possibly replace it with another solo/intrigue antag role in the future.

hugmeorelse:
Yes, I have yet to ever see anyone ever roleplay as Maniac. I do think an antagonist that focuses on driving the townsfolk insane is great, but this is an angle rarely ever done by the antagonist itself and is done by the players. Overall, I think this is the angle we should be playing into, less of the fragging conversely.

I wonder if may be beneficial to maybe tie them into a sort of wizard gone mad with the intent of making others see things in his own way to spread havoc.

Limeon666:
While I am not the biggest fan of the maniac I do think straight up cutting content is a shame. It's all good to say 'we'll replace this with another antag later' but in all likelyhood that's a lot of work and won't happen, or won't happen for a long time.

I think in particular the creations they make are very interesting and fit the setting, and them being around in some other form (perhaps in certain areas of the map like the Zizo church) would be a good way to retain content.

wow50000:
Dogshit antag, it's like if an antag randomly round ended on a D20

EvenInDeathIStillServe (approved):

badassbilly:
Commenting out code is hack, by the way. Just delete it and learn to use git.

PR #2221, opened 2025-01-15
Leather coats are now cheaper

john-pawson:
## About The Pull Request

While i was making the leather coats, i made them expensive to make, a bit... too expensive actually.
The normal leather coat needed so much cured hide, that most seamsters didn't even know it existed since it never showed up on the crafting menus, not only that, but it had the highest amount of needed hides out of all the other things.
And the heavy variant needed 4 furs, which i did not realise made it the same or even more expensive than FULL STEEL PLATE ARMOR when it was time to go ahead and buy the materials.
So yeah, hopefully this'll make more of an option, instead of a fancy way of throwing your mammon away and confusing the seamster after they make a cloak thinking you just mispelled it when making an order.

## Why It's Good For The Game

More options for the light armor enjoyers that aren't ridiculously expensive or obscure now, and also more variety on what clothing people wear.

### DIFF
--- a/code/modules/roguetown/roguecrafting/sewing.dm
+++ b/code/modules/roguetown/roguecrafting/sewing.dm
@@ -585,10 +585,10 @@
 	skill_level = 3
 
 /datum/crafting_recipe/roguetown/sewing/leathercoat/heavy
-	name = "heavy leather coat (3 furs, leather, fiber; EXPERT)"
+	name = "heavy leather coat (2 furs, 2 leather, fiber; EXPERT)"
 	result = /obj/item/clothing/suit/roguetown/armor/leathercoat/heavy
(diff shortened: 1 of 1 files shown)


radarseas:
I'd lessen it to two furs honestly and up the required cured leather to 2 as well.

it makes no sense its almost entirely made of fur.

john-pawson:
> I'd lessen it to two furs honestly and up the required cured leather to 2 as well.
> 
> it makes no sense its almost entirely made of fur.

Well tbh i considered it to be the heavy version because it was made almost entirely of fur, like the heavy leather gloves. But to be honest, it might be better this way like you said. It's not *that* good anyways, compared to the other armors.

EvenInDeathIStillServe (approved):

PR #2228, opened 2025-01-15
Climbable tent walls and expensive twig platforms

Hmeister-real:
## About The Pull Request

Both are now significantly weaker, tent walls having 150 integ and twig platforms having a measely 80, making them extremely easy to break and deal with quickly. Tent walls need journeyman to climb, meaning that Guards, bandits and rogues can climb them, but not regular towners. Twig platforms are craftable again, and now cost 3 twigs + rope since you aren't bundling twigs together with a single piece of fiber, should make them significantly harder to craft and more costly without outright removing them.

## Why It's Good For The Game

Allows for more divesity when traversing the map and helps bandits be more sneaky again instead of being a frag antag

### DIFF
--- a/code/game/turfs/closed/wall/roguewalls.dm
+++ b/code/game/turfs/closed/wall/roguewalls.dm
@@ -144,14 +144,14 @@
 	icon_state = "tent"
 	smooth = SMOOTH_FALSE
 	blade_dulling = DULLING_BASHCHOP
-	max_integrity = 300
+	max_integrity = 150
 	break_sound = 'sound/combat/hits/onwood/destroywalldoor.ogg'
 	attacked_sound = list('sound/combat/hits/onwood/woodimpact (1).ogg','sound/combat/hits/onwood/woodimpact (2).ogg')
 //	canSmoothWith = list(/turf/closed/wall/mineral/rogue/wood, /obj/structure/roguewindow, /turf/closed/wall/mineral/rogue/wooddark)
 	above_floor = /turf/open/floor/rogue/twig
 	baseturfs = list(/turf/open/floor/rogue/twig)
 	neighborlay = "dirtedge"
-	climbdiff = 99 //how are you even supposed to climb a tent??
+	climbdiff = 3
 	explosion_block = 0
 	hardness = 70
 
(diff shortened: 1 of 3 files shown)


RadiantFlash7:
tent wall climbing was removed because it made walls trivial. I don't really think re-adding it is a great idea, honestly.

Hmeister-real:
> tent wall climbing was removed because it made walls trivial. I don't really think re-adding it is a great idea, honestly.

They can be broken in like 4 mace swings and you can't get into places with them anywhere except for very specific locations, without at least one platform alongside it

RadiantFlash7:
> > tent wall climbing was removed because it made walls trivial. I don't really think re-adding it is a great idea, honestly.
> 
> They can be broken in like 4 mace swings and you can't get into places with them anywhere except for very specific locations, without at least one platform alongside it

Point is, it made walls completely trivial. It doesn't matter how strong it is, the point of it is that it was used to scale walls. Every time. Over any other options, when available. Just because it can be broken easily after the fact, doesn't change that.
You can easily slap twig platforms and tent walls together to make a staircase to climb litterally anywhere, with the most basic ingredients- It's fairly egregious. Roof of the merchants, roof of the church, scaling the keep's walls, scaling the town wall, the former bastion.

Xaytan:
Yeah uh, don't do that lmao.

Walls shouldn't be trivial to circumvent like this.

Hmeister-real:
Tent walls need Journeyman climbing to get use out of, most people only get apprentice. Even Rogue cutthroats don't get past Apprentice, so the amount of roles that can get use out of it are very limited, intended for bandits to have an easier time being sneaky while also allowing guards to follow them (guards get journeyman)

john-pawson:
> Tent walls need Journeyman climbing to get use out of, most people only get apprentice. Even Rogue cutthroats don't get past Apprentice, so the amount of roles that can get use out of it are very limited, intended for bandits to have an easier time being sneaky while also allowing guards to follow them (guards get journeyman)

Yeah no, same thing as the others said, maybe it's better if building a tower to climb up a wall needs at least a bit of effort, and don't forget, you can lower the amount of skill needed by just getting a chair.

Hmeister-real:
> I think it would be better to have the tent walls climbing and the twig platform PR split into two since the tent wall PR does come with a few problems with balance.

Il move the platforms to a new PR if this one gets closed

Hmeister-real:
Will make one for just platforms instead

PR #2233, opened 2025-01-16
Aasimar's customization 

SayonaraEcho:
## About The Pull Request

- Adds Aasimars some feathery wings.
- Adds new colour skin.

image

## Why It's Good For The Game

More customization for aasimars players because they're not unique and looks like human but with different colour. This PR helps aasimars be more special on this build. 

Requested by Anya Dawnheart and Twilight Fortress community.

### DIFF
--- a/code/modules/surgery/organs/feature_organs/wings.dm
+++ b/code/modules/surgery/organs/feature_organs/wings.dm
@@ -25,19 +25,15 @@
 /obj/item/organ/wings/anthro
 	name = "wild-kin wings"
 
-/obj/item/organ/wings/flight
-	unconditional_flight = TRUE
-	can_open = TRUE
-
-/obj/item/organ/wings/flight/aasimar
+/obj/item/organ/wings/aasimar
 	name = "angel wings"
 	desc = "A pair of magnificent, feathery wings. They look strong enough to lift you up in the air."
 
-/obj/item/organ/wings/flight/dragon
+/obj/item/organ/wings/dragon
 	name = "dragon wings"
 	desc = "A pair of intimidating, membranous wings. They look strong enough to lift you up in the air."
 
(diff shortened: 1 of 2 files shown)


EvenInDeathIStillServe:
The wings are marked as "flight"? I'd rather if they were regular wings. I don't want aasimar being able to just fly around.

Kevak:
Don't use the /flight wings. Use the /huge wings. They use the same sprites but the /huge wings don't call flight code at all. See how I handled it in [2234](https://github.com/Rotwood-Vale/Ratwood-Keep/pull/2234)

code/modules/surgery/organs/feature_organs/wings.dm
```diff
@@ -29,7 +29,7 @@
 	unconditional_flight = TRUE
 	can_open = TRUE
 
-/obj/item/organ/wings/flight/angel
+/obj/item/organ/wings/flight/aasimar
 	name = "angel wings"
 	desc = "A pair of magnificent, feathery wings. They look strong enough to lift you up in the air."
```
Spaghetti-bit, line 34:
```suggestion
/obj/item/organ/wings/aasimar
	name = "angel wings"
	desc = "A pair of magnificent, feathery wings. They look strong enough to lift you up in the air."
```

Aasimar should not be given wings that have flight. The code for flight isn't even in the code, should probably nuke those vars out as well.

image
image




Otherwise, code looks fine.

Be sure to update the pathing on their wings else where as well.

Kevak:
Use the wing paths I used in [2234](https://github.com/Rotwood-Vale/Ratwood-Keep/pull/2234), that'll integrate much better

Kevak:
In particular the /huge wings will give them the option to flare their wings out with *wing

Kevak:
You can just reopen your pr I'm p sure

PR #2235, opened 2025-01-16
Bedsheet Covers The Character Lying Down

SmiLeYre:
## About The Pull Request

Bedsheet Covers The Character Lying Down, but not covers standing character. The bedsheet is lost from view if it is hidden behind the visible FOV. This can be considered a bug and I haven't figured out how to solve it yet.

## Why It's Good For The Game

You can hide a character under a blanket. This change will add immersiveness to the build.


EvenInDeathIStillServe (approved):

ophaq:
Finally I can cover cute kobolds under the bedsheets.

PR #2244, opened 2025-01-17
SEELIE FUNCTIONALITY UPDOOT

Grigorix:
## About The Pull Request
Gives the seelie the jester and rogues ability to jump up a zlevel since climbing doesnt work for them and they have increased interaction timers regardless. Gives them the ability to pickpocket somewhat.

## Why It's Good For The Game
Allows seelies to transition zlevels which they currently are incapable of doing along with giving them the ability to at least attempt to steal some items and cause further mischief. Tough even 3 steeling is still a rather low chance to actually pull it off.


Spaghetti-bit (approved):
Code wise approved.

Not on balance.

Limeon666:
I love seelies, but I think they're already quite strong, and they easily have the most racial traits of any race. They also already start with climbing.

Grigorix:
might need to just buff their climbing skill if this is too strong as they can climb trees and rocks just not structures

PR #2255, opened 2025-01-19
Magical Item Framework + Silver Refactor

Xaytan:
## About The Pull Request

Creates a framework for enchanted items, as well as other special effects we may want to add such as special materials. Refactors silver to use this framework.

Opening as a draft for feedback, and because I know somebody was planning to use this framework in their own project.

To Do:

- [x] Create Framework
- [x] Move silver code into framework
- [x] Create handling for adding enchantments
- [ ] Create handling for removing enchantments
- [ ] Refactor silver code to be less terrible

## Why It's Good For The Game

Creates a framework for others to build on. Allows for a number of unique and interesting options in a modular and coherent way.

Stops silver being copy-pasted onto what, eight different items? That was what started this.


ophaq:
It would be cool if we had a magic properties/enchantment thing the artificer could do to apply onto weapons/armor at expensive costs. Like enchanting a sword to make people light on fire, higher durability armor enchantment, etc.

RadiantFlash7:
> It would be cool if we had a magic properties/enchantment thing the artificer could do to apply onto weapons/armor at expensive costs. Like enchanting a sword to make people light on fire, higher durability armor enchantment, etc. I can see it getting OP though, but artificer needs something more to do since barely anyone visits them.

At the moment, I'm planning on using this system for mages funnily enough. Plans are to have mages be able to make scrolls of enchanting in my magegameplay loop PR

code/datums/components/magic_item.dm
```diff
@@ -0,0 +1,33 @@
+//The base component for magical items.
+//Allows for item agnostic handling, and avoids needing to change the base item code
+//Most of our logic should be contained in the actual effects
+
+/datum/component/magic_item
```
dwasint, line 5: I would suggest a signal for adding an effect post creation so silver weapons could say be enchanted,

dwasint: also there already is an enchantment system in the game that supports weighted enchantments and stuff from the tarkov inventory

RadiantFlash7: The inventory PR thats getting reverted?

dwasint: I assumed it was still in because there are lingering files left over. But yea original point stands it should have a signal to add new enchantments. The alternative of reading the component is weird, I don't even really know why its a component to begin with tbh

Xaytan:
Included in mage gameplay loop PR

PR #2259, opened 2025-01-19
Twig platforms craftable again, but harder

Hmeister-real:
## About The Pull Request

Twig platforms can be crafted again, this time needing 3x twigs and 1x rope, along with Apprentice crafting. They are also much easier to destroy, so think of them as a sort of rope bridge/weak foundation

## Why It's Good For The Game

Them being gone made being sneaky a pain, also made building roofs for houses a massive pain and also made making scaffolding for 2+ story buildings rediculously more difficult without a team of masons/woodworkers, ALA The Moars, that you just cannot do as a solo Architect/Lumberjack.

image

Video proof. 7 axe chops with 9 strength to break it afterwards

https://github.com/user-attachments/assets/6da5ac9b-f9b3-4f95-8e64-ebbfb17871bc

### DIFF
--- a/code/modules/roguetown/roguecrafting/turfs.dm
+++ b/code/modules/roguetown/roguecrafting/turfs.dm
@@ -193,6 +193,7 @@
 	if(isclosedturf(T))
 		return
 	if(!istype(T, /turf/open/transparent/openspace))
+		return
 	return TRUE
 
 /datum/crafting_recipe/roguetown/turfs/tentwall


Spaghetti-bit (changes requested):
Include evidence of testing, a screenshot of a work platform / video of one being crafted will suffice.

Spaghetti-bit (approved):

PR #2260, opened 2025-01-20
Revival Semi-Salt PR

MarioTheShiba:
## About The Pull Request

Buffs revival window from 5 minutes to 10 minutes.

## Why It's Good For The Game

No one wants to be RR'd due to unforeseen circumstances, like being unable to find the head, the time it is required to move a body from a battle to a doctor, and the time it is taken to mend + attach a head back on and everything, it is time consuming.
This is technically a salt PR because I don't like seeing people get RR'd because their head decided to blend in with the ground, or because rot decided to kick in for no reason because it's tickbased, for some reason?????

Anyway, this should give more lee-way to allow physicians of all kinds, bandits, townies and court ones, to revive someone without actually getting fucked by external factors.

### DIFF
--- a/code/modules/spells/roguetown/acolyte/sawbones.dm
+++ b/code/modules/spells/roguetown/acolyte/sawbones.dm
@@ -94,7 +94,7 @@
 		to_chat(user, span_warning("The life essence was sucked out of this body."))
 		revert_cast()
 		return FALSE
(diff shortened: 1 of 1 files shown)


greed23:
Also restore the Priest being able to revive without a zizoid sacrifice, smh

PR #2268, opened 2025-01-20
New World Map Firearm Mapfix PR

Teragenic:
It's just Willy's PR's Firearm fix.


Teragenic:
It has conflicts due to Jager PR not being merged. Once that's merged there should be no conflicts.

PR #2271, opened 2025-01-21
Defiance Descriptor Tweak + Combat Indicator

Spaghetti-bit:
## About The Pull Request

Tweaks defiant to stand out more:
image
Adds a combat descriptor when combat mode is active:
image

Adds an overlay for when combat mode is active:
image

## Why It's Good For The Game

- Defiance: Needs to be something that really sticks out.'
- Combat Mode, Descriptor: Besides the `--` dashes added to a person's speech and the inability to take their armor off, there isn't much description for someone being 'combat-ready'
- Combat Mode, Indicator: Visual indicator that makes it easier to identify who is in combat. **This will need a vote by le community, and will not change the rules of combat, it's just a mechanical indicator.

## Testing

Loaded in, enabled combat mode, disabled it.
Died, yielded, anything that would disable combat mode.

All worked fine.

### DIFF
--- a/code/modules/mob/combat_indicator.dm
+++ b/code/modules/mob/combat_indicator.dm
@@ -16,13 +16,14 @@
 
 /mob/living/toggle_cmode()
 	..()
+	log_message("[src] has " + (cmode ? "enabled" : "disabled") + " combat mode.", LOG_ATTACK)
 	toggle_combat_indicator()
 
 /datum/mob_descriptor/combat_mode
 	name = "Combat Mode"
 	slot = MOB_DESCRIPTOR_SLOT_NOTHING
 	verbage = "looks"
-	describe = span_narsie("ready for combat!")
+	describe = span_artery("ready for combat!")
 	show_obscured = TRUE
 
 /datum/mob_descriptor/combat_mode/can_describe(mob/living/described)


Hmeister-real:
I don't like this at ALL
As someone who keeps combat mode on to dodge mid-type grabbers this is just a big "I am valid" logo above my head to anyone who wants to fight me. This is especially bad if paired with the fact the the hunted flaw already makes you valid, a merc walks up to you and sees that you go combative would be in the right, mechanically, to attack you on the spot. You can already tell they are tense you don't need a big flashing VALID marking, it's just not immersive or fun

Spaghetti-bit:
Waiting on more votes. But it seems like the combat indicator will not be added. The descriptor should stay however.

> I don't like this at ALL As someone who keeps combat mode on to dodge mid-type grabbers this is just a big "I am valid" logo above my head to anyone who wants to fight me. This is especially bad if paired with the fact the the hunted flaw already makes you valid, a merc walks up to you and sees that you go combative would be in the right, mechanically, to attack you on the spot. You can already tell they are tense you don't need a big flashing VALID marking, it's just not immersive or fun

DelineFortune:
Seeing that person in combat mode will make them auto valid or understand that they are running away from fight for other valid hunters

Xaytan:
Please do not do this. Combat mode is critical for defensive purposes in this codebase. Adding an indicator means that the second someone turns it on they become valid. You'll see a lot more fragging.

Hmeister-real:
> This just adds the ability to see that someone is prepared for combat WITHOUT having to wait on them to speak. Or having to see that they're stopping grabs / stripping. Especially helpful if they're already down. This doesn't prevent any manner of actual RP, of course, because without escalation, they're still grounds for administrative action. If someone claims the indicator is 'escalation', we already have combat indicators in the form of chat notifications "--", so that's nonsensical. Makes it easier to determine, sure, but it's still just as easily handled.
> 
> While I can see both sides of the argument, I genuinely don't see it causing too many issues. Perhaps you could argue that it 'lets you prepare in advance', but so does chat.
> 
> That said, add logging to combat mode.

the "--" is good because it is only shown post speaking, which is perfect for actually beginning combat. If you see someone speaking with the combat indicator already above their head you are obviously going to pre-emptively ready yourself for combat. On top of this, the admins already have enough strain, this would 100% result in a massive increase in fragging and RR, which is already a massive admin issue by itself, but especially if people have the excuse of "Oh, but there was a massive red marker telling me that they were entering combat". You are just fixing a non-issue, I'm guessing for your own personal reasons.

Hmeister-real:
> > the "--" is good because it is only shown post speaking, which is perfect for actually beginning combat. If you see someone speaking with the combat indicator already above their head you are obviously going to pre-emptively ready yourself for combat. On top of this, the admins already have enough strain, this would 100% result in a massive increase in fragging and RR, which is already a massive admin issue by itself, but especially if people have the excuse of "Oh, but there was a massive red marker telling me that they were entering combat". You are just fixing a non-issue, I'm guessing for your own personal reasons.
> 
> My own personal reasons? This isn't my PR.

Change "You are" to "This is", meant to direct it at the creator not you sorry

hugmeorelse:
Going into "combat mode" is a necessary due to how grabbing and how combat works on the server. If you're not in combat mode and you get grabbed, it's game over for you. Anyone that attacks anyone just for being in combat mode ought to be slapped silly, and this is something that can be drilled into staff.

Ultimately, I do see both sides. I do like this, as having a visual indicator when someone is tensed up is nice. However, I also understand the concerns players have that this change will cause them to be seen as valid. We do have a subsection of players that operate in bad faith, looking for any excuse to unleash their mechanical kung-fu on someone. 

However, this problem can be resolved resolutely, as it's only a valid concern due to a lack of persistent moderation. I only interject by saying this was artificially induced by the leadership on the server. All the lessons that were learned from last autumn were thrown away at whim, seemingly because there were sudden insecurities about not being grimdark enough. My guess is someone got into their ear, not realizing that they were being misled. It does not help that there is a constant push to be imitate Lifeweb either.

Anyways, this may be fine in the future once these issues are resolved. However, until then, I do believe it is a legitimate concern from the playerbase. So, I do think we should hold off, up until we're in a good place as a server.

hugmeorelse:
My other concern is that having an indicator up there may not be the greatest for immersion. 

I have some alternatives, @ProbablyCarl @Spaghetti-bit, that come to mind that may fit the setting better.

What if we did something more subtle like having the character pixel shift back and forth ever so slightly as if they're active on their feet. Anyone who uses that as an excuse to frag will look like a moron, and it doesn't have a big indicator hovering over a player. Yet, it should give enough for a player to know they're "tensing" up to be in combat.

Or maybe a textual output like, "Player is tensing up...", whenever they enter combat mode. That should suffice too.

Let me know.

Koteshkin:
- [x] We have added defiant feature, but its just OOC feature, you still can rape people even with defiant on. We arent changing rules about it.
- [x] We have finished defiant and you no longer can rape defiants. If you do, you break rules.
- [x] You can freely kill people as long you let them know you are going to murder them and having enough reason.
- [x] You have killed a person for a bounty of 500 coins? Banned.
- [x] We have added combat indicator feature, but its just OOC QOL feature. We arent changing rules about it.
**WE ARE HERE**
- [ ] Combat indicator rules.

PR #2288, opened 2025-01-23
Bandits "uncapped"

notafurryy:
## About The Pull Request

Changes the bandit cap from a random number between 4 and 8 to a static 8, also changes the bandit to pop ratio to 1:6, meaning it caps out at 8 bandits at 72 player pop removing the random chance to have 4 at high pop, also brings back bandit migrant waves.

## Why It's Good For The Game

With all the nerfs the bandits got and the major upgrade to town security with most important things being moved into town and the bastion gate being next door to the retinue eager to reinforce them I think it's fair to bulk the bandit numbers up a bit, since right now they are not really a major threat, with their low numbers and the lack of good hoardemaster gear after the recent changes.

### DIFF
--- a/code/game/gamemodes/roguetown/roguetown.dm
+++ b/code/game/gamemodes/roguetown/roguetown.dm
@@ -230,11 +230,11 @@ var/global/list/roguegamemodes = list("Rebellion", "Vampires and Werewolves", "N
 	"Goblin Smith",
 	"Goblin Shaman")
 	var/num_bandits = 0
-	var/limit_bandits = 15
-	if(num_players() >= 6)
-		// 1 bandit per 6 players,
-		num_bandits = round(num_players() / 6)
-		if(num_bandits >= limit_bandits)	//caps bandits at 15
+	var/limit_bandits = 8
+	if(num_players() >= 9)
+		// 1 bandit per 9 players,
+		num_bandits = round(num_players() / 9)
+		if(num_bandits >= limit_bandits)	//caps bandits at 8
 			num_bandits = limit_bandits
 		var/datum/job/bandit_job = SSjob.GetJob("Bandit")
 		bandit_job.total_positions = num_bandits


radarseas:
yeah _this is a_ good change. Especally with how easily 4-5 bandits get destroyed with no counterplay when keep decides to send an army for the guy that stole someone's 6 mammons and how they constantly keep splitting up.

Its hard enough to even get a sawbones on your team so hopefully we'll see more people rolling that with this.

notafurryy:
> Bandits are, quite literally, without any hyperbole, impossible to detect currently. They don't need additional numbers, because they're not outright getting caught and gutted in the street nonstop, as prior, because "MASK!!!!!!!!!!" doesn't work anymore. Just because they lost a lot of their meta gear doesn't mean they're worthless now, either.
> 
> Additionally, the lack of numbers is intentional. We don't want them as a siege antag. Their entire purpose is to harass and steal. Not march around as a giant, impossible to combat blob of men who'll instafuck anything they come across. Even at ninety pop, with players spread out as they are, there's no genuine way this would be good for server health in any capacity.
> 
> We're working on fixing the exploit where you use 'known people' to detect bandits, atop that. The only way, beyond using the nervelock confirmation, interrogation or guessing, to confirm that they're a bandit.
> 
> On another note, however, bandit waves were intended to be brought back at one point. Their return was oh-so-coincidentally left out of a PR that undid the gutting of bandits. Not pointing fingers. They were forgotten about, albeit with vestigial elements left in the code for a long time. Would be nice to see how the waves perform if returned, admittedly, but upping their numbers isn't it.

A random chance to cap bandits at 4 with 100 people on the server isn't good either though, given that most bandits die 30 minutes into the round, also didn't realize the gutting pr removed the examine text, I think giving them some higher numbers corresponding to the server population still makes sense, even if 15 may be a bit too much.

notafurryy:
> That's not a 'bit too much'. That's an obscene number.
> If your issue is with the, admittedly, cancerous pick rate? Fix that, my brother in the Inhumen.

And done! Changed it to the higher end of what bandits could roll before and made the ratio a bit less generous

Spaghetti-bit (approved):

Spaghetti-bit:
@notafurryy make sure to account for the subclass system for the bandit raids. We may have cases where all the subclass positions are filled.

notafurryy:
> make sure to account for the subclass system for the bandit raids. We may have cases where all the subclass positions are filled.

I am pretty sure that only sawbones, hedgeknight and rogue mage have limited slot numbers, other subclasses don't seem to have those

PR #2247, opened 2025-01-17
Gives nightmaster expert knives instead of expert crossbows

Coolgat3:
Because it just feels like he should be able to use daggers well idk?

PR #2291, opened 2025-01-24
Revert "Gives nightmaster expert knives instead of expert crossbows"

Spaghetti-bit:
Reverts Rotwood-Vale/Ratwood-Keep#2247


Kevak (approved):
Looks like a functional revert. We need to yell at the person who didn't test their shit though nor provide proof.

PR #2293, opened 2025-01-24
Readds tipping knives, also adds salt water turfs

Kevak:
## About The Pull Request

This does a couple things:

-Atomizes and readds the ability to tip knives (and only knives) with various reagents. Something that was stealthmerged with the gridventory and then reverted alongside it. it also ensures it logs properly for admins- unlike Gyran's iteration which did _not_ log for admins at _all._
-You can once again poison knives by dipping them in reagent containers.
-Adds two new turf types for mappers, shallow saltwater and deep saltwater. Both can have salt water extracted from them, you can get salt by reacting it in an alembic. 
-Alembics can pull reagents into them via the fill intent. 
-Most alchemy reactions now must occur within an alembic or they will not react. 
-Oh and poisoned knives lose the reagent after a bit, so best taint them shortly before a fight.

Note: Alembic code is jank moricode and due a refactor. I had to use a janky ass workaround of allowing alembics to directly fill themselves to get the salt reaction working inside the alembic. Couldn't enable filling directly via bottle- because again, janky moricode. Gonna refactor it later.
Note: Did not add salt turfs anywhere. That's just for mappers to use.

## Why It's Good For The Game

Readds a feature that was lost with the removal of the buggy as fuck gridventory, also properly atomizes what should have been atomized in the first darn place instead of snuck into gridventory without mention.

Also options for mappers to make the sea not clean water. Among other uses.

## Proof Of Testing

https://github.com/user-attachments/assets/77eae45a-1134-4f11-8fcc-d544024c2de0

Screenshot 2025-01-23 195426
Screenshot 2025-01-23 185909
Screenshot 2025-01-23 195314

https://github.com/user-attachments/assets/51e3f45c-ee9f-4e59-83e0-7a2772f5d040

### DIFF
--- a/code/modules/spells/roguetown/acolyte/sawbones.dm
+++ b/code/modules/spells/roguetown/acolyte/sawbones.dm
@@ -1316,7 +1316,6 @@ end recipe count: 8 ash, 8 minced meat, 4 swampweed, 2 poisonberry to make 1 bot
 	brewing_started = FALSE // Reset brewing status after brewing completes
 
 /obj/item/reagent_containers/glass/alembic/attackby(obj/item/I, mob/user, params)
-	//This is a touch jank of a snowflake proc. I'll prolly redo it when I refactor alchemy later, but it's a solid stopgap for now ~Neri
 	if(istype(I, /obj/item))
 		if(user.mind.get_skill_level(/datum/skill/misc/treatment) <= 2)
 			to_chat(user, span_warning("I don't know how this works."))


notafurryy:
Making goblin finally useful? Waow

ophaq:
Does drinking salt water dehydrate you? I think it should decrease your thirst level further when drunk.

Kevak:
> Does drinking salt water dehydrate you? I think it should decrease your thirst level further when drunk.

It does yes! Doesn’t have any alcohol reaction though. But yes, it makes you rather dehydrated rather fast.

Spaghetti-bit (approved):

PR #2297, opened 2025-01-24
Shunned races DLC. Forbidden surgeries, artefacts crafts and organs craft 

DelineFortune:
## About The Pull Request

All shunned races going to get natural dark magic resistance because of their origin. From graggar horc to erp tiefling maid
It gives them some resist (less permanent dmg) while they are a subject of dark miracles, surgeries and will allow them to use dark artefacts and craft them 

Shunned races surgery list:

This shit takes 120 seconds to perform, so they can do it they could simply kill you already and hide your head/body

Lobotomy - takes -5 of their int, gives them +1 cons +1 end, makes them *droll randonly -15 stress). Cannot become rebels or bandits after it.
Pacification - same but +pacifist trait 
Arcane block - you cannot use magic spells for 30 minutes 
Memory extract - if they are bandit or goblin you are getting special item to use it and get knowledge about their hideout
Memory vanish - removes one skill from target (I want to allow to make it sell to transer to other players so you can use combat roles as experement's material) 
Vivisection = 20 permanent dmg per 120 seconds, 10 if they are shunned.  It takes 10 minutes to kill a person but i guess leaving your opponent alive, with permanent wounds and able to move is better than killing them sometimes and random cleric/doctor will not spam red/middle mouse button to heal them
Shylock removal = removes skylock implant from you
Extract soul = takes 3 mintes and makes specific soul depends of pacients race. 25 dmg if not shunned and 15 if shunned. 
Main thing to pay in dark rituals/make heretic items like artefacts/organs+

Organs craft:
From now i have heart that puts 15u of red into your blood once you are getting crit
Mind parasite - slowly kills people if they have no special drug inside them

Artefacts craft:
Rings that can make you understand other people depends of soul you had to put into it before
Kobold soul? Draconic. And etc. 

Rituals:
Ill still @Gyrandola's sacrifice sigil 
Dark marriage - both of partners are getting skill that can be used even if they are dead/knocked
It will teleport them on their partner's tile gibbing their partner in process
Naked and fully healed 

STILL LOOKING FOR IDEAS!!

## Why It's Good For The Game

i have no idea why shunned races are shunned at all no unique content 100000000000000000000000000 coded job/content locks

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/furry/anthromorphsmall.dm
+++ b/code/modules/mob/living/carbon/human/species_types/furry/anthromorphsmall.dm
@@ -12,7 +12,7 @@
 		LIPS,
 		HAIR,
 	)
-	inherent_traits = list(TRAIT_NOMOBSWAP)
+	inherent_traits = list(TRAIT_NOMOBSWAP, TRAIT_DARK_RESIST_NATURE)
 	inherent_biotypes = MOB_ORGANIC|MOB_HUMANOID
 	attack_verb = "slash"
 	possible_ages = ALL_AGES_LIST
--- a/code/modules/mob/living/carbon/human/species_types/furry/kobold.dm
+++ b/code/modules/mob/living/carbon/human/species_types/furry/kobold.dm
@@ -7,7 +7,7 @@
 	desc = "<b>Kobold</b><br>\
 	In time you will learn the history of this race."
 	species_traits = list(EYECOLOR,LIPS,STUBBLE,MUTCOLORS)
-	inherent_traits = list(TRAIT_NOMOBSWAP)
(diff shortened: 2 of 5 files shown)


Coolgat3:
WE WANT LOBOTOMY!
WE WANT LOBOTOMY!
WE WANT LOBOTOMY!

ShiningM00N:
My main concern reading these is justifications for the shunned races knowing any of these. Unless this is some sort of role/subclass thing they can specifically roll or something.

BTW, try making the harlequin merc a thing now.

Coolgat3:
> It gives them some resist (less permanent dmg) while they are a subject of dark miracles, surgeries and will allow them to use dark artefacts and craft them

"It gives them some resist (less permanent dmg) while they are a subject of dark miracles, surgeries and will allow them to use dark artefacts and craft them" I believe it's just surgeries and dark miracles doable ON them from the way it's written, no?

PR #2299, opened 2025-01-25
Letting carpenters do the their job

notafurryy:
## About The Pull Request

gives the woodcutter a handsaw

## Why It's Good For The Game

woodcutters are supposed to be ALSO carpenters, which the ability to do so diminished after the pr that added planks and stone bricks as needed elements for construction. 

It's a single changed line


ophaq:
Hammer is mainly used to smash walls but to be honest it almost never gets used. I'd rather have a saw so good PR.

Elli-Skala:
you could also move the axe to the back and just give them a hand saw also

Spaghetti-bit (approved):

PR #2306, opened 2025-01-25
Runtime Fixes

Spaghetti-bit:
## About The Pull Request

Fixes Icon runtime - This would generate.... something like 2 thousand runtimes? Easily.
Fixes SQL out of bounds query
Fixes AI runtime - Rats will no longer become the plague of humanity (my eyes)
Fixes trait runtime
Fixes sewing runtime
Fixes organ slot runtime
Fixes reader runtime - Map loading no longer causes 2 MILLIIIOOOON runtimes.
Fixes observer verb runtime
Fixes wound runtimes
Fixes smelter runtime

## Why It's Good For The Game

Runtimes are not good.

Compiled.
Sewing tested, no longer run times.
Organ slot no longer goes out of bounds.
Wounds tested on death.
Excommunicated special no longer runtimes.
AI... has not been tested but it didn't break anything? (Will need to be tested further)
SQL is backend, compiled.
Icon no longer runtimes.
Observer no longer runtimes.

image

### DIFF
--- a/code/game/objects/items/rogueitems/needle.dm
+++ b/code/game/objects/items/rogueitems/needle.dm
@@ -69,7 +69,7 @@
 
 
 /obj/item/needle/attack_obj(obj/O, mob/living/user)
-	if(isnull(O) || !istype(O, /obj/item))
+	if(isnull(O))
 		return
 	var/obj/item/I = O
 	if(can_repair)
--- a/code/game/objects/items.dm
+++ b/code/game/objects/items.dm
@@ -193,8 +193,6 @@ GLOBAL_DATUM_INIT(fire_overlay, /mutable_appearance, mutable_appearance('icons/e
 
 	///The appropriate skill to repair this obj/item. If null, our object cannot be placed on an anvil to be repaired
 	var/anvilrepair
-	//this should be true or false
-	var/sewrepair
 
 	var/breakpath
 
--- a/code/game/objects/objs.dm
+++ b/code/game/objects/objs.dm
@@ -46,6 +46,9 @@
 	/// Amont of multiplicative slowdown applied if pulled. >1 makes you slower, <1 makes you faster.
 	var/drag_slowdown 
 
+	///If the item can be repaired by sewing.
+	var/sewrepair = FALSE
+
 	var/blade_dulling = DULLING_BASHCHOP
 
 	var/debris = null


code/__HELPERS/icons.dm
```diff
@@ -790,9 +790,10 @@ world
 		var/list/icon_dimensions = get_icon_dimensions(curicon)
 		var/icon_width = icon_dimensions["width"]
 		var/icon_height = icon_dimensions["height"]
-		if(icon_width != 32 || icon_height != 32)
+		if((icon_width != 32 || icon_height != 32) && (icon_width != 0 && icon_height != 0))
 			flat.Scale(icon_width, icon_height)
-
+		if(icon_width == 0 && icon_height == 0)
```
Cyprex, line 795: What is passing icons with 0 width and height into this, and is that proper- or does this runtime because of an issue in code calling this?

code/game/objects/items/rogueitems/needle.dm
```diff
@@ -70,6 +70,8 @@
 
 /obj/item/needle/attack_obj(obj/O, mob/living/user)
 	var/obj/item/I = O
+	if(isnull(I) || istype(I, /obj/item))
```
Cyprex, line 73: Feel free to rename I to target_object(makes this whole proc more readable imo). I think you can simply do istype(I), but more importantly- do you actually want to return here if the target is an item? That seems incorrect to me- items to be repaired are items, no?

Cyprex:
```suggestion
/obj/item/needle/attack_obj(obj/item/target, mob/living/user)
	if(!istype(target))
```

Cyprex: Nevermind- after discussion on Discord it's probably better to keep the first argument inline w/ the parent proc

code/__HELPERS/icons.dm
Spaghetti-bit: Scaling with 0,0 is an invalid operation.

code/game/objects/items/rogueitems/needle.dm
Spaghetti-bit: Fixed.

code/modules/mapping/reader.dm
```diff
@@ -325,6 +325,9 @@
 	var/first_turf_index = 1
 	while(!ispath(members[first_turf_index], /turf)) //find first /turf object in members
 		first_turf_index++
+		if(first_turf_index > members.len)
+			first_turf_index = members.len
+			break
```
Cyprex, line 330: This is original tgcode, definitely odd that this runtimed.

Spaghetti-bit: It runs out of bounds.

Cyprex (approved):

ophaq:
Did this touch autoclosing character face/body window on movement cause that's whats happening if it did.

PR #2308, opened 2025-01-25
We do a little preloading (Preloading Improvements)

BlackjackIsTired:
## About The Pull Request

Added missing area music files to preload.
Added better, smoother preloading logic.
Added automatic preloading upon client connection with automatic emergency stop if the server's lagging too much.

## Why It's Good For The Game

Good performance is good for the soul. Healthy souls will make players more pious and spiritual, thereby improving immersion by making everyone a little bit more immersed in their ingame religious choice. (But seriously, better performance speaks for itself.)

Compiled.
Screenshot_11


Spaghetti-bit (changes requested):
Include the other combat sound files as well.

Love seeing this, thank you for doing it. <3

Hyro-Fox:
Where's Carl on this one, I want to listen to combat music without muting my game completely just so I get a smooth experience without freezing every 0000000.01 pico seconds.
(I fucking die when my screen unfreezes.l)

BlackjackIsTired:
> Include the other combat sound files as well. 
> Love seeing this, thank you for doing it. <3
@Spaghetti-bit 

Not really a good idea without implementing a system that only loads the music specifically applied to your mob. Otherwise you'd be loading like 11 more different combat music tracks that you'd never hear anyway. I've tested it and it makes the stutter way worse.

Regaaaaaardless. I've compressed and lowered the bitrate of all the music files that were **BLOATED**. You can't hear the difference at all, but you can NOTICE the difference in the stutter during preload and even when your music isn't preloaded. It loads on the spot much faster as a result, so cmode music doesn't even necessarily need to be preloaded, but it would help for smoothness at a later date.

(I unironically had to learn about audio engineering to reduce the size of the file without lowering quality in Audacity for this. It wasn't fun. But here's the results of my pain. Enjoy.)

PR #2311, opened 2025-01-25
Fatigue & Stamina Rework

BlackjackIsTired:
## About The Pull Request

Modeled with Real Life examples in mind. Yes, I did do research. Yes, I have sources.
Intense bouts followed by short moments of recovery.
Energy no longer materializes from thin air. Your energy drains from your Nutrition and Hydration pools equally. The more energy you use up, the more you'll need to eat and hydrate.(Which I think you'll agree is actually frickin' awesome.)
This also means that when you go to sleep, you burn a lot of your nutrition and hydration on recovering your energy. So you wake up with an empty stomach! Just in time for breakfast!
Reworked stam costs to be in line with the system. Swift is a totally valid strategy, but you'll burn out fast. The less energy you have, the slower your stamina regens, which means if someone manages to hold you off and you've burnt yourself out on SWIFT Intent. You'll lose the battle of attrition.
Laying down anywhere isn't so great for regenerating your energy anymore. Chairs are a good factor, yes they have a function now. Beds are the biggest factor, but you should also sleep if you want fast regeneration. I didn't wanna remove the opportunity sleeping gives an opponent, but I also didn't wanna make it so you have to sleep constantly. You can just use a comfy chair instead.
Running out of Stam is still very punishing, it keeps you stunned for 5 seconds. But the opportunity to knockdown with a kick is only 2.5 seconds. If you land a kick after 2.5 seconds while they're still stunned, it won't topple them. So while not managing your stamina properly or being exhausted is still very punishing. It's not necessarily fight winning or losing. By the time you're no longer stunned, you'll generally have about half your stamina bar back. There's also the added skill and reaction time factor involved in getting that early knockdown by landing a kick.
An important thing to note, Energy pools and Stamina pools were reworked with again, Real Life examples in mind. I'm very happy with the outcome. Endurance also increases the size of these pools.
Also, parrying is always half as expensive as attacking.
I also tweaked the energy bar displays a bit to make them more accurate. So you don't get scammed and out of stam'd when it looks like you should've had more.
Reduced some RNG in Grappling Stamina Costs.

## Why It's Good For The Game

It's REAAAAAAAAALLY FUN. You'll see in the clip, you can take on massive battles and it's SO INTENSE, IT FEELS SO GOOD. I've had a few testers agree as well. It makes combat much more engaging. Much more realistic and will also result in longer battles.

Combat Test: https://files.catbox.moe/beu4ef.mp4

Combat Test is too big for Github, but at least you get a chair demo:

https://github.com/user-attachments/assets/863f696c-280e-4a59-a1eb-1ad9517650da

### DIFF
--- a/code/modules/mob/living/carbon/life.dm
+++ b/code/modules/mob/living/carbon/life.dm
@@ -740,6 +740,9 @@ GLOBAL_LIST_INIT(ballmer_windows_me_msg, list("Yo man, what if, we like, uh, put
 	if(HAS_TRAIT(src, TRAIT_NOSTAMINA))
 		energy = max_energy
 		return
+	if(HAS_TRAIT(src, TRAIT_NOHUNGER))
+		energy = max_energy
+		return
 	if(energy >= max_energy)
 		return
 	if(nutrition <= 0 && !HAS_TRAIT(src, TRAIT_NOHUNGER))
@@ -789,8 +792,8 @@ GLOBAL_LIST_INIT(ballmer_windows_me_msg, list("Yo man, what if, we like, uh, put
 
 	var halfcost = round(regenround / 2) // On testing, I think having it drain from one source first and then the other causes it to drain WAY TOO FAST
 
-	var nutritioncost = min(halfcost, nutrition) // Makes it so it doesn't drain more than we have
-	var hydrationcost = min(halfcost, hydration) // Makes it so it doesn't drain more than we have
+	var nutritioncost = round(min(halfcost, nutrition) * 0.67) // Makes it so it doesn't drain more than we have
+	var hydrationcost = round(min(halfcost, hydration) * 0.67) // Makes it so it doesn't drain more than we have
 
 	var splitcost = nutritioncost + hydrationcost // So we split the cost, add them together and that is the pool we draw energy from.
 


Limeon666:
I'm skeptical of this, but it seems like it could be a good change. Testmerge and see how it is, maybe?

Spaghetti-bit:
After the current round of TMs I want this TM'd, see how it plays out.

BlackjackIsTired:
> This ain't it, chief. Hunger and water goes too quickly, and at the WORST times.

Made the changes we spoke about and fixed the NOHUNGER issue. It's ready for another TM.

Gyrandola:
Seems you addressed the previous issues, let's give this another try.

BlackjackIsTired:
> Its a bit too much. I ate and drank, tackled a single skeleton and decapped it easily and I was suddenly hungry and thirsty.

There's some clarification needed here, because there "hungry" and then there's "peckish". The first tier of alerts aren't debuffs, they're just to inform you that if you get the chance to eat, it's probably a good idea. Imo, it's better than suddenly being hungry and being hit with debuffs seemingly out of nowhere. To outline it clearly, the first debuffs you get for hunger and thirst are only alerts, have no debuffs. Iirc, you should be able to check that if you shift click on them.

Also, how much you eat and drink obviously matters.

PR #2312, opened 2025-01-25
Poison Essence is Poison

Xaytan:
## About The Pull Request

Makes poison essence, refined in an alembic, actually poisonous. 

It is now 30% more poisonous than unrefined berry poison. Which makes it pretty damn poisonous.

## Why It's Good For The Game

It makes sense. It's clearly intended to be poisonous. It has code for saying it's poisonous it's just... Very much not poisonous because I guess it was coded wrong as a joke???????

image


Xaytan:
It looks like some things slipped into this I didn't mean to

PR #2314, opened 2025-01-26
Requested rot changes

wow50000:
## About The Pull Request
Requested by carl on vc

## Why It's Good For The Game
fuck rot timers

ONE LINE CHANGE
{2A0F17B9-9C9F-48FA-9C8C-0188D649E709}

### DIFF
--- a/code/datums/components/rotting.dm
+++ b/code/datums/components/rotting.dm
@@ -58,7 +58,7 @@
 		qdel(src)
 		return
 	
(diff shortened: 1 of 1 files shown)


Dilligentdog:
oh, does this change it from being tick based? If so i heavily support this

Spaghetti-bit (changes requested):
Requires a small refactor where instead of amount we use world time so that it isn't tick based.

PR #2317, opened 2025-01-26
less broken strengthen undead

notafurryy:
## About The Pull Request

makes the strengthen undead give a debuff instead of being a stronger kneestinger stun

## Why It's Good For The Game

The strengthen undead spell currently is the most powerful stun spell that lasts longer than the kneestinger shock while not requiring a direct line to the target from the caster, being silent and having a really short cooldown making it pretty borked.

https://github.com/user-attachments/assets/3c1cea3c-62a5-41db-a75e-8814171d7a2e

### DIFF
--- a/code/datums/status_effects/rogue/debuff.dm
+++ b/code/datums/status_effects/rogue/debuff.dm
@@ -325,3 +325,15 @@
 	name = "Migraine"
 	icon_state = "muscles"
 	desc = "My head is pounding, I need to get away from the light and rest a while!"
+
+/// Strengthen undead debuff
+/datum/status_effect/debuff/weaken_living
+	id = "weaken_living"
+	alert_type = /atom/movable/screen/alert/status_effect/debuff/weaken_living
+	effectedstats = list("speed" = -3, "constitution" = -3)
+	duration = 25 SECONDS
+
+/atom/movable/screen/alert/status_effect/debuff/weaken_living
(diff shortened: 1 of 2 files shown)


notafurryy:
> You describe the issues of the spell, but instead of correcting those, you instead nerf the spell's effects? RAAAAAAA!!!! I get why, mind, but there's better ways to do it instead of number changes. Is this the final version of your change? No further adjustments?

The issue is I don't have the skills to make it interesting without making it a 1 to 1 copy of the lightning bolt with different flavouring, and even then it would diminish it's intended use, healing undead. I am open to suggestions though

radarseas:
personally I think its better just doing a debuff instead of a stun.

why? _its a you only need to see them, stun._ Not fun to deal with, at least lightning can't hit the back of a crowd.

I say this as someone that's heavily abused this as occultist to escape impossible fights I totally deserved to die in.

notafurryy:
> personally I think its better just doing a debuff instead of a stun.
> 
> why? _its a you only need to see them, stun._ Not fun to deal with, at least lightning can't hit the back of a crowd.
> 
> I say this as someone that's heavily abused this as occultist to escape impossible fights I totally deserved to die in.

that's a good idea, I might make it more similar to ravox's burden

Spaghetti-bit (approved):

PR #2320, opened 2025-01-26
Fixes Potential Necromancy Exploit

Kevak:
## About The Pull Request

Although I was unable to confirm the exploit when testing local on my own, (Likely due to a TM issue) it was reported to me that raise lesser undead and raise undead could instakill vampires by turning them into skeletons. This PR adds a sanity check to prevent that niche case, as well as slapping the sanity check flag on lich for good measure so we don't have situations of liches and necromancers oneclick killing other liches.

## Why It's Good For The Game

Instakill exploits are bad.

Screenshot 2025-01-26 065222

### DIFF
--- a/code/__DEFINES/traits.dm
+++ b/code/__DEFINES/traits.dm
@@ -73,7 +73,7 @@
 #define TRAIT_GOBLINCAMP "goblincamp"
 #define TRAIT_VAMPMANSION "vampiremansion"
 #define TRAIT_VAMP_DREAMS "vamp_dreams"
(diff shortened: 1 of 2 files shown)


notafurryy:
pretty sure the build check does the same thing as you sending proof of it compiling

Kevak:
> pretty sure the build check does the same thing as you sending proof of it compiling

It does but it's still good form, even if redundant. Proof of testing always. No exceptions.

Spaghetti-bit (changes requested):
Needs to be applied to death knights as well, iirc.

Kevak:
Oh right good point. I forgot those existed. Gimme a few

Kevak:
Did that @Spaghetti-bit :3

Spaghetti-bit (approved):

PR #2328, opened 2025-01-28
WW's buff (armor's hp, wounds licking, +20 pen)

DelineFortune:
#2257 but better 

also found that sometimes wws are not strong enough to go through armor (if they are kobold or any low str) so +20 pen

### DIFF
--- a/code/modules/antagonists/roguetown/villain/werewolf/werewolf.dm
+++ b/code/modules/antagonists/roguetown/villain/werewolf/werewolf.dm
@@ -127,7 +127,7 @@
 	blocksound = SOFTHIT
 	blade_dulling = DULLING_BASHCHOP
 	sewrepair = FALSE
-	max_integrity = 250
+	max_integrity = 1500
 	item_flags = DROPDEL
 
 /datum/intent/simple/werewolf
@@ -137,7 +137,7 @@
 	attack_verb = list("claws", "mauls", "eviscerates")
 	animname = "chop"
 	hitsound = "genslash"
-	penfactor = 30
+	penfactor = 50
 	candodge = TRUE
 	canparry = TRUE
 	miss_text = "slashes the air!"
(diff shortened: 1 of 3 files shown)


divifest:
ww's shouldnt instantly go through armor. they should need to break it down a bit.

DelineFortune:
old wws could go through armor as well 
3-4 hits to break bervor and decap
new? cannot do anything if they are not meta race or got low str, get random crit from towner with 0 skill die

notafurryy:
> ww's shouldnt instantly go through armor. they should need to break it down a bit.

with how big ww's are and strong you'd think the whole paw could act as a blunt weapon even if it doesn't pierce

Genevesh:
This looks GREAT; Werewolves these days are basically forced to keep their asses as far away as possible from any real combatants which just results in them either becoming friendly or having to buddy-up with other antags which can HARDLY be justified usually. Werewolves are supposed to be lone fighters that snowball the more people they infect, eventually growing into a whole pack. Since they can't scale their personal strength otherwise, it's really vital they don't just die to the first person with a spear who randomly crits them on third hit after breaking the "skin armor" in two stabs.

RadiantFlash7:
...1500 Integrity for armor is a bit absurd. I agree 250 is way too low, But this is a bit too much.
I believe there was discussion in dev chat about making a way to heal armor.
Currently, werevolves can bite limbs/organs to full heal, but it doesn't restore armor.
Can you potentially put this armor at around 500-600, and see about making it so werevolves can also restore armor by eating limbs/organs? @DelineFortune

RadiantFlash7:
Specifically https://github.com/Rotwood-Vale/Ratwood-Keep/blob/0b5369578e94edc2f5d4ad77165209bcc4c522a1/code/modules/surgery/bodyparts/_bodyparts.dm#L167 has the bite section for fullhealing were wolves. See about making this repair their armor as well, if you could.

Issue #2253, opened 2025-01-18
[REQUEST] Remove Wooden Planks from the Stockpile


Frienderanian:
## Is your feature request related to a problem? Please describe. 

![Image](https://github.com/user-attachments/assets/09bfe3ae-ec3b-4044-be97-31ed04f1fce6)

There is never ANY wood in the stockpile because you get double the money for cutting a short log into a plank and putting it in instead. Planks are literally used for nothing practical for 90% of the playerbase. If I need to make a cross, a drying rack, a Mortar and Pestle, a Bucket, Palisades or literally anything else with wood I am fucking shit out of luck. 
Any class or craft that needs wooden planks can just buy a short log and cut the plank themselves.

## Describe the solution you'd like

Remove Wooden Planks from the stockpile OR make Wooden Planks work in ALL RECIPES that require short logs and make the price for the both of them the same and also make wooden planks add more than 5 minutes to fires when used as fuel. Also, [fix this issue](https://github.com/Rotwood-Vale/Ratwood-Keep/issues/2115) because it isn't a problem in Vanderlin.

## Additional context

It's bleak.

Keazra:
Agreed. The cconomy needs a nerf, too, and this might help.

r4iser:
It seems that this issue can be closed @Frienderanian

PR #2330, opened 2025-01-29
Removes wooden planks from the stockpile

john-pawson:
## About The Pull Request

This PR removes planks from the stockpile for two main reasons:
1. It prevents people that need to buy wood from doing so, since the woodcutters simply turn them into planks to win more money. Because for some reason they're worth more. Leaving the rest of the server without wood, but with a lot of planks that'll end up unused most of the time, and just serve the purpose of draining the treasury from money quicker.

2. The ones that do need planks can simply create them themselves by sawing the wooden log, it's actually cheaper than buying the wooden planks, so those that DO need the planks don't even buy them.

Should probably also mention issue https://github.com/Rotwood-Vale/Ratwood-Keep/issues/2253

## Why It's Good For The Game

Removes an unused material from the stockpile, and stops people from turning the usable material to a semi-useless one simply for 2 extra mammons from the stockpile.

Begone planks 
image
It compiles
image

### DIFF
--- a/code/modules/roguetown/roguestock/stockpile.dm
+++ b/code/modules/roguetown/roguestock/stockpile.dm
@@ -9,18 +9,6 @@
 	export_price = 4
 	importexport_amt = 10
 
-/datum/roguestock/stockpile/wood/plank
-	name = "Wood Plank"
-	desc = "Wooden planks ready to be worked."
-	item_type = /obj/item/natural/wood/plank
-	held_items = list(2, 2)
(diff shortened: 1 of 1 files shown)


Spaghetti-bit (approved):

Elli-Skala:
Wood logs were lowered from 3 mammon to 2 mammon due to wood planks as a refinement idea, now we're removing wood planks from said stock pile

PR #2333, opened 2025-01-29
Making planks now correctly uses carpentry

john-pawson:
## About The Pull Request

For some reason, cutting planks needed carpentry to do, and it gave you experience on lumberjacking. Now it gives you experience and uses lumberjacking correctly.

## Why It's Good For The Game
Bug fix!

image

### DIFF
--- a/code/game/objects/items/rogueitems/natural/wood.dm
+++ b/code/game/objects/items/rogueitems/natural/wood.dm
@@ -84,7 +84,7 @@
 	if(item_flags & IN_STORAGE)
 		return
 	user.changeNext_move(CLICK_CD_MELEE)
-	var/skill_level = user.mind.get_skill_level(/datum/skill/labor/lumberjacking)
+	var/skill_level = user.mind.get_skill_level(/datum/skill/craft/carpentry)
 	var/planking_time = (45 - (skill_level * 5))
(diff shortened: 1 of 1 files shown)


Actuallygoinginsanern:
This is not a bug! It is designed to use carpentry for MANY reasons. The real fix would be making it give carpentry XP, rather than lumberjacking. It was CHANGED to carpentry early on, mainly due to builder jobs not having any lumberjacking.

Which defeated their point.

As I said, correct change would be making planks give carpentry XP.

Spaghetti-bit (approved):

Spaghetti-bit:
Code is fine. I'd look at the comment that @Actuallygoinginsanern left, @john-pawson .

It should be a carpentry skill.

john-pawson:
> This is not a bug! It is designed to use carpentry for MANY reasons. The real fix would be making it give carpentry XP, rather than lumberjacking. It was CHANGED to carpentry early on, mainly due to builder jobs not having any lumberjacking.
> 
> Which defeated their point.
> 
> As I said, correct change would be making planks give carpentry XP.

Well, i was told otherwise by someone else, they said that the intended use was lumberjacking but they were given carpentry by mistake. Changing it!

PR #2336, opened 2025-01-30
Firearm Touchup

ProbablyCarl:
## About The Pull Request
Arquebus pistol made into a one-handed weapon once again. You can't AIM with it(At least so it seemed in testing.), but you can instantly fire without issue. No accuracy, at the tradeoff of instant response. Returning to how it used to be. Mind, that's how it worked when you wielded it, but you could actually aim in that case. This has NO aiming whatsoever. Period.

Adjusts the ammo pouch size so that it can be shoved into bags and the such, as the same size as pouches.

Atop that, it gives the Inquisitor their proper runelock pouch, removing the old three-round pouch.

## Why It's Good For The Game
Arquebus wasn't intended to be a 2 handed weapon and was fucked by collapsing all firearms under a single parent object. This is an objective nerf and makes it pointless outside of directly clicking on someone, or if RNG is in your favor to an extreme degree. I'll probably let this PR sit for a bit and just return the old intents, because of this. We'll see.

Pouches is something I've had requested repeatedly.

image
image

### DIFF
--- a/code/game/objects/items/rogueweapons/ranged/gunpowder/firearm.dm
+++ b/code/game/objects/items/rogueweapons/ranged/gunpowder/firearm.dm
@@ -309,11 +309,15 @@
 
 /obj/item/gun/ballistic/firearm/arquebus_pistol
 	name = "arquebus pistol"
-	desc = "A small gunpowder weapon that shoots an armor piercing metal ball. While it has the same power as a regular handgonne, it is balanced to be used accurately in a single hand."
+	desc = "A small gunpowder weapon that shoots an armor piercing metal ball. \
+	While it has the same power as a regular handgonne, it is balanced to be used in a single hand. \
+	Of course, this comes at the cost of accuracy."
 	icon = 'icons/roguetown/weapons/32.dmi'
 	icon_state = "pistol"
 	item_state = "pistol"
 	force = 10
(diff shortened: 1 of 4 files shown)


Teragenic:
You could always wield it no? Also. You could always instantly fire it. I have a video on the original Jager PR showing this alongside a video... 

Screenshot_2025-01-31-01-55-02-12_40deb401b9ffe8e1df2f1cc5ba480b12

Nothing in its

Here. https://files.catbox.moe/inifqe.mp4

It was always like that.

Looking at the possible intents, yeah that's my bad. Do make sure that the pistol cannot be wielded.

Teragenic:
Gripped intent = Null

Might break it? I'm not sure?

Limeon666:
> Adjusts the ammo pouch size so that it can be shoved into bags and the such, as the same size as pouches.

THANK YOU SO MUCH.

Heheniy:
> You can't AIM with it(At least so it seemed in testing.), but you can instantly fire without issue. No accuracy, at the tradeoff of instant response.

Just make it the longer you holding aim the better accuracy

because this is bleak
image

Gyrandola:
Closing this for 2 months of inactivity. Contact to reopen.

PR #2350, opened 2025-02-01
Underground bog village

DelineFortune:
current low town is still in place (some people asked to keep it from start at least)

no scom
no laws
no gods
no masters

image

image


DelineFortune:
image

DelineFortune:
yes i removed late join point south of farm too because not cool to latejoin in the middle of fight and meat rush antags tanking dmg for 40 seconds

PR #2354, opened 2025-02-02
Makes Estoc impale not a charged attack

Coolgat3:
## About The Pull Request

This makes Estoc's Impale attack, the only 2 reach one not require like a 1 second charge.

## Why It's Good For The Game

Estoc feels really unused and kinda underpowered right now? The only 2 reach attack on a sword that handles like zwei and GS is for some reason a 5 charge time?... This isn't a flail or a bow or whatever just let me click like you click on zwei stab and GS stab, bruh.

Can give if really required for a change /this/ miniscule

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/polearms.dm
+++ b/code/game/objects/items/rogueweapons/melee/polearms.dm
@@ -537,6 +537,5 @@
 	reach = 2
 	penfactor = 30
 	damfactor = 1.2
-	chargetime = 5
 	recovery = 20
 	clickcd = 10


badassbilly:
Why? The impale attack is extremely powerful.

john-pawson:
> Why? The impale attack is extremely powerful.

Ok, i'd like to disagree with this one, since the impale attack not only does the same damage as a greatsword's normal attack, but has almost the same penetration (Less pen than the normal estoc stab, somehow), double the recovery time, and the 0.5 seconds chargetime. I don't see how you'd be able to call this extremely powerful when it's an objectively worse spear stab.

PR #2357, opened 2025-02-02
Fixes werevolf infection

Hmeister-real:
## About The Pull Request

Makes artery bites a 100% chance to infect

## Why It's Good For The Game

Bugged infection, not isn't

image
image

### DIFF
--- a/code/datums/wounds/types/arteries.dm
+++ b/code/datums/wounds/types/arteries.dm
@@ -21,7 +21,7 @@
 	embed_chance = 40
 
 	zombie_infection_probability = 80 
-	werewolf_infection_probability = 100 //used to be a 4% chance, lmao
+	werewolf_infection_probability = 100 
 /datum/wound/artery/can_stack_with(datum/wound/other)
 	if(istype(other, /datum/wound/artery) && (type == other.type))
 		return FALSE


code/_onclick/other_mobs.dm
```diff
@@ -245,7 +245,7 @@
 		if(ishuman(src) && user.mind)
 			if(user.mind.has_antag_datum(/datum/antagonist/werewolf))
 				caused_wound?.werewolf_infect_attempt()
-				if(prob(30))
+				if(prob(100))
```
john-pawson, line 248: Shouldn't you just remove the if() if it's always going to succeed? After all, there's no need to check anything, it'll happen 100% of the time

Hmeister-real: it's so it can be easily modified if people are unhappy with the new infection rates

john-pawson: > it's so it can be easily modified if people are unhappy with the new infection rates

You should probably just comment it out if that's the case, no reason to write bad code so the next coder has to avoid writing `if(prob())`

Hmeister-real: fair enough, fine

hugmeorelse:
This doesn't just do it for werewolves but zombies as well. @Hmeister-real

PR #2362, opened 2025-02-03
Slight Miner QoL, Adds Hammer/Chisel

ophaq:
## About The Pull Request
**If you don't like anything let me know, or what could better improve the miner role.**

Added several things to the backpack for miner. Their knife (which started in their mouth), a hammer and chisel like prospector. Removed some combat skills they didn't really need and buffed maces.

## Why It's Good For The Game
Miners lacked a hammer/chisel to do masonry work with. Prospectors got it, and they were pretty much a better miner/smith mix. It makes sense to me for miners to have tools to work with stone.

Also there was some skill bloat and the miner is already pretty strong, so I also nerfed some things that felt redundant and buffed a few others to help with their role. I'll change things up if you feel the changes I made are terrible but honestly I think these changes will be useful for miner players.

miner

### DIFF
--- a/code/modules/jobs/job_types/roguetown/towners/types/miner.dm
+++ b/code/modules/jobs/job_types/roguetown/towners/types/miner.dm
@@ -26,7 +26,7 @@
 		pants = /obj/item/clothing/under/roguetown/trou
 		shirt = /obj/item/clothing/suit/roguetown/shirt/undershirt/random
 	if(H.mind)
-		H.mind.adjust_skillrank(/datum/skill/combat/maces, 3, TRUE)
+		H.mind.adjust_skillrank(/datum/skill/combat/maces, 2, TRUE)
 		H.mind.adjust_skillrank(/datum/skill/misc/athletics, 4, TRUE) // Tough. Well fed. The strongest of the strong.
 		H.mind.adjust_skillrank(/datum/skill/combat/wrestling, 2, TRUE)
 		H.mind.adjust_skillrank(/datum/skill/combat/unarmed, 3, TRUE)


wow50000:
Miners are good where they are, do not give them more self sufficiency.

ophaq:
> Miners are good where they are, do not give them more self sufficiency.

Prospector is basically better than miners at their job. This PR merely makes them about even.

wow50000:
> > Miners are good where they are, do not give them more self sufficiency.
> 
> Prospector is basically better than miners at their job plus they can be a smith and mason. This PR merely makes them about even. The difference being a miner isn't going to be a smith and focus more on mining and masonry. The only self sufficiency this gives is 1 blacksmithing for repairing their own pick since the smith is often too busy to bother (and prospectors can do that anyway), and the fact they get a chisel and hammer to do masonry work. The rest of it is actually nerfs. Honestly I think some of the miner skills could be nerfed a bit further.

Prospector was a shitty addition done by gyran, end of discussion.

code/modules/jobs/job_types/roguetown/towners/types/miner.dm
```diff
@@ -27,12 +26,10 @@
 		pants = /obj/item/clothing/under/roguetown/trou
 		shirt = /obj/item/clothing/suit/roguetown/shirt/undershirt/random
 	if(H.mind)
-		H.mind.adjust_skillrank(/datum/skill/combat/maces, 2, TRUE)
+		H.mind.adjust_skillrank(/datum/skill/combat/maces, 3, TRUE)
```
badassbilly, line 29: Why buff their weapon skill? They're a miner, not a combat role. City guard gets competent mace, miner doesn't need it too. Please god stop the skill bloat.

ophaq: I'm trying to remove skill bloat. People complained that because I removed spears and knives the mace wasn't enough so I made it 3. I'd rather leave it at 2 as well.

badassbilly: They can complain, bro. I'm with you on keeping it at 2. Miners don't need combat skills equal to any of the combat roles. Honestly most of the combat roles should be nerfed imo but that's outside the scope. Just keep them at 2

ophaq: It's been done.

PR #2366, opened 2025-02-04
Sneaky councillors

notafurryy:
## About The Pull Request

Makes Councillors more akin to Spymasters than brawlers?

## Why It's Good For The Game

There is no real reason for them to have those skills as fragile nobles, and courtiers which would more so benefit from guile and intrigue than brawn, and they get a dagger.

image

### DIFF
--- a/code/modules/jobs/job_types/roguetown/courtier/councillor.dm
+++ b/code/modules/jobs/job_types/roguetown/courtier/councillor.dm
@@ -32,7 +32,7 @@
 	beltr = /obj/item/storage/belt/rogue/pouch/coins/rich
 	backpack_contents = list(/obj/item/rogueweapon/huntingknife/idagger/steel = 1)
 	if(H.mind)
-		H.mind.adjust_skillrank(/datum/skill/combat/wrestling, 2, TRUE)
+		H.mind.adjust_skillrank(/datum/skill/combat/wrestling, 1, TRUE)
 		H.mind.adjust_skillrank(/datum/skill/combat/unarmed, 2, TRUE)
 		H.mind.adjust_skillrank(/datum/skill/combat/knives, 3, TRUE)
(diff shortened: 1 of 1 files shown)


radarseas:
I'd bump their wrestling down to 1 honestly, these guys already have the garrison at their side and the weight of authority. Give townsfolk a chance during rebelions to kill these people _if they're stupid enough to get involved directly._

honestly we need more impact to the court for losing shylocked nobles, rather than oh no, well now we don't have to pay them. They should be encouraged if anything to keep nobles alive.

Limeon666:
I'm all for councilor stats being changed but giving them in rogue stats feels weird.

As an example the main skills a noble would learn are riding, reading, and archery for sport. That being reflected would be more interesting.

Also learning swimming was actually seen as unusual for nobles.

notafurryy:
> I'm all for councilor stats being changed but giving them in rogue stats feels weird.
> 
> As an example the main skills a noble would learn are riding, reading, and archery for sport. That being reflected would be more interesting.
> 
> Also learning swimming was actually seen as unusual for nobles.
> 

They have riding and I lowered swimming?

Limeon666:
My bad, but the point is it's weird they even have swimming. Also I agree with removing their wrestling.

I still stand by though that from what I can glean a courtier was not a spy master in most cases, but maybe this could work as a subclass?

At the very least the current tutorial for councilor would not reflect this change at all, so if you go through with it I'd ask you include the fact you gather intel in that.

notafurryy:
> My bad, but the point is it's weird they even have swimming. Also I agree with removing their wrestling.
> 
> I still stand by though that from what I can glean a courtier was not a spy master in most cases, but maybe this could work as a subclass?
> 
> At the very least the current tutorial for councilor would not reflect this change at all, so if you go through with it I'd ask you include the fact you gather intel in that.

Oh, sure! I thought that was implied with being a Councillor but I can do that.

PR #2370, opened 2025-02-05
Remove the seemingly arbitrary do_after for looking up or down, lowers the cooldown for looking up or down

polymathPretender:
## About The Pull Request

I really don't know what to say here that isn't captured by the title.

## Why It's Good For The Game

The current method of looking up and down is clunky and janky for seemingly no reason, and this makes it more responsive and intuitive to use. You're tilting your head slightly, not astral projecting, it shouldn't take that much effort or time.

Down the line, if it's absolutely imperative that looking at a different elevation is a pain in the ass, maybe we can add some behavior to heavy armor helmets to slow it down. But for now, I think we should free the people's necks and let them look up and down as they please.

https://github.com/user-attachments/assets/db5f3bca-3d75-4df6-9ddb-c9fac7b76620

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -1872,18 +1872,11 @@
 
 	if(!OS)
 		return
-	var/ttime = 10
-	if(STAPER > 5)
-		ttime = 10 - (STAPER - 5)
-		if(ttime < 0)
-			ttime = 0
 
 	visible_message(span_info("[src] looks down through [T]."))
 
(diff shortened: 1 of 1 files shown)


Limeon666:
Love this, but I do actually think it'd be for the best if heavy helmets retained this limitation as you suggested because looking up and down or to the side gives you a 360 degree view.

CalamaBanana:
The do_after loadingbar had the benefit of being able to guess if someone is looking up to you, though. Some other indicator would be neat if that is gone.

PR #2380, opened 2025-02-09
Update archer.dm

Dilligentdog:
## About The Pull Request
Buff MAA archer to make them at least more on par with vanguard archers, who were just superior in every way for some reason

## Why It's Good For The Game

The trained soldier should be better, if not equal, to a peasant levy assigned to man the outer gate

image

### DIFF
--- a/code/modules/jobs/job_types/roguetown/garrison/manatarms/types/archer.dm
+++ b/code/modules/jobs/job_types/roguetown/garrison/manatarms/types/archer.dm
@@ -20,10 +20,10 @@
 		H.mind.adjust_skillrank(/datum/skill/combat/polearms, 2, TRUE)
 		H.mind.adjust_skillrank(/datum/skill/combat/swords, 2, TRUE)
 		H.mind.adjust_skillrank(/datum/skill/combat/maces, 3, TRUE)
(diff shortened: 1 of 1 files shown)


Teragenic:
Already a PR up to nerf Vanguard stats across the board.

AbsorbentHeart5:
A few things:
- I don't agree with them getting master crossbows, ideally crossbow 5 should only be for the gatemaster.
- Speed should be lowered to +1 and then bump con to either 0 or +1, why would a guard get bad con to begin with? It's also the only guard role to get - con.
- I would also say to increase firearm skill to 4 so they're on par with those powerful jager mercs, but you'd probably get a lot more backlash. 

Other than that the PR is good, MAA archer really needed a buff so that they're not just a useless member of the retinue. Currently they're hardly ever picked despite range weapons being insanely good, why play MAA archer when you can play vanguard and be a lot better.

> Already a PR up to nerf Vanguard stats across the board.

Now this just isn't true at all, have you even looked at PRs that are currently open? The last PR that attempted this got shut down immediately by the community.

RadiantFlash7:
This was more or less handled in another PR, Vanguard archers recieved a stat and skill nerf, MAA got a stat buff.

PR #2382, opened 2025-02-09
Implements bounty redemption via stocks

Arturlang:
## About The Pull Request
This PR adds redeeming bounties via simply putting people on the stocks, which will give you 50 mammon out of the bounty every 5 minutes, there is also, a way to speed this up if you wish to take up the initiative.

## Why It's Good For The Game
It seems there's interest in a better way to redeem bounties that does not require the death of the victim.

image
image

### DIFF
--- a/code/__DEFINES/dcs/signals/signals_mob/signals_human.dm
+++ b/code/__DEFINES/dcs/signals/signals_mob/signals_human.dm
@@ -1,0 +1,2 @@
+#define COMSIG_HUMAN_LEWD_FINISH "human_lewd_finish"
+#define COMSIG_HUMAN_LEWD_FINISHED_ON "human_lewd_finished_on"
--- a/code/datums/sexcon/sexcon.dm
+++ b/code/datums/sexcon/sexcon.dm
@@ -228,7 +228,9 @@
 		user.apply_status_effect(/datum/status_effect/debuff/cumbrained)
 	SSticker.cums++
 	cuckold_check()
-
+	if(target && target)
+		SEND_SIGNAL(user, COMSIG_HUMAN_LEWD_FINISH, target, src)
+		SEND_SIGNAL(target, COMSIG_HUMAN_LEWD_FINISHED_ON, user, src)
 
 /datum/sex_controller/proc/after_milking()
 	set_arousal(80)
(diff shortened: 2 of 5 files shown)


Koteshkin:
Another step to hugbox.
*sigh

john-pawson:
5 minutes per 50 mammon seems a bit low, doesn't it? Not only that, but it opens up people claiming a friend's bounty and then sharing it with the hunted simply to get a, for example, 200 mammon head start. At the cost of one of them being AFK for 20 minutes at the start of the round.

That and, unless for some reason the duchy has decided to start paying off bounties of people that are on the pillory, the bounty money shouldn't come from the treasury, since the duchy isn't the one that put in the bounty specifically. The money has already been inserted to the excidium when the bounty was made, it shouldn't be draining anything. 
The only thing the treasury has related to the excidium is the tax amount that it wins from bounties being placed on another player.

I'd also like to ask about the, uh, "Violation bonus".
And another thing, looking at the code, this works on EVERY pillory for some reason? Does that mean that someone can just craft a pillory in the middle of nowhere and it'll somehow have all of the machinery to detect and pay out a bounty?

Actuallygoinginsanern:
A "Violation bonus" is insane. Whats that one dexter clip "What is it with you and rape?"

...it makes no sense.

EvenInDeathIStillServe:
This is unwanted due to potential for abuse (and not in the good way). A talking pillory ain't it either.

Any "non-fatal" ways to satisfy bounties will need to have actual, severe repercussions.

hugmeorelse:
> Another step to hugbox. People chose hunted flaw by their own or taking a risk by clicknig special. Its their own fault. Stop pitying them. *sigh

You're clearly misunderstanding it. If people want to have a non-lethal bounty, this is an option you can take. This is not a replacement, merely an addition to a myriad of options. There's nothing wrong with that.

Arturlang:
To note, this was done as a bounty.

notafurryy:
> I think this will be a good option to have on hand.

what about.. Taking a hand instead?

john-pawson:
> > I think this will be a good option to have on hand.
> 
> what about.. Taking a hand instead?

Good idea, one problem, replacing it is extremely easy, so it doesn't really work as a severe punishment, as the punishment becomes "Kill a beggar, steal its hand, and go to the church to attach it"

PR #2385, opened 2025-02-10
Rockhill Tweaks Set One

Alanii:
## Tweaking Rockhill Set One

- Resolves missing pathset for firearms in Keep Armory.
image
image
- Adds broken bridge to Bum-Area and access river point.
image
- Dries up the initial Bog-Lowtown river entirely making it a bit simpler to cross.
image

## Why Good?

Fixing missing item pathing

Decreases the tediousness of crossing the initial river into low town. 

Originally, the river was the main obstacle, however with the "main gate" being placed at the town itself, the river does not serve its purpose well. Instead it hampers any venturing into the bog and makes getting there a bit more tedious and difficult with the broken bridge. This new setup should make it simpler to cross for both antagonists and regular players alike. 

Reminder, this is more of a tweak and not an actual major change. I'll discuss with "maptainers" ways to possible remove the river entirely and opening up the bog more in the future.


Alanii:
Closing to redo

PR #2387, opened 2025-02-10
Robust Hunger and Thirst

josephpatrickgreen:
## About The Pull Request

This PR makes it so that high strength/muscle mass characters require more food and high constitution/blood volume characters require more hydration with the modification calculation being (([stat]/ 10) + 1) / 2. This calculation assumes 10 of strength and constitution is an average character. This also works in reverse for ascetic characters lowering their hunger and thirst.

## Why It's Good For The Game

Adds an element of realism, stat maxers need to do slightly more to avoid debuffs, more support/food/drink required. Support classes require less sustenance so they can keep supporting longer.

Not much testing required, it's just math, here I will give examples of input numbers using the calculation in code

base hunger = 0.15 (global variable)

strength 5 : hunger = 0.1125
strength 10 : hunger = 0.15
strength 15 : hunger = 0.1875

--------------------

base thirst = 0.15 (global variable)

constitution 5 : thirst =  0.1125
constitution 10 : thirst = 0.15
constitution 15 : thirst = 0.1875

### DIFF
--- a/code/modules/mob/living/carbon/human/species.dm
+++ b/code/modules/mob/living/carbon/human/species.dm
@@ -956,6 +956,11 @@ GLOBAL_LIST_EMPTY(roundstart_races)
 				H.Jitter(5)
 			hunger_rate = 10 * HUNGER_FACTOR*/
 //		hunger_rate *= H.physiology.hunger_mod
+
+		//hunger modifier based on strength/muscle mass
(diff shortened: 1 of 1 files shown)


hugmeorelse:
Makes sense.

EvenInDeathIStillServe (approved):

Genevesh:
Maybe some parts of realism shouldn't be that sought-after; I'd much rather realistic combat than having to micromanage my character even more.

PR #2303, opened 2025-01-25
Refugee Foreigner Classes Part1

MrAsianFunTime:
With the intended goal of adding in Refugee Foreigner Classes from different parts of the world and places in the Setting.

First we have the two.
Gronn Refugees
Tribals
Silks Merchant
Spice Lord
Imperial Handhonner

Otavan Refugees
Otavan Cossack
Otavan Noble
Otavan Peasant
Otavan Orthodox Priest

## Proof of Testing (Required)
https://gyazo.com/f5f553c77d95ba35c6d3135ef21f7ac2

PR #2390, opened 2025-02-11
Underdwellers!

notafurryy:
## About The Pull Request

Re-adds underdwellers from https://github.com/Rotwood-Vale/Ratwood-Keep/pull/2303 with better balance?

## Why It's Good For The Game

Shunned races content good!! Especially for the most shunned and forgotten of them all, kobolds and verminfolk.

image
image

### DIFF
--- a/code/modules/jobs/job_types/roguetown/refugees/types/underdwellers.dm
+++ b/code/modules/jobs/job_types/roguetown/refugees/types/underdwellers.dm
@@ -68,7 +68,7 @@
 			beltr = /obj/item/gun/ballistic/revolver/grenadelauncher/bow
 			backl = /obj/item/storage/backpack/rogue/satchel
 			l_hand = /obj/item/ammo_holder/quiver/arrows
-			cloak = /obj/item/clothing/cloak/tribal
+			cloak = /obj/item/clothing/cloak/volfmantle
 			H.change_stat("strength", 1)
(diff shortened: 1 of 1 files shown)


ophaq:
I love this.

Purpleturtle3:
Compared to the other race locked refugees, these are rather weak.

notafurryy:
> Compared to the other race locked refugees, these are rather weak.

Thats because the other refugees are busted

notafurryy:
> I love this. Is the shaman patron locked by chance and will they get a healing/revive spell like the goblin shaman or town doctors? Since they don't have much in the way of town doctor things either in terms of healing other than their tools.

They are not. Also, yeah I will add that, thought it would be a lil too much for a shaman but it seems not, with even gobs having those spells

Purpleturtle3:
> > Compared to the other race locked refugees, these are rather weak.
> 
> Thats because the other refugees are busted

I get where you are coming from, but people are already saying refugees should be buffed and all...

notafurryy:
> I get where you are coming from, but people are already saying refugees should be buffed and all...

buffed does not mean merc/yeomen level. Also the underdwellers are meant to be fun to play thematic roles, not the best in their class, you aren't supposed to be comparable to the town smith as a shunned tinker-er

EvenInDeathIStillServe (approved):

PR #2391, opened 2025-02-11
MAA Gunner

Enite:
## About The Pull Request
Replaces the Retinue Archer with a Gunner. He gets a gun. Reduces old archery and crossbow skills to 3, bumps firearms up to 4. Removes -1 constitution. Bumps up swords from 2->3.

## Why It's Good For The Game
MAA Archer is just a worse Vanguard Archer. It's bad, everyone knows it. This gives it something the other guard roles don't have, and if you want or need to use a bow or crossbow you still have the skills for it.

image


Limeon666:
Them all spawning with arquebus now seems overtuned. Also why'd this balance change remove their -endurance?

Limeon666:
> The Arquebus is already a roundstart armory spawn in the Keep, the type path is just bugged. If you give them another gun they'll just come and grab one from the armory instead, once the PR to fix that is merged.

Well in the first place that's balanced around keep not having people with more than three gun skills at round start, and that scenario would still be better than MAA spawning with the best gun in the game.

You should have to go through a merchant or an artificer to get said better gun instead of spawning with it. That's how they get gameplay.

Fact is arquebus is meant to be a very limited round start resource and this shifts the balance of that.

TheKaiser2233:
Just give us a subclass for MAA archer instead.

Enite:
Balance freeze, other PRs touching on the same subject. Might reopen again later

Issue #2061, opened 2024-12-27
[BUG] Paper is fucked


Frienderanian:
I have no idea what is going on with paper, but when you view it the font is massive, unreadable and you can't scroll. The font is super low resolution too. I think it only happens when you write on the paper with a quill. Makes writing extremely long suicide notes impossible and makes them illegible. Really a massive hamper to communication when I can't know for sure if the letter I am sending will actually be legible. But who cares, the Hermes system in this game is a joke, no one gives a shit and I am wasting my life giving a shit too.

JagOfTroy:
Thanks for making us aware of it.
Could you provide some screen captures of what it looks like?

Frienderanian:
![Screenshot 2024-12-29 144102](https://github.com/user-attachments/assets/686d485c-2525-49fe-b9be-1ae52278be9a)
![Screenshot 2024-12-29 113903](https://github.com/user-attachments/assets/c18c26f5-b73e-44c0-af73-fb506faf53a9)

JagOfTroy:
paper.dm,
`/obj/item/paper/proc/format_browse(t, mob/user)
	user << browse_rsc('html/book.png')
	var/dat = {"<!DOCTYPE HTML PUBLIC \"-//W3C//DTD HTML 4.01 Transitional//EN\" \"http://www.w3.org/TR/html4/loose.dtd\">
			<html><head><style type=\"text/css\">
			body { background-image:url('book.png');background-repeat: repeat; }</style></head><body scroll=yes>"}
	dat += "[t]<br>"
	dat += "<a href='?src=[REF(src)];close=1' style='position:absolute;right:50px'>Close</a>"
	dat += "</body></html>"
	user << browse(dat, "window=reading;size=500x400;can_close=1;can_minimize=0;can_maximize=0;can_resize=1;titlebar=0;border=0")`
	
	Will have to do some trial and error testing. There's quite a bit of limits placed upon paper too(max length is 2000 characters) but it makes sense that your not gonna be able to write an incredibly long message on one sheet of paper alone.

PR #2397, opened 2025-02-12
Reverts new scroll text

john-pawson:
## About The Pull Request

Gyran, while making the docks, decided to go ahead and change how scroll text was read, this led to a text that was pixelated, low quality, unreadable sometimes, and broke the moment you would write a slightly long message. This simply reverts those changes and leaves the scrolls back how like they were, so they'll read the same as a parchment, but with a little overlay behind.

Here's a pretty decent explanation: https://github.com/Rotwood-Vale/Ratwood-Keep/issues/2061

## Why It's Good For The Game
Most people liked the text before this change, and a lot complained when it was added. So hopefully this'll make people use scrolls for messages again, instead of having to search for a parchment since they were just objectively better for messages.

image

### DIFF
--- a/code/modules/paperwork/rogue.dm
+++ b/code/modules/paperwork/rogue.dm
@@ -61,11 +61,14 @@
 	if(in_range(user, src) || isobserver(user))
 		user.hud_used.reads.icon_state = "scroll"
 		user.hud_used.reads.show()
-		user.hud_used.reads.maptext = info
-		user.hud_used.reads.maptext_width = 230
-		user.hud_used.reads.maptext_height = 200
-		user.hud_used.reads.maptext_y = 150
-		user.hud_used.reads.maptext_x = 120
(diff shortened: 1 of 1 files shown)


dwasint:
atleast position the browser so that its not hanging off the side of the scroll and change its background texture to match

john-pawson:
> atleast position the browser so that its not hanging off the side of the scroll and change its background texture to match

Well, i would, but the offsets don't seem to work for some reason. If you know how to do it, i'd more than gladly implement it into the code, but i'd rather it have the backround be a little bit offset until i can find a way to fix that too rather than keep having the current one.

john-pawson:
> I've had luck making the text readable by making it bold using ^BOLD^. Could force scroll text to always be bold if you think it looks readable enough too

Yeah, but that doesn't fix the issue of the very low character max, and the inability to do things such as, for example, scroll down, change the size of the window for easier readability, or select the text if you wish to copy it.

EvenInDeathIStillServe (approved):

Issue #2185, opened 2025-01-10
[BUG] "State Laws" Command does not work


Githubaccount222:
Going up to a SCOM and saying "State laws" does not make the SCOM state the laws.
It should state the laws when this is done.

## Steps to Reproduce:
Stand in front of a SCOM and say "state laws"

PR #2398, opened 2025-02-12
Fixes "say laws" command on SCOM

john-pawson:
## About The Pull Request

While creating jabberlines, the creator probably accidentally deleted the "state laws" and "say laws" command that let people that can't read know the laws of the land. This fixes it!

Issue that let me know it: https://github.com/Rotwood-Vale/Ratwood-Keep/issues/2185 (Thanks for reporting it)

## Why It's Good For The Game

Bug fix!

image

### DIFF
--- a/code/modules/roguetown/roguemachine/scomm.dm
+++ b/code/modules/roguetown/roguemachine/scomm.dm
@@ -230,7 +230,7 @@
 			if(calling.calling == src)
 				calling.repeat_message(raw_message, src, usedcolor, message_language)
 			return
-		if(lowertext(raw_message) == pick("say laws.", "state laws."))
+		if(lowertext(raw_message) in list("say laws.", "state laws."))
 			dictate_laws()
 			return
 		for(var/obj/structure/roguemachine/scomm/S in SSroguemachine.scomm_machines)


code/modules/roguetown/roguemachine/scomm.dm
```diff
@@ -230,6 +230,9 @@
 			if(calling.calling == src)
 				calling.repeat_message(raw_message, src, usedcolor, message_language)
 			return
+		if(lowertext(raw_message) == pick("say laws.", "state laws."))
```
EvenInDeathIStillServe, line 233: This code only works half the time because of the pick().

Try this instead. Not sure if it'll work but I assume so.
`if(lowertext(raw_message) in list("say laws.", "state laws."))`

EvenInDeathIStillServe (changes requested):

code/modules/roguetown/roguemachine/scomm.dm
john-pawson: It worked with both on my testing, both say laws and state laws, i simply copied the already existing code from before.
But i don't see why i shoulndn't make this change.

john-pawson: Changed, tested, it works.
image

EvenInDeathIStillServe (approved):

PR #2420, opened 2025-02-17
Azure - Vanderlin Alchemy Port

Alanii:
# This will remain DRAFT until it is ready for test merge.
This will most likely be tested after the Mage PR.

## About The Pull Request
Brings over Azure's port of Vanderlin Alchemy System.

Realistically all this does is removes a lot of the "moribunn" changes to the alchemy system and replaces it with a system similar to Vanderlin and Azure's Alchemy system.

## Why It's Good For The Game

Replaces the barebones alchemy system with something that is more complete.

As to whether this is good or not, I have no idea. People wanted this system and I don't really play with alchemy a whole lot.

image

### DIFF
--- a/code/game/objects/effects/spawners/roguemapgen.dm
+++ b/code/game/objects/effects/spawners/roguemapgen.dm
@@ -78,7 +78,7 @@
 	probby = 50
 	spawned = list(/obj/item/restraints/legcuffs/beartrap/armed/camouflage)
 
-/obj/effect/spawner/lootdrop/Initialize(mapload)
+/*obj/effect/spawner/lootdrop/Initialize(mapload)
 	..()
 	do_spawn()
 	return INITIALIZE_HINT_QDEL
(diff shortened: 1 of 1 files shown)


Alanii:
image

ophaq:
Awesome, been wanting this for a while. We will need the growable plants/herbs that can be harvested in the wild as well unless that is included with the port. Vanderlin's alchemy is also based off of alchemy skill instead of treatment IIRC so unless that is changed to stick with treatment, classes will need to be rebalanced with the new alchemy skill.

Alanii:
> Awesome, been wanting this for a while. We will need the growable plants/herbs that can be harvested in the wild as well unless that is included with the port. Vanderlin's alchemy is also based off of alchemy skill instead of treatment IIRC so unless that is changed to stick with treatment, classes will need to be rebalanced with the new alchemy skill.

Due to the amount if items the alchemy system searches for, this PR may be expanded in scope. It calls for specific items Azure-Peak added through previous additions. Due to these additions, they make sense to be pulled to Ratwood as well to provide more items and content.

I will need to speak with Useroth on if such a thing would be fine. This also means this PR will receive a name change.

Casteroki:
Honestly as someone who is already really into alchemy. VERY excited to have alchemy expanded!!

PR #2422, opened 2025-02-17
Bow QoL

NPC1314:
## About The Pull Request

Makes shooting arrows less cancer by the simple measure of making some of the transparent pixels clickable on bows and arrows. You can´t see the change, its transparent. Only will notice less annoyance.


notafurryy:
> Makes shooting arrows less cancer by the simple measure of making some of the transparent pixels clickable on bows and arrows. You can´t see the change, its transparent. Only will notice less annoyance.

You do know you can just click the quiver to quick reload? Though this will be a nice change for crossbows

EvenInDeathIStillServe (approved):

PR #2425, opened 2025-02-18
Allows broken armour to be repaired

Leudoberct1:
## About The Pull Request

This PR allows damaged armour to be repaired from 0% durability.

## Why It's Good For The Game

The intent of this PR is to ease the burden from smith players- at present, any damaged armour has to be resmelted, and if it costs more than one bar, is going to require new steel. This makes it so that every time someone gets into a fight and has some armour broken, a smith has to have them all over again- keeping them stuck at the smithy through most of the round. Discussed this with Eora in the Discord, and decided to try it out.

https://gyazo.com/135d78d67a4a764cbe62957cedbd266b

Successful compile, tested ingame and the armour can be repaired and loses the damaged appearance upon repair.

### DIFF
--- a/code/modules/clothing/clothing.dm
+++ b/code/modules/clothing/clothing.dm
@@ -352,6 +352,11 @@
 		. += how_cool_are_your_threads.Join()
 */
 
+/obj/item/clothing/proc/obj_fix(damage_flag)
+	obj_broken = FALSE
+	if(damaged_clothes)
+		update_clothes_damaged_state(FALSE) 
+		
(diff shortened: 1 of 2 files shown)


Silver-Paws:
Maybe it's worth to add a need to spend 1 material ingot? To make the repairability available. Just to make sure some fellas think twice before ragecharging into battle with thought "Eh. My smith gonna fix it anyway".

Ingot cost = + repair cost.

EvenInDeathIStillServe (approved):

PR #2427, opened 2025-02-18
Healing Overhaul Part 1: Wound Infections

Xaytan:
## About The Pull Request

This is the start of an overhaul to healing in Ratwood, to make the process of dealing with wounds more involved and interesting. Characters will need to clean their wounds and stay well fed while healing, lest they succumb to infection.

PLEASE DO NOT FULL MERGE. TEST MERGE AT MOST.

- Wounds can now become infected. Wounds slowly accrue infection if left untreated, with speed based on whether the wound is clean or not.
- Infected wounds slowly deal toxin damage and no longer heal passively.
- Wounds eventually become gangrenous, causing the limb they are on to rot.
- Sleeping now drains nutrition to heal your wounds. Running out of nutrition will cause your wounds to stop healing.
- Walking through murk has a small chance to make your wounds filthy.
- Walking through sewage makes your wounds filthy, and causes a small amount of toxin damage if you have open wounds
- Ash Syrum treats wounds and now actually does what it's supposed to do as well
- Oh yeah, I made berry poison syrum actually toxic like it's clearly supposed to be

## Why It's Good For The Game

Makes dealing with the aftermath of fights a little more interesting, and increases the gritty realism we're allegedly going for.

Sets the groundwork for later additions that will make healing and medicine more interesting with the aim of making acolytes and doctors and other healing types feel like real jobs instead of MMB simulators.

Oh yeah, and I fixed a bug where the "Purify" reagent doesn't actually process.

TBD

### DIFF
--- a/code/__DEFINES/medical.dm
+++ b/code/__DEFINES/medical.dm
@@ -144,6 +144,6 @@
 /// Wound has been cleaned and will become infected more slowly
 #define WOUND_CLEANLINESS_CLEAN 1
 /// Wound is still dirty and needs to be cleaned
-#define WOUND_CLEANLINESS_DIRTY 3
+#define WOUND_CLEANLINESS_DIRTY 5
 /// Wound has become filthy and really really needs to be cleaned
-#define WOUND_CLEANLINESS_FILTHY 4
+#define WOUND_CLEANLINESS_FILTHY 10


Limeon666:
Another Xaytan banger.

A bit dreading the idea of bog spawn becoming even more dangerous though. Simplemob nerf when?

notafurryy:
> * Sleeping now drains nutrition to heal your wounds. Running out of nutrition will cause your wounds to stop healing.

isn't that already how it works?

Hmeister-real:
So swimming in murk is now a death sentance with open wounds? Holy Based

CalamaBanana:
You should sneak in some kind of weak disinfectant that can be bush-crafted, maybe from bad berries, otherwise bog gameplay is certain death the second one wolf warps behind you. I could only spot potato and oat alcohol with enough booze power to disinfect, and those seeds are quite rare. I'd recommend lowering it far enough so it includes wine. 

Or what would be even more neat but more effort, allow applying any alcohol to a cloth to be used as wound wrapping, and it slowly fights against the infection based on booze power.

notafurryy:
> You should sneak in some kind of weak disinfectant that can be bush-crafted, maybe from bad berries, otherwise bog gameplay is certain death the second one wolf warps behind you. I could only spot potato and oat alcohol with enough booze power to disinfect, and those seeds are quite rare. I'd recommend lowering it far enough so it includes wine.
> 
> Or what would be even more neat but more effort, allow applying any alcohol to a cloth to be used as wound wrapping, and it slowly fights against the infection based on booze power.

We already have bandages that are cloth and ashes, could use that as a disinfectant, or even cauterizing the dead flesh with a torch.

githubuser4141:
I think it would be interesting if minor infections on a healthy host had a very good chance to either stay as minor infections that don't do lethal damage, or have a smaller chance to cure themselves? It can easily be themed as people having better immune systems / the environment isn't as harsh, so germs aren't as tough / competitive as they are now / the presence of magic somehow weakens germs. Rather than just have minor infection > 100% chance of evolving to lethal infection in 5 minutes > 75 toxin damage > "Dead"

Xaytan:
> So here's a idea, infections take a bit of time to actually occur in real life so maybe make it where if a wound gets dirty and isn't cleaned with in a day or so it will actually start become infected and there would be some early signs of the infection starting to begin

That's what it does. That's literally what the PR does.

RadiantFlash7:
Fix merge conflicts and finish this PR, Raaagh!!!

Gyrandola:
Is this being worked on? It will be closed for inactivity soon.

Xaytan:
Shit got wild IRL for a bit there, I'm going to close this and reopen at a later date.

PR #2430, opened 2025-02-18
Liches bleed to death often

### LINKED ISSUE #1857: [BUG] Liches bleed to death often
## Describe the bug

When an arrow hits a lich, they start bleeding to death. When you graft a non-skeletenized limb on a lich, they bleed to crit.

## Round ID:

Whenever

## Steps to Reproduce:

get shot by arrow as a lich, crit. Graft on humen limb to yourself as a lich, bleed to crit

## Additional context

They never die from bleeding, only enter crit

polvohime:
## About The Pull Request

Fixes #1857 hopefully, regarding the question of liches bleeding to death and having blood at all (they are skellies)

Basically, this adds a new trait to the game called NO_BLOOD which is applied to liches only. This will change a few things regarding blood in general as well as the heart_hud we have.

The heart is used to track your blood volume, and since liches do not have blood we will be using **LIMB DAMAGE** instead to track how hurt you are. It has a few other changes too in regards to text, but nothing too important to mention here!!!

## Why It's Good For The Game

I don't want my skeleton evil wizards to bleed to death or bleed at all, in fact i just removed their blood entirely.

my gyazo is not working so you will have to trust me on this
image
image
image

### DIFF
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -455,42 +455,67 @@
 		return
 	else
 		if(hud_used.bloods)
-			var/bloodloss = ((BLOOD_VOLUME_NORMAL - blood_volume) / BLOOD_VOLUME_NORMAL) * 100
+			// small check for no_blood trait (currently for liches)
+			if(HAS_TRAIT(src, TRAIT_NO_BLOOD))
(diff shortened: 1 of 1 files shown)


Eriripng:
i sohuld mention that this may or may not make liches harder to kill(?) 

personally i never seen one but im sure our QA (players) will let us know 🥺

EvenInDeathIStillServe (approved):

PR #2439, opened 2025-02-19
Makes, jackberries more dangerous than not when foraging.

ARefrigerator:
## About The Pull Request
Nerfs the ability to pull free food (that's right, free!) from the bushes.
Just changes the ratio so the odds of you getting safe berries is a lot lower.

## Why It's Good For The Game
The beggars have had it too good for too long. This will at least encourage people to feast on the other already abundant foods players create or at least have to cook all their berries via some method to ensure they are safe. Or they can in desperation attempt to eat what was foraged.

It's just the ratio of the list changed. You can see it in the code.

### DIFF
--- a/code/game/objects/structures/roguetown/rogueflora.dm
+++ b/code/game/objects/structures/roguetown/rogueflora.dm
@@ -303,8 +303,8 @@
 	var/bushtype
 	//Why bother generating the same list for every single bush? Static list time.
 	var/static/list/bush_options = list(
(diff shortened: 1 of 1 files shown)


Limeon666:
Berries were already nerfed when in town recently. It should still be possible to get jacks berries for wine and such. This is just annoying. If you want to nerf them further make them give less nutrition.

ARefrigerator:
Oh I see. I assumed that when all the berries were cooked in some manner they became clean to eat but that does not seem to be the case. I suppose I could try a new PR with that idea.

PR #2441, opened 2025-02-19
Nerfs berries nutriment.

ARefrigerator:
## About The Pull Request
Drops the berries nutriment.

## Why It's Good For The Game
It's too strong! Also puts it more in line with other servers (i.e. Vanderlin uses this value too)

image

### DIFF
--- a/code/modules/farming/produce.dm
+++ b/code/modules/farming/produce.dm
@@ -150,7 +150,7 @@
 	icon_state = "berries"
 	tastes = list("berry" = 1)
 	bitesize = 5
-	list_reagents = list(/datum/reagent/consumable/nutriment = 0.5)
+	list_reagents = list(/datum/reagent/consumable/nutriment = 1)
 	dropshrink = 0.75
 	var/color_index = "good"
 	can_distill = TRUE


john-pawson:
Ok, i'll have to admit.. this is maybe a little bit /too low/, since it'd barely have any kind of effect on your nutrition.
I'd say about 1 nutriment would probably be more fitting. 
It'd take 6 full berries to even come close to an apple, and it'll probably just feel like berries got bugged and don't give nutriments at all for ones playing the game that aren't keeping up with all the updates.

Limeon666:
You can nerf something without nerfing it this hard.
 

Good change but just reduce it to 2 or something.

ARefrigerator:
Apple nutriment is 3. Based on that I'd be fine with 1. Anything higher seems a little too much given how easy to obtain the berries are.

Limeon666:
> Apple nutriment is 3. Based on that I'd be fine with 1. Anything higher seems a little too much given how easy to obtain the berries are.

A bunch of berries irl actually has more calories than an apple, lol

EvenInDeathIStillServe:
The chance of finding berries within town has already been lowered to almost none. There is no need for this.

Xaytan:
Clearly we just need to nerf apples too.

PR #2442, opened 2025-02-20
Fixes unearthing stumps, changed to use shovel

ARefrigerator:
## About The Pull Request
Fixes the stump removal code. 

Previously it seemed to go via the 'attacked_by' proc which only calls during combat mode. This makes it use 'attackby' which does not require combat mode activated. It also ports the code from Vanderlin using the shovel instead. (currently always gives stump even if burnt) Works with all shovel sub types as well.

## Why It's Good For The Game
2 people and I tried to remove stumps for about 15 minutes not understanding how it was done. When I looked at the code I noticed the issue. I think not only does this make the most sense for removing a stump but it also aligns with the method in other servers meaning it'll be easier to transfer knowledge and also keep some familiarity in the system. It also fixes the need to use combat mode. While it no longer requires combat mode it still does work while in it. 

https://github.com/user-attachments/assets/990e1768-7bc7-4ad9-9846-6f8235951872

### DIFF
--- a/code/game/objects/structures/roguetown/rogueflora.dm
+++ b/code/game/objects/structures/roguetown/rogueflora.dm
@@ -179,16 +179,31 @@
 	var/lumber = /obj/item/grown/log/tree/small
 
 /obj/structure/table/roguetree/stump/attackby(obj/item/I, mob/living/user)
-	if(istype(I, /obj/item/rogueweapon/shovel))
-		to_chat(user, "I start unearthing the stump...")
-		playsound(loc,'sound/items/dig_shovel.ogg', 100, TRUE)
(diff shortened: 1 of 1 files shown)


ophaq:
Just keep it to axe and fix it so that you don't need combat mode for removing them. Woodcutters chop stumps too and its a pain in the ass carrying a shovel on you with your already limited space.

ARefrigerator:
Okay I put back the original axe code into the function. Now it's just in the attackby function so there's no combat mode confusion. Still have to use the chop intent.

EvenInDeathIStillServe (approved):

PR #2445, opened 2025-02-20
Bottle Bomb: lowers chance of fail, fixes qdel issue.

ARefrigerator:
## About The Pull Request
Bottle bomb for exploding is dreadful. What this does
- Lowers the default probability of failure to 1 (1%?)
- Moves the qdel around so it only deletes when it explodes or smashes apart.

## Why It's Good For The Game
The bottle bomb's rate of failure, along with its ability to delete itself without any warning frustrates many. It's a problem. This makes them a lot more likely to succeed. 

Should a bottle bomb fail to explode it will lay on the floor defused. This will allow you to try again or anyone else to pick it up, however the fuse is now a lot shorter so be careful! There is also a chance (very rare) the bottle bomb will snuff out in your hand. In this case the bottle bomb no longer has a fuse and would ignite on being exposed to a flame. This would be a pretty rare circumstance but I also think it would be funny to see that so I left it in.

https://github.com/user-attachments/assets/21d467f0-f760-403c-a11d-2e26253b286e


ARefrigerator:
wait I fucked up somehow. let me do this again.

PR #2446, opened 2025-02-20
 Bottle Bomb: lowers chance of fail, fixes qdel issue.

ARefrigerator:
## About The Pull Request
Bottle bomb for exploding is dreadful. What this does

- Lowers the default probability of failure to 1 (1%?)
- Moves the qdel around so it only deletes when it explodes or smashes apart.

## Why It's Good For The Game

The bottle bomb's rate of failure, along with its ability to delete itself without any warning frustrates many. It's a problem. This makes them a lot more likely to succeed.

Should a bottle bomb fail to explode it will lay on the floor defused. This will allow you to try again or anyone else to pick it up, however the fuse is now a lot shorter so be careful! There is also a chance (very rare) the bottle bomb will snuff out in your hand. In this case the bottle bomb no longer has a fuse and would ignite on being exposed to a flame. This would be a pretty rare circumstance but I also think it would be funny to see that so I left it in.

https://github.com/user-attachments/assets/f3e6b957-ca66-44dd-8af7-c7d0836bbd54

### DIFF
--- a/code/game/objects/items/rogueitems/bombs.dm
+++ b/code/game/objects/items/rogueitems/bombs.dm
@@ -11,7 +11,7 @@
 	throw_speed = 0.5
 	var/fuze = 50
 	var/lit = FALSE
-	var/prob2fail = 23
+	var/prob2fail = 1 //23 was lunacy
 
 /obj/item/bomb/spark_act()
 	light()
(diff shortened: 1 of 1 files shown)


Actuallygoinginsanern:
I still think if a bottle bomb fails to explode it should disappear, or at the least shatter, you're still chucking a glass bottle at the ground with all your might, it doesn't exactly make sense for it to just dink the ground and be completely fine.

Gyrandola:
Was tempted to test merge this but it is honestly just a fix.

Gyrandola:
> I still think if a bottle bomb fails to explode it should disappear, or at the least shatter, you're still chucking a glass bottle at the ground with all your might, it doesn't exactly make sense for it to just dink the ground and be completely fine.

Yeah thats a good idea too.

Gyrandola (approved):

PR #2404, opened 2025-02-13
No Walls, Only Mountain - Bastion Gate Remake + Town Fixes

ophaq:
## About The Pull Request

Adds the following:
- A new towner house near the well behind smithing guild with a scenic view and inside the mountain.
- The towner house near the church that looked shitty near the stairs? Has a little cave entrance now. It leads nowhere but it is neat. Feels more like a slummy towner house.
- Removed scholar's closet starting essences and drugs in trade for a westleach and swampleaf so they can consider gardening or giving it to soilsons. The court physician gets these and it gives them something to on the side. The fish and meat essences stayed though.
- Numerous small fixes and changes, such as merchant window fixes and removing an extra vomitorium in scholar lab.
- Many area reconversions to mountain area as it should be done with this new update adding more mountain. Some were wrong too so I took the liberty to fix those up. Only thing this affects is sound effects and game recognizing it is a 'mountain' zone for those curious.
- Some fixes in the druid grove, aesthetics like will-o-wisp mote lights/waterfall changes.

Biggest Change:
- Bastion walls are now mountain. The surrounding mountains now 'hug' the bastion gate towers as if the town was built into a large alcove within a mountain. Although we still have the ocean nearby to the north so not quite like that but you get the picture.
- No more pit, and a single gate at the front that is PREOPENED round start. Towners rejoice, no more waiting for people to roll it up on lowpop or because they're lazy.

## Why It's Good For The Game
I've heard of people voting or suggesting this. There was a PR the other day that did this poorly so I took some time out of my day to do a proper looking version. You can dig around town now if you want to slip by sneakily. Gate sucked anyway and a lot of people hated it, but at least it's only one now and no more autism pit that didn't get used. You can properly siege the town now without falling in said pit.

## Proof of Testing (Required)
Main changes, should capture a lot of what I did:
bastion1
bastion2
bastion3

PR #2452, opened 2025-02-21
Tavern map changes

john-pawson:
## About The Pull Request

This PR changes the tavern to be more of a social place, rather than.. whatever it was supposed to be before. It adds more tables small where people can sit around and talk with eachother, adds some cards and dice for people to play, adds a couple of tables outside so people can sit around there while watching whats happens around on the street or watching the local execution, adds a couple of windows so people can take a peek inside or look at what's happening outside without having to leave or enter, adds shutters to the bar for when the barkeep isn't around, and changes the kitchen layout, so noticing when a cook is around is easier and the cook can more easily put multpile dishes without struggling to fit it all in one table.

Photos are on the proof of testing.

## Why It's Good For The Game

Hopefully it'll make the tavern more of a social place where people can hang around and chat, like it was before, giving more chances to roleplay and do some activities with the dice and cards.

image
image
image
image
image


ophaq:
This is good but wait until this is merged:

https://github.com/Rotwood-Vale/Ratwood-Keep/pull/2404

RadiantFlash7:
This is unfortunately severely out of date with the current map.

john-pawson:
Map updated, and game launched, no warnings and doesn't seem to have any conflicts either.
Should be ready!
image

Gyrandola (approved):

PR #2465, opened 2025-02-24
Gives Dungeoneer Bed Breaker

Coolgat3:
## About The Pull Request

As title

## Why It's Good For The Game

Because if bandits are evil enough to be able to break your rib cage with their massive cock, how come the dungeoneer isn't?

Single line skill addition, it works because that's **exactly** how traits are applied for other jobs

### DIFF
--- a/code/modules/jobs/job_types/roguetown/garrison/dungeoneer.dm
+++ b/code/modules/jobs/job_types/roguetown/garrison/dungeoneer.dm
@@ -59,4 +59,3 @@
 	H.verbs |= /mob/living/carbon/human/proc/torture_victim
 	ADD_TRAIT(H, TRAIT_MEDIUMARMOR, TRAIT_GENERIC)
 	ADD_TRAIT(H, TRAIT_DEATHBYSNUSNU, TRAIT_GENERIC)
-


code/modules/jobs/job_types/roguetown/garrison/dungeoneer.dm
```diff
@@ -58,3 +58,4 @@
 	H.dna.species.soundpack_m = new /datum/voicepack/male/warrior()
 	H.verbs |= /mob/living/carbon/human/proc/torture_victim
 	ADD_TRAIT(H, TRAIT_MEDIUMARMOR, TRAIT_GENERIC)
+	ADD_TRAIT(H, TRAIT_DEATHBYSNUSNU, TRAIT_GENERIC)
```
EvenInDeathIStillServe, line 61: add a newline at the end of the file

EvenInDeathIStillServe (changes requested):

code/modules/jobs/job_types/roguetown/garrison/dungeoneer.dm
Coolgat3: well. there is a new line when i try to edit it, i dunno.

PR #2466, opened 2025-02-24
Inquisition ship

Dilligentdog:
## About The Pull Request

Moves the Inquisition onto a ship added into the harbor, as well as locks the doors on the ship behind the Puritan key instead of the Confessional key.
I have not done anything to the church dungeon yet, because i want to figure out with people what the best thing to do

## Why It's Good For The Game

The church has a habit of white knighting anytime the inquisitor/confessor tries to do any type of torture towards a heretic to get a confession. I feel like moving the Inquisition out of the church basement will decrease the amount of church people harassing the inquisition for doing their job

Torture room, cells, and storeroom
image
Main deck, as well as the Inquisitor's bedroom
image
Confessor's bedrooms and their desk
image


Hmeister-real:
don't like this, that ship is a very nice location to do criminal things out of the way, while still in town as a regular role.

Limeon666:
Idk why this is controversial. I'd like to see it tm'd and if it's bad in game it can be reverted, but I think it's neat.

Dilligentdog:
> don't like this, that ship is a very nice location to do criminal things out of the way, while still in town as a regular role.

Not sure at all what this means, but i am also designing a second change where the inquisition will be on the left side of the town is a ruined manor instead

Kregle:
It is incredibly bland for a faction with such incredible personality.

RadiantFlash7 (changes requested):
PR is out of date, needs to resolve conflict merges.
All and all, it looks good otherwise.

Dilligentdog:
> PR is out of date, needs to resolve conflict merges. All and all, it looks good otherwise.

how does one exactly resolve a mapping conflict? Never done it before

Dilligentdog:
alright, i seem to have made the conflicts worse, so i will just redo this PR as fixing a PR this outdated will be more annoying then just redoing it

PR #2474, opened 2025-02-25
Fireman changes

Rayforx:
## About The Pull Request
Changes fireman (and hard grabbed people) to soft grab when they are moved up a staircase

## Why It's Good For The Game
People did not like fireman carried people to default to hard grab when carrying people up stairs.

https://github.com/user-attachments/assets/7a32f4c2-0c07-4de8-b3ac-32f5a6e08421

### DIFF
--- a/code/game/objects/structures/stairs.dm
+++ b/code/game/objects/structures/stairs.dm
@@ -148,6 +148,4 @@
 		pulling.forceMove(newtarg)
 		L.start_pulling(pulling, supress_message = TRUE)
 		if(was_pulled_buckled) // Assume this was a fireman carry since piggybacking is not a thing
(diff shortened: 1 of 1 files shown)


ophaq:
Can you change it so that if you are hard grabbing them with both hands your required strength is 8 to move at normal speed? And if doing one handed then required strength is say, 12? I don't know if this is already a feature but I think Azure has something similar and currently just fireman carrying people is ass. Might as well drag them.

EvenInDeathIStillServe (approved):

Kevak:
@EvenInDeathIStillServe Why was this full merged without a testmerge? Any significant mechanical adjustment needs to be testmerged. Come now.

PR #2478, opened 2025-02-26
Gives Smash Intent To Tome of Justice

EvenInDeathIStillServe:
## About The Pull Request

gives a weak smash intent to the tome of justice

## Why It's Good For The Game

justice comes swift and hard

image


ophaq:
We need this ASAP so I can hit people with a book. It would be funny if it used mace skill as well.

RadiantFlash7 (approved):
An amusing, rather small change.

PR #2479, opened 2025-02-26
Do you remember? You have known people's species in your memory now.

Kepteyn:
Simply makes it so you remember the species of people you know.

## Why It's Good For The Game

You can remember people's names, age, gender, and job with just a glance, but somehow whether or not the person you're looking at is a human, or really short, hisses or is an actual dog just keeps slipping your mind? No longer. Your memory has been IMPROVED.

Untitled

Tested by spawning in mobs via game panel. I have not been able to test this with other, actual players, but it should work just fine.

### DIFF
--- a/code/datums/mind.dm
+++ b/code/datums/mind.dm
@@ -143,6 +143,7 @@
 			used_title = "unknown"
 		known_people[H.real_name]["FJOB"] = used_title
 		known_people[H.real_name]["FGENDER"] = H.gender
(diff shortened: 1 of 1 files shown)


Limeon666:
Godly

Kepteyn:
The position of the species name can be moved by the way, but I put it after the job as I figured it'd be one of the most notable features of a person.

PR #2482, opened 2025-02-28
Adds protection values to inspect [Azure port]

john-pawson:
## About The Pull Request

This partially ports this PR: https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/1294 
Specifically, the ability to know the amount of damage certain armors absorb after inspecting them.

Now, apart from saying which bodyparts they cover and what they prevent crits from, inspecting armor also tells you the amount of damage it can absorb from certain damage types, by using a letter tier system:
F = No protection
D = 1-19
D+ = 20-39
C = 40-49
C+ = 50-59
B = 60-69
B+ = 70-79
A = 80-89
A+ = 90-99
S = 100

## Why It's Good For The Game

This may come to a shock to us code-dwellers, but currently, there's no way to know anything about damage absorption in the game! 
Unless you manually test out and know the exact amounts of damages and penetration on certain weapons, there's no way to know which armor protects what amount, there's no way to know that hardened leather is weaker to stabbing than normal leather, there's no way to know that brigandines have higher blunt protection but lower stabbing and cutting protection, there's no way to know that certain armors have more protection against arrows and bolts than others!

Wich has given me a reminder that, there's a very high chance that a lot of people playing the game just don't know any of this stuff about armors, or that it was here on the first place.
But now they can! And they can know what are the pros and cons, and what armors protect against what things, compare different protections on different armors, etc.

image
image
image
image
image
image

### DIFF
--- a/code/game/objects/items.dm
+++ b/code/game/objects/items.dm
@@ -1253,48 +1253,48 @@ GLOBAL_DATUM_INIT(fire_overlay, /mutable_appearance, mutable_appearance('icons/e
 		if(C.armor)
 			var/defense = "<u><b>ABSORPTION: </b></u><br>"
 			var/datum/armor/def_armor = C.armor
-			defense += "[colorgrade_rating("BLUNT", def_armor.blunt, elaborate = TRUE)] | "
-			defense += "[colorgrade_rating("SLASH", def_armor.slash, elaborate = TRUE)] | "
-			defense += "[colorgrade_rating("STAB", def_armor.stab, elaborate = TRUE)] | "
-			defense += "[colorgrade_rating("PROJECTILE", def_armor.bullet, elaborate = TRUE)] "
+			defense += "[colorgrade_rating("BLUNT", def_armor.blunt)] | "
+			defense += "[colorgrade_rating("SLASH", def_armor.slash)] | "
+			defense += "[colorgrade_rating("STAB", def_armor.stab)] | "
+			defense += "[colorgrade_rating("PROJECTILE", def_armor.bullet)] "
 			str += "[defense]<br>"
 		else
 			str += "NO DEFENSE"
 	return str
 
-/proc/colorgrade_rating(input, rating, elaborate = FALSE)
+/proc/colorgrade_rating(input, rating)
 	var/str
 	switch(rating)
 		if(0)
 			var/color = "#f81a1a"
-			str = elaborate ? "<font color = '[color]'>[input] (F)</font>" : "<font color = '[color]'>[input] (F)</font>"
+			str = "<font color = '[color]'>[input] (F)</font>"
 		if(1 to 19)
 			var/color = "#680d0d"
-			str = elaborate ? "<font color = '[color]'>[input] (D)</font>" : "<font color = '[color]'>[input] (D)</font>"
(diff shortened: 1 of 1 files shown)


Radioaction-Cyber:
I don't think this is a good idea, just because this will encourage minmaxing and gaming the system even further. People will just now find the best possible armor and use it. Rather than this, it might just be better to change the armor list so we don't have 568125 different armors that are simply unused.

john-pawson:
> I don't think this is a good idea, just because this will encourage minmaxing and gaming the system even further. People will just now find the best possible armor and use it. Rather than this, it might just be better to change the armor list so we don't have 568125 different armors that are simply unused.

While i do see what you mean, people that were already looking for the best values of armor probably can, and already have been looking at the armor values for the best combinations on the code. 
Not only that but it'd also make getting feedback in terms of balancing certain armors that might get picked a lot easier too, some people didn't know that the best armors were, well, the best at absolutely everything, and letting people know this is a good way of getting them to say "Hey, this is currently too OP, you should probably nerf it", since now they know that's the reason.

notafurryy:
> I don't think this is a good idea, just because this will encourage minmaxing and gaming the system even further. People will just now find the best possible armor and use it. Rather than this, it might just be better to change the armor list so we don't have 568125 different armors that are simply unused.

most people already know the vague protections, also you can just think. For the most part it's obvious that things that cost more, or are plate rather than chain or leather will be better.

RadiantFlash7 (approved):
A good PR- I'm a fan of providing more transparency to players so they know what exactly is what, without code diving. My only concern, as shared with other maintainers is that the system of identifying the quality of armor may be a bit too 'gamey'/ 'jrpg-y'.
However, none of us have any other ideas in how to improve that. As it is, This is ready for a test merge, and to get player feedback.

code/game/objects/items.dm
```diff
@@ -1242,3 +1246,55 @@ GLOBAL_DATUM_INIT(fire_overlay, /mutable_appearance, mutable_appearance('icons/e
...
+			defense += "[colorgrade_rating("BLUNT", def_armor.blunt, elaborate = TRUE)] | "
+			defense += "[colorgrade_rating("SLASH", def_armor.slash, elaborate = TRUE)] | "
+			defense += "[colorgrade_rating("STAB", def_armor.stab, elaborate = TRUE)] | "
+			defense += "[colorgrade_rating("PROJECTILE", def_armor.bullet, elaborate = TRUE)] "
+			str += "[defense]<br>"
+		else
+			str += "NO DEFENSE"
+	return str
+
+/proc/colorgrade_rating(input, rating, elaborate = FALSE)
```
FreeStylaLT, line 1265: Heya! Original coder here, you may want to get rid of the `elaborate` parameter, unless you wish to make the readout more "vague" under some or all conditions. keep in mind that it only changes the letter grades and doesn't change the actual color, so it's a token obscurement at best.

john-pawson: Oh yes! I don't know how i didn't notice that little detail, but considering how the idea is to see the little differences between armors, i might go ahead and remove it, considering how there's a big amount of armors which the entire difference is that one is D (Lower side) and the other is D+ (Higher side). Thanks for pointing it out.

PR #2485, opened 2025-03-01
Arquebus pistol now can be used one-handed

john-pawson:
## About The Pull Request

Does as the title says, it was probably overlooked by the creator for the firearm rework, so this simply makes the arequebus pistol work exactly as it did before, but it can't be wielded anymore and it can shoot while one-handed.

## Why It's Good For The Game
Bug fix! Pistol is now a one handed weapon, as it should be.

image
image

### DIFF
--- a/code/game/objects/items/rogueweapons/ranged/gunpowder/firearm.dm
+++ b/code/game/objects/items/rogueweapons/ranged/gunpowder/firearm.dm
@@ -100,12 +100,6 @@
 /datum/intent/shoot/firarm
 	chargedrain = 0
 
-/datum/intent/shoot/firearm/can_charge()
-	if(mastermob && masteritem.wielded)
-		if(!masteritem.wielded)
-			return FALSE
-		return TRUE
-
 /datum/intent/shoot/firearm/get_chargetime()
 	if(mastermob && chargetime)
 		var/newtime = chargetime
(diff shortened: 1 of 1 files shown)


Radioaction-Cyber:
I think this is good. Arquebus is in a good spot right now- considering the Duke is the only one who spawns with it. Gives the Duke a better chance at self defense in a desperate situation.

notafurryy:
> I think this is good. Arquebus is in a good spot right now- considering the Duke is the only one who spawns with it. Gives the Duke a better chance at self defense in a desperate situation.

Why would they do anything but spam the master rod for self defense though

john-pawson:
> Why would they do anything but spam the master rod for self defense though

Bandits tend to usually lack a nervelock implant.
And most of the deadites that tend to like attacking the keep.

notafurryy:
> Bandits tend to usually lack a nervelock implant.
> And most of the deadites that tend to like attacking the keep.

in both cases you usually get swarmed, and that one shot isn't saving you realistically

john-pawson:
> > Bandits tend to usually lack a nervelock implant.
> > And most of the deadites that tend to like attacking the keep.
> 
> in both cases you usually get swarmed, and that one shot isn't saving you realistically

Better than nothing. That and it helps with executions now that the pit isn't here.

EvenInDeathIStillServe (approved):

PR #2492, opened 2025-03-01
Wanted bandits

Rayforx:
## About The Pull Request
**This PR need to be TM'ed for very clear balance reasons.**

This adds the "WANTED" trait, a optional play style for bandits that would rather start out as wanted, but with more skills.
This should not be a straight up "buff" for bandits, but the bandits choosing to have a "curse" to get better stats.

This helps with a few things: 
1. Bandits right now are usually seen as very boring as passive, this was introduced when the bandit tag was removed, for better or worse.
2. A lot of keep players have a bad problem with meta, it's a huge problem that came up after the bandit tag was removed.
3. We are adding more play styles for the bandits
4. We are introducing the gate having to check people faces at the gate, something, that was kind of removed and left the vanguard with only "asking" for coin, and not doing anything else at the gate.

In terms of stats:
If you choose not to be wanted: You get a poor pouch with coins. Something, that was lacking beforehand so they could not enter the town.

A change for the hedgeknight will be made, as he will be wanted by default now and instead of 30 PQ then it will be 15 PQ (still a lot of PQ, but not insanely so)
Rogue Mage will be getting armor by default (something he have not had for some reason) and a leather belt
```
armor = /obj/item/clothing/suit/roguetown/shirt/robe/newmage/adept
belt = /obj/item/storage/belt/rogue/leather
```
If wanted: 
brigand: 
```
axes, 1
maces, 1
whipsflails, 1
unarmed, 1
```
foresworn
```
polearms, 1
wrestling, 1
unarmed, 1
swords, 1
crossbows, 1
```
knave
```
endurance 2
constitution 1
sneaking, 1
athletics, 1
```
sawbone
```
knives, 1
swords, 3
wrestling, 1
athletics, 1
treatment, 1
```
rogue mage
```
backpack_contents = list(/obj/item/book/granter/spellbook/expert = 1)
intelligence 2
arcane, 1
```
hedge knight
```
intelligence 1
speed 1
```

## Why It's Good For The Game
More options for play styles
Introducing more shenanigans that should make for less boring rounds
Adding more for the roles to do, while it being "important" to check if people are bandits again

Skærmbillede 2025-03-01 150921
Skærmbillede 2025-03-01 134912
Skærmbillede 2025-03-01 134955

https://youtu.be/QSSCM68oApI


Radioaction-Cyber:
I can see the idea and concept, but the application would be horrible. As someone who regularly plays both bandit and retinue, this would be horrid. The idea of an opt-in challenge could be good, but increasing stats/skills in such a way would horribly skew game balance.

The buffs would need to be limited to 1 bandit per round, like Hedge Knight is currently limited to 1 HK, or the numbers need to be heavily nerfed.

OspreyIsHere:
> I can see the idea and concept, but the application would be horrible. As someone who regularly plays both bandit and retinue, this would be horrid. The idea of an opt-in challenge could be good, but increasing stats/skills in such a way would horribly skew game balance.
> 
> The buffs would need to be limited to 1 bandit per round, like Hedge Knight is currently limited to 1 HK, or the numbers need to be heavily nerfed.

I don't think it necessarily speaks of it being horrible. If you wish to be subtle and (try) to blend-in, you can choose not to! You may be less skilled than a wanted (therefore possibly a 'veteran') brigand, but you are still a threat. The strength of the Free Men is the fact they can always know who they are.

The 'passive' bandit players (steal vault 10 mins before round end) will remain passive. But those who want to be a threat can and will stand up as such. And upping skills to lvl 4 in some cases isn't terrible when most of the retinue has Expert on at least 1 weapon + wrestling + numbers and the good nature of townsfolks willing to help!

Kevak:
Armor better than light armor is literally useless for the rogue mage. Having medium or better armor drastically amplifies stamdrain from spellcasting

RadiantFlash7:
Commenting on this: The rogue mage should probably get an adept book vs a expert one. but other then that, seems fine with the chalkremoved.

EvenInDeathIStillServe (approved):

PR #2234, opened 2025-01-16
Wednesday Night Bugfixes/QoL

Kevak:
## About The Pull Request

This PR fixes a number of bugs, provides several QoL features, and readds an emote that was broken by one of the cleanup PR's.

-Fixes MMB Instant Casting - Credit [Azure Peak 1138 ](https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/1138)
-Fixes several missing female sprites
-Gives Axions a shark tail that was in the code but wasn't accessable to them
-Adds two additional shark tail types - Credit Bubber/whatever server the tails originated on.
-Adds an additional angel wing sprite - Credit patches1775 (Anya Dawnheart)
-Restores the ability of those with valid wing sprites to open and close their wings using *wing similar to how wagging one's tail works. (Note, not many wings have this functionality.)
-Moves wingflapping to *flap
-Gives the signal horn a 30 second cooldown, to avoid spam.
-I think there were like one or two other quick bugfixes but I forgot what they were cuz they went fast and weren't fiddly or annoying and thus did not stick in my kittybrain.

For those who want wings that open and close with *wing, the ones that currently do it are those in the /wings/huge filepath. If ya'll want other wings that do it, gimmeow more splayed open wing sprites! 

Lemme know if I forgot to strip out a debug message anywhere, I don't think I did but I might have.

## Why It's Good For The Game

Bugfixes, spritefixes, and QoL is all good

##

Screenshot 2025-01-16 024439
Screenshot 2025-01-16 024454
Screenshot 2025-01-16 024502
Screenshot 2025-01-16 024514

PR #2500, opened 2025-03-03
Fixes Leash Belt Bug. Again.

Kevak:
## About The Pull Request

This is the third time I've fixed this exact same identical bug. I do not know how it keeps getting stealth reverted. I am a mite bit irate.

## Why It's Good For The Game

Bugfix good. Even if it's fixing the same bug repeatedly.

https://github.com/user-attachments/assets/694b615b-2ac1-4f7d-afce-9c69fac4bbb9

### DIFF
--- a/code/game/objects/items/rogueitems/leash.dm
+++ b/code/game/objects/items/rogueitems/leash.dm
@@ -58,7 +58,7 @@
 	equip_sound = 'sound/foley/equip/rummaging-01.ogg'
 	drop_sound = 'sound/foley/dropsound/cloth_drop.ogg'
 	throw_range = 4
-	slot_flags = ITEM_SLOT_BELT|ITEM_SLOT_HIP
+	slot_flags = ITEM_SLOT_HIP
 	force = 1
 	throwforce = 1
 	w_class = WEIGHT_CLASS_SMALL


EvenInDeathIStillServe:
Is there an issue with having the ability to wrap a leash around your waist? I don't get it.

john-pawson:
> Is there an issue with having the ability to wrap a leash around your waist? I don't get it.

Unless the item you're putting on a belt slot is specifically one with a storage capacity, it breaks, if you're able to take it off the belt slot is permanently bugged and you can't put anything else on it, and the leash also gets a bit fucked too.
The leash doesn't have a storage capacity, so it breaks when you put it on. It does also runtime, so it means that whoever reverted it, if they did so intentionally, didn't test it.

So yes. Big issue.

john-pawson:
Actually, looking at the commit history, it looks like the first bugfix didn't exactly fix the bug, since it only deleted it from being worn on the hip, not the belt (The problematic slot) https://github.com/Rotwood-Vale/Ratwood-Keep/pull/2215
And then, a second bugfix went ahead and re-added the hip slot. Because, well, the hip slot doesn't have any problems, it's the belt slot that breaks things. https://github.com/Rotwood-Vale/Ratwood-Keep/pull/2234

So. It looks like the person that has been stealth-reverting and not-fixing your stuff was you all along!
image

Kevak:
> Actually, looking at the commit history, it looks like the first bugfix didn't exactly fix the bug, since it only deleted it from being worn on the hip, not the belt (The problematic slot) #2215 And then, a second bugfix went ahead and re-added the hip slot. Because, well, the hip slot doesn't have any problems, it's the belt slot that breaks things. #2234
> 
> So. It looks like the person that has been stealth-reverting and not-fixing your stuff was you all along! image

Damnit me

EvenInDeathIStillServe (approved):

Kevak:
> Is there an issue with having the ability to wrap a leash around your waist? I don't get it.

Yeah for some dumb reason beltslot breaks if a belt slot object lacks storage. It needs a refactor but is low priority

PR #2501, opened 2025-03-03
New mail sound

Purpleturtle3:
## About The Pull Request

What it says on the title, believe it or not.

## Why It's Good For The Game

Makes personal mail harder to miss

https://imgur.com/hwoaeFg
image

### DIFF
--- a/code/modules/roguetown/roguemachine/mail.dm
+++ b/code/modules/roguetown/roguemachine/mail.dm
@@ -173,7 +173,10 @@
 				else
 					visible_message(span_warning("[user] sends something."))
 					playsound(loc, 'sound/misc/disposalflush.ogg', 100, FALSE, -1)
-					send_ooc_note(span_boldwarning("New letter from <b>[sentfrom].</b>"), name = send2place)
+					for (var/mob/living/carbon/human/H in world)  
+						if (H.real_name == send2place)  
+							to_chat(H, span_notice("A new letter from [sentfrom] has arrived!"))  
+							playsound(H, 'sound/misc/mail.ogg', 100, FALSE, -1)  
 					return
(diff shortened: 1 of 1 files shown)


code/modules/roguetown/roguemachine/mail.dm
```diff
@@ -173,7 +173,10 @@
 				else
 					visible_message(span_warning("[user] sends something."))
 					playsound(loc, 'sound/misc/disposalflush.ogg', 100, FALSE, -1)
-					send_ooc_note(span_boldwarning("New letter from <b>[sentfrom].</b>"), name = send2place)
+					for (var/mob/living/carbon/human/H in world)
```
EvenInDeathIStillServe, line 176: Don't do this. Iterating through all the mobs in the WORLD is super slow.

Stick the sound in the `send_ooc_note` proc, ideally as a new argument for a sound effect to play for the receiver.

EvenInDeathIStillServe (changes requested):

ophaq:
You should change it to the AOL 'you got mail' sound.

code/modules/roguetown/roguemachine/mail.dm
Purpleturtle3: The thing is that the send ooc note does nothing right now, that's why I changed it, people are not getting notifications. I am not the best coder ever so I don't a better way to do it.

EvenInDeathIStillServe: Would you let me make a PR myself to fix send_ooc_note and add the sound to it?

Purpleturtle3: Alright.

PR #2508, opened 2025-03-05
Removes prob chance to fail cleaning floors (Why???)

ARefrigerator:
## About The Pull Request
Removes the probability check that makes you fail cleaning turfs

## Why It's Good For The Game
It never made sense to begin with, there's no message to tell you it failed and there's no cleaning skill associated with it anyway. Until there is it's such a simple task it should always succeed, no? Anyone who brings it up gets mad.

https://github.com/user-attachments/assets/85661608-bf3e-4ed7-847a-b508eebbd3b9


Radioaction-Cyber:
I didn't even realize it was a chance to fail. Also, why the fuck is there a chance to fail?
Good PR.

Hmeister-real:
it's so that people have to keep scrubbing, it's annoying but it adds soul

ARefrigerator:
I mean if you really want 'soul' I could try to think of some alternative.

PR #2236, opened 2025-01-16
[MAP REWORK] A Whole (Almost) New World

ophaq:
## About The Pull Request

Contrary to the title, this PR primarily touches up on and reworks hightown, lowtown, and sewers. The docks have been moved closer, the ship has been removed, and public stockpiles are now (temporarily) removed with the tavern, smithy, keep, and the merchant possessing them as part of an economic experiment.

**Now, before you go flaming this I want you to know this has been like four days in the making all by myself. Please give me CONSTRUCTIVE FEEDBACK. I know some of these changes are going to be controversial for the economy but things can get changed around later.**

I'll go into detail on what I touched up on but I'll tell you this, I am going to miss out listing on things I touched on because I'll probably forget to. I'll just say what I remembered I did and the main features. This is a PR that has been in the works and discussed through discord with coordination of maptainers and developers.

**If you have any bugs or issues list them here or through the mapping discord channel.**

I went through this map with a keen eye and tested it several times but who knows what shit I didn't notice. Please report weird stuff if you find any.

## Why It's Good For The Game

Now, onto the good stuff. This list is gonna be a bit long so buckle your seat belts.

- Rearranged the high town. There is now a central market area with stalls to peddle your goods. You can buy a key from the main peddler for 10 mammon.
- Smithy is located on east town now as well as artificer guild next to the market.
- Physician house sits in front of the church at the market as well and acts as an apothecary when not taking clients because I know you people are just gonna go to the church anyway to play medbay but now it's not out of the way, shoved in some corner of town.
- Every single roof in hightown and lowtown has been redone to look more aesthetically pleasing. Pipes have been added between some roofs to allow sneaky individuals better trekking across them.
- Lighting has been overhauled in some areas to make them more dark and atmospheric. Less torches overall in some places. Go find some in a chest or make a torch if you really want light. Or be a man and go through the dark alley ways to get mugged.
- Lighthouse has been relocated to the entrance of the previous docks. You go through the underpass to access it or through an alley. A nice shoreline and fisherman house awaits you below that. Also the lighthouse has a few hidden surprises in the basement.
[...]

PR #2509, opened 2025-03-05
Makes brick walls climbable

notafurryy:
## About The Pull Request

The stone brick and decorated wall were spread all across the town in: https://github.com/Rotwood-Vale/Ratwood-Keep/pull/2236 

## Why It's Good For The Game

The walls mentioned were used as decorative walls and with how prevalent they are they are limiting the possibilities for climbing and exploration quite a bit and the author of that pr agreed they should be climbable.
image

image


ophaq:
Starcraft Terran Commander: "Hell, it's about time."

Alanii:
I assume this just makes them climbable dependent on skill and not generic climbing, yes?

notafurryy:
> I assume this just makes them climbable dependent on skill and not generic climbing, yes?

It removes the var making it unclimbable

PR #2514, opened 2025-03-06
Mage/magic tweaks and summon fixes,

RadiantFlash7:
## About The Pull Request
Ports the Vanderlin changes to Find Familiar. Spell now spawns the familiar on the same tile as the caster, but it will target whoever was targetted with find familiar spell cast. Also uses new AI so a fair bit more dangerious.
All summon drops and mage items (where appropriate) are now fire proof instead of flammable. (incorrect flags were applied)
Adds Anti-magic checks for all wizard spells missing anti-magic checks.
Summoning a mob with a summoning circle will cause nearby murderholes to crumble, as an anti-cheese measure.
Further more,  summoned mobs now break and damage railings and palisades they're ontop of when trying to reach a target.
Summoned mob object damage has been increased as well.
Behemoth's now use quake properly.
Force wall spell increased to 2 points, wall integrity upped to 200.
Heartfelt Magos loadout updated to use new mage system

## Why It's Good For The Game
Anti-cheese additions prevent cheesey gaming of the system.
Fixes for mage spells not being blocked by anti-magic effects like the null ring.
Summoned loot drops no longer catching fire is also rather useful consider mages could blow up their loot from a spell they cast in combat.
Find familiar now plays and feels much better.

image
image
image


notafurryy:
> Find Familiar. Spell now spawns the familiar on the same tile as the caster, but it will target whoever was targetted with find familiar spell cast

FINALLY i cant count how many times I got friendly fired by those

Radioaction-Cyber:
Yippee. Another Vanderlin ripoff.
Mages don't need another buff.

notafurryy:
> Yippee. Another Vanderlin ripoff. Mages don't need another buff.

What? It just fixed things?

Limeon666:
> Yippee. Another Vanderlin ripoff. Mages don't need another buff.

That was my first reaction, but then I learned that the wolf now spawns on the mage. It's a good change overall because what made it strong before is you could spawn volfs in people's fortifications and such.

RadiantFlash7:
> Yippee. Another Vanderlin ripoff.
> 
> Mages don't need another buff.

In truth, it doesn't seem like a massive buff. You can no longer spawn familiars where ever, meaning you lose a lot of utility. In exchange for both better AI, and the ability to directly give it a target, rather then having it target an ally on spawn.

Gyrandola (approved):

PR #2520, opened 2025-03-07
Gave gambeson protection to grenz pants

Jackofashepard:
## About The Pull Request

Simply gives gambeson protection to grenzelhoft pants

## Why It's Good For The Game

Its kinda sad seeing grenzels throw away nearly every piece of their uniform so they don't get critted in a single stab. Its not so much an issue with the shirt since its covered by the cuirass but the pants leaves the legs pretty much unprotected. Its still not great at all but it at least holds up decently well now to not make them throw it away immediately for something like leather pants.

image

### DIFF
--- a/code/modules/clothing/rogueclothes/pants/pants.dm
+++ b/code/modules/clothing/rogueclothes/pants/pants.dm
@@ -205,8 +205,11 @@
 	item_state = "grenzelpants"
 	sleeved = 'icons/roguetown/clothing/onmob/helpers/stonekeep_merc.dmi'
 	detail_tag = "_detail"
-	armor = list("blunt" = 15, "slash" = 15, "stab" = 15, "bullet" = 0, "laser" = 0,"energy" = 0, "bomb" = 0, "bio" = 0, "rad" = 0, "fire" = 0, "acid" = 0)
-	prevent_crits = list(BCLASS_BLUNT)
+	armor = list("blunt" = 45, "slash" = 30, "stab" = 35, "bullet" = 15, "laser" = 0,"energy" = 0, "bomb" = 0, "bio" = 0, "rad" = 0, "fire" = 0, "acid" = 0)
+	prevent_crits = list(BCLASS_CUT, BCLASS_BLUNT)
+	blocksound = SOFTUNDERHIT
+	blade_dulling = DULLING_BASHCHOP
+	armor_class = ARMOR_CLASS_LIGHT
 	colorgrenz = TRUE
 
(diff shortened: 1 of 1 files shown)


RadiantFlash7:
In the future, try to make proof of testing an actual screenshot of the item in game.
Even if it's hard/impossible to show the changes, since it's all stat ones; We already know that it compiles because it passes the compile check. If we see that it's still present in game, It's better then nothing.

Jackofashepard:
> In the future, try to make proof of testing an actual screenshot of the item in game. Even if it's hard/impossible to show the changes, since it's all stat ones; We already know that it compiles because it passes the compile check. If we see that it's still present in game, It's better then nothing.
> 
> As for the change itself, I'm somewhat skeptical. On one hand, I'm all for encouraging grenzel's to maintain their outfit. On the other, Them having weaker armor everywhere else, is the trade off for having expert skill, MAA level stats, and the best chest armor in the game aside from plate armor.

Its really not that much of a buff, a Gambeson is nice for not getting 1 shot critted and that's about it. There are also a decent amount of roles with way better skills and armour than a grenzel: Knight and the retinue veteran to name just two. I don't think its a good idea to have that kinda mindset and to rather look at it from a larger perspective of neutral/antag/duchy aligned groups and the total power in each. Grenzels are neutral and can do just about anything and when looking at the neutral group as a whole it wouldnt be much of a offset in power. In comparison the duchy is so strong that it would be actual suicide to take it on without really smart tactics/overwhelming numbers/gear (and this is just assuming they dont gear their MMA's to knight tier).

As for the testing, bro, you can see the code. Nothing is removed, its a literal stat change, its the equivalent of renaming the pants to "Jack's booty protector" and asking me please take a screenshot in game.

Jackofashepard:
I would also like to add that the biggest thing on this server is having any armor at all rather than armor itself. Having a full set of knight armor is great and all but if even one armour slot is empty you might as well not be wearing anything at all due to how crits and armour are handled. For example this grenzel pants gambeson only covers crotch and legs. The entire head, eyes and nose are still completely exposed in addition to a lesser extent the hands and feet. Even 1 hit can kill this grenzel easily. All this PR really does is give them some "I will not die in one crit to the legs" protection so grenzels aren't as bothered to immediately throw most of their uniform away because even leather pants would be an upgrade. Do note this doesnt even protect from crits inflicted by stabs so guns, arrows and a stab intent will still absolutely kill this grenzel in a single hit if lucky.

Gyrandola (approved):

PR #2521, opened 2025-03-07
Each race's traits, followers of ten/inhumen traits, ducal family and clergy traits 

DelineFortune:
tools

## About The Pull Request

for church and shunned races content (for me)
also to add church mechanics between heretics and followers of ten
between church and keep
interactions with gods instead of lesser miracle my chest thanks

## Why It's Good For The Game

tools that I and other people can use to make their features
(we all know that 9 of 10 prs are ''im playing X job so my pr is to buff X job str +2 -> +3 but I have my hopes that someone will use it)

image


notafurryy:
there are other ways to check for a characters species, not sure why they need traits

DelineFortune:
way better than is*** race or h.job+h.job

notafurryy:
> way better than is*** race or h.job+h.job

how is it better?

PR #2530, opened 2025-03-08
Kneestinger immunity to Druid

Limeon666:
## About The Pull Request
Simple one line change that adds the kneestinger immunity traits to druids.

## Why It's Good For The Game

This was originally removed from druid in a very controversial PR with nineteen downvotes, and only three upvotes. ([#960](https://github.com/Rotwood-Vale/Ratwood-Keep/pull/960)). Druids no longer can summon kneestingers, and it's not very abuseable except against simplemobs, which can be killed with kneestingers even without this trait. The kneestinger immunity trait is also going unused at the moment which is sad because it's a flavorful inclusion for a druid.

image

### DIFF
--- a/code/modules/jobs/job_types/roguetown/church/druid.dm
+++ b/code/modules/jobs/job_types/roguetown/church/druid.dm
@@ -56,6 +56,7 @@
 		H.change_stat("perception", -1)
 
 	ADD_TRAIT(H, TRAIT_SEEDKNOW, TRAIT_GENERIC)
+	ADD_TRAIT(H, TRAIT_KNEESTINGER_IMMUNITY, TRAIT_GENERIC)
 	var/datum/devotion/C = new /datum/devotion(H, H.patron)
 	C.grant_spells(H)
 	H.verbs += list(/mob/living/carbon/human/proc/devotionreport, /mob/living/carbon/human/proc/clericpray)


CalamaBanana:
Would be neat to have that back for all Dendorites, not just for Druids.  There was no reason to remove it entirely with that self-merge salt PR back then.

radarseas:
It was removed because cleric dendorites would spam them in fullplate and decimate every single threat in the game and solo vampire lords while being cancerously impossible to escape if they floored you, because of how first-hit crit can break your legs, knock you out or floor you somehow.

that's not a thing anymore, _let alone even close to that level of armor_, **thank fuck.**

But I think its fine being exclusive to just druids and maybe werevolves to give them a tiny bit more mobility? Means there can be some cool caves or some shit mapped in exclusively as shortcuts for these roles off-the-bat.

CalamaBanana:
> t was removed because cleric dendorites would spam them in fullplate and decimate every single threat in the game and solo vampire lords while being cancerously impossible to escape if they floored you, because of how first-hit crit can break your legs, knock you out or floor you somehow.

Yes, the issue was both the spell and the easy access to it through adventurer classes. The spell was removed. Without the spell, kneestingers are not an issue anymore, and the immunity can be returned.

RadiantFlash7 (approved):
A small change, and one thats probably a good idea given we made the druid's home surrounded by kneestingers.
As dendorite miracles no longer include kneestinger spam, this should be good for merging.

PR #2536, opened 2025-03-08
Fixes decaptitation breaking re-entering bodies

RadiantFlash7:
## About The Pull Request
Fixes Decaptitation breaking re-entering bodies.
You now go back into your head on re-entering corpse.

## Why It's Good For The Game
Decapitation should no longer prevent re-entering your body now.
TLDR: Decapitating people prevented revivals. This fixes it.

image
image

### DIFF
--- a/code/modules/surgery/bodyparts/bodypart_dismemberment.dm
+++ b/code/modules/surgery/bodyparts/bodypart_dismemberment.dm
@@ -287,9 +287,9 @@
 		for(var/obj/item/worn_item in worn_items)
 			owner.dropItemToGround(worn_item, force = TRUE)
 
-	owner.ghostize(1)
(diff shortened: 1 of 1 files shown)


notafurryy:
Does it fix the skill loss issue?

john-pawson:
> > Does it fix the skill loss issue?
> 
> Theoretically it might, But It's impossible for me to solocheck, due to the way mind and ckeys work. That said This was primarily patching the lack of revivability with decaps, over the skill loss issue.

You could go ahead and launch a server through daemon, and then launch a second client without having BYOND working (Blue icon, just kill it through task manager) to enter as a guest, practically having two clients at the same time. Really useful for checking stuff like this that requires two people.

PR #2537, opened 2025-03-08
Death Consequences 2.0

notafurryy:
## About The Pull Request

Death consequences but a bit more forgiving giving you an additional chance, even if severely crippled.

## Why It's Good For The Game

Death consequences got removed with a pr that removed the standard ways of reviving people and were not readded when the pr was mostly reverted, this brings it back. Also it's weird that you can be revived 5 times and still be fully combat capable.

image
image

### DIFF
--- a/code/modules/spells/roguetown/acolyte/sawbones.dm
+++ b/code/modules/spells/roguetown/acolyte/sawbones.dm
@@ -119,6 +119,9 @@
 		to_chat(user, span_warning("They need to be mended more."))
 		revert_cast()
 		return FALSE
+	if(target.has_status_effect(/datum/status_effect/debuff/death_claimed))
+		to_chat(user, span_warning("They are too far gone."))
+		revert_cast()
 
 	var/mob/living/carbon/spirit/underworld_spirit = target.get_spirit()
 	//GET OVER HERE!


Limeon666:
This is unfathomably based

Purpleturtle3:
What about adding back the ability priests had to remove the buff from people? It's not like people play priest anyway...

Alanii:
This is fine, yes. I was curious if the original stat debuffs for being revived were removed or not, because soldiers were so quick to get back into combat.

It was weird.

notafurryy:
> It was weird. However make sure these have an expiry time.

Should they expire? I'd imagine dying would be a big deal even if you can be revived, and you can die twice now unlike the old pr which only gave you one chance. Also them expiring would kinda undermine the perma death since you could just sit and do nothing until it expires effectively never reaching the point where you can't be revived.

Purpleturtle3:
Yes, well... the book that only the priest can use was the way the debuff could be removed before...

notafurryy:
> Yes, well... the book that only the priest can use was the way the debuff could be removed before...

Thats a boring way to remove it though, its free, you should make a sacrifice if you want to recover. Maybe even dabble in heresy @DelineFortune

Limeon666:
Tbh looking at the code I think the stat reductions are too much. Limiting to only 2 revives is already a lot. Not saying there can't be stat reductions, but maybe make it -1 to all, then -2 to all instead?

notafurryy:
> Tbh looking at the code I think the stat reductions are too much. Limiting to only 2 revives is already a lot. Not saying there can't be stat reductions, but maybe make it -1 to all, then -2 to all instead?

even after dying once most combat roles still should be able to handle towners not really threats, though after you die twice you are supposed to just flat out stop combat with anyone, that's just there for you to still interact with people as a traumatised veteran.

wow50000:
not a fan of this tbh, I would rather have no revives or ways to re enter the round than forcing someone to be useless mostly.

notafurryy:
> not a fan of this tbh, I would rather have no revives or ways to re enter the round than forcing someone to be useless mostly.

You can just go to the underworld, no one is forcing you to stay, it's a choice.

Limeon666:
> > not a fan of this tbh, I would rather have no revives or ways to re enter the round than forcing someone to be useless mostly.
> 
> You can just go to the underworld, no one is forcing you to stay, it's a choice.
> 

You know revives pull you of the underworld right? There isn't a pop-up asking if you want to be revived.

But my suggestion was just to tone down the debuff a little bit because this is already a big change, and yeah it would suck having to play the game with that much stat reduction.

notafurryy:
> But my suggestion was just to tone down the debuff a little bit because this is already a big change, and yeah it would suck having to play the game with that much stat reduction.

-2 stats is still bearable, it doesn't even offset most of the role stats they get, though, there is always suicide, like infirm people sometimes resort to. Which fits since that's exactly what happens to you

Dilligentdog:
So reading through the comments, is it a unlimited revive time again but with limited revives? I'm at work and can't filly read through the code to figure it out

notafurryy:
> So reading through the comments, is it a unlimited revive time again but with limited revives? I'm at work and can't filly read through the code to figure it out

it's old death consequences, removing stat points but with a single additional chance if you die, but your stats are fucked even more.

Dilligentdog:
> > So reading through the comments, is it a unlimited revive time again but with limited revives? I'm at work and can't filly read through the code to figure it out
> 
> it's old death consequences, removing stat points but with a single additional chance if you die, but your stats are fucked even more.

I'm referencing specifically physician revival, how you used to be able to revive a person any time after death, has that returned? Also are you still able to just remove the debuff with new organs like you used to?

notafurryy:
> I'm referencing specifically physician revival, how you used to be able to revive a person any time after death, has that returned? Also are you still able to just remove the debuff with new organs like you used to?

This pr doesn't touch physician stuff, those debuffs are applied on death and can't be removed

Foxentail:
I agree with the debuff but I feel like there should be /some/ way to remove it. Whether its via priest with the book like is used to be, or having to get new organs.  Otherwise I feel like suicide rates are going to increase lol

notafurryy:
> why not just make a bedridden debuff so people can do nurse rp / have to rest and be fed which both creates interactions and keeps revives from instantly rejoining combat instead of a flat stat debuff that makes us not want to be revived in the first place.

That should be the case for heavy injuries and crits, not death.

Gyrandola:
This has been test merged for over a week

Gyrandola (approved):

Foxentail:
Make it so there's some way to recover! There mind as well not be any revive if you're fucked for the entire round without any recourse, your stats get shafted so hard. I got attacked by wolves and survived, then after I was so weak and had so little end that I could literally do nothing. I was worse then a beggar.

PR #2405, opened 2025-02-13
Retinue stat changes - MAA, Vanguard & Watchman

AbsorbentHeart5:
## About The Pull Request

Turns MAA archer into a limited roll, capped at 2 slots. Lowers archery and crossbows to level 3 in skill. Also increases some other skills to level 3 such as Climbing, wrestling and unarmed. Additionally creates two new sub classes:
- Gunner, Increases firearm skill by 1 and spawns with a steel sword and an arquebus.
- Archer, increases archery and crossbow skill by 1 and spawns with a steel sword and a bow.

Turns MAA veteran into a 1 slot roll instead of 2 and allows middle aged characters to play the role. Also reduces strength to 2 instead of 3, bumps endurance to 2 and adds a single (1) point into intelligence.

Edit 3: As requested by a maintainer, I have given all of the guard roles (MAA, Vanguard & Watchmen) a total of 9 stat points, where 1 point in either strength or speed counts as 2. I have additionally tweaked some skills, notably for vanguard. Lowering Vanguard footman bow skill to 2 and Vanguard archer crossbow skill to 4. Below will list the new stats for the guard roles:

Vanguard Footman:
Strength - 2 
Constitution - 1
Endurance - 2
Speed - 1

Vanguard Archer:
Perception - 3
Speed - 2
Endurance - 2

Watchman: 
Strength - 2
Perception - 1
Constitution - 1
Endurance - 1
Speed - 1 

MAA Footman:
Strength - 2
Constitution - 3
Endurance - 2

MAA Halberdier:
Strength - 2
Perception - 1
Constitution - 2
Endurance - 2

MAA Archer:
Strength - 1
Perception - 3
Endurance - 2
Speed - 1

MAA Veteran:
Strength - 2
Perception - 1
Constitution - 1
Endurance - 2
Intelligence - 1

## Why It's Good For The Game

MAA archers are currently extremely weak compared to other ranged based classes and roles to the point where you can have a full roster of MAA and not one has picked the archer class. This PR aims to make archer a support role that isn't entirely useless. Furthermore, the class is limited to 2 slots meaning keep can only have 2 guards with guns at most. This is 2 less than Jagers who have a better stat and skill line up.

MAA veteran is rarely played as you might as well just play the actual Veteran job, which has legendary weapon skills, if you're going to play an old character. Making it so that middle aged characters can roll the job will hopefully see for the job being rolled more often. 

## Proof of Testing (Required)

image

PR #2538, opened 2025-03-08
Less MAA slots

notafurryy:
## About The Pull Request

Removes 2 slots from maa's

## Why It's Good For The Game
Men At Arms got a decent buff in https://github.com/Rotwood-Vale/Ratwood-Keep/pull/2405, making them more powerful than they need to be currently, this lowers their numbers and makes them more of a resource than a throwaway guard.

Also it encourages people to play other guard roles.

image

### DIFF
--- a/code/modules/jobs/job_types/roguetown/garrison/manatarms/manorguard.dm
+++ b/code/modules/jobs/job_types/roguetown/garrison/manatarms/manorguard.dm
@@ -3,8 +3,8 @@
 	flag = MANATARMS
 	department_flag = GARRISON
 	faction = "Station"
-	total_positions = 6
-	spawn_positions = 6
+	total_positions = 4
+	spawn_positions = 4
 
 	allowed_sexes = list(MALE, FEMALE)
 	allowed_races = RACES_TOLERATED_UP
@@ -19,7 +19,7 @@
 	give_bank_account = 22
 	min_pq = 6
 	max_pq = null
-	subclass_cat_rolls = list(CTAG_MAA = 6)
+	subclass_cat_rolls = list(CTAG_MAA = 4)
 
 	cmode_music = 'sound/music/combat_guard2.ogg'
 


Alanii:
I thought the limit was four... What the fuck?

AbsorbentHeart5:
I don't believe this to be a good idea considering there can be 6+ bandit's, all who can have expert weapon skills and 10+ additional stats who will smith in the bog for the majority of the round and then frag every guard they see. Also, it doesn't seem incredibly difficult for bandits to acquire the 1.4-2k+ dragon necklace considering I've been seeing it a lot recently, therefore meaning they have another extra step above MAA. 
Bandits shouldn't even be going up against MAA and the main retinue to begin with. Sure it's annoying when you raid low town and they arrive but again you're bandits not an invasion force. You are a MINOR antag, not a major one.

notafurryy:
> I don't believe this to be a good idea considering there can be 6+ bandit's, all who can have expert weapon skills and 10+ additional stats who will smith in the bog for the majority of the round and then frag every guard they see. Also, it doesn't seem incredibly difficult for bandits to acquire the 1.4-2k+ dragon necklace considering I've been seeing it a lot recently, therefore meaning they have another extra step above MAA. 
> Bandits shouldn't even be going up against MAA and the main retinue to begin with. Sure it's annoying when you raid low town and they arrive but again you're bandits not an invasion force. You are a MINOR antag, not a major one. 

Having bandits smith is pretty rare, as is them getting the necklace, while you consistently have maa slots filled.

The way things are set up you don't really have much else to do as a bandit, you can't be a highwayman if there is no highway, and no people travelling; forcing you to attack instead. And the only choke point has people stationed at it. Also with the pretty huge buff to men at arms the remedy for maa's going out into the bog and hunting antags also got removed.

ALSO bandits are really the only viable threat at the moment, most rounds the lich and vamps are inconsequential at best, either dying early on or not getting enough momentum to even breach the gate, and bandits only rarely manage even that.

Cyprex (approved):

PR #2549, opened 2025-03-10
Reworks platter code

ARefrigerator:
## About The Pull Request
Platters currently only work for certain foods, they have specific sprites, and then return specific platters on consumption. This is disastrous as they only return wooden platters when other types exists meaning your old platter will forever become a wooden one once used.

This PR changes the code to allow any foods to be put on a platter and does the following:
1. Creates a datum of sprites to use when certain types are used on platters. This is an associated list which uses the path type as a key and the platter sprite's name as the value. 
2. Allows any food/snacks subtype to be allowed into a platter's contents.
3. uses this contents item to create a vis_contents for the platter.
4. renames the platter to be 'platter of x'.
5. gives the platter the item's description.
6. Adds the bonus nutriments platter food gets to any placed on platters (probably needs to be changed to a buff or something, currently it's the only incentive to use platters outside of RP).
7. Right clicking will allow you to remove anything on the platter (so long as your hand is free). Also removes bonus nutriment
8. comments out all the previous attackby code for platters
9. adds platter sprites for previously existing sprites which had them (removes platter from sprite)
10. Adds platter to the list of things that prevent food rot (tables prevent it too?? Seems like only chests should but it can be changed later if needed.)

I tested this quite a bit and haven't been able to find anything wrong with it but I'm willing to fix any issues it produces in the future if someone lets me know about them.
_I'll make a commit soon to clean up the comments..._

## Why It's Good For The Game
Platters simply don't work outside of weird manually coded exceptions and even then they break the subtypes of platters. This will allow people to put any single food on platters, get a small benefit from it, and keep the platter they started with. 

https://github.com/user-attachments/assets/eae08100-f7f9-4d1f-95c0-faea5e44edaf


ophaq:
Finally, we can put most foods on platters. Now we just need forks to work.

ARefrigerator:
I'm planning on getting to forks too. I wanted to keep them in separate PRs just in case anything goes wonky.

Gyrandola (approved):

ARefrigerator:
Ah shit did I fuck up this fast foward... Maybe I'll redo the PR.

PR #1914, opened 2024-12-15
Pinning Nerf

notafurryy:
## About The Pull Request

makes pinning actually cost stamina preventing spam pinning with 0 drawbacks, also makes pinning two handed again, also also makes the pinning tome a bit shorter and favor strength more

## Why It's Good For The Game

Pinning as it is sucks, you can get tackled and pinned and chained in 5 seconds regardless of your strength or skill level, this makes it so that getting a single pin on your target doesn't automatically win a fight and forces you to rely on others for help if you pin someone, as it was intended.

PR #2550, opened 2025-03-10
Pinning Stamina Cost

notafurryy:
## About The Pull Request
Makes pinning people drain stamina, you can still keep someone pinned for pretty long if you have above average strength and end.

## Why It's Good For The Game
pinning can remove people from a fight completely, it's a bit weird for it to also be pretty much free and work through any armor.

https://github.com/user-attachments/assets/e02c61eb-74e7-44df-bf66-354fbcb33f6d

### DIFF
--- a/code/modules/mob/living/grabbing.dm
+++ b/code/modules/mob/living/grabbing.dm
@@ -204,7 +204,7 @@
 				if(user.loc != M.loc)
 					to_chat(user, span_warning("I must be on top of them."))
 					return
-				user.stamina_add(rand(20,30))
+				user.stamina_add(rand(10,15))
 				M.visible_message(span_danger("[user] pins [M] to the ground!"), \
 								span_userdanger("[user] pins me to the ground!"), span_hear("I hear a sickening sound of pugilism!"), COMBAT_MESSAGE_RANGE)
 				M.Stun(max(((65 + (skill_diff * 10) + (user.STASTR * 5) - (M.STASTR * 5)) * combat_modifier), 20))


john-pawson:
I'd reccomend another solution if possible, i can already see people simply shooting themselves on the foot by pinning someone that's already on the ground and stamina critting themselves accidentally with such a high cost of stamina. 
Perhaps make it so instead of having to click every now and then to re-apply the pin stun, since you can't know for how long your opponent is stunned, it simply gives a short stun that auto-applies again after the average stun period is done. Kind of how autosmithing works. So it'll consume stamina in multiple little amounts, rather than one single click giving you a massive drain.

Rayforx:
PR before, that covers the same https://github.com/Rotwood-Vale/Ratwood-Keep/pull/1914
I still stand by the same point as last time it came up, nerfing pinning like this will increase lethality, more fragging behavior and less kidnapping being done. You have to loose multiple checks (grab, reinforce, tackle, get over the person, tackle again) while they can kick them away, counter grab, resist spam, before they start a pin chain against the other person.

Crowbar764:
If people cannot non-lethal contain someone, then they will be forced to lethally do it instead. 👎

Hmeister-real:
> If people cannot non-lethaly contain someone, then they will be forced to lethally do it instead. 👎

You can still really easily pin with this, you just can't spam it more than 3-4 times without taking a breather, more than enough to chain someone and you can always just escelate and break legs if needed or have someone else take over since most pinning is done by the retinue anyways.

Rayforx:
And this is why we do not do this, every action takes stamina. And the video only showed when the person was already tackled. This is a balance nightmare, and the current system is working fine.
For this scenario both people had combat mode on, you would have to use even more stamina if the other person was resisting.

https://github.com/user-attachments/assets/34da4913-0737-4f28-9564-89c6bab69e4d

Hmeister-real:
> And this is why we do not do this, every action takes stamina. And the video only showed when the person was already tackled. This is a balance nightmare, and the current system is working fine. For this scenario both people had combat mode on, you would have to use even more stamina if the other person was resisting.
> 
> The watchman are the aggressor, on the other side is a brigand.
> 
>  2025-03-11.12-37-23.mp4

This is literally the worst example you could show. Bandit Brigands and Knights are the only ones who would have this specific issue, Knaves, Mages and sawbones would not have this happen since you could instantly tackle them as they have lower strength and wrestling.

Not being able to tackle someone who is a combat role made to be equal to you should be an intended thing and grabbing should not be encouraged over regular combat. The fact a watchmen can't tackle a bandit brigand is good, because that means they have to consciously decide what to do instead of mindlessly grab, tackle, pinning and be strategic on what they do. But tackling someone like a rogue, or a knave or anyone whose main counter is grabbing /has/ low wrestling and strength as it's traded for speed, meaning that this PR wouldn't really effect them unless the guard tackling didn't conserve stamina well enough

Alanii (changes requested):
Reduce 20,30 to 10,15

I do not want changes like this to be done this drasticaly.

Alanii (approved):
Testable

Gyrandola:
Testing

ToothyDeer:
i like this change. pinning someone for free for 5 minutes is retarded. this needed 2 happen, grapple combat needs 2 be less relevant. a majority of the time a pin leads to death via skull-crack or beheading. i also don't know if its still around but i know u used to be able to pin w/ 1 hand and then in ur off-hand use a weapon. that was retarded, too.

Jackofashepard:
Makes pinning useless to anyone but a knight. People just break legs and heads now

notafurryy:
> used to be able to pin w/ 1 hand and then in ur off-hand use a weapon. that was retarded, too.

You can still do that

notafurryy:
> Makes pinning useless to anyone but a knight. People just break legs and heads now

People always did that.

Jackofashepard:
Sure, just dont complain when you get murdered because no one can pin you

Gyrandola (approved):

PR #2551, opened 2025-03-11
Mount nerfs and changes

notafurryy:
## About The Pull Request

Removes the ability to use grab fu from mounts, As a stand in for the inability to reach people on the ground you are unable to shove them or move them.

Removes exp gain from buckling and unbuckling from a saiga.

Vanguard riding skill reduced by 2.

Man at Arms riding skill increased to 2 and 3 for the sword wielding archetype.

## Why It's Good For The Game
Saigas are way too fast currently and they let you move even with the immobilized debuff making running out of stamina on them inconsequential while being pretty fast and ignoring terrain difficulty. This changes those things. Also being able to wrestle people while not even being able to reach them is odd.

https://github.com/user-attachments/assets/d473667d-7c74-4bb8-a7c4-876e4ab1b5a2

### DIFF
--- a/code/modules/jobs/job_types/roguetown/garrison/bogguard.dm
+++ b/code/modules/jobs/job_types/roguetown/garrison/bogguard.dm
@@ -118,7 +118,7 @@
 	bogger.mind.adjust_skillrank(/datum/skill/misc/sneaking, 2, TRUE)
 	bogger.mind.adjust_skillrank(/datum/skill/misc/swimming, 2, TRUE)
 	bogger.mind.adjust_skillrank(/datum/skill/misc/treatment, 1, TRUE)
-	bogger.mind.adjust_skillrank(/datum/skill/misc/riding, 2, TRUE)
+	bogger.mind.adjust_skillrank(/datum/skill/misc/riding, 3, TRUE)
 	bogger.mind.adjust_skillrank(/datum/skill/misc/reading, 1, TRUE)
 	bogger.change_stat("strength", 2)
 	bogger.change_stat("constitution", 1)
--- a/code/modules/jobs/job_types/roguetown/mercenaries/types/desertrider.dm
+++ b/code/modules/jobs/job_types/roguetown/mercenaries/types/desertrider.dm
@@ -57,7 +57,7 @@
 		H.mind.adjust_skillrank(/datum/skill/combat/knives, 3, TRUE)
 		H.mind.adjust_skillrank(/datum/skill/misc/reading, 1, TRUE)
 		H.mind.adjust_skillrank(/datum/skill/misc/athletics, 3, TRUE)
-		H.mind.adjust_skillrank(/datum/skill/misc/riding, 2, TRUE)
+		H.mind.adjust_skillrank(/datum/skill/misc/riding, 4, TRUE)
 		H.change_stat("strength", 2)
 		H.change_stat("endurance", 2)
 		H.change_stat("constitution", 1)
--- a/code/modules/mob/living/grabbing.dm
+++ b/code/modules/mob/living/grabbing.dm
@@ -210,7 +210,7 @@
 				M.Stun(max(((65 + (skill_diff * 10) + (user.STASTR * 5) - (M.STASTR * 5)) * combat_modifier), 20))
 				user.Immobilize(20 - skill_diff)
 			if(usr.buckled)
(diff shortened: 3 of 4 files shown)


Alanii (changes requested):
Remove the saiga speed nerf for now.

I would want to test the changes without it.

notafurryy:
> Remove the saiga speed nerf for now.
> 
> I would want to test the changes without it.

Done, the speed is back to what it was.

Pikadoko:
boooo, you already can't dodge or parry on a saiga (and it doesn't take much to make the saiga absolutely useless by wounding it, which can only be fixed with miracles). This is removing the only thing that makes Vanguard playable without sitting in the smithy queue for 5 hours.

Limeon666:
I'm fine with this cause I think grabbing people from a saiga is pretty silly. It would be nice if there was someway to put someone on the back of a saiga, though, as a substitute for dragging them along. Even just being able to attach a cart to the back of a saiga with a rope would massively make up for this change.

But for another thing Vanguard should have good riding skill over MAA imo, since they are actually in the bog, and would have more need for the extra agility. -2 seems like a big nerf for them.

notafurryy:
> But for another thing Vanguard should have good riding skill over MAA imo, since they are actually in the bog, and would have more need for the extra agility.

polearms and walking sticks exists, also they do have more speed and endurance than city watch and MAA's in there

notafurryy:
> boooo, you already can't dodge or parry on a saiga

That doesnt matter if you are a saiga archer, which most vanguards are, and you can't fight back against that since not even running out of stamina stops them.

Actuallygoinginsanern:
while you're at it, fix the bug that lets you train riding without needing to sleep.

You can currently train riding by just jumping up and down on a saiga to master in like 15 minutes.

Crowbar764:
Can you make it so tackling someone on a saiga would dismount them, otherwise the only way to get them off is to kill the saiga, or to basically kill them.

Alanii (approved):
Cool.

notafurryy:
> Can you make it so tackling someone on a saiga would dismount them, otherwise the only way to get them off is to kill the saiga, or to basically kill them.

I can't exactly get that to work right now, sadly.

AbsorbentHeart5:
I think you should totally try and add ways to parry on saiga back. No dodging though because that wouldn't make any sense.

Pikadoko:
Saigas are the only way to use bows, they drain all your stamina in about four shots. They're also useless against undead with masks. It's insane to me that you're trying to further gimp Vanguard when the average refugee shoots 63 fireballs in a series and mace users rng win battles in 5 hits due to knockouts. Looks like a salt PR to me.

notafurryy:
> Saigas are the only way to use bows, they drain all your stamina in about four shots. They're also useless against undead with masks. It's insane to me that you're trying to further gimp Vanguard when the average refugee shoots 63 fireballs in a series and mace users rng win battles in 5 hits due to knockouts. Looks like a salt PR to me.

Saigas make them unstoppable, there is no counter to them since you can't even get close enough to them. Being able to remove people in 1 hit from the battle is powerful enough you just have to be smart about it, use vantage points and the arc intent to not hit your colleagues, instead of flanking the enemy in no time and then running to the other side of the map with no risk or penalty, and they still can use saigas.

RadiantFlash7 (changes requested):
Glancing this over. I have mild concern that 2 riding skill might be too low. It's meant to reflect apprentice level. If anything they should probably be 3. Though if it proves to still be too much dropping it to 2 may suffice.

From a balance standpoint, dragging people while on saigaback is indeed something that was in need of a nerf.
Since you're adjusting riding skill on roles, I'd ask that you increase Desert riders riding skill to 4. Riding is their whole gimmick lorewise, and they've been sorely lacking.

As from a coding standpoint- Make it have better feedback.
 If you're grabbing someone and can't move because of it, have it say that. Don't have it say 'I can't be buckled'. Something like, 'I can't  pull/haul while seated' or some such.

Gyrandola:
I'll TM this while waiting for a final review from Radiant.

atomicplan666:
makes sense not not allow wrestling while riding

RadiantFlash7 (approved):
Haven't seen any issues within our test merge so far,
Code seems fine at the moment.
Ready to merge.

PR #2552, opened 2025-03-11
Werewolf Upgrade

MickDick1:
## About The Pull Request

Adds the Bog Trekking trait to werewolves. This allows them to move through muddy tiles without slowdown.
Increases werewolf hide from 250 integrity to 400. 
Adds smash resist bclass to the werewolf hide.

## Why It's Good For The Game

I believe the bog trekking trait speaks for itself.
Increasing hide integrity was requested. 250 is nothing for the entire body. And as far as I could see, there is no rejuvenating it without waiting until the next transform.

Smash resist is added as the hide already has chop resist, which is actually what's weird here, and is the axe equivalent of smash. Beyond that, for specific reasons smash resist was added as it's terribly easy to just walk up to a werewolf, then aim leg or groin with a lucerne or other smash weapons and immediately win.

Werewolves are not immune to pain, mind. They can and will pain stun and slowdown from being injured still. The damage you do that doesn't crit DOES MATTER.

Everyone pretty much agreed werewolves are painfully weak. This a bump. We'll see if it makes the difference.

Tested bog trekking on local. Works. Doesn't show up in trait list.

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/roguetown/other/werewolf.dm
+++ b/code/modules/mob/living/carbon/human/species_types/roguetown/other/werewolf.dm
@@ -29,7 +29,8 @@
 		TRAIT_CRITICAL_RESISTANCE,
 		TRAIT_HARDDISMEMBER, //Decapping Volfs causes them to bug out, badly, and need admin intervention to fix. Bandaid fix.
 		TRAIT_PIERCEIMMUNE, //Prevents weapon dusting and caltrop effects due to them transforming when killed/stepping on shards.
-		TRAIT_IGNORESLOWDOWN
+		TRAIT_IGNORESLOWDOWN,
+		TRAIT_BOG_TREKKING
 	)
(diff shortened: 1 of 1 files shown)


DelineFortune:
https://github.com/Rotwood-Vale/Ratwood-Keep/pull/2328

DelineFortune:
And this one https://github.com/Rotwood-Vale/Ratwood-Keep/pull/2257

Hmeister-real:
Could always take some of the stuff from this list and implement it also
https://discord.com/channels/1240735983276654733/1342468694260977745/1342468694260977745

notafurryy:
Would be cool if werewolves were a bit stronger too, since they don't have access to anything that pierces armor.

Alanii:
I have no issue if this is Test Merged.

RadiantFlash7:
> I have no issue if this is Test Merged.

I think we should see about test merging #2328 with some additions from this PR, personally.

PR #2559, opened 2025-03-12
Goblin Prisoner and Unbanditifying Prisoner Roles

MickDick1:
## About The Pull Request

Allows goblins to be bog and rockhill prisoners.
Removes the bandit camp trait from bog and rockhill prisoners.
Ensures goblins spawn with graggar as patron.

## Why It's Good For The Game

No real reason to not let goblins roll prisoner.
Giving prisoners the bandit camp trait is just an opening for gamey exploitation which has occurred already in the past. It also forces the characters to background-wise be bandits after all.

image
image


Alanii:
I will be honest...

This will not go well at all.

MickDick1:
> I will be honest...
> 
> This will not go well at all.

Would you like to elaborate?
The way I see it, this role will be a net positive.
The free goblins in the camp no longer have to put up with the obvious bait ones primarily.

PR #2543, opened 2025-03-09
Preists can perform sermons now!

DelineFortune:
Preist seems consistently weaker than your run of the mill brigand, I consistently spawn with 13-14 str while brigand seems to sway between 13-15. I suggest a buff.. It's hard to wrestle brigands even when they're naked in the current state.

spawning with 12-13 strength is equal to being a fucking vanguard. please stop.
Look at bandit stats, my roleplay in the dungeon has to stop at considering unchaining them to let them do anything at all, beacuse you people think the preist is a combat role or something and giving him strength to wrestle prisoners is going to ruin the game.

But it's alright. You people know an absolute nothing about balance, considering you still call polearms the meta weapon, then spam ahelps when 2 bandits rape you with maces, half the drugs continue to be insanely fucking strong without a nerf, and many others i'm not even going to start mentioning because it will all boil down to the same cancer dogshit from you about how wrong I am, when it's YOU you daft motherfuckers that have the IQ of a fucking time period accurate peasant.

You lpay here, probably too much, like me, yet you can't look around and voice your opinion on roles you probably don't even fucking touch, you ignorant fucking lemmings get out of my game. Give me a break ratwood community, buy a fucking brain I am tired of listening to your nimrod fucking knight main takes. Dumbest fking playerbase so far and this time it's not even the furries this time that make me want to donate to fucking birth control research.

Also, it is in fact a small buff, you will not overexaggerate how much a point of stat does when it's solely to bring him up to par to PEOPLE HE IS SUPPOSED TO EASILY CONTROL IN THE DUNGEONS. Stop being a fucking idiot.

ok, go serious 
Preist get new button
Be in the church
Press it. Wait 120 second
give -5 stress and +1 fortune +1 con to everyone (people on your screen) for 20 minutes 

image

image

image

DONT LISTEN SERMON AS INHUMEN OR YOU WILL GET DEBUFF!!!

PR #2560, opened 2025-03-12
Preist's rework (sermons+excom and apostasy)

DelineFortune:
Preists can perform sermons now! https://github.com/Rotwood-Vale/Ratwood-Keep/pull/2543

Heretic and excommunication rework
One day someone will rework inquisitor to add them heretic mark blah blah

image
Apostasy = old excommunication mark but changed
Excommunicate = old heretic mark but changed

Apostasy = works as old ex but worse for clergy
image

Excommunicate = debuff for followers of ten/psydonites and buff for followers of four
Also it block you from getting healing from lesser miracle/miracle/cure rot/anastasia
image
image
image
image
image
image
image

If they act against church (follow their objective) they are supposed to get something instead of being valid 
image
image

Can you combine debuffs against clergy? Sure
image

If follower of ten excommunicated: 
cannot get get help from miracles, 
+5 stress -2 int -2 luck -2 per

If inhumen excommunicated:
cannot get get help from miracles
+1 speed -5 stress +1 int

If cleregy is getting Apostasy mark
+5 stress 
Cannot use or get help from miracles 
-2 int
-5 luck
-2 per 

SERMON EFFECTS:
image

image

### DIFF
--- a/code/modules/jobs/job_types/roguetown/church/priest.dm
+++ b/code/modules/jobs/job_types/roguetown/church/priest.dm
@@ -266,12 +266,12 @@ GLOBAL_LIST_EMPTY(heretical_players)
     if(!istype(get_area(src), /area/rogue/indoors/town/church/chapel))
         to_chat(src, span_warning("I need to do this in the chapel."))
         return FALSE
-    visible_message(src, span_notice("[src] begins preaching a sermon..."))
+    src.visible_message(span_notice("[src] begins preaching a sermon..."))
     if(!do_after(src, 1200, target = src)) // 120 seconds
-        visible_message(src, span_warning("[src] stops preaching."))
+        src.visible_message(span_warning("[src] stops preaching."))
         return
 
-    visible_message(src, span_notice("[src] finishes the sermon, inspiring those nearby!"))
+    src.visible_message(span_notice("[src] finishes the sermon, inspiring those nearby!"))
     for(var/mob/living/carbon/human/H in view(7, src)) // all mobs within 7 tiles
         if(HAS_TRAIT(H, TRAIT_ORGAN_EATER) || HAS_TRAIT(H, TRAIT_COMMIE) || HAS_TRAIT(H, TRAIT_MATTHIOS_EYES) || HAS_TRAIT(H, TRAIT_CRACKHEAD) || HAS_TRAIT(H, TRAIT_ZIZO_MARKED))
             H.add_stress(/datum/stressevent/heretic_on_sermon)
--- a/code/datums/stress/negative_events.dm
+++ b/code/datums/stress/negative_events.dm
@@ -363,7 +363,7 @@
 	stressadd = 3
 	desc = span_boldred("The apostasy's mark is upon me!")
 
-	/datum/stressevent/heretic_on_sermon
+/datum/stressevent/heretic_on_sermon
 
 	stressadd = 5
 	desc = span_red("My PATRON is NOT PROUD of ME!")


DelineFortune:
image

Alanii:
Change apostasy mark to apostate instead when examined upon on players.

DelineFortune:
> Change apostasy mark to apostate instead when examined upon on players.

image

ophaq:
Cool PR. The debuffs are pretty big but it makes sense to have them.

On another note, could you please activate Windows?

Limeon666:
I personally do not like the priest being able to physically reduce your stats of all things if they know your name.

I also don't really like how these are two separate things when in most instances the priest will just apply them both to someone.

Also apostacy mark is very weird on a cleric. They mechanically are not an apostate because they had to have been devout to their God to get that power. A heretical clergy is very possible, but an apostate clergy? Not really. The word picked just doesn't make sense since apostates are explicitly somebody switching gods.

Jackofashepard:
Psydonites should be immune to Excommunicate but also get no benefits from the other effects since they are not a part of the church of the ten

radarseas:
I think only the inquisition should be affected by excommunication, since its their objective to attempt to guide a flock that's misled in their eyes and root out sinners, being excommunicated, that and to **highly encourage them to work with the church.**

Other Psydonites shouldn't give a crap either way for all they need is **PSYDON**, not any of these false gods that defile his will as mere aspects of it, they serve one shepard, **accept there is only one god.**

DelineFortune:
Its preist pr...
Psydonite part is for someone that is going to rework inquisitor and psydonite faith as well
I gave reworked code for them, they may change it as well

code/modules/jobs/job_types/roguetown/church/priest.dm
```diff
@@ -240,3 +257,25 @@ GLOBAL_LIST_EMPTY(heretical_players)
...
+        return
+    if(!istype(get_area(src), /area/rogue/indoors/town/church/chapel))
+        to_chat(src, span_warning("I need to do this in the chapel."))
+        return FALSE
+    visible_message("<span class='notice'>[src] begins preaching a sermon...</span>")
+    if(!do_after(src, 1200, target = src)) // 120 seconds
+        visible_message("<span class='warning'>[src] stops preaching.</span>")
+        return
+
+    visible_message("<span class='notice'>[src] finishes the sermon, inspiring those nearby!</span>")
```
Gyrandola, line 274: Ugh please spannify and I'll merge.

Gyrandola (changes requested):

Teragenic:
Hey, uh I should point out that THIS may be a problem later on. Inhumen do get access to their own miracles. Even if only minor healing, would this PR make that an issue?

You should not get PER or Int damage. You don't get Dumber. Maximum stress should already do a lot and -2 LCK is with that. I think I'd prefer if it was END, (Slightly easier to get tired).

Otherwise I really make sense?

DelineFortune:
> Hey, uh I should point out that THIS may be a problem later on. Inhumen do get access to their own miracles. Even if only minor healing, would this PR make that an issue?
> 
> You should not get PER or Int damage. You don't get Dumber. Maximum stress should already do a lot and -2 LCK is with that. I think I'd prefer if it was END, (Slightly easier to get tired).
> 
> Otherwise I really make sense?

I guess its just more simple to split inhumen miracles into their own category to make it work under the mark

Why int per and luck
Not a fun of touching combat stats a lot, do your best, try to fight against the preist
It breaks some exploit meta (no im not talking about swampweed+spices)
Its more non combat related. You are not supposed to live your life in peace being excommunicated.

Gyrandola (approved):

Jackofashepard:
I still think this makes no sense for psydonites

PR #2565, opened 2025-03-13
Allows re-entering body when decapitated, I think.

ARefrigerator:
## About The Pull Request
This might fix people's ability to enter their body when decapped. 
I don't know what exactly this block of code is doing but removing it allows you to re-enter your body after decapping. I assume somewhere under these piles of procs is something nulling an important variable (like a ckey) breaking everything.

With it commented out it works?? I haven't been able to test every aspect of this so someone else might need to take this up if they feel like investigating deeper.

**NOTE:** decapped people don't get a ghost sprite. I might try to look into this but so far it seems purely aesthetic as they adhere to all spirit logic (walls, doors, etc) so I am less concerned.

Things tested:

- Re-entering body after decap (works)
- leaving to underworld (works)

Things to be tested:

- Someone being able to re-enter a body once the head is attached by another person in game
- Someone being able to re-enter said body if brought back to life.

**I would strongly suggest Test Merging this for a while because while I don't think it should break anything but I wasn't able to test enough to be sure.**

## Why It's Good For The Game
It seems like originally one wanted decapitation to be a permanent solution. As servers moved away from this concept this block of code has remained a menace halting anything about that. Hopefully it means admins no longer have to force people back into bodies after decap revivals but we shall see...

https://github.com/user-attachments/assets/7960b438-ad82-4e9b-af75-e289cd47f5b1

### DIFF
--- a/code/modules/surgery/bodyparts/bodypart_dismemberment.dm
+++ b/code/modules/surgery/bodyparts/bodypart_dismemberment.dm
@@ -295,12 +295,21 @@
 
 	name = "[owner.real_name]'s head"
 	. = ..()
+	/* Evil cursed code, contain it in comments! 
+	NOTE:	I'm not entirely sure why this is breaking re-entering
+			bodies as a ghost but it seems to be the culprit. It shouldn't
+			affect anything else though? 
+
+			You do not get a ghost sprite if you are decapped, may need to fix later,
+			but that's nothing compared to decap nulling your body!!
+	
 	if(brainmob)
 		QDEL_NULL(brainmob)
 	var/obj/item/organ/brain/BR = locate(/obj/item/organ/brain) in contents
 	if(BR)
 		if(BR.brainmob)
 			QDEL_NULL(BR.brainmob)
+	*/
 
 //Attach a limb to a human and drop any existing limb of that type.
 /obj/item/bodypart/proc/replace_limb(mob/living/carbon/C, special)


ARefrigerator:
Note: this does not solve the skill issue. I genuinely believe that to be the `/datum/mind/proc/tansfer_to` proc that causes it. The function appears to have nothing in it which transfers the skills or abilities the players have when moving the mind around and I'm pretty sure that is the function it calls when you decap or have to move your mind around. If you want I can try getting around to that eventually because unless someone else knows what the issue is it definitely looks like that proc to me. 

Otherwise I'll close this if someone else moves the mob info back to the body.

RadiantFlash7:
I actually did this exact PR days back, but got told on the discord to move the ckey back to the chest instead. 
Even so, I’m happy this is getting approved now.

ARefrigerator:
I feel pretty confident in what's causing the skill issue. I can try adding it here or in another PR later on. Maye here since it is related and it's a big deal...

RadiantFlash7:
> I feel pretty confident in what's causing the skill issue. I can try adding it here or in another PR later on. Maye here since it is related and it's a big deal...

Right, I've been essentially code diving into this issue for the past few hours.
Transfer_to isn't the problem as far as I can tell. Datum/mind holds the known_skills list perfectly fine when transfered,
Because all it's doing is attaching itself to a new mob's mind var. 

From what I can tell, our issue currently lies somewhere within zombie code, in that on death, skills are moved from the mind to the antag datum for zombies- and then at somepoint during decapitation, it's not getting transfered back (Normally on revival, the zombie antag is deleted, and skills restored.) I'll keep taking a look into it and see if I can track down the exact problematic code myself and throw up a fix, as part of my lich phylactory fix.

RadiantFlash7:
I've fixed the root issue with decapitation this is trying to fix in #2574 @ARefrigerator 
The issue with skills should be resolved in #2571  by removing skill removal during zombification.

Gyrandola:
Closing this as I believe Radiant fixed it.

PR #2568, opened 2025-03-14
Fix for Millstone breaking

GravyMaker:
## About The Pull Request

This PR fixes an issue with millstones, where placing sugar (and potentially other items) into the millstone would cause it to be unusable - it could be interacted with, but would stop successfully processing items for the rest of the round.

This issue was caused by a combination of two apparent errors:
- Firstly, when considering whether an item is acceptable to place in the millstone, the original code checked the object's "grind_result" value, which appears to have been copied from similar code for the grindstone. However, this value was otherwise unused by the millstone. This line has been modified to instead check for a valid mill_result, which is the value which actually determines the object created by the millstone.
- Secondly, when processing the item in grindUp, the code did not check the validity of mill_result before attempting to instantiate from it. As mill_result's default value is null, this could lead, in cases where an object's grind_result was defined but its mill_result was not, to an attempted instantiation of a null type, causing the subsequent code (which removed the current item from the millstone's internal queue) not to execute. Though it should no longer be necessary due to the change described in the previous bulletpoint, a further failsafe has been added in the form of a null check against mill_result; if the variable is non-null, the code will function as before. If it is null, however, the millstone will instead produce a copy of the original item.

## Why It's Good For The Game

The issue doesn't affect any functionality other than fixing a bug. The bug in question was unlikely to be triggered by an experienced cook - as far as I know, it would only occur when putting sugar in the millstone. However, once triggered - whether accidentally or deliberately - the millstone would be rendered unusable for the remainder of the round, which could have a significant impact on roles which interact heavily with cooking mechanics such as Maid. By fixing the issue this is no longer a possibility, with no other consequences as far as I'm aware.

Video was too large to submit on Github, so I've uploaded it to Youtube instead.
https://www.youtube.com/watch?v=TjWivNw38mU

The video shows me placing all valid items - that I know of - into the millstone and processing them. I then attempt to place sugar in the millstone, showing that it is no longer accepted. I then try to place various clearly invalid items in the millstone (such as a rolling pin) to confirm that these still aren't accepted.

## Addendum
Though I'm an experienced programmer, I'm entirely new to BYOND development - this is my first ever contribution to Ratwood, SS13, or any similar project (as you can see, I'm taking the classic "learn by fixing minor bugs" approach). As such, please (gently) let me know any issues, even if they wouldn't ordinarily be enough to prevent the PR from being accepted (eg. if there is a more idiomatic way of doing what I've done).

### DIFF
--- a/code/modules/roguetown/roguejobs/cook/tools/milling.dm
+++ b/code/modules/roguetown/roguejobs/cook/tools/milling.dm
@@ -25,11 +25,12 @@
 /obj/structure/fluff/millstone/proc/grindUp(list/obj/item/to_grind, mob/user)
 	for(var/obj/item/itemtogrind in to_grind)
 		if(do_after(user, 8, target = src))
+			to_grind -= itemtogrind
 			if (!isnull(itemtogrind.mill_result)) // prevent the millstone breaking if no mill_result is defined
 				new itemtogrind.mill_result(src.loc)
+				qdel(itemtogrind)
 			else
 				itemtogrind.loc = src.loc
-			to_grind -= itemtogrind
 			apply_farming_fatigue(user, 5) //replace with herbalism/alchemy version if/when it's added
 			playsound(src, 'sound/items/wood_sharpen.ogg', 100, TRUE)
 		else


code/modules/roguetown/roguejobs/cook/tools/milling.dm
```diff
@@ -25,7 +25,10 @@
 /obj/structure/fluff/millstone/proc/grindUp(list/obj/item/to_grind, mob/user)
 	for(var/obj/item/itemtogrind in to_grind)
 		if(do_after(user, 8, target = src))
-			new itemtogrind.mill_result(src.loc)
+			if (!isnull(itemtogrind.mill_result)) // prevent the millstone breaking if no mill_result is defined
+				new itemtogrind.mill_result(src.loc)
+			else
+				new itemtogrind.type(src.loc)
 			to_grind -= itemtogrind
```
EvenInDeathIStillServe, line 32: I don't see anywhere in this file that actually _deletes_ the item that's been ground. This is also a problem. Please add a qdel here for it.

code/modules/roguetown/roguejobs/cook/tools/milling.dm
```diff
@@ -25,7 +25,10 @@
 /obj/structure/fluff/millstone/proc/grindUp(list/obj/item/to_grind, mob/user)
 	for(var/obj/item/itemtogrind in to_grind)
 		if(do_after(user, 8, target = src))
-			new itemtogrind.mill_result(src.loc)
+			if (!isnull(itemtogrind.mill_result)) // prevent the millstone breaking if no mill_result is defined
+				new itemtogrind.mill_result(src.loc)
+			else
+				new itemtogrind.type(src.loc)
```
EvenInDeathIStillServe, line 31: The itemtogrind already exists. Just move it back to src.loc

EvenInDeathIStillServe (approved):

EvenInDeathIStillServe:
Thank you for your first PR. Please amend what I've brought up.

Looking forward to further contributions.

GravyMaker:
Thanks for the feedback - the recent commits should hopefully resolve the issues. They don't result in any visible change in the game, so I've just captured the output from Dream Maker showing that it compiles.

image

EvenInDeathIStillServe (approved):

PR #2570, opened 2025-03-15
Refugee wave removal

Rayforx:
## About The Pull Request
A unlimited amount of refugee can be spawned which means that this wave was never used outside of events. The code was changed so that a refugee wave could be spawned in by a admin, while removing a wave that nobody really used.

## Why It's Good For The Game
Refugee waves are not used by anyone, and people have their time wasted while refugee wave that nobody plays goes by them.
If someone really wanted a refugee wave, then it's usually because of a event.

billede

### DIFF
--- a/code/datums/migrants/migrant_wave.dm
+++ b/code/datums/migrants/migrant_wave.dm
@@ -38,6 +38,7 @@
 /datum/migrant_wave/refugee
 	name = "Adventure Party"
 	downgrade_wave = /datum/migrant_wave/refugee_down_one
+	can_roll = FALSE
 	roles = list(
 		/datum/migrant_role/refugee = 4,
 	)


Limeon666:
Refugee waves are still useful if you want to spawn in together as refugees. The problem here is not that refugee wave exists, but that there's no way to vote on migrant waves. 

Honestly I'm not against removing this regardless, but the migrant wave DESPERATELY needs an overhaul. Without an admin getting a lot of the waves is borderline impossible.

Rayforx:
> Refugee waves are still useful if you want to spawn in together as refugees. The problem here is not that refugee wave exists, but that there's no way to vote on migrant waves.
> 
> Honestly I'm not against removing this regardless, but the migrant wave system DESPERATELY needs an overhaul. Without an admin getting a specific wave can be borderline impossible.

Yeah, that is also why I made it so that the admins can still spawn in the wave. 
Admins can also still /choose/ to spawn in specific waves, if you ahelp for it and they agree.

EvenInDeathIStillServe (approved):

PR #2571, opened 2025-03-16
Lich Fixes, Spell fixes, Delimb through armor prevention, skill loss fix

RadiantFlash7:
## About The Pull Request
This PR refractors a bit of lich code, ported from Azure peak, so the lich will ressurect with their starting equipment at their phylactory.
It also fixes their phylactories to work again.
Also grants the lich blood magic skills, given some of their spells are actually blood magic, not arcyne.
PR removes skill-loss upon zombie antag gain for all mobs. 
Adds a check during Delimb to check for crit protecting armor.
Adds missing anti-magic check to blade burst.
Ensures Lightning lure and lightning bolt now respect shockimmune traits.

## Why It's Good For The Game
Lich is broken, This should fix it for the most part,
We don't intend on ever having player Deadites again, so there is no point in removing player skills upon death, and re-adding them later. Especially as this breaks on mind transfering without revival, and in some cases revivals after decapitation.
I've been told repeatedly that people can get declimbed through armor. with enough luck and damage roll. Shouldn't happen any more.
Fixes missing anti-magic check for blade burst, and lightning spells now proper respect 'shockimmune' trait.

image
image

### DIFF
--- a/code/modules/spells/roguetown/wizard.dm
+++ b/code/modules/spells/roguetown/wizard.dm
@@ -1028,8 +1028,8 @@ Unless of course, they went heavy into the gameplay loop, and got a better book.
 	playsound(T,'sound/magic/charged.ogg', 80, TRUE)
 	for(var/mob/living/L in T.contents)
 		if(L.anti_magic_check())
-			visible_message(span_warning("The blades dispel when they near [target]!"))
-			playsound(get_turf(target), 'sound/magic/magic_nulled.ogg', 100)
+			visible_message(span_warning("The blades dispel when they near [L]!"))
+			playsound(get_turf(L), 'sound/magic/magic_nulled.ogg', 100)
 			qdel(src)
 			return BULLET_ACT_BLOCK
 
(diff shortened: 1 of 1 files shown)


EvenInDeathIStillServe:
Atomise your PRs, please. Don't put bugfixes and balance changes in the same PR.

RadiantFlash7:
> Atomise your PRs, please. Don't put bugfixes and balance changes in the same PR.

The only thing in this PR that I consider properly a balance change is giving the lich blood magic skill.
Which I /can/ remove and atomize if you'd like

Lichs respawning properly at their phylactory with their gear was originally intended on the initial rendition of lich, where they would keep the things they were actively wearing. At some point that stopped working, or was torn out. I yanked some of the refractored code from Azure Peak to fix that.
If you'd like I can atomize that as well, but that seems like just fixing the initial mechanic for liches?

EvenInDeathIStillServe (approved):

PR #2574, opened 2025-03-16
Decapitation fixes

RadiantFlash7:
## About The Pull Request
Removes mind transfer to brain

## Why It's Good For The Game
This code was commented out by azarak 9 months ago, and re-enabled by gyran during his cleanup PRs. This is what causes the mind to transfer to the brain rather then remain in the body. Removing should fix majority of decapitation issues.

This fixes the root problem that  #2565 is trying to address.

image
image


ARefrigerator:
Ah that would explain things. It transfers it to the brainmob and then also deletes it in the decap code. I couldn't figure out where that was happening.

Gyrandola (approved):

SmiLeYre:
WTH.

PR #2580, opened 2025-03-18
Reworks platter code

ARefrigerator:
## About The Pull Request
(Updated Re-submission of https://github.com/Rotwood-Vale/Ratwood-Keep/pull/2549)
Platters currently only work for certain foods, they have specific sprites, and then return specific platters on consumption. This is disastrous as they only return wooden platters when other types exists meaning your old platter will forever become a wooden one once used.

This PR changes the code to allow any foods to be put on a platter and does the following:
- Creates a datum of sprites to use when certain types are used on platters. This is an associated list which uses the path type as a key and the platter sprite's name as the value.
- Allows any food/snacks subtype to be allowed into a platter's contents.
- uses this contents item to create a vis_contents for the platter.
- renames the platter to be 'platter of x'.
- gives the platter the item's description.
- Adds the bonus nutriments platter food gets to any placed on platters (probably needs to be changed to a buff or something, currently it's the only incentive to use platters outside of RP).
- Right clicking will allow you to remove anything on the platter (so long as your hand is free). Also removes bonus nutriment
- comments out all the previous attackby code for platters
- adds platter sprites for previously existing sprites which had them (removes platter from sprite)
- Adds platter to the list of things that prevent food rot (tables prevent it too?? Seems like only chests should but it can be changed later if needed.)

I tested this quite a bit and haven't been able to find anything wrong with it but I'm willing to fix any issues it produces in the future if someone lets me know about them.

## Why It's Good For The Game
Platters simply don't work outside of weird manually coded exceptions and even then they break the subtypes of platters. This will allow people to put any single food on platters, get a small benefit from it, and keep the platter they started with.

https://github.com/user-attachments/assets/3b3b1f98-a7de-4c4e-891e-5f75ceeab0c7

### DIFF
--- a/modular/Neu_Food/code/NeuFood.dm
+++ b/modular/Neu_Food/code/NeuFood.dm
@@ -599,7 +599,8 @@ What it does:
 		name = "platter of [contents[1].name]"
 		desc = contents[1].desc
 		//Need something better than this in future like a buff
-		contents[1].bonus_reagents = list(/datum/reagent/consumable/nutriment = 2)
+		if(istype(contents[1],  /obj/item/reagent_containers/food/snacks/))
+			contents[1].bonus_reagents = list(/datum/reagent/consumable/nutriment = 2)
 	else
 		vis_contents = 0
 		name = initial(name)


ARefrigerator:
The tray is what deals with moving large amounts of wood while the platters are individual food specific. I do intend to make an update to fix trays though because currently due to their storage nature they spill bowls and glasses which I would like to change, I also don't like that they stack items, among other things.

I can also make an update for reducing how much time it takes based on cooking skills as I would also like to make cooking skills more useful. Perhaps in a future PR.

ARefrigerator:
When this gets merged I'll make forks functional and work off eating from platters and food sources. I did it for Azure recently so it seems a trivial addition but I'd rather it be its own PR for atomizing sake.

Gyrandola (approved):

PR #2581, opened 2025-03-18
Werewolf armor fix

GravyMaker:
## About The Pull Request
Corrects an oversight which led to werewolves' innate armour not preventing critical hits.

The checkcritarmor proc in human_defense.dm defines a list of slots to check when considering whether a critical hit can be applied to a given body part. However, this list didn't include the "skin_armor" slot (which is where werewolf "armour" is applied). This change corrects that oversight, meaning that werewolf armour - and any other armour applied to the skin slot - will now correctly prevent critical hits. This slot was already considered elsewhere (eg. in "checkarmor" in human_defense.dm).

As a side note, this should affect vampire susceptibility to critical hits as well, as they also have a defined skin_armor. However, their innate armour is more potent, and so this issue would be somewhat less visible with vampires.

## Why It's Good For The Game

As a physically-imposing antagonist, werewolves are naturally supposed to be sturdy, which is defeated by potentially being incapacitated by a single critical hit. Innate armour is pretty much a necessity for this, since werewolves shed all equipment when they turn. Consequently, this change makes them better able to terrorise non-antagonists, as antagonists are meant to do. (Additionally, this isn't a balance decision on my part - just a correction of an oversight.)

image
Since this change only prevents something that was never guaranteed in the first place, it's difficult to prove with footage, especially as the werewolf transformation removes weapon skills (making it very difficult to deal damage to itself through the armour). If higher standards of evidence are needed, I'd be happy to test alongside somebody else.

### DIFF
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -53,7 +53,7 @@
 	if(isbodypart(def_zone))
 		var/obj/item/bodypart/CBP = def_zone
 		def_zone = CBP.body_zone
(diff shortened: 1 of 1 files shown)


RadiantFlash7 (approved):
Much appreciated. Should fix vampyre armor of darkness AND werewolf skin from crits bleeding through.

PR #2582, opened 2025-03-18
Adds restrictions to fishable turfs.

ARefrigerator:
## About The Pull Request
Limits where you can fish.

- Adds a global proc and defines.
- Puts the check in fishing rods and fishing cages.
- Also reordered the fishing cage's turf check before the action so it checks before trying.

Allowed list:
```
#define VALID_FISHING_SPOTS list(\
	/turf/open/water/river,\
	/turf/open/water/cleanshallow,\
	/turf/open/water/sea,\
	/turf/open/water/sea/deep,\
	/turf/open/water/swamp,\
	/turf/open/water/swamp/deep )
```
I want to eventually remove the swamp water from this since that includes bog water and 'murk' tiles which I think shouldn't have any fishing abilities. This however makes the bog very oppressive for fishers so I will need to work on some alternatives before I remove it. I might also try allowing fishing but getting non fish items more often depending on the water turf type.

## Why It's Good For The Game
Currently you can fish on any water tile even in the temple's bath pools!! This makes it only possible in places that make more sense for such a thing. At the very least a few people might have to walk a little further to go fishing. (and possibly more in future if murk gets removed from the list)

https://github.com/user-attachments/assets/4d58eda1-e50b-4629-9c15-8e79845a7e2a

### DIFF
--- a/code/__HELPERS/unsorted.dm
+++ b/code/__HELPERS/unsorted.dm
@@ -1532,6 +1532,22 @@ GLOBAL_DATUM_INIT(dview_mob, /mob/dview, new)
 
 // Roguetown
 
+#define VALID_FISHING_SPOTS list(\
+	/turf/open/water/river,\
+	/turf/open/water/cleanshallow,\
+	/turf/open/water/sea,\
+	/turf/open/water/sea/deep,\
+	/turf/open/water/swamp,\
+	/turf/open/water/swamp/deep )
+
+//Valid spots for fishing add to it if there's more.
+/proc/is_valid_fishing_spot(turf/T)
+	for(var/i in VALID_FISHING_SPOTS)
+		if(istype(T, i))
+			return TRUE
+	return FALSE
+
+
 /proc/minone(input)
 	if(!input)
 		return FALSE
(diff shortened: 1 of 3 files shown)


RadiantFlash7:
Reasonable pr- Though I don't know if I would block out fishing in swamp tiles entirely in the future. After all, fish /do/ live in swamps. Even if it's just crawfish.

ARefrigerator:
> Reasonable pr- Though I don't know if I would block out fishing in swamp tiles entirely in the future. After all, fish /do/ live in swamps. Even if it's just crawfish.

Swamps maybe but isn't the town connected specifically to a bog? It's more frogs and turtles if anything in the water. I would also like to push the idea of it being truly an oppressive area slowly being stripped of normal life (due to rot) but that could just be me.

> Finally, no more bandit fisheries
Unfortunately you would need to change the water here to bath water for that to be true. Someone can do it in a future PR if they care about it but with this update it will be trivial to edit that out.
image

john-pawson:
Perhaps instead of making all swamp tiles fishable or no swamp tiles fishable, put in a couple of clean water spots in the bog, to signalize a clean lake with fish, so only specific parts that you have to search are fishable, it still takes a little bit of effort to find the right spot and you'd also be able to choose how far it is from town, directly controlling how convinient the spot is.

PR #2587, opened 2025-03-18
Fix for over/underflowing stat boosts causing round-permanent stat changes

GravyMaker:
## About The Pull Request

This PR fixes an issue where applying stat boosts or reductions which take a stat above or below the maximum or minimum values (20 and 1, respectively) could sometimes cause the stat to permanently change after the stat change is removed.

As part of the fix, I've also taken the time to refactor the code - the same code was duplicated for each stat, while the new code is wrapped in a single proc. I've also removed an unused "statbuf" variable.

One detail is that I've added a variable, "tempskill". This variable is of type /obj/statdata, which is defined at the end of the file, to allow code reuse rather than having to duplicate the stat change code for each stat. In principle, this could (probably? I don't know the details of BYOND's threading model) be replaced with a single global piece of data, since it's only used to allow the stat change code to return two variables at once. Please let me know if there's a better way of achieving this effect and I'll be happy to change that detail of the PR.

## Why It's Good For The Game

The previous stat boost logic could cause stat reductions for the rest of the round if a player wasn't careful with which buffs they were applying and when, which could lead to frustration - a friend of mine mentioned his intelligence becoming too low for certain magos activities as a result of the issue, for instance. This fix prevents that frustrating situation.

Image prior to any buffs, showing intelligence at 19:
image

Image captured following use of spice, showing intelligence at 20. If not for the cap, it would instead to 24:
image

Image captured after additionally taking swampweed, boosting intelligence by another 2, showing intelligence still at 20, plus an image showing the background BUFINT variable, which stores the amount we currently exceed the limit by (6).
image
image

Image after spice has worn off, showing BUFINT of 1, and intelligence still at 20:
image

Finally, after both buffs have worn off, intelligence is back to its original value of 19:
image

### DIFF
--- a/code/modules/mob/living/stats.dm
+++ b/code/modules/mob/living/stats.dm
@@ -135,7 +135,7 @@
 			BUFCON = tempskill.buffer
 
 		if("endurance")
-			tempskill.modifystat(STAINT, BUFINT, amt)
+			tempskill.modifystat(STAEND, BUFEND, amt)
 			STAEND = tempskill.value
 			BUFEND = tempskill.buffer
 
@@ -145,7 +145,7 @@
 			BUFSPE = tempskill.buffer
 
 		if("fortune")
-			tempskill.modifystat(STAINT, BUFINT, amt)
+			tempskill.modifystat(STALUC, BUFLUC, amt)
 			STALUC = tempskill.value
 			BUFLUC = tempskill.buffer
 


Purpleturtle3:
I think this is causing issues with stats, I had 7 end as a male sissean miner, got the tired debuff, and my end was still 7

GravyMaker:
> I think this is causing issues with stats, I had 7 end as a male sissean miner, got the tired debuff, and my end was still 7

Looking into it, you're correct - I made an error when copying the code for those stats. I'll get a fix in for that immediately.

Gyrandola:
Updated

Gyrandola (approved):

PR #2584, opened 2025-03-18
Alchemy: Herbal Remedies and Poisons

notafurryy:
## About The Pull Request
Alchemy reimagined! Toss away those rusty alembics and grab an iron pot, there is alot to brew ahead of you.

A whole **_10_** new herbs that spawn in the bog

5 new potions and 2 new poisons

New booze type!

New poison arrow variant

New alchemical equipment, a hourglass and an alchemical hearth, so that you don't burn your potions

Yes, most of this was stolen from kcd.

TO BE ADDED:

Lore friendly pamphlets with recipes

## Why It's Good For The Game
Currently alchemy is awful and not really that interesting.

## Proof of Testing (Required)
image
image

PR #2588, opened 2025-03-19
Health potions toxin fix

Rayforx:
## About The Pull Request
Health potion fixing toxin damage (0.25 from bad, 0.5 from normal, 1.25 from the good one)

Balance things of note from this would be the sickness ray, which does:
```
/datum/reagent/toxin
	name = "Toxin"
	var/toxpwr = 1.5

```
1.5 damage per tick means that even if you take any of the health potions, you will still be kicking the bucket.
```
/datum/reagent/berrypoison/on_mob_life(mob/living/carbon/M)
		M.adjustToxLoss(3, 0)
```
And berry poison which does 3 per tick.

## Why It's Good For The Game
An alternative to trying to fix a leech while in the middle of nowhere when a person is in critical. While it is pretty slow with the bad health potions, but a bit better with a normal health pot.

People kind of expect health potions to "fix" most damage, which leads to them dumping a health potion into a person while expecting that they are fixed. It does feel like a bit hug box to allow it to fix toxin damage, but I am hoping that giving a little bit of healing will help in places where a person can't find a leech in time. 

image

### DIFF
--- a/code/modules/roguetown/roguejobs/alchemist/reagents.dm
+++ b/code/modules/roguetown/roguejobs/alchemist/reagents.dm
@@ -22,7 +22,7 @@
 	M.adjustBruteLoss(-1*REM, 0) // 45u = 15 oz = 120~ points of healing
 	M.adjustFireLoss(-1*REM, 0)
 	M.adjustOxyLoss(-1, 0)
-	M.adjustToxLoss(-0.5*REM, 0)
+	M.adjustToxLoss(-0.2*REM, 0)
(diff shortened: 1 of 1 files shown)


notafurryy:
I think health potions not having tox damage removal is intentional. This makes leeches pretty much obsolote. Also health potions might need to be nerfed a little with the pr I made https://github.com/Rotwood-Vale/Ratwood-Keep/pull/2584

DelineFortune:
>>I am hoping that giving a little bit of healing will help in places where a person can't find a leech in time.

Put some efforts into the game and get a leech from start
Health potions are fallout stims already

PR #2594, opened 2025-03-21
Removes mog from the filter

wow50000:
## About The Pull Request
See title

## Why It's Good For The Game
There's one guy called Mog Hog which is a valid name and not a reference to mogging and it should be an admin issue and very unimersive that we can't call him Mog hog and call him degrade hog instead

I'm actually gonna rope myself if this doesn't work

### DIFF
--- a/strings/accent_universal.json
+++ b/strings/accent_universal.json
@@ -81,6 +81,7 @@
 		"clickbait": "deceiving",
 		"soyjak": "eunuch",
 		"soyjack": "eunuch",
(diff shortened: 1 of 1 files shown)


Gyrandola:
Just remove mog then.

horcporker:
Thank you boobear

wow50000:
> Just remove mog then.

done

Gyrandola (approved):

PR #2596, opened 2025-03-21
Makes forks (utensil) functional.

ARefrigerator:
## About The Pull Request
Allows you to use forks to eat food. Added functionality into:
```/obj/item/cooking/platter/attackby```
and 
```/obj/item/reagent_containers/food/snacks/attackby```

At some point someone might need to merge the two fork types into one, I don't think the metal forks have any special functionality that makes them distinct.

## Why It's Good For The Game
Forks exist but you can't do anything with them. Plus this will allow you to keep the food on the table and eat it like a proper meal (for the RP)

https://github.com/user-attachments/assets/01c54094-ad8b-42ee-beed-74ea0e7ba90f

### DIFF
--- a/code/modules/food_and_drinks/food/snacks.dm
+++ b/code/modules/food_and_drinks/food/snacks.dm
@@ -361,6 +361,7 @@ All foods are distributed among various categories. Use common sense.
 	if(istype(W, /obj/item/kitchen/fork/) || istype(W, /obj/item/kitchen/ironfork/)) //why is this a different type????
 		if(do_after(user, 0.5 SECONDS))
 			attack(user, user, user.zone_selected)
+			return ..()
 
 	if(istype(W, /obj/item/storage))
 		..() // -> item/attackby()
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: code/modules/food_and_drinks/food/snacks.dm [hunk 1]
```diff
@@ -357,6 +357,11 @@ All foods are distributed among various categories. Use common sense.
 				. += "[src] was bitten multiple times!"
 
 /obj/item/reagent_containers/food/snacks/attackby(obj/item/W, mob/user, params)
+
```
Gyrandola, line 360: You could probably so an early `return ..()` on those checks

Gyrandola (approved):

code/modules/food_and_drinks/food/snacks.dm
ARefrigerator: Alright I'll do it soon.

PR #2601, opened 2025-03-22
Novitiate fix  (broken latejoin)

DelineFortune:
renamed from Novitiate back to Churchling

The problem is 
image
 I guess 
image
image


Gyrandola:
So you identified the problem and instead of fixing it, you changed everything else. Why couldn't you just put Notivitiate in the youngfolk list?

DelineFortune:
> So you identified the problem and instead of fixing it, you changed everything else. Why couldn't you just put Notivitiate in the youngfolk list?

fine ok...........
Didnt touch their 	C.grant_spells_churchling(H) because people use it right now in their prs 
image

Gyrandola (approved):

PR #2602, opened 2025-03-23
Expands goldface access.

ARefrigerator:
## About The Pull Request
Allows anyone who can read, and has access to an unlocked goldface, to access the secret menus. 

## Why It's Good For The Game
It's a pain for a solo shop hand to run the merchant if the merchant dies or is away. It would also make it more fun if people could also mess with the menus for their own benefits. Of course they'd still need a key. it might also make merchant slightly more exciting if people find out they could access certain things this way without your explicit permission. Up to you. If you want I can just edit it to include shop hands but I think anyone being able to access it works well.

Not sure if this breaks anything else but the portable gem device seems to be unused so it should be fine?

https://github.com/user-attachments/assets/6039f434-c7c9-4cfa-8d7e-a877513c8e60

### DIFF
--- a/code/modules/roguetown/roguemachine/merchant.dm
+++ b/code/modules/roguetown/roguemachine/merchant.dm
@@ -307,14 +307,13 @@
 	contents = "<center>GOLDFACE - In the name of greed.<BR>"
 	contents += "<a href='?src=[REF(src)];change=1'>MAMMON LOADED:</a> [budget]<BR>"
 
-	/*var/mob/living/carbon/human/H = user
-	if(H.job == "Merchant")
+	var/mob/living/carbon/human/H = user
+	if(H.job == "Merchant" || H.job == "Shophand")
 		if(canread)
-		contents += "<a href='?src=[REF(src)];secrets=1'>Secrets</a>"
+			contents += "<a href='?src=[REF(src)];secrets=1'>Secrets</a>"
 		else
 			contents += "<a href='?src=[REF(src)];secrets=1'>[stars("Secrets")]</a>"
-	*/
-	contents += "<a href='?src=[REF(src)];secrets=1'>Secrets</a>"
+
 	contents += "</center><BR>"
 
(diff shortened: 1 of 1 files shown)


RadiantFlash7:
I'm not sure this is a good idea, frankly.
The point of having secret's locked to merchant is so the merchant isn't immediately replaceable when they irritate someone. Be it by their own shophands, or the duke ordering them removed.

ARefrigerator:
I could reduce it back down to just the shop hands. The issue is if the merchant dies that's it. If he didn't unlock anything the store has only a basic list of goods. The other roles all seem to have ways to replace a head if they go missing.

RadiantFlash7:
> I could reduce it back down to just the shop hands. The issue is if the merchant dies that's it. If he didn't unlock anything the store has only a basic list of goods. The other roles all seem to have ways to replace a head if they go missing.

At a minimum shophands only. No one else.
Even then, I'm /still/ a touch skeptical- but I suppose it's a reason for merchants to not miss-treat their shophands.

ARefrigerator:
Maybe I could think of something for a merchant to get in response to this, or split licenses and secrets so shop hands could still buy licenses while all the other stuff is merchant only.

Elli-Skala:
one idea which i imagine would be a pain to actually set up is that any one can buy the licencing, but its tied to the PERSON and not the gold face, and give the merchant round start access to the licences. making them have a actuall purpose as a middle man instead of objectively better dead then they are now

ARefrigerator:
Okay I set it back and included only shophands to the access under the assumption that working under the merchant means they would also have access to their machine should anything happens. I know the merchant in general needs other work so I might focus on trying to make that better (i.e. fixing the prices so he's competitive with a smith who might just give things away for free when they get a pile of ore) The economy in general is a bit wonky.
shophand

RadiantFlash7 (approved):
This has been test merged long enough. Good to go.

PR #2606, opened 2025-03-23
Gives turfs their ImmediateCalculateAdjacentTurfs() proc on init

ARefrigerator:
## About The Pull Request
Gives the ```ImmediateCalculateAdjacentTurfs()``` proc back to turfs. Seems like whatever ripped out atmos forgot about this. This proc seems to have almost no impact on loading time either so it should be a trivial to allow it to run on init.

If someone thinks I should rewrite smoke to not use this as there's some other function making a list of adjacent tiles let me know and I could try changing the smoke procs to use that instead but this was just the easiest solution for what I wanted.

Not sure what else uses this but it would possibly fix them too as before this whole list was always null. 

**EDIT**: HOWEVER if you make a new turf it calls this proc anyway so smoke spreads on say stone floors or twig floors you make but nothing else by default! Messed up!

## Why It's Good For The Game
Turfs knowing their adjacent turfs is useful for many things but it would also, and more importantly what I wanted to do, is fix smoke. It currently does not spread. This returns that functionality back to smoke.

smoke_fix

### DIFF
--- a/code/game/turfs/turf.dm
+++ b/code/game/turfs/turf.dm
@@ -105,7 +105,7 @@
 		has_opaque_atom = TRUE
 
 	ComponentInitialize()
-
+	ImmediateCalculateAdjacentTurfs()
 	queue_smooth_neighbors(src)
 
 	return INITIALIZE_HINT_NORMAL


Gyrandola:
I have this feeling this isn't the optimal solution but I'll keep it TMed

ARefrigerator:
Maybe but it's still called when you craft turfs anyway so either it needs to be removed completely or allowed to function and I chose the lazier option. I doubt the list sitting there adds much to the memory and as I stated before there's a a chance other things _might_ be trying to use this since it's a rather old list that was used to get adjacent tiles.

Gyrandola (approved):

PR #2613, opened 2025-03-24
Uses update_icon() to better give feedback to washbins and containers

ARefrigerator:
…s being filled/emptied.

## About The Pull Request
Uses update_icon() to better give feedback to wash bins and container when filling emptying. Simply puts the update_icons() in the do_after loop.

I don't think there's anything that could be called to update icon that could give a runtime here but it's possible _something_ could but it seemed to work fine on buckets vs. buckets and the wash bin.

## Why It's Good For The Game
It's a minor graphic issue that can make buckets look full when in fact they are empty. It also is a bit annoying to have to put something in the washbin to update its sprite. This changes that.

https://github.com/user-attachments/assets/84113c27-88f7-4519-80e0-5ee1a512f059

### DIFF
--- a/code/modules/farming/bin.dm
+++ b/code/modules/farming/bin.dm
@@ -188,6 +188,7 @@
 					reagents.add_reagent(/datum/reagent/water/gross, amount_to_dirty)
 			update_icon()
 			return
+	update_icon()
 	. = ..()
 
 /obj/item/roguebin/trash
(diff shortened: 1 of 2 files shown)


ARefrigerator:
hang on. I might have to make this specific. I had a barrel turn invisible on me...

ARefrigerator:
Yeah I might have to work on this more before I bring the PR in. for some reason barrels do not like having their update_icon called and hide away. I might make it more specific to the wash bin or something or make sure things don't turn invisible.

ARefrigerator:
I see why. The update_icon for barrels is messed up because it's using a system that doesn't appear to be used at all?? Maybe I'll have to fix that too... It's turtles all the way down.

PR #2615, opened 2025-03-24
Adds Minstrel towner role.

ARefrigerator:
## About The Pull Request
People like to play a musician but currently the only way is to be a refugee. This allows them to spawn in as a towner. Plus towners need some love so I might try to put more roles in for them in future.

## Why It's Good For The Game
The alternative is being a foreigner and traveling a bog to get to town just to play songs and RP. This will allow a little more music around town while also allowing those who want to be a minstrel to do so at roundstart rather than wait for the fog to clear.

https://github.com/user-attachments/assets/3fcb9f53-e948-4f79-8234-a3d2bc8df616

### DIFF
--- a/code/modules/jobs/job_types/roguetown/towners/types/minstrel.dm
+++ b/code/modules/jobs/job_types/roguetown/towners/types/minstrel.dm
@@ -12,25 +12,23 @@
     ..()
     H.mind.adjust_skillrank(/datum/skill/misc/music, 4, TRUE)
     H.mind.adjust_skillrank(/datum/skill/misc/reading, 3, TRUE)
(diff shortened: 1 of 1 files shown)


ARefrigerator:
Okay I'll adjust the values.

RadiantFlash7 (approved):
This PR is in line with other towner stats. Having knife skill be apprentice is fine for a non combat role, especially as it's their only weapon skill as a towner. 
Hurray for the return of minstrels!

ARefrigerator:
Oh wait this is still open I can do it here...

ARefrigerator:
Some adjustments to the towner minstrel. Also gave them a second instrument too. Is a little tricky but it should insure they always get 2 random items and if they get the same it will pick the next one over in case. Preview: 
image

PR #2623, opened 2025-03-25
Adds generic proc for delays via skill level.

ARefrigerator:
## About The Pull Request
Allows more tuning via a global proc one can call for getting delays for do_after (or anything else) for skills.

example usage in skill via code:
```
var/delay = get_skill_delay(user.mind.get_skill_level(/datum/skill/craft/cooking), 0.5, 4)
if(do_after(user, delay, target = src))
```
and video below

## Why It's Good For The Game
Would make coding this a little easier. You'd no longer have to code anything too magical manually every time. You just set the skill, the fastest and slowest possible times and then the function will use your skill to figure out where between those two values your character should take to do something (via LERP)

EXAMPLE USE CASE, COOKING SKILLS FOR SETTING PLATTERS:

https://github.com/user-attachments/assets/3663f7f7-d009-48a6-ba8b-9c7bdcb69a28

### DIFF
--- a/code/__HELPERS/unsorted.dm
+++ b/code/__HELPERS/unsorted.dm
@@ -1557,7 +1557,14 @@ proc/is_valid_hunting_area(area/A)
 			return TRUE
 	return FALSE
 
-
+// How long an action (e.g. do_after) can takes IN SECONDS by using skill checks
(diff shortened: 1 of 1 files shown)


RadiantFlash7 (approved):
This is a backend proc which will be VERY useful indeed. Thank you, fridge.

PR #2625, opened 2025-03-25
Fixes head attachment surgery disconnects.

ARefrigerator:
## About The Pull Request
Stops player from disconnecting when head is reattached. Will it break other things? Maybe! I only saw one runtime and that's what the other file fixes. If this seems glaringly messed up or you know this code is breaking something let me know.

The runtime I was getting was about getting turf lum counts but I think it only was an issue from attaching the head to the surgery but this could potentially be breaking how mob_movement can get a turf?? It's the only proc that was giving me this issue so maybe the way that specific proc tries to call turf is messed up? Not sure...

Since the brainmob thing isn't used the issue was caused by it nulling out the already existing player's ckey in the body.

## Why It's Good For The Game
people get disconnected and sent back to lobby when a physician tries to save them. it's ass. please test merge ASAP

https://github.com/user-attachments/assets/bce837bb-8d5f-4a86-9d8d-9e56d2c2dc0b

### DIFF
--- a/code/modules/mob/mob_movement.dm
+++ b/code/modules/mob/mob_movement.dm
@@ -607,7 +607,12 @@
 		rogue_sneaking = TRUE
 		return
 	var/turf/T = get_turf(src)
-	var/light_amount = T.get_lumcount()
+	// This is hacky but it's the only runtime that fixing decap gives
+	// please forgive me...
+	var/light_amount = 0
+	if(T != null)
+		light_amount = T.get_lumcount()
+		
 	var/used_time = 50
 	if(mind)
 		used_time = max(used_time - (mind.get_skill_level(/datum/skill/misc/sneaking) * 8), 0)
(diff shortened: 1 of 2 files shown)


ARefrigerator:
Also while on the subject of the bug. I think it goes away when you take a step. There's something about going from being attached before taking your first step where I guess your character doesn't know where it is on the ground until it moves? I think that might be the issue so with any luck this might be a solved issue going forward, finally.

RadiantFlash7:
> Also while on the subject of the bug. I think it goes away when you take a step. There's something about going from being attached before taking your first step where I guess your character doesn't know where it is on the ground until it moves? I think that might be the issue so with any luck this might be a solved issue going forward, finally.

Looking this over- I thought we prevented the player being transfered to the brain at all. Shouldn't they remain within a players body now when decapitated?
I haven't experienced this issue personally, so I'm a touch confused on the circumstances you're describing fixing. 

Are you saying players are remaining in their body until their head is re-attached, at which point it crashes them and when they come back, they're in the lobby?

ARefrigerator:
Being sent back to the lobby seems to be directly caused by the ckey being null, if you null someone's ckey they will disconnect and the game will act like they never joined sending them to the menu. The brainmob var (null) was trying to overwrite the mob's current ckey.

The other issue I am explaining is specifically when you reattach the head until the mob moves: it seems that one specific proc did not understand where the mob was. I think moving fixes it though and I doubt many will be trying to sneak/hide in shadows in that time so I think it's okay.

Gyrandola (approved):

PR #2626, opened 2025-03-26
Dark Elf Adjustments And Additions

Xaytan:
## About The Pull Request

A pull request in response to ongoing feedback regarding the Dark Elf darkling effect. This PR is intended to improve playability, reduce IRL eyestrain, and add new features for drow characters to enjoy in making them feel more like the horrible little creatures of the dark they are.

- Adjusts eye strain and migraine thresholds to be slightly more forgiving.
- Replaces white screen flash effect with blur effect to reduce IRL eyestrain. We don't want to be THAT immersive.
- Removes blur effect from migraine, this was having the unintentional effect of removing the ability to read. Oops!
- Dark Elf hoods will now equip themselves in the mask slot instead of just spawning in your hand or at your feet.
- Darkling now properly removed on losing the Dark Elf species.
- Slightly improved night vision, dark elves can now see very well in the dark.
- Adjusted feedback message to be more _immersive_ when being blinded by bright lights.
- Added "dark elf sunshields". A type of spectacles that protect from the hateful sun. Wearing them removes your darkvision.
- Added sunshields to starting loadout of Dark Elf Hands and Merchants. Check your backpack!
- Added sunshields to items that can be purchased from the merchant. Wardrobe, 150 mammons. Pricey!

## Why It's Good For The Game

Changes made in response to feedback. Flash was too bright, eye strain occurred too quickly. Sunglasses allow certain roles like the Hand to function, and creates a distinction between wealthy dark elves and broke ones. It also helps if you want to be able to show your face around town, literally. You got that reference picture after all!

image
image
image

### DIFF
--- a/code/datums/components/darkling.dm
+++ b/code/datums/components/darkling.dm
@@ -49,7 +49,7 @@
 	current_light_stress = clamp(current_light_stress + 5, 0, max_light_stress)
 	darkling.blur_eyes(15)
 	next_blind = world.time + (rand(30 SECONDS, 60 SECONDS) * (darkling.STACON/10)) //Con determines how frequently you can get blinded
-	to_chat(parent, span_danger("My eyes water at the sudden exposure to bright light! Maybe I should "))
+	to_chat(parent, span_danger("My eyes water at the sudden exposure to bright light!"))
 
 //Applies the effects of our current light stress accumulation threshold
 /datum/component/darkling/proc/apply_stress_effects(var/mob/living/carbon/darkling)
--- a/code/modules/jobs/job_types/roguetown/yeomen/merchant.dm
+++ b/code/modules/jobs/job_types/roguetown/yeomen/merchant.dm
@@ -51,7 +51,8 @@
 	backl = /obj/item/storage/backpack/rogue/satchel
 	id = /obj/item/clothing/ring/gold
 	backpack_contents = list(/obj/item/storage/keyring/merchant)
-
+	if(isdarkelf(H))
+		backpack_contents.Add(/obj/item/clothing/mask/rogue/spectacles/delf)
 	if(H.gender == MALE)
 		shoes = /obj/item/clothing/shoes/roguetown/armor/leather
 		H.dna.species.soundpack_m = new /datum/voicepack/male/wizard()


ARefrigerator:
I was thinking about the something like sunshields the other day. Good stuff!

Hyperoit:
a heads up when going knave bandit im missing my steel mask due to the hood , please do fix that i miss my mask.

ARefrigerator:
If the mask slot is full it might be best to put it into the satchel. Most jobs start with it empty or pretty close to empty. If the job doesn't have that maybe put it into the belt next (if it fits), then the right hand (usually also empty). As a last resort you could default back to the floor drop as a final solution to ensure they keep it.

Would be a tricky series of checks though but you could ensure they have everything plus the hood.

Xaytan:
> If the mask slot is full it might be best to put it into the satchel. Most jobs start with it empty or pretty close to empty. If the job doesn't have that maybe put it into the belt next (if it fits), then the right hand (usually also empty). As a last resort you could default back to the floor drop as a final solution to ensure they keep it.
> 
> Would be a tricky series of checks though but you could ensure they have everything plus the hood.

It's tricky because I'm applying the mask when the species is selected, which means that it's placed there before job code spawns its masks. I'll have to figure out a better way to accomplish it. Maybe make it so they can retrieve the mask from a tree same way the specials work, I don't know.

ARefrigerator:
You can access the mob in the `pre_equip()`, `post_equip` and the `apply_character_post_equipment` procs. I think despite it being called pre_equip it happens after the mob spawns with it stuff so maybe either place works.

```
/datum/subclass/proc/equipme(mob/living/carbon/human/H)
	// input sleeps....
	set waitfor = FALSE
	if(!H)
		return FALSE

	if(outfit)
		H.equipOutfit(outfit)

	post_equip(H)

	H.advjob = name

	//sleep(1)
	//testing("[H] spawn troch")
	if(torch)
		var/obj/item/flashlight/flare/torch/T = new()
		T.spark_act()
		H.put_in_hands(T, forced = TRUE)

	var/turf/TU = get_turf(H)
	if(TU)
		if(horse)
			new horse(TU)

	// After the end of adv class equipping, apply a SPECIAL trait if able
	apply_character_post_equipment(H)
```

the post equipment one is pretty simple you could probably do the species check there:
```
/proc/apply_character_post_equipment(mob/living/carbon/human/character, client/player)
	if(!player)
		player = character.client
	apply_charflaw_equipment(character, player)
	apply_prefs_special(character, player)
```

Gyrandola (approved):

PR #2631, opened 2025-03-27
Dye bin: fix multiple insertion bug, updates window GUI.

ARefrigerator:
## About The Pull Request
Dye bins did not check for contents before inserting items. This adds the check as well as changes the GUI for the window to also update the color when selected (image below)

## Why It's Good For The Game
Dye bins confuse people when they allow multiple insertions because it only treats the most recent insertion as the only thing in it. This prevents that. The GUI color didn't update because it was using an ascii symbol that didn't respect the color change so now it's more obvious what you selected.

dyebin_01
dyebin_02


ARefrigerator:
Made the window smaller height wise so there's no big space.
dyebin_03

RadiantFlash7 (approved):
Code changes seem fine. A old bug thats been needing fixing. Well done.

PR #2632, opened 2025-03-27
Gives priest cure rot spell.

ARefrigerator:
## About The Pull Request
What it says in the title.

## Why It's Good For The Game
If you play priest, don't have a pestra acolyte (it happens, especially on low pop) you cannot save a deadite at all. It also makes sense a priest would probably know this. 

cure_rot

### DIFF
--- a/code/modules/jobs/job_types/roguetown/church/priest.dm
+++ b/code/modules/jobs/job_types/roguetown/church/priest.dm
@@ -82,7 +82,7 @@ GLOBAL_LIST_EMPTY(heretical_players)
 	var/datum/devotion/C = new /datum/devotion(H, H.patron) // This creates the cleric holder used for devotion spells
 	C.grant_spells_priest(H)
 	H.verbs += list(/mob/living/carbon/human/proc/devotionreport, /mob/living/carbon/human/proc/clericpray)
-
+	H.mind.AddSpell(new /obj/effect/proc_holder/spell/invoked/cure_rot)
 	H.verbs |= /mob/living/carbon/human/proc/coronate_lord
 	H.verbs |= /mob/living/carbon/human/proc/churcheapostasy
 	H.verbs |= /mob/living/carbon/human/proc/churchannouncement


DelineFortune:
Whats point of being a pestran then...
And I dont that you are supposed to save everyone 
Accept death
Cannot always win

ARefrigerator:
It was a suggestion passed on to me and it took 5 minutes to do so I did it. 

I think the general argument was that "_this is something that the leader of the house of the Ten should be able to do since they are meant to be a head role that tries to function as all their underlings in some fashion._" The other argument was that "_it was already added into other servers and works fine._"

Part of the suggestion also included removing cure rot from physicians so I could add that if you wish alongside it or in another PR. I know people have mentioned that and I'd be fine with it. Seems more like the church's job.

gamevoin:
Having to rely on luck of the draw to have a player playing Pestran Acolyte to perform what's generally considered one of the main Church functions isn't ideal, especially during low-pop or medium-pop rounds where there's very often no Pestran or even Acolytes around. Having something like this as a guarantee for Priest not only makes thematic sense, but will also add a tad more popularity to the role and church in general

The Church's healing is already vastly out-performed by decent Town/Court Physicians (especially so if their De-Rotting Ability gets fixed and starts functioning again), it'd be nice to have a little something IMO, and a Pestran Acolyte is still going to have a lot more freedom of movement since the Priest is a very targeted role - when there's both a Pestran & Priest around, that'd mean you could spread out the workload and the former would have more to do than just be a de-rotting bot roaming around, while the latter might see a tad more traffic to interact with even if they play it safe and hole up in their base

ophaq:
Might as well remove the rot removal spell from physician while you're at it since it doesn't work anyway.

RadiantFlash7:
Hm. I'm not sure about this honestly.
There's a lot of issues with the church and physician from a design standpoint.

As a gameplay concept the church acolyte were designed to be the healers of the server, split based upon their patron.
As people added more content, this shifted dramatically with the allowance of so many other patrons in the Acolyte role, dramatically reducing the chances of getting a pestran. (An astratans already guaranteed to be there due to the priest.)

From a design perspective, allowing other patrons to be acolytes was probably a major mistake- Yet it's one the community really seems to enjoy. 
To compound the issue- Physicians are pretty much superior to the church at present. They are able to revive both rotted and dead people, as well as tend to folks, where as the church is directly locked by patron as to what they can do, requiring differing patrons to handle different situations.

Or atleast they would be, if Physicians Cure rot weren't bugged at the moment.
I wonder if a better solution would be to add a hospitaller role similar to mortician that acts as a guaranteed pestran spawn a la morticians (Morticians are guaranteed necrans).

Purpleturtle3:
The reason we have druids and morticians is to avoid having acolytes that weren't astratan or pestran, and now...

AukariBunn:
> The reason we have druids and morticians is to avoid having acolytes that weren't astratan or pestran, and now...

And now the problem is that nobody plays acolytes. The church overally needs a huge overhaul because nobody plays it.

gamevoin:
> I wonder if a better solution would be to add a hospitaller role similar to mortician that acts as a guaranteed pestran spawn a la morticians (Morticians are guaranteed necrans). To those unaware, Morticians act as a secondline combat role against antagonists, behind templars. They have very high wrestling/unarmed skill along with access to rebuking the undead.

That's a good thought about separating the generic Acolyte into a couple more limited, yet far more evocative roles. As it stands, every single Acolyte has the same skill set (with very high wrestling & unarmed just like Mortician) with equally high Treatment levels. Being their own unique role would be appreciated not only by the Pestran, but especially so by Malum Acolytes who have so little to work with at the moment - this isn't a matter of only skill levels & mechanical oddities, but rather about giving them a notable and visible role within the Church itself, maybe with special rooms and equipment contributing to their work and rituals.

The Hammerites from Thief or The Builders from The Dark Mod would be great inspiration for a Malum role, Pestra can become a proper Hospitalier, surely something can be figured out for Eora, and Acolyte can go back to either being an Astrata-only helpful cleric or be elevated into a Hierophant-like role to stand out among the other Patron roles. Novitiates will still be there as a beginner-friendly role without any particular responsibility.

But those PRs and ideas are a while away yet. Giving Cure Rot to the Priest may be a band-aid fix, but I have a feeling it'd be fairly effective at what it's intended to do. There'll be less incentive to gravitate exclusively toward Pestran Acolyte when they're not the only ones who have access to such a sought-after ability. (I'm presuming Physicians will have their De-Rot ability removed, since it's bugged anyway and they already have so much else going for them)

ARefrigerator:
Right. I feel this is more of a stop gap until some other things get sorted.

Gyrandola:
Not a fan of this but I think everyone knows that.

Frienderanian:
> Not a fan of this but I think everyone knows that.

Genuine question, what would you prefer?

Gyrandola:
> > Not a fan of this but I think everyone knows that.
> 
> Genuine question, what would you prefer?

I would prefer rot being uncurable since it is the almost world ending scenario the setting takes place in.

RadiantFlash7:
Riight. After discussion between maintainers on this:

Balancing for the server is mostly done for 'high/average pop rounds', Where you would /hopefully/ have a pestran acolyte around. With that said, We're intent on keeping each church role patron spells to their patron only, and moving cure rot to the priest would put a pretty hard hit to this themeing. 

I'm aware theres an issue with non pestrans taking up all the acolyte slots, and thats something that will be addressed fairly soon, But as it is, Giving the priest cure rot is a no go. Notably, the same PR that will address the issue mentioned with pestrans will also address physicians being a fix-all role.

PR #2634, opened 2025-03-28
Nerfs non steel spears, buffs flails a bit, murders leather bracer exploit

Purpleturtle3:
## About The Pull Request

All non **STEEL** spears have 30 ap while the rest has 50, flails now have one (1) more force so it is twenty one (21) instead of twenty (20), and the stamina drain has been reduced a bit... this is nowhere near how op they used to be. Also, kills the exploit of smelting leather into steel, Malum weeps.

## Why It's Good For The Game

Balancejack, exploit bad

imagen
imagen

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/polearms.dm
+++ b/code/game/objects/items/rogueweapons/melee/polearms.dm
@@ -9,7 +9,7 @@
 	chargetime = 1.5
 	warnie = "mobwarning"
 	hitsound = list('sound/combat/hits/bladed/genstab (1).ogg', 'sound/combat/hits/bladed/genstab (2).ogg', 'sound/combat/hits/bladed/genstab (3).ogg')
-	penfactor = 50
+	penfactor = 30
 	item_d_type = "stab"
 
 /datum/intent/spear/bash


RadiantFlash7:
In the future atomize your PRs!!! Keep fixes seperate from balance changes, and additional content where able. That said, I'll get around to reviewing this tomorrow.

Purpleturtle3:
Sorry I need to hold the pr hostage with a singular (1) bugfix

RadiantFlash7 (changes requested):
> All non steel spears have 50 ap while the rest has 30

Right. This is actually backwards of what was discussed in discord I'm afraid. Iron spears having more penetration then steel ones is the yikes. Since Iron is easier to acquire and find, having them pierce far better then steel spears is- problematic.

If you're able to invert that, by instead giving Iron spears 30 AP, and steel ones 50, that would work.

RadiantFlash7 (approved):
Seems fairly good.
Thank you for adjusting the pen for the spears.
Flail adjustments seem reasonable (They were probably nerfed too hard to begin with back in the day)
and the Exploit fix is always welcome.

PR #2637, opened 2025-03-28
Makes cooking skills matter more: delay time, extra slice, buffs

ARefrigerator:
_Full title_
Makes cooking skills matter a little more: do_after time, extra slices adding buffs to plattered food.

## About The Pull Request
Does the following

- Gives people with better cooking skills a noticeably lower time to do actions (plating, slicing, combining)
- Adds skill gain to cutting food
- Adds buff to plated foods if your skill is expert or higher
- If your skill is expert or higher you gain an extra slice of whatever you cut.
- Uses new delay proc to make reading easier, and delay time adjustments easier to read

## Why It's Good For The Game
Cooking skill feels like it doesn't matter much. This will be a small part in helping it get to where I'd hopefully like it to be one day where people with high cooking skills will be valued. The buff is probably a little janky so if you want I can remove that until I get a better idea but there's only perhaps 2 roles in an entire game who can do such a thing right now (tavern cook, servant chef)

https://github.com/user-attachments/assets/7a19d1df-5d89-474c-8810-69b42669483b


ARefrigerator:
Actually I'm gonna come back to this. Making some sort of proc to handle the food first might be better because it will consolidate where all the do_after and buff/nerfs go making it less of a pain to add later.

PR #2643, opened 2025-03-29
Gives bookworm heir magic book and 'learn a new spell'

ARefrigerator:
## About The Pull Request
Puts the book into their satchel and gives them 'learn new spell' option

## Why It's Good For The Game
Quote from original post on discord from a heir main who picks it (Vyutriska): 

> TL;DR Give Bookworm-subclass Heir the Learn A Spell spell, and an apprentice level book to read from. I play this role and subclass a lot, I am very biased in my desire of it but I think it would be good and give the role a little more life, and round presence.

> Bookworm is my favorite subclass- I love the idea of a noble with a passive interest in magic, and it's basically the only subclass I ever play. I rarely ever see others play it, though, and I think the reason why is that it's a little underwhelming. You only get four spells: Prestidigitation, Message, Fetch, and Darkvision. They're all cute spells and have their niche uses, but they aren't really game-changers. 

> I think that giving the Learn a Spell activity and a book to use it with will give the Bookworm Heiress a great impetus to be more of an active participant in the round. Much like how Militant Heir can go and speak to the Blacksmiths for a good set of armor and a weapon, and bother the retinue to train their skills, the Bookworm Heir could get meaningfully involved in the magos tower and get a great book to learn lots of useful spells!

> I think that encouraging that would make the rounds a little more dynamic, while letting them have a book to start gives them enough independence to where they wouldn't feel the need to bother the Magos every single round in order to make the most of their role. But they could if they wanted to, and there are many avenues of approach they could take to make the round more interesting.

> For example, a humble heir might flavor themselves as a fellow student among the apprentices- a privileged one, but a peer nonetheless. Maybe he views the tower as an escape from boring court life. Whereas a more haughty heir might try to abuse their privilege to get the magos to focus on them exclusively.

I personally don't play the role but it does seem like an oversight that they don't have these in any capacity.

**NOTE:** Apparently there's only ever one bookworm heir so you don't have to worry about multiples of them or it being some balance issue.

https://github.com/user-attachments/assets/ff84df7d-400f-4ae0-ab7b-6bb95dbcb7c7

### DIFF
--- a/code/modules/jobs/job_types/roguetown/youngfolk/prince.dm
+++ b/code/modules/jobs/job_types/roguetown/youngfolk/prince.dm
@@ -171,8 +171,12 @@
 		H.change_stat("constitution", -1)
 		H.change_stat("strength", -2)
 		H.change_stat("endurance", -2)
+		H.mind.AddSpell(new /obj/effect/proc_holder/spell/invoked/learnspell)
+		backpack_contents = list(/obj/item/book/granter/spellbook/apprentice = 1)
 	ADD_TRAIT(H, TRAIT_NOBLE, TRAIT_GENERIC)
 	ADD_TRAIT(H, TRAIT_SEEPRICES, TRAIT_GENERIC)
+	
+	
 
 /datum/subclass/prince/inbred
 	name = "Inbred"


Hmeister-real:
learn new spell should be behind all the other spells on the left

Purpleturtle3:
It's not an oversight, bookworm heir not being able to learn spells is 100% intended

RadiantFlash7:
As the self proclaimed 'Magic maintainer' I've been more or less handling changes and additions related to magic.
When a 'magic heir' was proposed I wasn't a fan. At all. At a minimum, I felt like they shouldn't be learning spells, as thats how you end up with heirs having an assortment of combat spells. Mage heirs getting added, while not getting a way to learn more spells was a compromise I made at the time to allow people to have their fun.

As it stands, I still don't particularly want bookworm heirs to be able to learn more magic. Their role is as an heir, I don't really want to shift an heir role into a proper mage role. I'll leave this PR open to discuss for a bit- But as it is, I'm heavily inclined to reject it.

Skyglazer:
As the original requester, I was unaware of the bookworm heir's history of intentionally not having the ability to learn spells. From my perspective playing lots of heir, I see my militant sibling (it's always a militant, every time) constantly getting strong armor and weapons, whereas I have to resign myself to my spells that are pretty much only for flavor, aside from Message. 

I'll not lie, one of the main things I wanted from the ability to learn spells was the ability to get a combat spell- namely Lightning Bolt, because the character I play is very bratty and stuck-up and would love to zap an uppity lowborn with lightning for talking back to her. Following discussions in the discord, I've since become a little more enlightened to just how much a careful balance is endeavored to be kept between the Keep and the other factions at play. I don't think that their access to spells would be immensely troublesome, especially if there were some way to limit how many combat spells they could choose-- their stats and skills are so low that they'd be pretty much helpless once a single grab lands on their mouth-- not to mention their spells having to be aimed, and having cooldowns.

That aside, I think it's a shame that RadiantFlash sees it as "Disney level theming," I think a noble interested in learning magic is perfectly reasonable for any level and theme of fantasy- dark or light, and that a royal successor that knows magic needn't be a princess frollicking in the meadows and attracting woodland creatures as a princess might. I think it can be taken seriously, and I think that giving them the capacity to learn spells would have lent a mechanical side into the subclass's background as an active studier of the arcyne.

I'd still like the subclass a little expanded in some way, if not with the ability to customize and express character through explicit spell selection, through the addition of maybe one or two more spells on the already quite underwhelming list. As I stated before, it just feels a little unfair that Militant heir gets this goal of gearing up to strive for, whereas Bookworm heir is apparently just expected to know its place as an uninfluential role.

So following that, I'd be willing to revise the proposal- perhaps after a discussion with RadiantFlash (I'm just assuming here that they are the one who would be ultimately deciding on this matter, though I could be wrong, I am not really familiar with how development works here) and/or other maintainers as to what spells would be acceptable, and then come back with something different that will hopefully leave me happier with the role that I think could use a little more love. Thanks for the time and consideration in the replies to this proposal here and on Discord!

RadiantFlash7:
> 
> Following discussions in the discord, I've since become a little more enlightened to just how much a careful balance is endeavored to be kept between the Keep and the other factions at play.

This ironically more then everything is in a constant pendellum with each change. It's a struggle for maintainers to balance what should be what, especially as players who play roles will always complain when their role gets nerfed, regardless of whether their complaint is legitimate or not.

> 
> That aside, I think it's a shame that RadiantFlash sees it as "Disney level theming," I think a noble interested in learning magic is perfectly reasonable for any level and theme of fantasy- dark or light, and that a royal successor that knows magic needn't be a princess frollicking in the meadows and attracting woodland creatures as a princess might. I think it can be taken seriously, and I think that giving them the capacity to learn spells would have lent a mechanical side into the subclass's background as an active studier of the arcyne.

The main reason I considered the magical heiress disney levels of themeing, is because the idea of magic within our lore, is that it takes many years to become an effective arcyne pracitioner. Usually studying underneath an accomplished/learned magos- and dedicated study at that.
For balance and gameplay reasons, mages get to learn spells throughout the week as time goes on. But for an heiress who has duties beyond studying magic, The idea of them being on par with apprentices and other mages of the sort, really doesn't stick super well with me.

> I'd still like the subclass a little expanded in some way, if not with the ability to customize and express character through explicit spell selection, through the addition of maybe one or two more spells on the already quite underwhelming list. As I stated before, it just feels a little unfair that Militant heir gets this goal of gearing up to strive for, whereas Bookworm heir is apparently just expected to know its place as an uninfluential role.

This I can understand. I honestly never looked particularly closely at the militant heir, theres plenty a chance It outshines bookworm heir due to my decision on it. But thats also because I wasn't particularly in charge of balancing it when it was added. A good option would be allowing the ability to pick differant spell loadout sets at round start as part of the loadout selection, sort of like how some roles can pick their tool or weapon.
> 
> So following that, I'd be willing to revise the proposal- perhaps after a discussion with RadiantFlash (I'm just assuming here that they are the one who would be ultimately deciding on this matter, though I could be wrong, I am not really familiar with how development works here) and/or other maintainers as to what spells would be acceptable, and then come back with something different that will hopefully leave me happier with the role that I think could use a little more love. Thanks for the time and consideration in the replies to this proposal here and on Discord!

Typically whichever maintainer assigns themselves to a PR is the one that has the final say on a subject. That said, I've been more or less curating what spells we have, how they work, the balance for them, and designed the entire mage gameplay loop. So if it's mage/spell related, odds are it'll be in my wheel house.

Purpleturtle3:
If anything militant heir should be nerfed.

AranyaPTH:
The idea of Bookworm heir being able to select a loadout selection of spells might not be awful? but it'd end up with one pick always being the Meta.

AukariBunn:
> It's not an oversight, bookworm heir not being able to learn spells is 100% intended

The developer of bookworms here
You're wrong, actually. The bookworm heir not being able to learn spells was a middleground that must've been changed after the mage loop is out. Im too lazy to find the convos in the dev channels, but I remember we've discussed that already.

AukariBunn:
> Mage heirs getting added, while not getting a way to learn more spells was a compromise I made at the time to allow people to have their fun.

As I remember, we agreed that they will be given their separate spell list for learning to maintain balance upon magic loop release.

GravyMaker:
I was thinking about this myself, and it seems to me that the ability to define separate spell lists would go a long way toward solving the problem - allow the bookworm heir to select from a "harmless" list of spells, and also potentially allow new spells to be restricted to specific roles in the future (eg. court magos-specific spells, say), whether or not that's something anyone wants to do right now.

If I were to submit a PR allowing role-specific spell lists and _this_ PR were reworked to make use of it, would anybody still object?

GravyMaker:
Well, my thoughts are that, if we can create separate spell lists and assign whichever we want to each class, we could have, say, a generic list, a combat-only list, a noncombat-only list... And various other lists if we wanted (eg. a lich-specific list with unique spells - just hypothetically.)

Alternatively, we could implement spell flags, so each spell is flagged as eg. combat, noncombat, necromancy, etc. allowing us to define _rules_ about which role gets what, rather than lists. But that means more initial work for relatively little gain, considering that updating the lists on the odd occasion a new spell is added isn't _that_ difficult.

RadiantFlash7:
Right. For now, I'll be closing this PR. Refer to the discussion above, but as it stands in this form it isn't going to be accepted.

PR #2646, opened 2025-03-30
Adds TRAIT_NO_SELF_UNEQUIP to items, and uses it in drop proc.

ARefrigerator:
## About The Pull Request
So this was originally meant to fix the cursed collar (it does that in part) but it also has introduced a new if tree in `/mob/proc/doUnEquip`
What does it do?

- TRAIT_NO_SELF_UNEQUIP is a new define for items explaining that you cannot self remove it once it is equiped

Adds the following tree to the proc `doUnEquip`
```
if(HAS_TRAIT(I, TRAIT_NO_SELF_UNEQUIP) && !force)
  if(istype(src, /mob/living/carbon/human))
    // Putting things into hands is a null loc ref why? no idea! But this works for now...
    // Because the only alternative is removing something from someone, which puts it on the floor.
    if(newloc == null) 
      var/mob/living/carbon/human/H = src
      for(var/Item in H.get_equipped_items()) // Find this item
        if(Item == I)
          to_chat(src, span_warn("I can't remove \the [I.name]!"))
          return FALSE
```

- Edits the cursed collar to use this new flag
- Removes the drop code for the cursed collar (it's ass???)

Turns out there is a distinction between removing an item yourself and someone else doing it: `newloc`. self removing something apparently makes newloc null (I assume because it doesn't really change position from you) while someone else removing things puts it onto the floor (newloc has a value!). This is a little hacky. I would have preferred to have an actual mob reference when this proc called but I assume this proc is 15 years old and I don't want to mess with its parameters unless I have to (I don't).

Also currently the only thing that would be using this is the cursed collar so any issues would be very localized to it until it expands. But it seems fine.

Tested but not shown in video: Removal on decapping (via guillotine, Works.)

## Why It's Good For The Game
The original code makes the collar stick to your hand, making you unable to use your hand, it breaks when you swap hands (this technically counts as a drop I guess), and the code it was trying to use was not designed for what it was trying to do. 

Also as an additional benefit: if this does turn out to work perfectly fine (I think it does) we could add other items that can only be removed by other people e.g. face masks to punish the person, heavy shoes (for slow down?) you get the idea.

https://github.com/user-attachments/assets/df3ff8db-1c57-4a30-9b94-a558fdc36123

### DIFF
--- a/code/__DEFINES/traits.dm
+++ b/code/__DEFINES/traits.dm
@@ -338,6 +338,7 @@ Remember to update _globalvars/traits.dm if you're adding/removing/renaming trai
 #define TRAIT_NODROP            "nodrop"
 #define TRAIT_NOEMBED			"noembed"
 #define TRAIT_NO_TELEPORT		"no-teleport" //you just can't
+#define TRAIT_NO_SELF_UNEQUIP			"no-self-unequip" // For items you can't remove on wearing but can still hold freely
 
 // common trait sources
 #define TRAIT_GENERIC "generic"
--- a/code/modules/clothing/rogueclothes/neck/neck.dm
+++ b/code/modules/clothing/rogueclothes/neck/neck.dm
@@ -377,17 +377,19 @@
 	. = ..()
 	name = "cursed collar"
 	resistance_flags = FIRE_PROOF
-	ADD_TRAIT(src, TRAIT_NODROP, CURSED_ITEM_TRAIT)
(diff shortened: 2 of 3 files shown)


ARefrigerator:
Sure. I'd have done it with this but it became such an ordeal that it essentially turned into its own feature. I am thinking of using it for a punishment mask that maybe the dungeon or keep can put on someone they want other's to know is bad.

Gyrandola (approved):

PR #2650, opened 2025-03-31
Makes feints work against dodge

Dilligentdog:
## About The Pull Request

Makes feinting work against dodge. Also slightly buffs feinting while on feint intent to make the immobilize 2 seconds instead of 1.2 seconds

## Why It's Good For The Game

Dodge based classes are notoriously bad to try and fight. They usually have more stamina than you and cannot be grabbed or attacked as they have a 90% chance to just dodge it. Now feints work exactly the same for if they are parrying or dodging

It was a single line change, nothing broke when tested.  
Proof it compiles
image

### DIFF
--- a/code/game/objects/items/rogueweapons/rmb_intents.dm
+++ b/code/game/objects/items/rogueweapons/rmb_intents.dm
@@ -32,8 +32,6 @@
 		
 
 
-	if(L.d_intent == INTENT_DODGE)
-		perc = 0
 	if(!L.cmode)
 		perc = 0
 	if(L.has_status_effect(/datum/status_effect/debuff/feinted))
@@ -45,8 +43,8 @@
 	if(prob(perc)) //feint intent increases the immobilize duration significantly
 		if(istype(user.rmb_intent, /datum/rmb_intent/feint))
 			L.apply_status_effect(/datum/status_effect/debuff/feinted)
-			L.changeNext_move(10)
-			L.Immobilize(12)
+			L.changeNext_move(20)
+			L.Immobilize(20)
 			to_chat(user, span_notice("[L] fell for my feint attack!"))
 			to_chat(L, span_danger("I fall for [user]'s feint attack!"))
 		else


Hyro-Fox:
💗

RadiantFlash7:
Probably an over due change, honestly. Dodging uses more stamina, but it can be very difficult for most players to deal with a dodge maxer. I'd be up for giving this a test merge.
( also like the buff to feinting in general)

ddsaw:
I remember buffing feint a lot months ago and it's still dogshit useless it seems. Maybe two seconds is not enough, considering I don't even notice I've been immobilized at the current 1.2.

RadiantFlash7 (approved):
This is pretty much ready for a test merge. I'd like to see how it plays out on server before a full merge.

ddsaw:
After codediving a bit regarding feints and how they work, feinting was/is supposed to disable parries for the duration of the feint debuff, not just immoblize.
You did not change it to disable dodges, only make it apply the feint debuff when the enemy is on dodgeintent and you successfully feint him, aswell as apply to demobilize.
In order to make this actually disable dodging for the time you will have to go into roguetownprocs, find the part of the horrible switch mess that handles feints disabling parries, and copy something similar to the horrible part that handles dodges.

Then again /datum/status_effect/debuff/feinted does not seem to fucking exist and feinting into nothing applies "curse of mundaneity" for some reason(maybe a fallback????) 
But feint behaviour does seem to function(this was annoying to test locally).

RadiantFlash7:
> After codediving a bit regarding feints and how they work, feinting was/is supposed to disable parries for the duration of the feint debuff, not just immoblize. You did not change it to disable dodges, only make it apply the feint debuff when the enemy is on dodgeintent and you successfully feint him, aswell as apply to demobilize. In order to make this actually disable dodging for the time you will have to go into roguetownprocs, find the part of the horrible switch mess that handles feints disabling parries, and copy something similar to the horrible part that handles dodges.
> 
This might be necessary, though holding a rogue in place is near as good.

> Then again /datum/status_effect/debuff/feinted does not seem to fucking exist and feinting into nothing applies "curse of mundaneity" for some reason(maybe a fallback????) But feint behaviour does seem to function(this was annoying to test locally).

https://github.com/Rotwood-Vale/Ratwood-Keep/blob/27d7c597a2878f2df43a6f2142fe2e47a80c2026/code/game/objects/items/rogueweapons/rmb_intents.dm#L87C1-L89C15 It exists here.

Actuallygoinginsanern:
after having played with this on both sides of the coin. I can pretty much say that this isn't a small nerf to dodging, but literally just kills it. Most builds that have dodge as their primary focus have low STR as a tradeoff.

So literally any feint that can freeze them in place for two seconds (100% grab success rate) will basically end combat for them.

Dodging needs to be nerfed, but this isn't a "nerf" to dodging, this is just killing it outright.

PR #2651, opened 2025-03-31
Makes mail notifcations more obvious, adds sound.

ARefrigerator:
## About The Pull Request
Reworks how you get notified about a letter (see video)

## Why It's Good For The Game
It gets lost in the sea of messages easily and the noise might help people realize more.

https://github.com/user-attachments/assets/9869844d-0491-41d7-a47f-c99f32998732


ARefrigerator:
Maybe. I think the color difference on the same line might be enough though. It's already pretty rare for any message to do it if people still complain I can do that in the future.

RadiantFlash7 (approved):
Looks good to me at the moment. More visible mail is always a good thing as long as it isn't obnoxiously large.

Gyrandola (approved):

PR #2656, opened 2025-04-01
Allows leaving via carriage and more...

ARefrigerator:
## About The Pull Request
What it does:

- Puts a door item on top of the carriage's door on new you can click on and examine.
- Clicking it gives you a prompt to leave
- Allows you to leave by clicking on the carriage, you enter after 5 seconds.
- Cannot leave if you are considered important jobs `(ruler, prince, consor, priest) `
- Checks for special items that should never be allowed to leave (master key,crown of rockhill, priest staff)
- Also checks the inventories to make sure you're not hiding it!
- leaving deletes the character and removes it from all the lists, leaves you a spirit (This was the easiest way to deal with a cooldown to keep it consistent)
- opens up the slot you took up and subclass (thanks Azure Peak)
- Logs who left and with what and what job to admins (So if some idiot abuses this you can call them out on it)

pls don't abuse this gamers :'(

## Why It's Good For The Game
People want to leave the round in a way that doesn't mean perma sleeping or killing themselves. This allows them to. I may need to adjust things going forward depending on what people do with this but I think we can trust most people to not abuse this. Especially considering how many people hate the underworld it would at least filter some people from trying to game this...

https://github.com/user-attachments/assets/4928354a-f6eb-43b2-b2cc-fc3fbe8d057b

### DIFF
--- a/code/modules/jobs/job_types/roguetown/garrison/bogmaster.dm
+++ b/code/modules/jobs/job_types/roguetown/garrison/bogmaster.dm
@@ -22,6 +22,7 @@
 	give_bank_account = 35
 	min_pq = 8
 	max_pq = null
+	can_leave_round = FALSE
 	cmode_music = 'sound/music/combat_bog.ogg'
 
 /datum/job/roguetown/bogmaster/after_spawn(mob/living/L, mob/M, latejoin = TRUE)
--- a/code/modules/jobs/job_types/roguetown/garrison/mayor.dm
+++ b/code/modules/jobs/job_types/roguetown/garrison/mayor.dm
@@ -16,6 +16,7 @@
 	display_order = JDO_MAYOR
 	min_pq = 5
 	max_pq = null
+	can_leave_round = FALSE
 	give_bank_account = 35
 
 	//cmode_music = 'sound/music/combat_bog.ogg'
--- a/code/modules/jobs/job_types/roguetown/garrison/sheriff.dm
+++ b/code/modules/jobs/job_types/roguetown/garrison/sheriff.dm
@@ -14,6 +14,7 @@
 	You rose up through the ranks as a watchman, and now rule over them - Ensure that they enforce the laws of this land properly. \
 	While you may work under the Mayor, everyone knows the Duke is the true authority, or is he?"
 	whitelist_req = FALSE
+	can_leave_round = FALSE
 	outfit = /datum/outfit/job/roguetown/sheriff
 	give_bank_account = 26
 	min_pq = 8
--- a/code/modules/jobs/job_types/roguetown/nobility/captain.dm
+++ b/code/modules/jobs/job_types/roguetown/nobility/captain.dm
@@ -24,6 +24,7 @@
 	give_bank_account = 26
 	min_pq = 8
 	max_pq = null
+	can_leave_round = FALSE
 
 	cmode_music = 'sound/music/combat_guard2.ogg'
 


code/modules/underworld/underworld.dm
```diff
@@ -39,6 +39,118 @@
...
+	else
+
+		var/datum/job/J = SSjob.name_occupations[user.job]
+		for(var/K in list(CANT_LEAVE_ROLES))
+			if(istype(J, K))
+				to_chat(user, span_warning("You are too important to leave!!"))
+				return
+
+		//Should this go somewhere??
+		var/list/no_leave_items = list(/obj/item/rogueweapon/woodstaff/aries, /obj/item/clothing/head/roguetown/crown, /obj/item/key/lord)
```
Gyrandola, line 78: #Define this

Gyrandola (changes requested):
Does this stop major antags or bandits from leaving?

ARefrigerator:
No it doesn't. I guess I could include to stop antags from leaving. I'll do the fixes to the code soon.

code/modules/underworld/underworld.dm
ARefrigerator: Defined it above the carriage door. If I can find a good place to drop it I'll move it there but it's just hanging for now.

ARefrigerator:
I'll get to adding antag checks soon so they can't just leave. Anything else that should be considered?

I know deline suggested the carriage should be down in the bog so you have to walk to it but right now people either close the window and walk away from the game or kill themselves so I think an IC way to leave a little more gracefully would be nice.

RadiantFlash7:
> I'll get to adding antag checks soon so they can't just leave. Anything else that should be considered?
> 
> I know deline suggested the carriage should be down in the bog so you have to walk to it but right now people either close the window and walk away from the game or kill themselves so I think an IC way to leave a little more gracefully would be nice.

We definitely do not want the spawn and exit point at the bog, no.

ARefrigerator:
I think I got it to stop most important antags from leaving. Had to make a define for it though. 
antag_roles

code/__DEFINES/roguetown.dm
```diff
@@ -167,7 +167,17 @@
 	/datum/species/goblin/moon,\
 )
 
-
+//I guess a list like this didn't exist until now?
```
noelle-lavenza, line 170: `special_roles_rogue` exists.

code/modules/underworld/underworld.dm
```diff
@@ -39,6 +39,130 @@
...
+			return
+
+	// No important antags can leave!
+	if(H.mind && H.mind.antag_datums)
+		var/datum/mind/M = H.mind
+		for(var/a in M.antag_datums)
+			for(var/b in ALL_ANTAG_TYPES)
+				if(istype(a, b)) 
+					to_chat(user, span_warning("I still have work to do here..."))
+					return
```
noelle-lavenza, line 93: My question for you: What antags *are* allowed to leave? Deadites? Or is that just an oversight?

code/modules/underworld/underworld.dm
```diff
@@ -39,6 +39,130 @@
...
+		if(SEND_SIGNAL(i, COMSIG_CONTAINS_STORAGE))
+			var/list/inv = list()
+
+			SEND_SIGNAL(i, COMSIG_TRY_STORAGE_RETURN_INVENTORY, inv, FALSE)
+
+			for(var/obj/item/L in inv)
+				for(var/M in list(CANT_LEAVE_ITEMS))
+					if(istype(L, M))
+						to_chat(user, span_notice("You can't leave with \the [L.name]!!"))
+						return
```
noelle-lavenza, line 114:
```suggestion
	var/list/user_items = H.get_all_gear()
	for(var/forbidden_type in list(CANT_LEAVE_ITEMS))
		var/obj/item/forbidden_item = locate(forbidden_type) in user_items
		if(forbidden_item)
			to_chat(user, span_notice("You can't leave with \the [forbidden_item]."))
			return
```

code/modules/underworld/underworld.dm
```diff
@@ -39,6 +39,130 @@
...
+					dat += " Nothing on them."
+				else
+					// WARNING: This doesn't currently list items in the storage
+					// But since we checked for shit that matters before already do we really care?
+					var/atom/movable/content = user.contents[1]
+					dat += " [content.name]"
+					for(var/i in 2 to length(user.contents))
+						content = user.contents[i]
+						dat += ", [content.name]"
+						dat += "."
```
noelle-lavenza, line 134:
```suggestion
				var/list/contained_items = user.get_all_gear()
				if(!length(contained_items))
					dat += " Nothing on them."
				else
					dat += " [english_list(user.get_all_gear())]"
```
This will list all items including those in storage.

code/modules/underworld/underworld.dm
```diff
@@ -39,6 +39,130 @@
...
+
+	var/mob/living/carbon/human/H
+	if(ishuman(user))
+		H = user
+
+	if(!isliving(user) || user.incapacitated())
+		return //No ghosts or incapacitated folk allowed to do this.
+
+	if(user.get_active_held_item())
+		to_chat(user, span_warn("I cant turn the handle. My hand has something in it!"))
```
noelle-lavenza, line 77:
```suggestion
		to_chat(user, span_warning("I can't turn the handle. My hand has something in it!"))
```
Also, is it even possible to use `attack_hand` with an item in your active hand? This should probably just be removed.

code/modules/underworld/underworld.dm
```diff
@@ -39,6 +39,130 @@
...
+		return
+
+	var/datum/job/J = SSjob.name_occupations[user.job]
+	for(var/K in list(CANT_LEAVE_ROLES))
+		if(istype(J, K))
+			to_chat(user, span_warning("You are too important to leave!!"))
+			return
```
noelle-lavenza, line 84: This should probably use a variable on the job definition rather than a hardcoded list.

code/__DEFINES/jobs.dm
```diff
@@ -342,6 +342,15 @@
...
 	/datum/job/roguetown/ruler
 
+
+// DUKE, CONSORT, HEIR, PRIEST
+#define CANT_LEAVE_ROLES \
+	/datum/job/roguetown/priest,\
+	/datum/job/roguetown/consort,\
+	/datum/job/roguetown/ruler,\
+	/datum/job/roguetown/prince,\
+	/datum/job/roguetown/hand
```
noelle-lavenza, line 352: Please define `can_carriage` or a similar variable on `/datum/job` or `/datum/job/roguetown` and check that instead of using this list.

code/modules/underworld/underworld.dm
noelle-lavenza: You should probably define a `can_carriage` var (or similar) on `/datum/antagonist` and define it on the relevant types, then check that.

noelle-lavenza (changes requested):
This PR needs some substantial revisions.

ARefrigerator:
Alright I'll work through all these issues and fix it.

code/__DEFINES/roguetown.dm
ARefrigerator: When I tried this list it wasn't allowing me to read it. But given the other suggestion for how to read this value I might end up removing it altogether anyway.

ARefrigerator:
Still have to work on the job datum part. Might conflict with the job PR being tested with now but I can add it in and fix it up later should anything occur.

ARefrigerator:
Sorry for the delays. I think I got this all up to date and it's finally working as I think everyone wants it to.

code/modules/underworld/underworld.dm
ARefrigerator: Made this check for length `(M.antag_datums)` Perhaps a little cheap but currently there's no known antag who should be leaving the round.

ARefrigerator:
This commit should fix the items being left behind here is my local test that shows cions in belt, keyring in hand in image 1 and image 2 showing nothing left behind.
image
image

Gyrandola (approved):

PR #2659, opened 2025-04-01
Allows cleaning remains away from Turfs.

ARefrigerator:
## About The Pull Request
Allows cleaning turfs to remove remain types.

## Why It's Good For The Game
You can't get rid of these things currently and they have no use as fast as I know. They can be an eye sore and it would make sense you could get rid of them by cleaning.

https://github.com/user-attachments/assets/01f21bc4-f722-42c9-a2ff-f8fd0abad38d

### DIFF
--- a/code/modules/spells/roguetown/wizard.dm
+++ b/code/modules/spells/roguetown/wizard.dm
@@ -702,7 +702,7 @@ Unless of course, they went heavy into the gameplay loop, and got a better book.
 	var/skill_level = user.mind?.get_skill_level(attached_spell.associated_skill)
 	cleanspeed = initial(cleanspeed) - (skill_level * 3) // 3 cleanspeed per skill level, from 35 down to a maximum of 17 (pretty quick)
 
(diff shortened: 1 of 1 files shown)


RadiantFlash7 (changes requested):
Hm.
I don't know if someone is planning on making it so you can get bones from remains, or any of the sort that this might mess up. But since no one that I know of is actively working on that, I don't have an issue with cleaning remains like so. 

Only one requested change, is to add this to prestidigiation here to keep all cleaning procs parallel:
https://github.com/Rotwood-Vale/Ratwood-Keep/blob/4eb54506db3cbf33f41a02dda5823c64ef6fff45/code/modules/spells/roguetown/wizard.dm#L698

ARefrigerator:
Wizard remains cleaning:

https://github.com/user-attachments/assets/86a05518-4723-4361-b4f2-39fc1e4ef17a

RadiantFlash7 (approved):
Looks good, thank you for the changes.

PR #2663, opened 2025-04-02
Removes TRAIT_NOSTAM from skeletons and converted peoples.

ARefrigerator:
## About The Pull Request
Removes the infinite stamina on skeletons (and hopefully the people converted into skeletons)

## Why It's Good For The Game
Infinite stamina seems fine on the head antagonists but there are complaints that on skeletons it is often too difficult to deal with especially if the enemy somehow gets the swift intent to stab.

I'm not too familiar with the change and what it means but I've seen about 3 people ask for it and it was simple enough to do it.

https://github.com/user-attachments/assets/3dc5ae3d-4703-4af8-8b39-b538ce6b85be


ARefrigerator:
Apparently skeletons lost their swift intent? I think that was originally the root issue. If so feel free to close this.

RadiantFlash7:
> Apparently skeletons lost their swift intent? I think that was originally the root issue. If so feel free to close this.

I believe theres an exploit where if the player is on swift intent before getting skeletoned, they keep swift intent.

As an aside: I think there miight be an issue with skeletoned guard roles. I'm fairly certain they keep their stats and skills from their prior job? So this means you can have really high Con skeletons which can be very hard to deal with when skeletons only die from decapitation or certain critical hits.

Purpleturtle3:
> > Apparently skeletons lost their swift intent? I think that was originally the root issue. If so feel free to close this.
> 
> I believe theres an exploit where if the player is on swift intent before getting skeletoned, they keep swift intent.
> 
> As an aside: I think there miight be an issue with skeletoned guard roles. I'm fairly certain they keep their stats and skills from their prior job? So this means you can have really high Con skeletons which can be very hard to deal with when skeletons only die from decapitation or certain critical hits. On top of guard role level wrestling being used as skeletons with infinite stamina.

Skeletons that were people keep all the intents they had, and they also keep their stats while gaining all the skeleton stuff skills, so the tabaxi longshoreman or whatever can use maces now on top of having inifite stam and keeping all their stats

RadiantFlash7:
Personally I think we should clear their skills/ stats and set it universally for raised undead. This stops horribly broken things like watchmen and vanguard skeletons infinite wrestling people, due to str stat and wrestling skill from their job being applied with no stam

ARefrigerator:
Alright I'll look into that instead.

PR #2664, opened 2025-04-02
Adds sally port to keep.

ARefrigerator:
## About The Pull Request

- Adds a door in the west part of the entrance so people can enter and leave while also able to keep the gate closed. 
- Changes gatehouse lockid to manor

## Why It's Good For The Game
Sometimes the gatemaster isn't around or is suicidal enough to not do their job. This would 1. allow someone else to easily do it (a servant perhaps) and 2. allow traffic of manor jobs to flow while the main gate remained inaccessible to common folk.

It might eventually need to go somewhere else (Could add one of those alarms if someone not allowed enters the area?) or worked a bit if people hate the 2 door system but it allows it to be at least a little more secure.

https://github.com/user-attachments/assets/a0c0f2ce-1bcd-43e6-8763-199c7cd9de26

### DIFF
--- a/_maps/map_files/rockhill/rockhill.dmm
+++ b/_maps/map_files/rockhill/rockhill.dmm
@@ -4315,11 +4315,12 @@
 /turf/open/water/swamp,
 /area/rogue/outdoors/town/harbor)
 "cls" = (
-/obj/structure/mineral_door/wood{
-	icon_state = "wcv";
+/obj/structure/mineral_door/wood/donjon{
+	dir = 1;
 	locked = 1;
 	lockid = "manor";
-	name = "Sally port"
+	name = "sally port";
+	desc = "sally port"
 	},
 /turf/open/floor/rogue/cobble,
 /area/rogue/outdoors/town)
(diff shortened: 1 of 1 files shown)


RadiantFlash7 (changes requested):
Glancing this over, a few things. If we're adding a sally port, Make sure it's dungeon-doors that can slide the slidable viewing port from the inside. These doors can be manor id'd. However, The gatehouse proper should still be kept at gatehouse id'd.

ARefrigerator:
Alright I made the changes.


https://github.com/user-attachments/assets/9fd4a772-a782-4cad-8f9d-072adfae5517

RadiantFlash7 (approved):
Perfect, Thank you.

Xaytan:
Make sure these have a good lock on them. Don't want it to be too trivial to lockpick your way into the keep.

ARefrigerator:
is that related to lock id or a different var?

ARefrigerator:
It currently has a lockdiff of2 which is equivalent to the knight's quarters. I think it's fine (considering you'd have to go through 2 doors) but if that's too low I could raise it up later.

E231kei500bandai:
this would basically defeat the gatemasters job as i imagine 90 percent of players would choose to just use this system instead. which is not that bad since gatemaster is kinda pointless and does nothing but inconvience you for a second.

PR #2678, opened 2025-04-05
Remove Cavernous Wombs (Mating Press Fix)

josephpatrickgreen:
## About The Pull Request
fixes glitch which inserts one character into another during mating press.


Gyrandola:
Closing this as its based on our pre-revert situation. If someone wants to bring back mating, they have to PR the whole code again (fixed).

PR #2682, opened 2025-04-06
Church role skill adjustment + Shovels using axe skill

Hmeister-real:
## About The Pull Request

 Morticians right now get 4 unarmed, 5 wrestling and 3 daggers. Monks get 4-4 in those skills and rogue stalker gets 3 daggers, on top of having blessings and having church protection if they get into trouble. They are busted for absolutely no reason. This lowers their Unarmed to 2, Wrestling to 3, Daggers to 1 and Axes to 2, also making shovels use axe skill. 

Also makes Acolytes get 3-2 and Druids get 3-3

## Why It's Good For The Game

Bad for balance

https://github.com/user-attachments/assets/1f0ad5c6-3443-4003-9353-b1877d6616ed


Hmeister-real:
> Unironically though as pointed out this majorly nerfs the Mortician compared to the rest of the church 

Adjusted the rest, Druids getting 3-3 since they actually get nothing

Purpleturtle3:
Just reduce wrestling by one and str by one if you want to nerf it instead of this straight up murder lol

RadiantFlash7:
Fundamentally I don't think this is an issue with the church's wrestling skill, nor there unarmed one.
All of these roles can't wear armor. The best they can use is a katar for parrying and attacking.

If anything, the issue lies fundamentally with wrestling at the moment. A better way to handle wrestling is make it only be able to tackle someone, they aren't holding non-large weapon or shield.

On top of that, the church serves a fairly important purpose in being both the spiritual center, as well as the healing center of the duchy when you have tons of heretics who are most likely more then happy to kill them. Both antagonists, and towner heretics.

For the moment, I'm leaning towards closing this PR.

Heheniy:
there is no such things as 'tons of heretics', and church always allied to keep and fights together with them

mortician being able to rotate and pin werewolf with one hand as it is some little puppy is NOT ok

RadiantFlash7:
> there is no such things as 'tons of heretics', and church always allied to keep and fights together with them
> 
> mortician being able to rotate and pin werewolf with one hand as it is some little puppy is NOT ok

Your average mortician is not wrestling a werewolf without getting decapitated.

DelineFortune:
Thanks but no
Acolytes's wrestling was always their defence, they have no armor, skills of stats to self defence 
Not our problem, that Neon turned wrestling into op thing in his salt
(You were supposed to grab chest of your enemy with both hands to tackle them and pin down, now it's one)

PR #2672, opened 2025-04-04
Adds icon to examine text of atoms.

ARefrigerator:
## About The Pull Request
Adds a little icon to the text box

## Why It's Good For The Game
I just think it's neat. Also might help a little in figuring out what you examine?

## Proof of Testing (Required)
image

Increases with the font size too!
image

PR #2684, opened 2025-04-06
Revert "Adds icon to examine text of atoms."

ARefrigerator:
Reverts Rotwood-Vale/Ratwood-Keep#2672


ARefrigerator:
You could try test merging it first to find out if it's the culprit of the examine delays (it might be). The only other culprit I can think of is admin music spam that chugs the server. I guess this would be a good way to know for sure.

Gyrandola (approved):

PR #2662, opened 2025-04-02
Physician roles, Acolyte, Hospitaller, Revival, Derot Overhaul

RadiantFlash7:
## About The Pull Request
This PR is meant to handle the current issues regarding physicians, acolytes, pestrans, revivals, and rot. It contains the following contains the following:

- Farming and climbing removed from towner physician.
- Removes all 'spells' from ALL physician roles with the exception of secular diagnosis.
- Removes 'De-rot' surgery.
- Adds recipes for 'embalming fluid' used to reset the rot timer when given to someone (Essence of honey and essence of earth). Toxic to living beings. 
- Physician self aid kits given embalming fluid by default, replacing medical book.

- Replaces CPR spell with a revival surgery. Ingredient to revive called 'lux' has to be harvested from living humans, (including deadites), and applied during a surgery to a corpse to revive. Harvesting lux causes stat penalties to those harvested from and effectively removes one of that players revives. (Borrowed from Azure peak and modified)

- Adds 'Hospitaller' role. A dedicated pestran role, like Dendor's druids and necra's  morticians. Hospitallers have higher treatment then acolytes, less holy skill and much less wrestling. They do however, have apprentice polearms. (Hopefully better sprites for their outfit pending.)

- Removes Cheele and Pestra's needle from surgical kits. Holy mother of power creep.
- Court physician granted it's needle of pestra once more in it's bag, 2 lux's in their chambers.
- Tweaks Eoran spells: Now actually cost devotion to use, and the Eoran Charm no longer hard stuns. Only pacification.
- Removes Miracle from Pestran. 

- Grants Pestrans Pestra's Mercy: A spell that requires 'channeling' next to the target. Target is given no-die trait as long as miracle is channeled, and oxyloss is slowly removed from target, with a minimum of blood added. Channeling the miracle slowly drains devotion. Other actions can not be taken while in use.

- Reworks Lesser miracle: Lesser Miracle now works as an 'over time' heal. Affects by various 'conditions' dependant on patron (Borrowed from Azure peak and modified).
- Removes Miracle from Astratans

- Grants Astratans Smite Unfaithful: A miracle that activates instantly on screen wide effects. Targets all excommunicated and apostates, setting them on fire, providing a chance to stun, and granting a debuff. This spell is twice as strong in the day time,  and during day time it will affect the undead including vampires, deadites, and skeletons.

[...]

PR #2685, opened 2025-04-07
Leech application speed fix

GravyMaker:
## About The Pull Request

This PR fixes two issues affecting medicine:
1) Detox and snake bite (in injector form) cannot be applied to targets which are immobile. This is due to an argument mismatch in the proc definition for attacking with medipens, which causes them to incorrectly check the target's movement flags (to determine, for example, that they're not paralysed) rather than the user's. This PR adjusts the proc's signature to include the expected parameters in the expected order, and updates the subsequent call to inject() to properly make use of them, allowing these reagents to function as intended. This shouldn't affect anything other than medipen-style drugs.
2) Leeches, when applied to a target other than the user, use the target's Treatment skill to determine application time, rather than the user's Treatment skill as one would expect; this means that applying a leech to a physician, for example, would be quicker than applying one to a soilson. This PR modifies the code to use the user's treatment skill instead, so the time is based purely on the user's skills rather than the target's.

## Why It's Good For The Game

I play a lot of Acolyte, and of late I've had several occasions where a poisoning/overdose victim has been brought to the church who wasn't able to be treated with detox due to this issue. While doing so works when the other player is conscious and mobile, in cases of poisoning or overdose, the problem is often only apparent to others once the victim has lost consciousness - meaning that, by the time its necessity is clear, it's become useless due to this bug. The fix returns it to its rightful state as a key tool against poisoning both in yourself and in others you're willing to use it on, even after they've passed out.

The leech issue, meanwhile, is simply unintuitive, and creates an unusual situation where, mechanically, your likelihood of being saved from poison is partly influenced by your own medical knowledge, rather than exclusively on your physician's. It doesn't matter too much - but as you can see in the GIFs below, the application time varies quite appreciably between no treatment and Legendary treatment, which would mean some roles would receive less timely aid than others; prior to this PR, these times would be based on the _victim's_ treatment, meaning a physician and, say, a soilson would take very different lengths of time to apply a leech to, even being treated by the same physician.

These issues are of particular importance given the possible removal of physician spells in PR #2662 - following such a change, the main alternative to Detox for rapidly curing another person of poisoning/overdose would be removed, and so it's critical that detox function correctly when used, whether by the victim or by somebody nearby _on_ the victim. The leech fix isn't quite as important - but nevertheless could potentially make a difference in a life-or-death poisoning scenario without access to the current physician spells.

Image showing me looming ominously over two debug-spawned unconscious (mindless) characters, snake bite in hand.
image

Image after having used the injector on the southernmost character. Note the injection text on the right, used state of the snake bite injector, and the lack of visual effects and buff indicators (indicating I haven't injected myself).
image

Image after having injected the detox into the northernmost character.
image

Image after injecting myself with snake bite. Note the visual effect and buff.
image

Applying a leech with no (0) treatment. (Note that cheeles use the exact same application time as leeches at all treatment levels.)
dreamseeker_zP91AzgN8n

Applying a leech with Expert (4) treatment.
dreamseeker_8Fy5VF9kDq

Applying a leech with Legendary (6) treatment.
dreamseeker_D9SP1osFQB

Finally, I don't have screenshots for this but I've also verified that leeches are correctly applied to the legs from time to time while walking through murk.

### DIFF
--- a/code/modules/reagents/reagent_containers/hypospray.dm
+++ b/code/modules/reagents/reagent_containers/hypospray.dm
@@ -144,12 +144,12 @@
 	playsound(src, 'modular/Smoker/sound/inject.ogg', 100, TRUE)
 	return TRUE
 
-/obj/item/reagent_containers/hypospray/medipen/sty/attack(mob/living/M, mob/user)
+/obj/item/reagent_containers/hypospray/medipen/sty/attack(mob/user)
 	if(user.canUseTopic(src, BE_CLOSE, FALSE, NO_TK))
 		if(has_cap)
 			to_chat(user, span_warning("[src] has a cap still! You need to remove it first."))
 			return
-		inject(M, user)
+		inject(user, user)
 
 /obj/item/reagent_containers/hypospray/medipen/sty/detox
 	name = "DETOX"


RadiantFlash7:
I will say; I completely rewrote the injection code for Mori's medical items including stys and bottles in #2662 Per administration request. So I'm afraid half this PR is a touch unnessicary >_>;

That said- the leech changes are good.

GravyMaker:
I can remove those changes, then, leaving just the leech changes, if you'd like?

RadiantFlash7:
> I can remove those changes, then, leaving just the leech changes, if you'd like?

That would preferable!

GravyMaker:
That file should now have been reverted to its old state. I've also renamed the PR accordingly.

code/modules/roguetown/roguejobs/fisher/leeches.dm
```diff
@@ -171,7 +171,7 @@
 		if(!get_location_accessible(H, check_zone(user.zone_selected)))
 			to_chat(user, span_warning("Something in the way."))
 			return
-		var/used_time = (70 - (H.mind.get_skill_level(/datum/skill/misc/treatment) * 10))/2
+		var/used_time = (70 - (user.mind.get_skill_level(/datum/skill/misc/treatment) * 10))/2
```
noelle-lavenza, line 174:
```suggestion
		var/used_time = 3.5 SECONDS - user.mind.get_skill_level(/datum/skill/misc/treatment)/2 SECONDS
```
While you're at it, it'd be nice if you made it clearer what the calculation is. You could also do `(3.5 - user.mind.get_skill_level(/datum/skill/misc/treatment)/2) SECONDS` instead if you think that's clearer.

RadiantFlash7: ...This winds up being a differant calculation then the one they have made, I think.

RadiantFlash7 (approved):
Leech tweaks application tweaks approved. Thank you.

code/modules/roguetown/roguejobs/fisher/leeches.dm
noelle-lavenza: It shouldn't be. `(70 - (treatment skill * 10))/2` is equivalent.
1. Through the distributive property, you get `70/2 - treatment_skill*10/2`.
2. Through substitution, we replace `*10` with `SECONDS`. `70/2 - treatment_skill SECONDS / 2`
3. Through the commutative property we reorder the second term. `70/2 - treatment_skill/2 SECONDS` (note that this is `treatment_skill / 2 * 10`, not `treatment_skill / (2 * 10)`.)
4. Through arithmetic we get `35 - treatment_skill / 2 SECONDS`.
5. Representing 35 in terms of seconds, we can use `3.5 SECONDS` instead of `35`.
So, if you want to make it clearer you can use parentheses for `(treatment_skill / 2) SECONDS`, but by order-of-operations rules they aren't strictly necessary.

PR #2690, opened 2025-04-08
Population locking Autostockpile

RadiantFlash7:
## About The Pull Request
Stops auto-stockpiling from occuring after 25 players are on.

## Why It's Good For The Game
With the rate at which things are used from the stockpile, theres very little reason for anyone to actually be /needed/ as towner roles most of the time, unless you're going hard into for example, smithing, or leatherworking. It will also reduce the wild amount of mammons some stewards can make by selling stuff thats auto-stockpiled for free.

Little hard to show this given pop require ment. However, I can show that lesser pop still autostockpiles easily enough:
image
image

### DIFF
--- a/code/controllers/subsystem/rogue/treasury.dm
+++ b/code/controllers/subsystem/rogue/treasury.dm
@@ -60,9 +60,10 @@ SUBSYSTEM_DEF(treasury)
 						X.demand += rand(5,15)
 					if(X.demand > initial(X.demand))
 						X.demand -= rand(5,15)
-		for(var/datum/roguestock/stockpile/A in stockpile_datums) //Generate some remote resources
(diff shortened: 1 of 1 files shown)


DelineFortune:
Autostockpile makes hunter/miners/farmers out of game in my opinion 

Instead of gathering rare resources you can just wait few minutes and buy it
(The difference for fur is +5 mammons, for real?)
You can also sell it as a steward and generate ~400 mammons per day

Gyrandola:
Wasn't this a thing already? I remember it being talked about way back

RadiantFlash7:
> Wasn't this a thing already? I remember it being talked about way back

You made this exact PR, but closed it when you left in January I believe.

Gyrandola (approved):

PR #2693, opened 2025-04-08
Glasses now actually helps

Heheniy:
## About The Pull Request
If you have perception < 5, spectacles will actually help you

## Why It's Good For The Game
Currently spectacles are totally useless apart if you have bad sight flav
This pr makes them actually useful if you got really shitty perception (ie revive debuff)

image
image
image

### DIFF
--- a/code/modules/clothing/rogueclothes/mask/mask.dm
+++ b/code/modules/clothing/rogueclothes/mask/mask.dm
@@ -48,18 +48,15 @@
 
 /obj/item/clothing/mask/rogue/spectacles/equipped(mob/living/carbon/human/user, slot)
 	. = ..()
-	if(!user.has_flaw(/datum/charflaw/badsight))
-		if(slot == SLOT_WEAR_MASK && user.STAPER < 5)
-			user.STAPER += 4
-			cured_eyesight = TRUE
-	user.update_fov_angles()
+	if(!user.has_flaw(/datum/charflaw/badsight) && slot == SLOT_WEAR_MASK && user.STAPER < 5)
+		user.change_stat("perception", 4, "spectacles")
+		cured_eyesight = TRUE
 
 /obj/item/clothing/mask/rogue/spectacles/dropped(mob/living/carbon/human/user)
 	. = ..()
 	if(cured_eyesight)
-		user.STAPER -= 4
+		user.change_stat("perception", 0, "spectacles")
 		cured_eyesight = FALSE
-	user.update_fov_angles()
 
 /obj/item/clothing/mask/rogue/spectacles/delf
 	name = "dark elf sunshields"


Gyrandola:
This assumes PERCEPTION is only tied to eyesight, what if someone has low PER because they just naturally aren't very attentive for whatever reason. Why would wearing glasses help them? I don't think this is good.

RadiantFlash7:
I don't think spectacles should provide perception inherently- not everyone has bad eyesight.
That said, I'm perfectly fine with them providing a +1 to perception IF they have bad eyesight, like how you get a +1 to endurance from masochist.

Heheniy:
they are providing buff only if you have <5 PER, giving maximum possible of 8 PER you can get with them, if you have 4 PER without glasses
and if you have bad sight flav glasses just prevents you from getting -10 PER and -3 SPD debuff

noelle-lavenza (changes requested):
I would really suggest just checking for equipped spectacles in `/mob/proc/update_fov_angles()` instead of giving a PER buff, and then calling `update_fov_angles()` on equip/unequip. You could even use a trait... If you don't want to do it, I'll do it for you.

Heheniy:
i want glasses actually help people seeing, not just remove fov, thats why i added buff

RadiantFlash7:
...This should be fine then. I don't know what scenarios a person has less then 5 PER without the bad sight flaw though; But it should be okay to add in.

code/modules/clothing/rogueclothes/mask/mask.dm
```diff
@@ -40,6 +40,20 @@
...
+			user.STAPER += 4
+			cured_eyesight = TRUE
+	user.update_fov_angles()
+
+/obj/item/clothing/mask/rogue/spectacles/dropped(mob/living/carbon/human/user)
+	. = ..()
+	if(cured_eyesight)
+		user.STAPER -= 4
+		cured_eyesight = FALSE
+	user.update_fov_angles()
```
noelle-lavenza, line 56:
```suggestion
			user.change_stat("perception", 0, "spectacles")
```
this just sets the spectacles bonus to 0 (undoes the existing bonus), and will call `update_fov_angles()`. it is totally safe if the bonus has not been given, and means you don't need the `cured_eyesight` var at all

code/modules/clothing/rogueclothes/mask/mask.dm
```diff
@@ -17,7 +17,7 @@
 	resistance_flags = FIRE_PROOF
 	body_parts_covered = EYES
 	anvilrepair = /datum/skill/craft/blacksmithing
-//	block2add = FOV_BEHIND
+	var/cured_eyesight = FALSE
```
noelle-lavenza, line 20: with the other changes i've suggested this var should not be necessary
```suggestion
```

code/modules/clothing/rogueclothes/mask/mask.dm
```diff
@@ -40,6 +40,20 @@
 		take_damage(11, BRUTE, "blunt", 1)
 	..()
 
+/obj/item/clothing/mask/rogue/spectacles/equipped(mob/living/carbon/human/user, slot)
+	. = ..()
+	if(!user.has_flaw(/datum/charflaw/badsight))
+		if(slot == SLOT_WEAR_MASK && user.STAPER < 5)
+			user.STAPER += 4
+			cured_eyesight = TRUE
+	user.update_fov_angles()
```
noelle-lavenza, line 49:
```suggestion
	if(!user.has_flaw(/datum/charflaw/badsight) && slot == SLOT_WEAR_MASK && user.STAPER < 5)
		user.change_stat("perception", 4, "spectacles")
```
this is safer, and `change_stat("perception")` already includes the `update_fov_angles()` call. also this can be written with only one condition

noelle-lavenza (changes requested):
alright, if maints are okay with it just giving a PER bonus, then here's the stuff you need to change to do it right.

code/modules/clothing/rogueclothes/mask/mask.dm
Heheniy: still need it for proper stat change in case user has multiple spectacles

noelle-lavenza: Yeah, you're right.

noelle-lavenza (approved):

RadiantFlash7 (approved):
Again, I'm... not actually sure what point this PR will ever be used, given having less then 5 PER is- almost impossible without the badsight flaw. But, I'm fine with having it as an edge case in the event it somehow comes up.

PR #2694, opened 2025-04-08
Lowers second death stats debuffs

Heheniy:
## About The Pull Request
Lowers second death stats debuffs from -3 to -1 (int -2)

## Why It's Good For The Game
Debuff on second death does not replace debuff on first death, which gave you -5 to all stats (-3 to int) if you die twice, which is making the rest your round almost unplayeble as your stats would be literally worse that admin punishment or leprosy
image

### DIFF
--- a/code/datums/status_effects/debuffs.dm
+++ b/code/datums/status_effects/debuffs.dm
@@ -149,7 +149,7 @@
 	id = "death_embrace"
 	alert_type = null
 	examine_text = span_notice("They appear not entirely whole, as if some part of them was left behind.")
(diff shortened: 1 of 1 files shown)


Gyrandola:
Isn't that the point? Death is already forgiving enough as it is IMO.

RadiantFlash7:
We don't really need to reduce death penalties at the moment I think. That said, this would conflict with the heavy tweaks that the physician Derot revive PR has. Once thats been fully test merged; we can return to potentially reducing some of the stat penalties; though I'm heavily skeptical about doing so.

PR #2695, opened 2025-04-08
10 players to 1 bandit

Rayforx:
## About The Pull Request
The bandits are pretty strong, and this feels like the best way to nerf them when they have a lot more tools to do their job with the new hoardmaster changes.

This means, that instead of 8 bandits being spawned at 72 players
Then they will become the maximum number of bandits at 80 players.

## Why It's Good For The Game
A lot less bandit shenanigans, when there is nobody to react to it.

Number changes, with the code being changed before, also, hard to spawn in 80 clients into a testing server.


RadiantFlash7:
I will point out there is a pending bandit nerf coming after the hoardmaster changes are finalized.
I'm not sure reducing the amount of bandits will be needed ontop of that.

Rayforx:
I'll close this, it's good for reference later on.

PR #2696, opened 2025-04-09
Adds the blacksmith mod menu.

ARefrigerator:
## About The Pull Request
This should work now.

- Makes the menu for blacksmith mods into a proc
- uses item right clicking to determine to name things.
- Checks if it's at full health and on an anvil to name it.

## Why It's Good For The Game
People want it, and it will be expanded in time to allow modifying an assortment of items.

image


ARefrigerator:
hang on it literally broke as I made a PR...

ARefrigerator:
Okay I think I got it. Also saw some other issues right clicking some items gave so I fixed it. Item modding should only work on anvils. I might eventually get the other functionality these items have to the blacksmith tool. In fact maybe I'll eventually make that proc generic since it probably can be (aka devoid of needing to be embedded in an item)

code/game/objects/items/rogueweapons/melee/blunt.dm
```diff
@@ -27,6 +27,8 @@
 	blade_dulling = DULLING_BASHCHOP
 
 /obj/item/rogueweapon/mace/attack_right(mob/user)
+	if(locate(/obj/machinery/anvil) in (loc))
+		return ..()
```
Gyrandola, line 31: Whats the reason why this cannot be put in the parent type?

ARefrigerator: This seems to be some special code specifically for this object to change its sprite when you right click it. Only some of them have it I believe.

Gyrandola (approved):

PR #2700, opened 2025-04-10
Adjusts cage speeds, message, and examine

ARefrigerator:
## About The Pull Request

- Changes the examine to tell players when something is inside it.
- Lowers the time it takes to catch something dramatically
- Changes the message when harvesting to be more generic

## Why It's Good For The Game
No one uses fishing cages, they take forever to get anything out of them and if you do catch something you could get junk. The entire fishing cycle is currently designed around multiple attempts to roll for something. Until that changes the speed these work needs to reflect that. 

These are also the only way to get lobsters and crabs!! Make them faster!!!

image

### DIFF
--- a/code/game/objects/items/rogueitems/fishing_cage.dm
+++ b/code/game/objects/items/rogueitems/fishing_cage.dm
@@ -10,7 +10,9 @@
 	var/deployed = 0
 	var/obj/item/caught
 	var/obj/item/bait
-	var/time2catch = 20 MINUTES
+	// Until crabs and lobsters can be caught in other ways
+	// This taking any length of time is insane
+	var/time2catch = 20 SECONDS 
 
 /obj/item/fishingcage/attack_self(mob/user)
 	. = ..()
@@ -22,7 +24,7 @@
 
 	user.visible_message(span_notice("[user] begins deploying the fishing cage..."), \
 						span_notice("I begin deploying the fishing cage..."))
-	var/deploy_speed = 15 SECONDS - (user.mind.get_skill_level(/datum/skill/labor/fishing) * 2 SECONDS)
+	var/deploy_speed = get_skill_delay(user.mind.get_skill_level(/datum/skill/labor/fishing), 1, slowest = 6) //in seconds
 
 	if(!is_valid_fishing_spot(T))
 		to_chat(user, span_warning("This body of water seems devoid of aquatic life..."))
@@ -40,10 +42,10 @@
 
 /obj/item/fishingcage/attack_hand(mob/user)
 	if(deployed)
-		var/deploy_speed = 15 SECONDS - (user.mind.get_skill_level(/datum/skill/labor/fishing) * 2 SECONDS)
+		var/deploy_speed = get_skill_delay(user.mind.get_skill_level(/datum/skill/labor/fishing), 1, slowest = 6) //in seconds
 		if(caught)
-			user.visible_message(span_notice("[user] begins to harvest the shellfish from the cage..."), \
-								span_notice("I begin harvesting the shellfish from the cage..."))
+			user.visible_message(span_notice("[user] begins to harvest from the cage..."), \
+								span_notice("I begin harvesting the from the cage..."))
 			if(do_after(user, deploy_speed, target = src))
 				STOP_PROCESSING(SSobj, src)
 				icon_state = "fishingcage_deployed"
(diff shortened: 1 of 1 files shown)


Genevesh:
Can we get the exact values of how much the time was reduced per fishing level?

ARefrigerator:
> Can we get the exact values of how much the time was reduced per fishing level?

Master level would be 5 seconds on a cage, no skill is 40 seconds. Which means a journeyman is somewhere in the middle of that at about 22 seconds? it uses a LERP and the values are between 0-6 I think.

Gyrandola (approved):

JohnnyJohnJohnson:
> > Can we get the exact values of how much the time was reduced per fishing level?
> 
> Master level would be 5 seconds on a cage, no skill is 40 seconds. Which means a journeyman is somewhere in the middle of that at about 22 seconds? it uses a LERP and the values are between 0-6 I think.

Bruh, the point of it taking a while if you have no skill is to prevent people with no fishing skill from just spamming these and getting lots of fish which invalidates the point of there being a Fisherman Towner class. The rates of catching stuff before were fine, the skill level that Fisherman spawns with makes it only take about 3 or 5 minutes to get a fish, and 90 seconds at Legendary level is already busted as shit if you have five cages.

JohnnyJohnJohnson:
> > > Can we get the exact values of how much the time was reduced per fishing level?
> > 
> > 
> > Master level would be 5 seconds on a cage, no skill is 40 seconds. Which means a journeyman is somewhere in the middle of that at about 22 seconds? it uses a LERP and the values are between 0-6 I think.
> 
> Bruh, the point of it taking a while if you have no skill is to prevent people with no fishing skill from just spamming these and getting lots of fish which invalidates the point of there being a Fisherman Towner class. The rates of catching stuff before were fine, the skill level that Fisherman spawns with makes it only take about 3 or 5 minutes to get a fish, and 90 seconds at Legendary level is already busted as shit if you have five cages.

Not to mention that fishing cages are meant to be a PASSIVE way of getting more fish, you can fish with the rod while you wait for the cages to catch something, you don't have to use just cages. Or, bait the cages, walk away, and go ACTUALLY ROLEPLAY with people or just do whatever else. The point isn't that they are speedy, but they ALWAYS catch something, the fish never gets away unlike with a fishing rod. This changes makes fishing cages way more OP than a fishing rod now. PLEASE revert this.

ARefrigerator:
Then I have to change what cages can catch if I increase the timer again because waiting 20 minutes to come back to an arrow was making me never even bother with cages.

JohnnyJohnJohnson:
> Then I have to change what cages can catch if I increase the timer again because waiting 20 minutes to come back to an arrow was making me never even bother with cages.

Or don't use cages if your character sucks at fishing. Use a rod to practice, that's much faster. The long wait time for having no skill was intentional so non-fishermen couldn't invalidate actual fishermen since cages never lose what they catch. Why the fuck did Gyran approve this?

Genevesh:
Yeah I can't say I like this change either; Not sure why it was speedmerged again

ARefrigerator:
> > Then I have to change what cages can catch if I increase the timer again because waiting 20 minutes to come back to an arrow was making me never even bother with cages.
> 
> Or don't use cages if your character sucks at fishing. Use a rod to practice, that's much faster. The long wait time for having no skill was intentional so non-fishermen couldn't invalidate actual fishermen since cages never lose what they catch. Why the fuck did Gyran approve this?
> 
> But I will say, cages are supposed to have a much lower chance of catching junk so either you've been unlucky or the values already weren't correct, in which case thank you for discovering this problem. But I promise the time to catch was intentional and thought-out.

Alright I can try looking more into it to see if maybe it makes more sense if it guarantees a catch but making it take longer again.

JohnnyJohnJohnson:
Yeah, cages are supposed to catch way less non-fish items than fishing rods. They're meant to be reliable, but slower than a rod. You'll get less jewelry and junk, they're mainly for food. They also provide people that don't know fishing a reliable, but slow way to level up their skill, but it's SUPPOSED to take long without skill because there is zero chance of failing to catch something. You are removing the risk factor in exchange for it taking longer, that's the entire point of the cages. If someone with no skill needs less than one minute to catch something with a cage, then what's the point in Fisherman being it's own Towner class?

ARefrigerator:
I mean anyone can do that with a rod...

JohnnyJohnJohnson:
> I mean anyone can do that with a rod...

With a rod you'll fail a lot and have to do it one at a time. Cages can be set up en-masse, hence the longer time to catch, that's why it was already fair.

PR #2704, opened 2025-04-10
Removes simple mob crits.

ARefrigerator:
## About The Pull Request
Removes the simple mob ability to crit.

## Why It's Good For The Game
This is the one thing about simple mobs people hate. Their crit rate seems absurdly high and this would solve it. Video below showing what it looks like.

https://github.com/user-attachments/assets/7a67630e-857a-470f-becf-f657879271b5

### DIFF
--- a/code/modules/surgery/bodyparts/bodypart_wounds.dm
+++ b/code/modules/surgery/bodyparts/bodypart_wounds.dm
@@ -143,6 +143,10 @@
 			dam += 10
 		if(istype(user.rmb_intent, /datum/rmb_intent/weak))
 			do_crit = FALSE
+
+	if(istype(user, /mob/living/simple_animal))
+		do_crit = FALSE // no more mob crits!!
+		
 	testing("bodypart_attacked_by() dam [dam]")
 	var/added_wound
 	switch(bclass) //do stuff but only when we are a blade that adds wounds


DelineFortune:
But mobs cannot do anything without crits...

Coolgat3:
> But mobs cannot do anything without crits...

and this is fine. If I have to die to bullshit let it be a player not a mob that can outrun you thanks to diagonal walking. I want interactions with players personally not mobs its not really fo13 where everyone is perpetually looting I guess is one way to put it.


I like this, please yes.

ARefrigerator:
I can lower the crit chance I suppose. I think the main point of this PR is just to mostly let people know where, deep down in this code, you actually go to change this because it was a doozy to find the right place.

Xaytan:
Damn. You can already stab all simple mobs to death without them ever being able to fight back, now they can't even kill you when they do fight you.

Sad day for the volfs. Rest in piss minotaurs.

RadiantFlash7:
We're not removing simple mobs ability to crit. They would never be able to kill anything otherwise.

That said; Something that was discussed was a minimum threshold of damage to a limb before a crit can take place. Something that would help with simplemob critting in one hit, while also helping in pvp issues.
If you want to do that instead; That would take the issue off of my plate. @ARefrigerator

Gyrandola:
Removing crits makes simple mobs trivial. Simple mobs shouldn't be trivial, traversing the bog shouldn't be trivial. I don't think this is the correct way to make combat feel better.

Radiant's suggestion is worth exploring.

PR #2711, opened 2025-04-12
Makes Druid's Tame and Seelie's Tame Awaken Animals

yupup123:
## About The Pull Request

- Makes Druid's Tame Beast and Seelie's Tame Rous awaken the targeted beast, letting players control them. One druid can only have two tamed beasts at a time. Seelie's can have one tamed rous.
- Makes the spells actually work how they are meant to by adding them to the `neutral` faction, making them no longer attack.
- Gives spiders, rous, and wolves their respective factions, which druids were already included in. I think they may have been intended to ignore druids because of this.
- Gives simplemobs some code to be able to have different languages. Gives volfs, spiders, saiga, and rous the ability to speak beast-speech, but not common.

## Why It's Good For The Game

- This will give Druids a fun gameplay niche, and reinforces their themes of "friend of beasts". Only the druids will be able to communicate with the awakened beasts.
- Makes the Tame spells much more useful, and actually work, even if it doesn't awaken.
- Druids will be able to go into the bog and be a little bit safer, no longer needing to worry about spiders, volfs, or rous, and maybe even make friends with them!
- It will be very easy to add more beasts to the list to be able to be awakened.
- Further functionality to the awakened animals can be added at some point, such as volfs being able to carry objects in their mouths, or spiders creating webs.

I would show more footage of testing, but this one shot took multiple takes with how buggy having two byond clients running at the same time is. 

https://github.com/user-attachments/assets/bb1b52f2-5fb5-4245-b019-e26fe507660b

### DIFF
--- a/code/modules/spells/roguetown/acolyte/dendor.dm
+++ b/code/modules/spells/roguetown/acolyte/dendor.dm
@@ -80,7 +80,7 @@
 	target.visible_message(span_warning("The [target.real_name]'s body is engulfed by a calming aura..."), runechat_message = TRUE)
 	// Kind of a hacky fix to make sure the ai doesn't attack people, but it works. 
 	target.faction = list("neutral")
-	target.tamed = TRUE
+	target.tame = TRUE
 	
 
 	// Poll for candidates to control the tamed animal


ARefrigerator:
Does the new AI affect this at all or override it?

yupup123:
> Does the new AI affect this at all or override it?

It's been tested rather thoroughly. The new AI is overridden if they are awakened. 

`ai_controller = new /datum/ai_controller/basic_controller(src)`

ophaq:
This is cool.

mkr3492:
Perhaps include goblins into the volf faction so that they remain on friendly terms with them, because as it stands volfs do not attack goblins and vice versa including player goblins.

noelle-lavenza:
#2713's fix may be preferable to adding tamed mobs to the `neutral` faction, since it'll ensure they only avoid attacking their owner rather than including a large number of other mobs too. Unsure which is desired.

Issue #2224, opened 2025-01-15
[BUG] Tame beast no longer works (on spiders)


shorkleshark:
## Describe the bug
Tame beast no longer stops animals from eating druids. (spider specifically) and when it DOES work on certain creatures, there is no feedback.
 

## Round ID:
Mountain1737-01.15.48

## Steps to Reproduce:
Play as a druid, find a spider, attempt to tame it.

## Additional context
we have a spider in the grove, tamed already.
![image_2025-01-15_065800296](https://github.com/user-attachments/assets/24a65344-eea8-4d97-ad1c-b94e75679d49)

Frienderanian:
Pretty sure Tame Rous doesn't work for Seelies either. This is so lame...

PR #2713, opened 2025-04-13
Fix tamed mobs attacking their owners.

noelle-lavenza:
- Tamed animals now have a visual indication of taming again.
- Tamed animals using the new AI system no longer attack their owners.

### DIFF
--- a/code/modules/mob/living/emote.dm
+++ b/code/modules/mob/living/emote.dm
@@ -741,6 +741,13 @@
 	message = "stares blankly."
 	message_param = "looks at %t."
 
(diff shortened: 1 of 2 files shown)


noelle-lavenza:
This should fix #2224, but I'm not 100% sure because I can't test it (being bad at the game and all). The issue is only for mobs that use the new AI controller system, though.

ARefrigerator:
The issue with mobs using the new AI is just needing the spell to perhaps add a variable to their blackboard they can read from.  The only issue I've had is finding the right place to put it because when I set the value in one of the behavior procs the mob would just stand there and break. I probably failed to call failure state of the tree maybe but they can respond to the spell if you set it. They just don't use the 'aggressive' var old AI uses.

PR #2714, opened 2025-04-13
Fix linter warning

noelle-lavenza:
Fixes a linter warning from undefining a define that was never defined. It was annoying me.


noelle-lavenza:
It also... adds the linter check to CI, because that's good to have.

noelle-lavenza:
this is going to take approximately ten billion years.

Gyrandola:
wtf

PR #2728, opened 2025-04-14
Templar helmet selection

GravyMaker:
## About The Pull Request
This PR adjusts Templar setup logic to remove the random helmet selection for Astratan and Necran Templars, replacing it with a menu offering a choice between the two on round start.

## Why It's Good For The Game
This PR was requested by a friend who sometimes plays an Astratan Templar. From looking at the code, and at the items in-game, there's no mechanical between the two helmets aside from one having a visor that can be lifted. Lifting the visor offers no combat advantage (and in fact disadvantages the wearer by reducing its protection), so the randomness appears not to have been a balance decision.

As such, it seems harmless to give players the option to choose a helmet according to their preference. In particular, I'm aware that at least some admins have, in the past, been willing to swap out the helmet in response to an ahelp - this will obviate that process, saving both player and admin time.

The menu:
image

Helmet slot after queueing as an Astratan Templar and choosing the visored helmet:
image

Helmet slot after queueing as an Astratan Templar and choosing the bucket helmet:
image

Helmet slot after queueing as a Necran Templar and choosing the visored helmet:
image

Helmet slot after queueing as a Necran Templar and choosing the bucket helmet:
image

I've also ensured that loading as Templars of other gods does _not_ create the menu - though if appropriate helmet choices are added for them the logic could easily be modified.

## Addendum
I'm not really that familiar with the proper terms for mediaeval and renaissance arms and armour, so if there would be a better way to refer to the visored helmet than as "Visored" in the menu (and likewise for the bucket helm), let me know and the string can be changed.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/church/templar.dm
+++ b/code/modules/jobs/job_types/roguetown/church/templar.dm
@@ -33,7 +33,7 @@
 		H.real_name = "[title] [prev_real_name]"
 		H.name = "[title] [prev_name]"
 		
-		addtimer(CALLBACK(L, TYPE_PROC_REF(/mob, templar_helmet_choice)), 50)
+		addtimer(CALLBACK(src, PROC_REF(templar_helmet_choice), H), 50)
 
 /datum/outfit/job/roguetown/templar/pre_equip(mob/living/carbon/human/H)
 	..()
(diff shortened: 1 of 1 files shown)


RadiantFlash7 (commented):
Hurray; getting to pick their helmet is rather nice; rather then it being RNG. Code looks solid; Though I wonder why set the proc base as mob, instead of basing the proc off the outfit/job? It works either way; But seemed a little odd.

Gyrandola (approved):

PR #2730, opened 2025-04-15
Accuracy adjustments, swords get a bonus to use

RadiantFlash7:
## About The Pull Request
This PR adjusts accuracy calculation; adding penalties for specific area targetting.
Aiming for the eyes, ears, nose, hands and feet are much harder then, per say, targetting the arms and head.
Chest remains the easiest hit; and on double accuracy fail, will still land blows there.
Single accuracy fails were given a boon, so you have a higher chance of landing your hit on the broader region rather then defaulting to the chest.
Swords also recieved a bit of a buff, getting an accuracy improvement.
Regular swords get the most accuracy; Long swords get none,
Sabers get half, and rapiers get a quarter. 
Short swords were also given the short weapon length, meaning they should have an accuracy benefit now.

## Why It's Good For The Game
The meta of targetting hyper specific places that are unarmored is pretty horrible for combat; Especially when several of these places can lead to instant kills.  I'd rather people be aiming for the head, legs, and arms natively; But we'll have to see how it goes.

image

image


Cyprex:
Had to resolve a merge conflict, gladius was changed by this, but removed on main

Cyprex (approved):

PR #2732, opened 2025-04-16
Allows low and medium priority duke rolls

ARefrigerator:
## About The Pull Request
This should be what allows people to roll as duke even with medium and low priority. I haven't quite tested it but it should be the only thing stopping a duke from not getting selected if it's on medium or low (because it's hard coded to not accept it)

## Why It's Good For The Game
Some people have duke as their second choice. They don't mind playing duke but they'd still like to roll for their other preferred job if given the choice. This would allow people who have a secondary preference for the role to get it. Perhaps making starting rounds a lot easier.

Will this make some dukes less ready for the role? Maybe. But it could be fun to see them play it regardless. Plus it might ease up the lobby begging for people to go duke if someone wants it and doesn't role. 

https://github.com/user-attachments/assets/da4d6c71-7aca-40e1-bd60-c18b5453f29a

### DIFF
--- a/code/controllers/subsystem/ticker.dm
+++ b/code/controllers/subsystem/ticker.dm
@@ -249,7 +249,7 @@ SUBSYSTEM_DEF(ticker)
 		for(var/mob/dead/new_player/player in GLOB.player_list)
 			if(!player)
 				continue
-			if(player.client.prefs.job_preferences[V] == JP_HIGH || player.client.prefs.job_preferences[V] == JP_MEDIUM )
+			if(player.client.prefs.job_preferences[V] == JP_HIGH || player.client.prefs.job_preferences[V] == JP_MEDIUM || player.client.prefs.job_preferences[V] == JP_LOW )
 				if(player.ready == PLAYER_READY_TO_PLAY)
 					if(player.client.prefs.lastclass == V)
 						if(player.IsJobUnavailable(V) != JOB_AVAILABLE)


ARefrigerator:
Now that I Think about it I think this could be a problem where if everyone has it set to medium or low and gets their first choice there might not be a duke... Unless there's somewhere else in the code where duke is always chosen to happen. This might explain why someone coded it so only high priority starts the game. I could look deeper into making it so a duke always checks for everything first and has some kind of priority for that, it it doesn't already.

E231kei500bandai:
duke is supposed be rolled for first before the other roles as it is a required roll

image

it would cause a SSticker.mode.setup_error but idk what that does

E231kei500bandai:
i believe this whole proc was added solely to prevent that error from occuring

it serves no puripose otherwise so you might as well remove the entire proc

E231kei500bandai:
can_continue = can_continue && SSjob.DivideOccupations(list()) 		


put "Duke" here

E231kei500bandai (changes requested):
You need to make it so that it only adds players who have it set to low, medium, or high. Right now, it would add players who don't have it set to anything.

This is making the entire function pointless since it will always return true no matter what.

Or, you could just remove the entire function if you wanted, but this could result in games that don't have a duke if literally nobody set it to low medium or high. Although, there is a different piece of code that should prevent this too, but I'm not really sure how exactly it works. https://github.com/Rotwood-Vale/Ratwood-Keep/blob/0325554be3079f3d0906fc60bd8143232137e7f0/code/controllers/subsystem/ticker.dm#L339

ARefrigerator:
I know people have flirted with the idea of allowing no dukes but it could affect things (probably causes omens to spawn 24/7)

But is that true it also includes players who have it set to nothing? I guess I Can write that one if statement back in to include all three.

Alanii:
Due to this PR existing, I assume some retard removed the ability to actually allow LOW set dukes to be pulled?

Originally, anyone could become duke if they had "LOW" on Duke role and no one readied up in MEDIUM or LOW.

E231kei500bandai:
> Due to this PR existing, I assume some retard removed the ability to actually allow LOW set dukes to be pulled?
> 
> Originally, anyone could become duke if they had "LOW" on Duke role and no one readied up in MEDIUM or LOW.

nobody direclty removed the abilty, the game just refusednto start if nobody had duke on high, making it impossible to get duke jb medium or low unless admins forcd started the game

ARefrigerator:
> > Due to this PR existing, I assume some retard removed the ability to actually allow LOW set dukes to be pulled?
> > Originally, anyone could become duke if they had "LOW" on Duke role and no one readied up in MEDIUM or LOW.
> 
> nobody directly removed the abilty, the game just refused to start if nobody had duke on high, making it impossible to get duke on medium or low unless admins force started the game. Whoever made the system didn't realize required jobs system existed and thought you had to have someone with duke on high to guarantee a duke.

If you know more about setting that up are you able to do it. It would involve me having to read through all the code first since I'm not familiar with it.

ARefrigerator:
For now I just set it back to where it was but also allowed medium and low priority via that method. It's not ideal but I guess it might work for now. I still gotta test it.

PR #2740, opened 2025-04-17
Fix invalid uses of Destroy and del

noelle-lavenza:
## About The Pull Request
Cleans up various incorrect uses of `Destroy()` and `del()`, replacing them with `RemoveComponent()`/`qdel()` where applicable.

## Why It's Good For The Game
You should never ever call Destroy() or del() directly.

See CI.


RadiantFlash7 (approved):
Good code refractor. Thank you for your work.

PR #2741, opened 2025-04-17
Clean up some stat-related code.

noelle-lavenza:
## About The Pull Request
Big PR, entirely backend.

Makes `change_stat()` use string defines because I keep typing `change_stat("luck"` instead of `change_stat("fortune"`, so now it uses `STAT_STRING_LUC` to go with `STALUC`. I'm willing to negotiate on the name of the define, it's easy enough to change at this point. `STAT_STR_LUC` was a candidate but then you get `STAT_STR_STR` and that's a little confusing...

Also makes some direct stat add/subtracts use `change_stat()` instead, letting us take advantage of the index system to make setting/unsetting them easier.

## Why It's Good For The Game
Cleaner code, more compile-time checking. Also, using the index-based thing makes sure that you can't accidentally gain or lose extra stat points, `change_stat(stat, 0, "index")` will always totally negate that index's buff/debuff, and do nothing if it hasn't been applied. Hooray!

Backend code, so it should just work if it compiles... I hope. I can hardly join as every job and species in the game...

### DIFF
--- a/code/datums/gods/patrons/curses.dm
+++ b/code/datums/gods/patrons/curses.dm
@@ -175,11 +175,11 @@
 
 /datum/curse/xylix/on_gain(mob/living/carbon/human/owner)
 	. = ..()
-	owner.STALUC -= 10
+	owner.change_stat(STAT_STRING_LUC, -10, "xylixcurse")
 
 /datum/curse/xylix/on_loss(mob/living/carbon/human/owner)
 	. = ..()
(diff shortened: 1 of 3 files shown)


Gyrandola:
Conflicts

Kush1Push1:
https://github.com/Rotwood-Vale/Ratwood-Keep/pull/2245
Same fate

PR #2747, opened 2025-04-18
Crit Protection No Longer Stops Wounds

Xaytan:
Non-Crit Wounds can be applied if damage gets through your armor.

## About The Pull Request

Attacks that successfully penetrate armor will now apply wounds, even if the armor prevents crits from that BCLASS.

Damage for the purpose of wounding is reduced by half if the armor has the appropriate wound protection.

Wound/crit damage math changed to apply armor consistently with all other damage in the game.

## Why It's Good For The Game

This makes bladed and penetrating weapons substantially more deadly, and makes armor penetrating weapons more capable of dealing meaningful damage to their target. It will make fights a little more bloody, and should see a good bit more wounds going around even when people aren't getting crit.

image
image
image

### DIFF
--- a/code/modules/mob/living/carbon/human/species.dm
+++ b/code/modules/mob/living/carbon/human/species.dm
@@ -1607,7 +1607,7 @@ GLOBAL_LIST_EMPTY(roundstart_races)
 			if(I)
 				I.take_damage(1, BRUTE, I.d_type)
 		if(!nodmg)
-			var/datum/wound/crit_wound = affecting.bodypart_attacked_by(blade_class, (Iforce * weakness) * ((100-(armor_block+armor))/100), user, selzone, crit_message = TRUE)
+			var/datum/wound/crit_wound = affecting.bodypart_attacked_by(blade_class, (Iforce * weakness) - (armor_block+armor), user, selzone, crit_message = TRUE)
 			if(should_embed_weapon(crit_wound, I))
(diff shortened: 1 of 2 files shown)


RadiantFlash7 (commented):
If we're going to make injuries punch through armor on penetration, we'll want to probably force it to use the minor wound specifically. Small slashes, small punctures, etc. Maybe if could use a larger one if it pierces by X amount over?

Xaytan:
> If we're going to make injuries punch through armor on penetration, we'll want to probably force it to use the minor wound specifically. Small slashes, small punctures, etc. Maybe if could use a larger one if it pierces by X amount over?

The calculation is done based on the damage after being reduced by armor (Note: This is done with a percentile reduction, not a flat one. I had thought it was a flat reduction!) and then reduced by a further 10 before wound type is selected. In the example I screenshotted, if the character had taken the hit without crit protection it would have been a gaping puncture even after armor but was reduced to a normal puncture instead.

The main concern I have is extremely high damage, high AP weapons like maces and warhammers. It may be time to nerf the AP on those so that they don't consistently leave massive hematomas through plate armor.

Xaytan:
Changes made to the math behind it.

RadiantFlash7:
I'll look at putting this for a TM after the other crit PR is finalized

CVSthePharmacy:
Muh mace stealthnerf

PR #2750, opened 2025-04-18
Removes Magos Refugee Patron Lock

yupup123:
## About The Pull Request

Allows magos refugees to have either Noc or Zizo as their patron.

## Why It's Good For The Game

Magos refugees are forced into only Noc (except for Battle Magos for some reason). This gives more opportunities for different flavors of magic users from afar. I made sure to check if there was a specific reason as to why they were locked only to Noc, but it doesn't seem like there is one.

Screenshot 2025-04-18 172617

### DIFF
--- a/code/modules/jobs/job_types/roguetown/refugees/types/magos.dm
+++ b/code/modules/jobs/job_types/roguetown/refugees/types/magos.dm
@@ -86,6 +86,11 @@
 			H.change_stat("endurance", -1)
 			H.mind.AddSpell(new SPELL_LEARNSPELL)
 			H.mind.AddSpell(new SPELL_PRESTIDIGITATION)
+			switch (H.patron?.type)
+				if(/datum/patron/zizo)
+					H.set_patron(/datum/patron/zizo)
+				else
+					H.set_patron(/datum/patron/divine/noc)
 			if(istype(H.dna.species, /datum/species/dwarf) || istype(H.dna.species,/datum/species/anthromorphsmall)|| istype(H.dna.species, /datum/species/kobold))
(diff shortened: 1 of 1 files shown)


RadiantFlash7:
Mages are typically locked to either Noc or Zizo; as Noc the patron of magic and knowledge and Zizo is- well, zizo.
Given magic requires intense study- It makes sense that they'd be worshipping them over any other patron.

RadiantFlash7:
> > Mages are typically locked to either Noc or Zizo; as Noc the patron of magic and knowledge and Zizo is- well, zizo. Given magic requires intense study- It makes sense that they'd be worshipping them over any other patron.
> 
> I agree, it definitely makes the most sense for those two, but it still feels pretty restricting. Just because it makes sense for the role doesn't mean it needs to be a requirement, right?
> 
> If it's a requirement for this to be merged, I'll make it so they can only be patrons of Zizo or Noc.

I'd prefer just these two for now. In game people tend to follow patrons related to their work due to benefits in doing so. This follows a fairly established theme- and while there could be more nuisance in having differant patrons, Our server tends to be fairly binary in this respect. It's why you don't have people with divine patrons often doing heretical acts, or heretical patrons doing 'good' acts.

RadiantFlash7:
> 
> Make sense. I do wish there could be a little more nuance in such things but, such is the way they are. I'll get around to changing this to be only Zizo and Noc then.
> 
> If this is the case however, it does make me wonder what sets the apprentice magician role apart from this theme...

..I thought mage apprentices could only be zizo and noc as well? If they can be other patrons that might be an issue.

RadiantFlash7 (approved):
Approved. Thank you for making the changes.

PR #2751, opened 2025-04-19
Fixed grabs not reducing blood loss on grabbed body parts

GravyMaker:
## About The Pull Request

This PR fixes an issue which causes grabbing a bleeding body part to not reduce the blood loss on that body part.

A body part's blood loss was intended to be reduced by 50% if that body part was grabbed - and the remainder by a further 50% for every additional grab (so, two grabs would be a reduction of 75%, three would be 87.5%, and so on). This code already existed, and in fact wasn't modified in this PR.

However, the code dealing with _grabs_ would fail to add the grab to the limb's "grabbedby" list. As a consequence, the code described above was essentially dead code - the condition for the blood loss reduction would never be fulfilled under any circumstances. This is what the PR modifies, by adding the grab to the body part's grabbedby list (and thereafter removing it at the appropriate time).

## Why It's Good For The Game

This is something that was always intended to work, and would be valuable in buying a little more time while taking heavily-wounded characters to the church or a physician, without actually doing anything to treat the wounds on its own. This is particularly valuable in light of the recent limits placed on character revival - a little more time to get somebody treated could go a long way.

It's a little tricky to show this in a way that's unambiguous as to the reduction in blood loss since, as far as I'm aware, there's no display that shows an actual numerical amount in-game. Likewise, I could do so much damage that I bled out rapidly, but this would mean I'd be doing a lot of prior damage to myself, meaning a small health pool to show the effects from. So, instead, I got in my test environment, wounded myself a bunch, and then compared the bleeding in the debugger before and after grabbing the affected body part, without otherwise modifying anything.

The wounds:
image

Before grab, 21 blood loss:
image

After grab, 10.5 blood loss:
image

I've additionally ensured that grabbing mobs still works correctly, as well as objects becoming embedded mid-combat.

### DIFF
--- a/code/modules/mob/living/grabbing.dm
+++ b/code/modules/mob/living/grabbing.dm
@@ -88,6 +88,17 @@
 	if(ismob(grabbed))
 		var/mob/M = grabbed
 		M.grabbedby -= src
+		if (iscarbon(M) && sublimb_grabbed)
+			var/mob/living/carbon/carbonmob = M
+			var/obj/item/bodypart/part = carbonmob.get_bodypart(sublimb_grabbed)
+
+			// Edge case: if a weapon becomes embedded in a mob, our "grab" will be destroyed...
+			// In this case, grabbed will be the mob, and sublimb_grabbed will be the weapon, rather than a bodypart
+			// This means we should skip any further processing for the bodypart
+			if (part)
+				part.grabbedby -= src
+				part = null
+				sublimb_grabbed = null
 	if(isturf(grabbed))
 		var/turf/T = grabbed
 		T.grabbedby -= src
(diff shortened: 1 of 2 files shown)


RadiantFlash7 (approved):
Thank you for fixing this issue. It's been one for- about 8-10 months? One I've known about but haven't had the time to track down myself. 😸

noelle-lavenza (approved):
An easier solution would have been to make wound code loop over the mob's grabbedby and check if limb_grabbed matches the bodypart the wound is on, but as far as I know this should work too, since it properly handles clearing the relevant vars on Destroy().

(That said, at some point this cleanup should probably be moved to a signal.)

PR #2754, opened 2025-04-20
PP Improvements

HelennaC:
## About The Pull Request

It pretty much substitutes the New Player Panel (NPP) with the Super Player Panel (SPP)
(I won't call it New New, or call the other one 'old' because of an OLD PP existing.)

The SPP will substitute the NPP.

## Why It's Good For The Game

Removed broken and crashing functions of the New Player Panel with a SUPER Player Panel
Still have plans to add more to it, but as of now it's way better than the New Player Panel.

That's the SUPER PP. Say hi!
image

[The right side options open up too]
image
image
image

### DIFF
--- a/code/modules/admin/admin.dm
+++ b/code/modules/admin/admin.dm
@@ -155,7 +155,7 @@
 	body += "<div id='right-header'>"
 	body += "<button onclick=\"toggleSection('skills')\">Skills</button>"
 	body += "<button onclick=\"toggleSection('languages')\">Languages</button>"
-	body += "<button onclick=\"toggleSection('stats')\">Stats</button>"
+	body += "<button onclick=\"toggleSection('stats')\">WIP</button>"
 	body += "</div>"
 
 


code/modules/admin/admin.dm
```diff
@@ -13,21 +13,50 @@
...
+	body += "    var activeSection = \"[clicked_flag]\";"
+	body += "    if (activeSection !== \"0\" && activeSection !== \"\") {"
+	body += "        toggleSection(activeSection);"
+	body += "    }"
+	body += "}"
+	body += "</script>"
+
+	body +="</head>"
+	body += "<body><div id='container'>"
+	body += "<div id='left'>"
```
dwasint, line 59: honestly not my best work, this should be wrapped in a {""} instead of like 40 lines of body +=

RadiantFlash7:
Someone threw this into a test merge without a review or label...

RadiantFlash7 (approved):
Haven't had any complaints from admins about bugs while using this while it's been test merged. Approved.

PR #2755, opened 2025-04-20
Food interaction restructure with datum recipes

ARefrigerator:
## About The Pull Request
Restructures all the slap crafting food interactions into datums
This allows us to have recipe datums referenced twice when the interaction is reverse without having to bring up the issue of manually writing attackbys for the slap craft objects. I tried to comment everything in case anyone in the future reads this and has no idea what it is (That person will be me)

### - Some info -
Recipes datums are laid out in the following way.
It has variables for:
-how long it takes to assemble (time_to_make)
- items used
- reagents (possibly for a future check, it's a remnant from my older code)
- the item you make (result)
- interaction type (this is so the datum can fire and know the difference between attackby and attackhand)
- slices (currently unused)
- sound to make on assembly (craft_sound)
- message to send to user (crafting_meesage)

They have the current procs:
**clear_items**
-  While loops through all the items in the recipe and qdels them from the world

**check_items_in_list**
- Creates an associated list and counts backwards to know you have enough items to actually make the recipe

**pre_check**
-  Simple proc you can use as a simple check for any special code a recipe might need (e.g. the peppermill needing to have reagents inside it to be used)
- Can be, and should be, overridden for these special cases

**post_handle**
-  Proc for anything that needs to happen after creation that is unique to that recipe (e.g. actually removing the reagent from the peppermill)

Adds the following global procs:
**select_interaction_recipe**
The real meat of this systems. Takes all the available recipes in the food interaction list, the items you want to check, the interaction type, and whether this is exact (it always is for now).

- Checks if you have enough items via check_items_in_list
- checks for the interaction type
- Runs through every recipe with these requirements
- Once it has a list it attempts to do the most complex recipe it can (AKA the one with most ingredients). Considering that most things are still 2 ingredients at most this wont matter for now but in the future it might mean you could say make a cake and skip the butterdough step if you have an egg also in you hand (provided it is written).

**food_handle_interaction**
Takes the source, user, items, and interaction type.
- Tries to find the recipe via previous proc
- Failure returns false (attackby and attack hand will call parent proc on failures)
- if it finds one it uses all the information from it to handle the actual making of the thing and then calls all the recipe's procs to clean up.

Attackby has changed for all `snacks` subtypes
- Looks for a table (currently all actions require a table)
- Attempts slice code first (I don't want to touch this yet and I don't think I have to)
- if slicing does not work we get a list of the player's holding items (active + inactive)
- add this to a list of to_check items with the original item hit.
- feed this into the previous proc to check
- on success we get a new meal / item
- on failure we try to call parent procs as a fallback.

attack_hand has changed for all `snacks` subtypes
- Find a table, if no table call parent proc
- if there is a table try the recipe interaction
- if the recipe fails try the parent proc.

Adds two defines for the interaction types 
`#define FOOD_INTERACTION_ITEM 1 // use for attackby proc`
`#define FOOD_INTERACTION_HAND 2 // use for attack_hand proc`

I think those are the only three things I gotta fix. I can do those after I get this PR posted up.

### UPDATE: 2025/05/19
**Cooking from Scratch (Lucky Lily)**
Adds:

- Venus cake (cake base + jacksberries)
- fruitcake (base + apples + berries (hold two in your hands, click third))
- plain cake (bake base)
- iced cake (base + sugar)
- fish fillet (cut again for mince)
- scalloped potatoes (fried potatoes + cheese wedge)
image

## Why It's Good For The Game
Ideally if everything is working the players should notice absolutely no difference in how cooking currently works. So from a user end experience this changes nothing.

However from a dev perspective this should allow making food slap crafting a lot more concise and clear as you now only need to list the items you want to use (1-3 of them) make a small pre_check proc for any special edge cases and then you no longer have to worry about inconsistent items not having matching attackby logic (e.g. currently a chicken cutlet onto fried potatoes will show message but the inverse does not.)

It may also help references in the future if say books want to keep a list of recipes, now it wont have to be manual because you could just load this whole datum into the book and save it there.

Me trying to make every recipe I could think of (they all seemed to work)
https://github.com/user-attachments/assets/977d14e7-c29b-4bed-952f-bf642f5bbbc6

example of wiener meal being made in either order
https://github.com/user-attachments/assets/8ca38574-ac4f-4b0c-96cd-e8638a809c5b

Example of extra items not being eaten by code.
https://github.com/user-attachments/assets/e86740de-7417-4be5-a41b-1af498255a2b

### DIFF
--- a/modular/Neu_Food/code/raw/NeuFood_dough.dm
+++ b/modular/Neu_Food/code/raw/NeuFood_dough.dm
@@ -536,6 +536,7 @@
 	foodtype = GRAIN | DAIRY | SUGAR
 	bitesize = 3
 	eat_effect = /datum/status_effect/buff/greatsnackbuff 
+	extra_eat_effect = /datum/status_effect/buff/sweet
 	rotprocess = SHELFLIFE_LONG
 
 
--- a/modular/Neu_Food/code/raw/NeuFood_veggies.dm
+++ b/modular/Neu_Food/code/raw/NeuFood_veggies.dm
@@ -62,6 +62,5 @@
 	name = "scalloped potatoes"
 	desc = "A cheesey covering of baked potatoes."
 	icon_state = "scalloped_potatoes"
-	eat_effect = null
 	tastes = list("potato" = 1, "cheese" = 1)
 	eat_effect = /datum/status_effect/buff/greatsnackbuff


ARefrigerator:
This does currently have a bug that I thought I fixed and it has returned. The bug is that one of the items does not delete on the loop to clear it. I think this could have been as a result of moving the code into unsorted but I'll see.

ARefrigerator:
Okay I think I might have fixed everything with this. Unless the linter shows any issues I guess it could be good to test merge it to see. Any missing recipes I might have missed would just be an issue of adding to the datum so it's less severe.

code/__HELPERS/unsorted.dm
```diff
@@ -1626,3 +1626,34 @@ GLOBAL_LIST_INIT(duplicate_forbidden_vars,list(
...
 	return O
+
+/*================
+handle interaction
+================*/
+/*	- Takes the current item in your active hand, inactive hand, and what was hit (the src)
+	  and feeds it through the food_interaction recipes to determine what to do. This way
+	  we no longer have to write an copy paste a bunch of attackby code everywhere. - */
+
+proc/food_handle_interaction(obj/item/source, mob/living/user, list/items, interaction_type)
```
dwasint, line 1637: should start with a slash

code/datums/food_handle_recipes.dm
```diff
@@ -0,0 +1,518 @@
...
+	for(var/obj/item/reagent_containers/R in to_check)
+		if(!istype(R, /obj/item/reagent_containers/powder/flour))
+			R.reagents.remove_reagent(/datum/reagent/water, 10)
+			break
+	..()
+
+// Specific code again to show overriding
+// This dough code has been the gold standard.
+// Please consult this when understanding how to deal with special cases.
+datum/food_handle_recipes/dough_wet/clear_items(list/itemlist)
```
dwasint, line 234: missing starting /

dwasint (commented):
Minor things I saw.

ARefrigerator:
> Minor things I saw.

Okay I think I got them all. This should be read for a test then. In the mean time I might expand on the craft_messages later while it waits.

RadiantFlash7:
I'll review this either today or tomorrow.

ARefrigerator:
Unless anything seems obviously missing from the lists or something is horribly wrong and I don't see it I feel more confident that this can be TM'd. Added sausage roll just as proof of concept on how you can expand this easily with no attackby code since all snack subtypes should use the same logic now.

ARefrigerator:
Had an idea I might try with this where I make the results a list. Will allow for making some recipes give multiple things. Could also mean we remove slices but I'd rather not refactor that deep. It'll just be basic for now.

RadiantFlash7:
> Had an idea I might try with this where I make the results a list. Will allow for making some recipes give multiple things. Could also mean we remove slices but I'd rather not refactor that deep. It'll just be basic for now.

This ready to under go testing?

modular/Neu_Food/code/raw/NeuFood_dough.dm
```diff
@@ -458,6 +459,129 @@
...
+	tastes = list("cake"=1, "sugar"=1)
+	foodtype = GRAIN | DAIRY | SUGAR
+	rotprocess = SHELFLIFE_LONG
+	slice_batch = TRUE
+	slice_sound = TRUE 
+	eat_effect = /datum/status_effect/buff/greatsnackbuff 
+	bitesize = 16
+
+/obj/item/reagent_containers/food/snacks/rogue/icedcake_slice
+	name = "fruit cake slice"
```
john-pawson, line 530: iced cake is called fruit cake slice.

modular/Neu_Food/code/raw/NeuFood_dough.dm
```diff
@@ -458,6 +459,129 @@
...
+	icon_state = "icedcake_slice"
+	slices_num = 0
+	list_reagents = list(/datum/reagent/consumable/nutriment = SNACK_DECENT-1)
+	w_class = WEIGHT_CLASS_NORMAL
+	cooked_type = null
+	foodtype = GRAIN | DAIRY | SUGAR
+	bitesize = 3
+	eat_effect = /datum/status_effect/buff/greatsnackbuff
```
john-pawson, line 538: No sweet buff? Intentional?

modular/Neu_Food/code/raw/NeuFood_veggies.dm
```diff
@@ -57,3 +57,11 @@
 	cooked_smell = /datum/pollutant/food/baked_potato
 	can_distill = TRUE
 	distill_amt = 8
+
+/obj/item/reagent_containers/food/snacks/rogue/veg/scalloped_potatoes
+	name = "scalloped potatoes"
+	desc = "A cheesey covering of baked potatoes."
+	icon_state = "scalloped_potatoes"
+	eat_effect = null
```
john-pawson, line 65: Two eat_effects

RadiantFlash7 (approved):
All reported issues fixed, this has been testmerged for- quite a long time. Approved.

PR #2756, opened 2025-04-21
reverts armor break code refractor

RadiantFlash7:
## About The Pull Request
This reverts a section of code related to armor breaking; the changes were made in PR #2719.
It should theoretically fix the issues with armor breaking early, and keeping crit protection- Along with the issue amongst our players of people not being able to be crit ever.

## Why It's Good For The Game
Fixes armor not properly breaking. This can be re-refractored later.

## Proof of Testing (Required)
image

PR #2757, opened 2025-04-21
Fix issues with armour (not) breaking

noelle-lavenza:
## About The Pull Request
- Fixes armour permanently losing its armour values after breaking. The code for this hadn't worked since BYOND 515.1641 (as mentioned in the comment there), but my rewritten version *did* work even on 1641; I didn't notice the difference because when I test on 515 instead of 516 I use 515.1637 because it's the installer I have downloaded. Evidently the game is and had always been designed around that particular bit of code not working, so removing it fixes it entirely.
- Due to the above, fixes armour never breaking enough to stop protecting against crits. Removing all the armour values means it will no longer receive damage because that code checks if it has an armour rating above 0.
- Fixes armour being 'partially broken' in that it was below the break threshold (10% integrity) but above the crit protection threshold (0.5% integrity, anything above 1). Now broken armour will always stop protecting against crits.
- Armour now stops protecting from damage when it breaks, rather than at 0 integrity.

## Why It's Good For The Game
Fixes issues with armour, more thoroughly than #2756.

image

### DIFF
--- a/code/modules/roguetown/roguejobs/blacksmith/tools.dm
+++ b/code/modules/roguetown/roguejobs/blacksmith/tools.dm
@@ -99,11 +99,10 @@
 
 	if(isitem(attacked_object) && !user.cmode)
 		var/obj/item/attacked_item = attacked_object
-		if(attacked_item.obj_integrity >= 0)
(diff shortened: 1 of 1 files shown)


RadiantFlash7 (approved):
Desperately needed fix. Going to TM.

PR #2758, opened 2025-04-23
Proper directions for sounds

Heheniy:
## About The Pull Request
Instead of calling get_dir to... get direction... of sound, which would get you this 
image
gets a direction based on the angle between user and source, and then converts it into direction, which gives this
image

Changed that for speaking/screaming (the arrows (⇗) thing) and signal horn. If there are any other things similar to those which needs to be changed, let me know, because i cant remember any more of them.

## Why It's Good For The Game
Proper directions of sounds

image
image


Gyrandola:
Probably isn't too expensive? TMing.

Gyrandola (approved):

PR #2761, opened 2025-04-24
Adds spike pit to game, recipe for making

ARefrigerator:
## About The Pull Request
Adds spike pit structure to code base along with recipe

spike pit:
- walking over it causes damage
- gives you a puncture wound
- buckles you to the pit
- paralyzes you based on Constitution
- unbuckle yourself by clicking on it with empty hand
- unbuckle others by clicking on it with empty hand
- destroyed with shovel on attackby
- Seelies ignoe / fly over them for now (I assume they're too small to suffer the effects of one)

Recipe:
- 3 stakes
- shovel needed
- dirt tile needed

## Why It's Good For The Game
Could be fun to make them I guess.

https://github.com/user-attachments/assets/adb042a3-1a98-43f2-b5ac-9a53c1703c86

### DIFF
--- a/code/modules/roguetown/roguejobs/gravedigger/hole.dm
+++ b/code/modules/roguetown/roguejobs/gravedigger/hole.dm
@@ -309,7 +309,7 @@
 	// TO DO: figure out how to either make jumping its own special proc
 	// Or read throw_at to understand how to parse it here to allow jumping
 	
-	if(is_living(AM))
+	if(isliving(AM))
 		var/mob/living/L = AM
 		if(L.movement_type & (FLYING|FLOATING)) //don't close the trap if they're flying/floating over it.
 			return ..()


code/modules/roguetown/roguejobs/gravedigger/hole.dm
```diff
@@ -290,3 +290,98 @@
...
+	max_integrity = 0
+	buckle_lying = 90
+	layer = 2.8
+
+/obj/structure/spike_pit/Crossed(atom/movable/AM)
+	. = ..()
+	var/hitsound = pick('sound/combat/hits/bladed/genstab (1).ogg', 'sound/combat/hits/bladed/genstab (2).ogg', 'sound/combat/hits/bladed/genstab (3).ogg')
+	// TO DO: figure out how to either make jumping its own special proc
+	// Or read throw_at to understand how to parse it here to allow jumping
+	if(isseelie(AM))
```
dwasint, line 312: should just add a flying movement check instead

dwasint:
Won't this just be like the defacto way to kill mobs since they aren't smart enough to path over them without tweaking your astar function to add weighted paths

ARefrigerator:
> Won't this just be like the defacto way to kill mobs since they aren't smart enough to path over them without tweaking your astar function to add weighted paths

It all comes back to the inbuilt pathfinding... I guess at some point I have to address them.

ARefrigerator:
> 
> I redid astar on vanderlin to use pathweight for astar pathing, you could look at that, it does have the side effect of making mobs respect z-levels though so they will chase you up stairs

I guess I should look at that. If it's the same system it might be easier than having to port my older one I tried making once. The one here I think currently uses inbuilt byond pathfinding. I'm not sure though.

dwasint:
> > I redid astar on vanderlin to use pathweight for astar pathing, you could look at that, it does have the side effect of making mobs respect z-levels though so they will chase you up stairs
> 
> I guess I should look at that. If it's the same system it might be easier than having to port my older one I tried making once. The one here I think currently uses inbuilt byond pathfinding. I'm not sure though.

I made ai_controller mobs use hybrid pathing which basically means dumb pathing for like finding objects, but astar for mob targetting.

RadiantFlash7:
> Won't this just be like the defacto way to kill mobs since they aren't smart enough to path over them without tweaking your astar function to add weighted paths

I don't have an issue with mobs running over them frankly. Mantraps and spike fits filling the same 'roll' per say is fine. I /do/ wonder whether or not not having the buckled icon is a good idea though.

RadiantFlash7:
> ## About The Pull Request
> Adds spike pit structure to code base along with recipe
> 
> spike pit:
> 
> * walking over it causes damage
> * gives you a puncture wound
> * buckles you to the pit
> * paralyzes you based on Constitution
> * unbuckle yourself by clicking on it with empty hand
> * unbuckle others by clicking on it with empty hand


Can you unbuckle from it by resisting with the resist hotkey?

ARefrigerator:
> > ## About The Pull Request
> > Adds spike pit structure to code base along with recipe
> > spike pit:
> > 
> > * walking over it causes damage
> > * gives you a puncture wound
> > * buckles you to the pit
> > * paralyzes you based on Constitution
> > * unbuckle yourself by clicking on it with empty hand
> > * unbuckle others by clicking on it with empty hand
> 
> Can you unbuckle from it by resisting with the resist hotkey?

Yes you can... I might need ti figure out a better way to do this but buckling works. Maybe there's a way to deny the resist button or make it work like cuffs...

RadiantFlash7 (approved):
I'm a leaning towards having this still add a 'buckled' pop up indicator to help make it more obvious to folks; But we can give it a try in this state and see how folks respond to it testmerged.

ARefrigerator:
Is there a way to change how fast unbuckling takes? If so I can change it to have the buckle icon again and change that value.

JohnnyJohnJohnson:
I hope this can be crafted using a spade instead of only needing a shovel. Hunters can craft their own hand spades, they don't spawn with shovels.

PR #2635, opened 2025-03-28
Artificer expansion pack, Now with food rotting

RadiantFlash7:
## About The Pull Request
This PR expands artificer with new content.
Adds in a grappling hook gadget, can grapple yourself, other people and objs up floors.
Self purifying waterskins, purifies water on fill.
Pyroclastic bolts and arrows(sets targets on fire) via artificing and pyro flowers (found in bog),
Adds Voltic gauntlets(shocks nearby people after delay, a la lightning lure spell; activated by right clicking).
Adds metallizer result to sconce, for making wall mounted lampterns,
Rebalances bronze weaponry. Bronze Weapons no longer are just extra integrity. Instead, bronze weapons have high AP, yet iron levels of durability and very low defense stats.
Rebalances some artificer recipes to make them a bit less egregiously expensive compared to smith counterparts,
Musketballs and grapeshot can now be made by artificers with tin.
Adds Artificer armor, A copper and bronze light armor with 45 armor stats at 300 integrity. This covers the chest, neck, groin, vitals. The armor can be upgraded with an arcane meld (Mage loot), allowing it to either enhance the wearer's magic, or their physical stats, toggleable between the two modes by an artificer with a hammer. (+3 Int +1 arcane magic, Or +2 str +2 end, +2 athletics).
Adds bronze whisperer Craftable communication  device contraption (chance of missfiring and has limited charges requiring cogs to power).
Adds steamshield (has a similar knockback effect to repulse in an AOE on trigger, though slower and telegraphed and no stun.)
Adds cooling table (Table that prevents all rotting food)
Adds cooling backpack (Backpack that will prevent things from rotting within)
Adds coal cracking for components
Adds gunpowder sticks (mostly structural damage, long delay before detonation)
Adds gunpowder satchels (Sticks together, meant to breaching walls predominantly.)
Adds various impact grenades. ((NOT MODERN GRENADES, refer to 1600s Period Grenades)

-TWEAK: RE ENABLES FOOD ROTTING on standard tables, albiet at a reduced rate.

Possible to come in far future but not included in this PR: 
Orderable construct minions,
Moving siege breaking bombs from church PR to Artificer,
larger humanoid constructs able to be build and awakened for player control.

## Why It's Good For The Game
Artificer is sorely lacking in content. This will hopefully generate a bit more interest in artificers makeable gear.

## Proof of Testing (Required)
image
image
image
image
image
image

PR #2604, opened 2025-03-23
Church rework [26.03.2025] bombs done, surgeries are next [27.03.2025]

DelineFortune:
To put shit together because I can miss some changes

☑ Preist's changes https://github.com/Rotwood-Vale/Ratwood-Keep/pull/2560
☑ Graggar's rework https://github.com/Rotwood-Vale/Ratwood-Keep/pull/2599
☑Rituals port from Azure  https://github.com/Rotwood-Vale/Ratwood-Keep/pull/2621 (taken from https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/1373)
☑ Secrets of gunpowders https://github.com/Rotwood-Vale/Ratwood-Keep/pull/2628
[In process] Surgeries and organs crafts https://github.com/Rotwood-Vale/Ratwood-Keep/pull/2542 (port from this and toss between PC's almost done)
✗Engraves
✗Artefacts
✗Leeches
✗Gods curse/blessing upon the city

PR #2765, opened 2025-04-25
Clean up impact grenade code

noelle-lavenza:
Cleans up impact grenade code.
- Changes the phrasing of certain strings.
- Makes the name properly `"impact grenade"` rather than `"Impact grenade"`.
- Adds the extra part of each description separately.
- Makes the smoke grenades use smoke systems, to stop them from going through windows.
- Makes the various poison gas grenades a type of smoke grenade.
- Makes impact grenades not use relatively pathed procs.

### DIFF
--- a/code/modules/roguetown/roguejobs/artificer/artificer recipes/artificer_recipes.dm
+++ b/code/modules/roguetown/roguejobs/artificer/artificer recipes/artificer_recipes.dm
@@ -643,11 +643,11 @@
 							/obj/item/rogueore/sulfur,
 						 	/obj/item/rogueore/selitre,
 						 	/obj/item/reagent_containers/powder/alch/berryp = 3)
-	created_item = /obj/item/impact_grenade/smoke
+	created_item = /obj/item/impact_grenade/smoke/poison_gas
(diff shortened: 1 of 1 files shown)


Heheniy:
crafting recepies needs to be fixed too, they all are giving regular smoke grenade

DelineFortune:
> crafting recepies needs to be fixed too, they all are giving regular smoke grenade

https://github.com/Rotwood-Vale/Ratwood-Keep/pull/2604/files#diff-b98f3b1fee46e62928ec821315b9d73340b736dcf8bcdabaa81694869c390547 

its not supposed 2 be...

noelle-lavenza:
> as far i can remember i had an idea that you are not supposed to see the name of the bomb you were supposed to EXAMINE it and get unique discription of each bomb

yes, that is what happens. because of inheritance they all have the name "impact grenade," they don't need to be redefined on everything. the unique description is also the same thing.

as for the recipes, those were changed in #2635 and are wrong, so i'll fix them

Gyrandola (approved):

PR #2770, opened 2025-04-26
Allows Create Item interface to search by item name, as well as type

GravyMaker:
## About The Pull Request
This PR modifies the Game Panel's Create Item and Quick Create Item interfaces to allow searching for items by in-game name, as well as by object type path. Two radio buttons have been added to the interface to allow toggling between the old and new behaviour, with the new behaviour (search by in-game name) being the default.

## Why It's Good For The Game
Spawning items is a function admins may occasionally need to do to aid players - helping recover from a bug that consumed an item, for example, or even in a more Game Master-like sense by creating an item in response to a prayer, or for an event. Most players won't interact with this functionality in the first place, but for those they do, being able to search for an item by its more familiar in-game name rather than its type path is convenient, particularly in cases where the names differ significantly between the two.

The search results for the new, search-by-in-game-name functionality:
image

The item having successfully been spawned after being searched for by name:
image

The same item being searched for by its type in code:
image

And the same item, having been spawned successfully after being searched for by tpye:
image

I've additionally tested both the Create Object and Quick Create Object interfaces with these changes, as well as testing spawning a number of items in both modes - but I don't want to drown this PR with dozens of screenshots.

### DIFF
--- a/code/modules/admin/create_object.dm
+++ b/code/modules/admin/create_object.dm
@@ -8,8 +8,23 @@
 	if (!create_object_html)
 		var/objectjs = null
 		objectjs = jointext(typesof(/obj), ";")
+
+		var/objectnames = null
+		var/list/object_name_list = list()
+		for (var/obj/object as anything in typesof(/obj))
+			// Null and empty names won't display; semicolon names will mess up the generated JS...
+			if (object.name && object.name != "" && object.name != ";")
+				object_name_list.Add(object.name)
+			else
+				// ...so display an appropriate "name" rather than removing the entry.
+				// We want these arrays to be the same length!
+				object_name_list.Add("(NO NAME)")
+		objectnames = jointext(object_name_list, ";")
+		objectnames = replacetext(objectnames, "\"", "\\\"") // Some names have quotation marks in them, so escape them...
+
 		create_object_html = file2text('html/create_object.html')
 		create_object_html = replacetext(create_object_html, "null /* object types */", "\"[objectjs]\"")
+		create_object_html = replacetext(create_object_html, "null /* object names */", "\"[objectnames]\"")
 
 	user << browse(create_panel_helper(create_object_html), "window=create_object;size=425x475")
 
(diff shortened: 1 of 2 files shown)


Skyglazer:
Thank you! This will make finding items to spawn in much easier, especially ones that have a name that doesn't match up with their name in the code, like the crimson fang. Makes it so you don't have to code dive to find something you know about, very good.

Red-Honk:
Bring it.

dwasint:
feel as if this may be better merged into one so searching returns both results sorted by typepath?

GravyMaker:
That change should be possible. If you could ask other admins their opinions - I don't really have a way to reach out to them collectively that isn't just a public chat where I'm likely to be drowned out or ignored - and they agree, I can see about modfiying the code further to behave that way, rather than based on a toggle as it currently is.

RadiantFlash7 (approved):
Better utility for admins is usually appreciated. Thank you.

noelle-lavenza (changes requested):
this is likely incredibly slow, i would like it if the object name list was cached because it'll always be the same afaict

GravyMaker:
You're absolutely right that this is a slow operation, and while there's a little that I could do to make it modestly faster, it ultimately needs to iterate over every object anyway - it's never going to be fast, unfortunately.

That said, the create object menu _does_ actually cache the data - just not the object_name_list directly. Instead, it maintains create_object_html as a _static_ variable, so whatever value is assigned to it actually survives between proc calls, even though it's declared within the proc itself. I didn't notice this at first, either, when I was working on it, and it was something I was planning to change - but it's there, and as a consequence, the page generation logic actually only runs once.

The _quick_ create object menu is similar but a little more complex - it does its generation and caching on a per-list basis: it will generate, say, an HTML form for selecting reagents the first time that option is selected and refer to it in future, and cache a separate HTML form for clothing the first time _that_ list is accessed. As in the regular create object menu, the HTML generation logic only runs for a list which doesn't already exist, as you'd hope.

So, ultimately - yes, the operation is expensive, but the code does actually already cache the data; it's just not obvious about it.

noelle-lavenza (approved):
> You're absolutely right that this is a slow operation, and while there's a little that I could do to make it modestly faster, it ultimately needs to iterate over every object anyway - it's never going to be fast, unfortunately.
> 
> That said, the create object menu _does_ actually cache the data - just not the object_name_list directly. Instead, it maintains create_object_html as a _static_ variable, so whatever value is assigned to it actually survives between proc calls, even though it's declared within the proc itself. I didn't notice this at first, either, when I was working on it, and it was something I was planning to change - but it's there, and as a consequence, the page generation logic actually only runs once.
> 
> The _quick_ create object menu is similar but a little more complex - it does its generation and caching on a per-list basis: it will generate, say, an HTML form for selecting reagents the first time that option is selected and refer to it in future, and cache a separate HTML form for clothing the first time _that_ list is accessed. As in the regular create object menu, the HTML generation logic only runs for a list which doesn't already exist, as you'd hope.
> 
> So, ultimately - yes, the operation is expensive, but the code does actually already cache the data; it just doesn't do it in quite the way you'd typically see.

oh, that's perfect then

RadiantFlash7 (approved):
No issues reported in test merge, code looks good enough for our expectations.

PR #2771, opened 2025-04-26
Ports Azure's text formatting.

ARefrigerator:
## About The Pull Request
Port's Free's text formatting over to here https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/1720

**HOW TO USE:**

- Requires a start and end example: `This is +bold text+` will output: This is **bold text**
- bold text symbol: `+`
- Italics text symbol: `|`
- underline text symbol: `_`

Note: The only issue I've seen (and Free also mentioned this) is that you cannot start a sentence with a bold word because it conflicts with using + for whispers. We _could_ change whisper to # but would need to make it very clear for people until they get used to this or find some way to check a whole block of text for a second + in some edge case. We could also change the bold char symbol but I'd rather keep it consistent between the servers for clarity sake (and because I play on both and would go mad trying to remember which is which)

## Why It's Good For The Game
Fromatting is fun.

image

### DIFF
--- a/code/__HELPERS/text.dm
+++ b/code/__HELPERS/text.dm
@@ -446,25 +446,25 @@ GLOBAL_LIST_INIT(binary, list("0","1"))
 			end = temp
 	return end
 
-/proc/parsemarkdown_basic_step1(t, limited=FALSE)
+/proc/parsemarkdown_basic_step1(t, limited=FALSE, barebones = FALSE)
 	if(length(t) <= 0)
 		return
 
 	// This parses markdown with no custom rules
 
 	// Escape backslashed
-
-	t = replacetext(t, "$", "$-")
-	t = replacetext(t, "\\\\", "$1")
-	t = replacetext(t, "\\**", "$2")
-	t = replacetext(t, "\\*", "$3")
-	t = replacetext(t, "\\__", "$4")
-	t = replacetext(t, "\\_", "$5")
-	t = replacetext(t, "\\^", "$6")
-	t = replacetext(t, "\\((", "$7")
-	t = replacetext(t, "\\))", "$8")
-	t = replacetext(t, "\\|", "$9")
-	t = replacetext(t, "\\%", "$0")
+	if(!barebones)
+		t = replacetext(t, "$", "$-")
+		t = replacetext(t, "\\\\", "$1")
+		t = replacetext(t, "\\**", "$2")
+		t = replacetext(t, "\\*", "$3")
+		t = replacetext(t, "\\__", "$4")
+		t = replacetext(t, "\\_", "$5")
+		t = replacetext(t, "\\^", "$6")
+		t = replacetext(t, "\\((", "$7")
+		t = replacetext(t, "\\))", "$8")
+		t = replacetext(t, "\\|", "$9")
+		t = replacetext(t, "\\%", "$0")
 
 	// Escape  single characters that will be used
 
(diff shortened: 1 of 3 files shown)


Skyglazer:
This is purely a personal preference, but I tend to use /forward slashes/ when doing text that I would write as italics in-game, and -hyphens- when I would do something that I'd put in text. It'd be nice if you could have them as additional formatters, so people could do it in multiple different ways. Getting bold off of + would also free up + in order to avoid the whisper problem. I understand you play on both servers though, so it's tough to make a suggestion that would fit your plight... I think doing a check for a second + would be reasonable-- if one is found, and no subsequent ones, then assume that the text was meant to be bolded. If you can find some way to do that, great- if not, I think that's fine. Not everyone uses say to do whispers, anyway, and those who use + can surely see the value markdown adds- having to get used to the switch will be a worthy sacrifice of effort.

ARefrigerator:
> This is purely a personal preference, but I tend to use /forward slashes/ when doing text that I would write as italics in-game, and -hyphens- when I would do something that I'd put in text. It'd be nice if you could have them as additional formatters, so people could do it in multiple different ways. Getting bold off of + would also free up + in order to avoid the whisper problem. I understand you play on both servers though, so it's tough to make a suggestion that would fit your plight... I think doing a check for a second + would be reasonable-- if one is found, and no subsequent ones, then assume that the text was meant to be bolded. If you can find some way to do that, great- if not, I think that's fine. Not everyone uses say to do whispers, anyway, and those who use + can surely see the value markdown adds- having to get used to the switch will be a worthy sacrifice of effort.

I probably don't mind that much when it comes to which keys are used. Which ever ones seem more common and right I'll try to use those ones. I don't think it will be that big of a deal so long as it's mentioned somewhere for people to remember.

GravyMaker:
Perhaps if this gets approval from a maintainer, it might be wise to have a staff member create an advisory poll to see what's most popular? Ordinarily I'd counsel that design considerations come first (and that's still true in this case - eg. don't present the community with options that are technically infeasible), but in this case getting people's opinions would probably be for the best.

ARefrigerator:
Zoni might be trying something and fixing some of the syntax choices among other things so ask him before doing anything with this PR.

GravyMaker:
The change I want to make won't be affected by this, having looked into it - so this can be merged or testmerged whenever. It works pretty much perfectly as-is, with one exception, which is that you can't have a message be bolded from the very beginning unless you add some leading whitespace to it.

The reason for this is that the bold delimiter, +, is the same as one of two whisper indicators (# and +). As such, a message beginning with the bold delimiter will be interpreted as a whisper, which happens to have a + in it.

This can be fixed in /mob/living/say(), defined at code/core/modules/mob/living/say.dm:84. Specifically, you're looking at the procs called on lines 115 (get_message_mode, which makes a message a whisper if you use the '#' indicator), and line 143 (check_whisper, which checks for '+' instead). Why these _both_ check whether the message is a whisper using different characters is beyond me, but there it is - either one of these (but obviously only one) can be removed. However, this might bother some people, so it may be wiser to change the delimiter instead from + to something else.

Unfortunately, using a single asterisk isn't possible, due to the emote-and-say functionality using it as a separator - there's no way to use it to indicate formatting _and_ to separate the emote from the message; it has to be one or the other.

Personally, I think Skyglazer's suggestion of forward-slashes to indicate italicisation is sensible and intuitive, as is the existing use of _ to delimit underscores. The only real problem is bold text - + would be fine, if not for it also being a whisper indicator. Again, I'd put it to a poll - but even as someone who prefers + over # for whispers, I'd _personally_ be willing to adapt in the name of having proper markdown functionality.

RadiantFlash7 (approved):
Neat change, hopefully lets folks be a bit more emphatic in their roleplay.

It might be down to some peoples personal preference as to what will format for what, but this set up is pretty standard to all ss13 servers. Might as well keep what folks are familiar with.

Keazra:
I kind'a don't like this. I'd rather just use ** or // or <> and who cares if people see them, they get the jist..

PR #2777, opened 2025-04-30
Update knave.dm

lalaria1:
## About The Pull Request
Buffs knave a bit

## Why It's Good For The Game
To my surprise knave was ridiculously weak, even refugee thief had better lockpicking than they do, so I brought them up to expert lockpicking. I also removed the -3 int, I do not understand why someone thought to punish knave so much with it, because it means any faint would go through and that means you cannot dodge. I gave them 1 point of endurance so you don't run out of breath so quickly when running since they have 0 end.

If someone knows why knave was so weak I'd be happy to listen, otherwise these changes seem reasonable to me.

I played the role and then ran some tests in my head, and it seems to do well.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/bandits/types/knave.dm
+++ b/code/modules/jobs/job_types/roguetown/bandits/types/knave.dm
@@ -28,7 +28,7 @@
 	H.mind.adjust_skillrank(/datum/skill/misc/sneaking, 4, TRUE)
 	H.mind.adjust_skillrank(/datum/skill/misc/stealing, 4, TRUE)
 	H.mind.adjust_skillrank(/datum/skill/misc/athletics, 3, TRUE)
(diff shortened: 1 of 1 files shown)


RadiantFlash7:
Bandit knave, disregarding the current test merged balance PR which this PR does not account for, is one of the STRONGEST bandit classes there is. Buffing them further, is definitely not happening.
Denied.

Vasili59:
He was intellectually weak due to his evasion skill, as well as his high proficiency in knife fighting, which allowed him to consistently stab eyes and blind his victims, turning them into punching bags. And let me clarify—the steel dagger had an increased armor-piercing chance, making masks useless against it. Meanwhile, his mastery of evasion made him a target that was nearly impossible to hit by conventional means. Before exhausting himself completely in the first assault, he managed to kill knights, the entire vanguard, and templars—twice over. (Such cases did occur.)

PR #2779, opened 2025-04-30
Datumizes stew code.

ARefrigerator:
## About The Pull Request
Does the following:
- makes pots 66oz (six bowls, requires 6 items, more on that later)
- new datum system for cooking stew recipes

The new data for the pot recipes makes it easy to edit and manipulate. You can also add non food items to create reagent types or which ever you'd like in the future.
The example here is `/datum/pot_recipe/poo` where it accepts an `/obj/item/natural` item

## Why It's Good For The Game
Stew has been broken in weird ways for a while. This allows people to make a filling meal, albeit not exactly great, by making stew once more with a straight forward system of 1 food item per bowl of stew. This is rather in line with what other food currently does due to a lot of it mostly existing to fill stomachs without giving any buffs.

This also allows us to try adding other things into stew that aren't traditionally food easily and may also allow us to make use of the `end_cooking` proc to perhaps spawn items on creation at the pot's location (it could be how we make spaghetti or rice if we ever add stuff like that.)

boil check
image

water check
image

6 bowls from one pot
image

### DIFF
--- a/code/modules/roguetown/roguejobs/cook/tools/pot.dm
+++ b/code/modules/roguetown/roguejobs/cook/tools/pot.dm
@@ -105,7 +105,7 @@ obj/item/reagent_containers/glass/bucket/pot/proc/boil()
 		var/list/things = STR.contents()
 		if(!length(things))
 			// Should be the termination point if a pot with no items is idle on a hearth
-			addtimer(CALLBACK(src, PROC_REF(boil)),4 SECONDS)
+			addtimer(CALLBACK(src, PROC_REF(boil)),1 SECONDS)
 			return
 
 		for(var/obj/item/I in things)


code/modules/roguetown/roguejobs/cook/tools/pot.dm
```diff
@@ -45,3 +51,185 @@
...
+		if(istype(I, R.input))
+			if(R.check_can_cook())
+				return R
+	return FALSE
+
+/*========
+Pot Recipe
+========*/
+/datum/pot_recipe
+	var/obj/item/reagent_containers/glass/bucket/pot/cooking_pot
```
noelle-lavenza, line 95: this variable is unused and should be removed if you're using a global recipe list

code/modules/roguetown/roguejobs/cook/tools/pot.dm
```diff
@@ -45,3 +51,185 @@
...
+			add_sleep_experience(user, /datum/skill/craft/cooking, user.STAINT)
+			qdel(input)
+			recipe.start_cooking(src, user)
+	else
+		to_chat(user, span_warning("There is not enough water in the pot."))
+
+
+// TO DO: Find a better place for this stuff eventually
+GLOBAL_LIST_INIT(pot_recipes, init_subtypes(/datum/pot_recipe))
+#define STEW_COOKING_TIME 60 SECONDS
```
noelle-lavenza, line 81: i don't see why this is a define if it's also a var on `/datum/pot_recipe`. could do `var/const/BASE_STEW_TIME = 1 MINUTES` *defined on `/datum/pot_recipe` if you really want but i think it's overkill

noelle-lavenza: also is it intended that you didn't copy the stew timings from the various recipes to the new system?

code/modules/roguetown/roguejobs/cook/tools/pot.dm
```diff
@@ -45,3 +51,185 @@
...
+		return
+	if(!pot.reagents.has_reagent(/datum/reagent/water, volume_to_replace)) //if we somehow snuck in or some water level changed before we fired just fucking DIE
+		end_cooking()
+		return
+
+	// One final sanity check in case some lunatic empties some water but still keeps cooking and didn't hit 0 water or null reagents
+	var/true_volume_to_remove = volume_to_replace
+	var/checker = pot.reagents.get_reagent_amount(/datum/reagent/water)
+	if(checker < volume_to_replace)
+		true_volume_to_remove = checker
```
noelle-lavenza, line 136: your code just does this but more verbose, i recommend using `min()` instead
```suggestion
	var/true_volume_to_remove = min(volume_to_replace, pot.reagents.get_reagent_amount(/datum/reagent/water))
```

code/game/objects/lighting/roguelighting.dm
```diff
@@ -635,104 +635,17 @@
...
-						pot.reagents.add_reagent(/datum/reagent/consumable/soup/stew/chicken, 18)
-						pot.reagents.remove_reagent(/datum/reagent/water, 1)
-					else
-						qdel(W)
-						sleep(900)
-						playsound(src, "bubbles", 30, TRUE)
-						pot.reagents.add_reagent(/datum/reagent/consumable/soup/stew/meat, 18)
-						pot.reagents.remove_reagent(/datum/reagent/water, 1)
-				return
+			if(P.reagents.chem_temp < 374)
```
noelle-lavenza, line 640: i would really suggest creating a `T100C` define in `code/__DEFINES/atmospherics.dm` right after `T0C` and `T20C`, then using that instead. `#define T100C 373.15`
no need to replace every use of `374` with that though, just the one here

code/game/objects/lighting/roguelighting.dm
```diff
@@ -635,104 +635,17 @@
...
+				to_chat(user, "<span class='notice'>The pot isn't boiling yet.</span>")
```
noelle-lavenza, line 641:
```suggestion
				to_chat(user, span_notice("The pot isn't boiling yet."))
```

code/modules/roguetown/roguejobs/cook/tools/pot.dm
```diff
@@ -45,3 +51,185 @@
...
+
+/* Handle recipes */
+/obj/item/reagent_containers/glass/bucket/pot/proc/check_for_recipe(obj/item/input, mob/living/user)
+	var/datum/pot_recipe/recipe = select_pot_recipe(GLOB.pot_recipes, input)
+	if(!recipe)
+		return FALSE
+	if(!recipe.output)
+		return FALSE
+
+	var/reserve_water = 0
```
noelle-lavenza, line 64: in the old system water was consumed when cooking starts (after the do_after finishes and when the ingredient is deleted), is there a reason this doesn't do that anymore?

noelle-lavenza (changes requested):

code/modules/roguetown/roguejobs/cook/tools/pot.dm
ARefrigerator: I didn't know they had different times. How long is sleep(600) is that still deciseconds?

ARefrigerator: I made the times back to what they were, assuming it's deciseconds.

code/modules/roguetown/roguejobs/cook/tools/pot.dm
ARefrigerator: I'm gonna mark this as resolved too unless something brings up a good reason for why it shouldn't be that way.

E231kei500bandai: you could pour out the pot but then have more soup randomly appear in the pot

E231kei500bandai:
there should also be a feedback message when something is cooking in the pot imo

code/modules/roguetown/roguejobs/cook/tools/pot.dm
E231kei500bandai: well technically you havent considered a lunatic pouring out the pot

and then filling it up again

and then havoung soup randomly appear

code/modules/roguetown/roguejobs/cook/tools/pot.dm
```diff
@@ -45,3 +51,185 @@
...
+	if(!pot.reagents.has_reagent(/datum/reagent/water, volume_to_replace)) //if we somehow snuck in or some water level changed before we fired just fucking DIE
+		end_cooking()
+		return
+
+	// One final sanity check in case some lunatic empties some water but still keeps cooking and didn't hit 0 water or null reagents
+	var/true_volume_to_remove =  min(volume_to_replace, pot.reagents.get_reagent_amount(/datum/reagent/water))
+
+	var/temp = pot.reagents.chem_temp
+	pot.reagents.remove_reagent(/datum/reagent/water, true_volume_to_remove)
+	pot.reagents.add_reagent(output, true_volume_to_remove, list(), temp) //last value should stop pot losing temp.
```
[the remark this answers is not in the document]
ARefrigerator: The problem with that, at least with the current system, is you would add so many things it would start to confuse people leaving them with half a pot of stew and half full of water unsure of why it was occurring.

ARefrigerator:
> and one last thing you could remove the pot from the heat source and it would still cook which is slightly unrealistic but not really that big of a deal and probably impossible to add

It isn't impossible but it would require the objects perhaps stay inside the pot, and then the pot itself calls the procs when hot enough. I think that makes sense but the system would then perhaps turn into more like the ovens where they increment over time until they hit a value and then fire. I don't mind that much. I have done that in the past so maybe once the datum is at least in I could think about rewriting it to be that way in the future.

ARefrigerator:
You know what maybe I might change it. This system is nice but it kind of doesn't work for the edge cases.

ARefrigerator:
I rewrote a bunch of stuff this may need to be checked again. Notably I had to add one check for the pot storage component to include ` /obj/machinery/light/rogue/hearth` to let it enter.

If anything looks terrible let me know. I'd also like to get this working maybe so people could get stuff out from a pot on a hearth without taking out a pot but right now it's probably fine.
```
/*
/obj/machinery/light/rogue/hearth/MouseDrop_T(atom/movable/O, mob/user)
	if(attachment)
		if(istype(attachment, /obj/item/reagent_containers/glass/bucket/pot))
			var/obj/item/reagent_containers/glass/bucket/pot/P = attachment
			var/datum/component/storage/STR = P.GetComponent(/datum/component/storage)
			if(STR)
				STR.user_show_to_mob(user, TRUE)
			return
	. = ..()
*/
```

ARefrigerator:
**This probably has runtime errors I gotta fix tomorrow probably involving items that have reagents but don't give results. I'm not sure I'll do proper checks then.**

Image of the inventory:
image

E231kei500bandai:
@ARefrigerator this system would be really good for making tea once its merged, you could even make it so that tea leaves can be reused if you add an override for the code that deletes it
just need tea sprites for the leaves and teapot

ARefrigerator:
> @ARefrigerator this system would be really good for making tea once its merged with a teapot that only accepts tea leaves to be stored into it, you could even make it so that tea leaves can be reused if you add an override for the code that deletes it just need tea sprites for the leaves and teapot

I'll definitely think about adding a teapot and expanding on some stuff like that. We do have tea currently it just had to be imported and then grinded up with a mortar and pestle. I'll get more into double checking this code today since I was fixing up the food one first.

dwasint:
> Only issue left with my stew code (it's not that major) is that you can see items in pots if they are on the hearth via mouse drop onto your mob but you can't take them out... I don't know what part of storage component procs I have to overwrite but if you do let me know and I can do it. otherwise I think this is pretty much done.

you'd likely need to modify your calls to COMSIG_TRY_STORAGE_TAKE or remove_from_storage depending on the route you go

ARefrigerator:
Okay I can try giving it a shot. I guess while the PR is also sitting here I could try making it so items with no stew datums could try rendering down into their reagents while the pot is active.

ARefrigerator:
> A few things I've found that may be related to this PR:
> 
>     * I once dropped my torch in a pot on a hearth and couldn't get it out even after throwing the pot on the floor.
> 
>     * The bubbling/boiling sound effect seems to be constantly running when you approach the pot, even after it's off the hearth. Stop sounds works temporarily but it starts up again.

boiling continues even after removing from the hearth until the pot cools back down enough to no longer be at a boil. It should eventually be stopping. I can check the torch thing.

ARefrigerator:
A few issues should be fixed now with the pot with the recent commit. Still not sure how to create the torch being stuck in it issue. Did you click on the torch icon itself? I got it out just fine.

ARefrigerator:
ok I think I fixed every issue someone could have with this regarding temperature heat.

RadiantFlash7 (approved):
Any issues found during the test merge have been resolved, and the PR is approved for merging.

PR #2781, opened 2025-05-01
Adds a new ears and snout sprite!

Fn-OS:
## About The Pull Request
Adds a new snout and ears sprite! Called Feline, short and Broad (ears). 

## Why It's Good For The Game
More customization! That's always good I think.

Screenshot (879)

### DIFF
--- a/code/modules/client/customizer/customizers/organ/ears.dm
+++ b/code/modules/client/customizer/customizers/organ/ears.dm
@@ -48,6 +48,7 @@
 		/datum/sprite_accessory/ears/cat_big,
 		/datum/sprite_accessory/ears/cat_normal,
 		/datum/sprite_accessory/ears/lynx,
+		/datum/sprite_accessory/ears/broad,
 		)
 
 /datum/customizer/organ/ears/axian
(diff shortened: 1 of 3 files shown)


Fn-OS:
I think this is good to go.

ARefrigerator:
Is the comma at the end of the lists causing any issue? I don't think it would but it's not making an empty entry on the end of every list is it?

Fn-OS:
> Is the comma at the end of the lists causing any issue? I don't think it would but it's not making an empty entry on the end of every list is it?

They're supposed to be there as far as I know. They got deleted in earlier testing but I did re-add them.

Fn-OS:
This is ready to merge. Code's been tested several times and it all works

noelle-lavenza:
> Is the comma at the end of the lists causing any issue? I don't think it would but it's not making an empty entry on the end of every list is it?

It doesn't do that (anymore), in the current BYOND version (and all the ones for the past several years) all it does is reduce diff noise when you add new entries, because you don't have to add a comma to any existing lines.

Gyrandola (approved):

PR #2780, opened 2025-05-01
Church's research wind sprites + floors + flowers to wear

DelineFortune:
## About The Pull Request

sprites and a bit empty code

## Why It's Good For The Game

things to add before my map prs (dungeons and church rework and antags rework)

## Proof of Testing (Required)

image

PR #2782, opened 2025-05-01
Sewers dungeon, church research wing, more tunnels around the city

DelineFortune:
after https://github.com/Rotwood-Vale/Ratwood-Keep/pull/2780

added: church statues of Ten (put valuable things inside to get blessing upon the city put shunned races blood to bring curses, its just sprites for now)
image
image
image
more tunnels to lurk around church
pile of power: 
image
its just a sprite from now but I want to let dukes to put money into and force buy antag waves
shit sewers dungeon will make it less slop if Ezra going to approve it
image


DelineFortune:
> Nice idea with adding more ways around town with tunnels, not sure how needed is big tunnel area in the west though. tunnels that go around gates to church look promising suggest thinking of some context and environment story telling for them and see how abusable they will be in the game. Not sure if adding more space to church is needed. Maybe this area with idols can be placed in bellow church in the library area instead?

People do use library area a lot to read books/have private conversations
Also its wide open. Statues are supposed to be protected because there will be a way to use shunned blood and corrupt statues. 
Their top side is also +2 spots for rogues to get into

NikolayByondPlayer:
> > Nice idea with adding more ways around town with tunnels, not sure how needed is big tunnel area in the west though. tunnels that go around gates to church look promising suggest thinking of some context and environment story telling for them and see how abusable they will be in the game. Not sure if adding more space to church is needed. Maybe this area with idols can be placed in bellow church in the library area instead?
> 
> People do use library area a lot to read books/have private conversations Also its wide open. Statues are supposed to be protected because there will be a way to use shunned blood and corrupt statues. Their top side is also +2 spots for rogues to get into Big west tunnel leads into the water. Its safe place to jump down

image
I mean this giant tunnels - if you want more tunnels and underground better dig even deeper one z level then current sewer area, also you can add more sewer areas around the current one 

Like here:
image
image
Also better to give some context to what this place is (just reminding I know it is a draft yet)

For the praying room maybe you can play it out like and old wing of church like this or smth
image
I still think that this wing area can be:

a) made smaller since it's a bit too big now and would be a bit of pain to patrol I think 
b) could have another exit that leads to the main church area if it won't be made smaller since one big Г moves can be a bit detrimental to flow I think

I also like your garden - maybe you can change this place to yours? 
image

PR #2621, opened 2025-03-25
Azure ritual port 

DelineFortune:
## About The Pull Request

https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/1373 same but no PACIFIST  in Eora ritual and no  longstrider in Dendor ritual

## Why It's Good For The Game

promised rituals 9 months back

## Proof of Testing (Required)

image

PR #2783, opened 2025-05-01
third time azure rituals port

DelineFortune:
my man why have i wait for a month to readd it  again its not funny
https://github.com/Rotwood-Vale/Ratwood-Keep/pull/2621

### DIFF
--- a/code/modules/rituals/ritual_types/divine/ritualcircles.dm
+++ b/code/modules/rituals/ritual_types/divine/ritualcircles.dm
@@ -14,14 +14,14 @@
 		playsound(loc, 'sound/foley/cloth_wipe (1).ogg', 100, TRUE)
 		qdel(src)
 
-// This'll be our tutorial ritual for those who want to make more later, let's go into details in comments, mm? - Onutsio 
+// This'll be our tutorial ritual for those who want to make more later, let's go into details in comments, mm? - Onutsio
 /obj/structure/ritualcircle/astrata
 	name = "Rune of the Sun" // defines name of the circle itself
-	icon_state = "astrata_chalky" // the icon state, so, the sprite the runes use on the floor. As of making, we have 6, each needs an active/inactive state. 
+	icon_state = "astrata_chalky" // the icon state, so, the sprite the runes use on the floor. As of making, we have 6, each needs an active/inactive state.
 	desc = "A Holy Rune of Astrata" // description on examine
 	var/solarrites = list("Guiding Light") // This is important - This is the var which stores every ritual option available to a ritualist - Ideally, we'd have like, 3 for each God. Right now, just 1.
 
-/obj/structure/ritualcircle/astrata/attack_hand(mob/living/user) 
+/obj/structure/ritualcircle/astrata/attack_hand(mob/living/user)
 	if((user.patron?.type) != /datum/patron/divine/astrata)
 		to_chat(user,span_smallred("I don't know the proper rites for this..."))
 		return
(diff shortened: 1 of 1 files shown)


DelineFortune:
proof that its fine 
image

DelineFortune:
better
sprite

DelineFortune:
image
better sprite ok

code/modules/rituals/ritual_types/divine/ritualcircles.dm
```diff
@@ -0,0 +1,290 @@
...
+							loc.visible_message(span_warning("[user] opens their mouth, disgorging a great swarm of flies!"))
+							playsound(loc, 'sound/misc/fliesloop.ogg', 100, FALSE, -1)
+							flylordstriage(src)
+							user.apply_status_effect(/datum/status_effect/debuff/ritesexpended)
+							spawn(120)
+								icon_state = "pestra_chalky"
+
+/obj/structure/ritualcircle/pestra/proc/flylordstriage(src)
+	var/ritualtargets = view(0, loc)
+	for(var/mob/living/carbon/human/target in ritualtargets)
```
noelle-lavenza, line 148:
```suggestion
	for(var/mob/living/carbon/human/target in view(0, src))
```
don't use a var for it, but also you could just do `in loc` if you're just doing view(0).

code/modules/rituals/ritual_types/divine/ritualcircles.dm
```diff
@@ -0,0 +1,290 @@
...
+							loc.visible_message(span_warning("[user] snaps their head upward, they let out a howl!"))
+							playsound(loc, 'sound/vo/mobs/wwolf/howl (2).ogg', 100, FALSE, -1)
+							lesserwolf(src)
+							user.apply_status_effect(/datum/status_effect/debuff/ritesexpended)
+							spawn(120)
+								icon_state = "dendor_chalky"
+
+/obj/structure/ritualcircle/dendor/proc/lesserwolf(src)
+	var/ritualtargets = view(1, loc)
+	for(var/mob/living/carbon/human/target in ritualtargets)
```
noelle-lavenza, line 196: once more. i'm not gonna go through and point out every case where this pattern is done but they do all need to be fixed.

noelle-lavenza (changes requested):
a lot of the do_after()s and spawn()s need to be changed to use the `SECONDS` macro. just divide by ten and use SECONDS, so `spawn(120)` is `spawn(12 SECONDS)`, etc. plus, don't store the results of `view()` in a list if you only use it to loop over, that's inefficient and a bad pattern.

Skyglazer:
Do you think that refugee ecclesiasts could get access to rituals too? Maybe they shouldn't spawn with a feather, but if they can get one (or have someone put down a rune somewhere), it would make sense for them to be able to do them. Only other role I can think to check is that the Heartfelt Prior (priest in the heartfelt migrant wave) should probably have a feather and ability to use it.

DelineFortune:
> Do you think that refugee ecclesiasts could get access to rituals too? Maybe they shouldn't spawn with a feather, but if they can get one (or have someone put down a rune somewhere), it would make sense for them to be able to do them. Only other role I can think to check is that the Heartfelt Prior (priest in the heartfelt migrant wave) should probably have a feather and ability to use it.

Maybe and adventurers clerics as well
I want to upgrade the system, rework cult with it later and make leathers craftable

RadiantFlash7 (approved):
Haven't had any issues with this test merge, Approved for merging.

PR #2784, opened 2025-05-01
Fixing proc params for equipped

ARefrigerator:
## About The Pull Request
Adds the two proc params to some calls I forgot earlier. These might fix some of the runtimes that show up. I'm not sure but the two examples I 

## Why It's Good For The Game
runtimes are bad.

I don't really have proof but hopefully it should stop any `runtime error: bad arg name 'silent'` from showing up now.

The two examples I did see seemed to be right clicking keyrings off the floor, which I think went through the not_handled statement and spectacles equip and I can't get them to runtime when I tried. 
image

### DIFF
--- a/code/modules/mob/living/carbon/human/inventory.dm
+++ b/code/modules/mob/living/carbon/human/inventory.dm
@@ -218,7 +218,7 @@
 
 	//Item is handled and in slot, valid to call callback, for this proc should always be true
 	if(!not_handled)
-		I.equipped(src, slot)
+		I.equipped(src, slot, initial)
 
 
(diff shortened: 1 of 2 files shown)


ARefrigerator:
Yeah I just tried with a different branch, I got the error:
image

This PR definitely fixes the runtime. Should be safe to merge but it's also probably small enough to TM for a bit first just in case.

noelle-lavenza (changes requested):
only that one edit was necessary

RadiantFlash7 (approved):
Runtime fixing. Approved!

PR #2785, opened 2025-05-01
Fixes division by zero runtime in lockpicking

ARefrigerator:
## About The Pull Request
lockpicking tries to divide by difficulty. Some doors do not have it. Causes a runtime. I just make those doors default to the easiest difficulty if it happens.

## Why It's Good For The Game
h8 runtimes

before:
image

after (no runtime):
image


ARefrigerator:
okay I added in the value change too. I'm keeping the check just in case something ever happens to make it zero though.

RadiantFlash7 (approved):
Standard fix for runtimes running on the server; Approved.

ARefrigerator:
Some new updates. turns out they divide against difficulty for fractions and multiply by ten when they see fit... this is the fastest solution but it's sloppy...

Sometimes they multiply the value even though 0 is the strongest and 6 is the weakest? Anyway this flips this value until something gets done. It should in theory fix lockpicks breaking all the time on easy locks.

PR #2786, opened 2025-05-01
adds budget2change ismob check

ARefrigerator:
## About The Pull Request
Simple check in `budget2change` to check if there's a mob before putting coins into hands

## Why It's Good For The Game
Should stop the balloon pad runtime error

I did...


noelle-lavenza (approved):
i don't think you even need an ismob, just `if(user)` is enough, but this'll do

RadiantFlash7 (approved):
Changes Approved. Good bugfix, fixes existing runtimes

PR #2791, opened 2025-05-03
Fix various grab-related bugs

noelle-lavenza:
## About The Pull Request
- Cleans up grabbing code, optimises it, removes some redundant/unused code, and makes it use lazylists.
- Fixes some incorrectly-written code that was *supposed* to make aggressive grabs give you perfect accuracy. May need balance tweaking but it was intended in the original Roguetown code so I figure it's probably fine.
- Fixes losing grabs to mobs moving diagonally, as a band-aid until simplemobs can be made to not move diagonally.

## Why It's Good For The Game
Fixes some grab bugs.

I tested these on Azure, don't have the screenshots/recordings on-hand though. It never got PR'd there because I quit around the same time.


RadiantFlash7 (approved):
No issues reported during test merge, code fixes seem relatively fine.

PR #2794, opened 2025-05-03
Buff NPC pathfinding and AI

noelle-lavenza:
## About The Pull Request
Buffs NPC pathfinding and AI.
- NPCs can now path up and down stairs.
- NPCs should no longer move diagonally (except when fleeing from being in too much pain, because that still uses BYOND's built-in pathfinding).
- NPCs will now take into account slowdown from turf type when pathing. This even includes using flowing water to travel upstream quicker, and avoiding entering it entirely if it's going the wrong way (unless there's no better option).
- NPCs fleeing in pain should work more reliably, and has a lower threshold.
- NPCs should now resist/stand up/etc. properly.
- Attacking an NPC interrupts their attempts to do timed actions like climbing (except for resisting or standing up).
- Humanmob NPCs can now path through unlocked doors.
- Mobs with the door-bashing trait can now path through doors and will bash them down, even if locked. Scary!
- **MOBS NOW ATTEMPT TO GRAB THEIR ENEMIES, UPGRADE GRABS, AND TWIST/SMASH/CHOKE/ETC.**
- Converted deadites will not have skills.
- Deadite AI has been completely rewritten because it was garbage before.
- In preparation for switching to the AI controller system for humanmobs, a lot of the AI logic has been pulled out into separate overrideable procs.
- Goblins now lose stamina.

Also some miscellaneous related tweaks:
- Resisting an embed-grab will return the weapon to the wielder on success and twist the weapon in the wound on failure. This is so that goblins can't run off with your sword in them, and to make up for embeds stopping bleed.
- Reduces the 6 second cooldown between opening separate doors to 0.3 seconds (there is still a 2 second cooldown between bumps on a single door). This is because that was stupid.
- Increases the maximum pain on chest and head, because otherwise NPCs had trouble hitting the pain threshold from attacks just targeting one zone.
- Adds additional woundpain to special wounds.
- Unarmed attacks now properly use stamina; stamina drain amounts were defined but not actually checked. Punches have had their stamina drain on hit lowered from 5 to 1 (10 to 5, for swift/strong punches) to compensate since the meta is currently "punches are good because they don't drain stam".

Also-also, fixes an issue with zwebs that was introduced when I optimised it on Azure. My bad.

## Why It's Good For The Game
No more stair cheese.

EiJXWdU27i-ezgif com-optimize

btLoOo3ZbY

### DIFF
--- a/code/_onclick/other_mobs.dm
+++ b/code/_onclick/other_mobs.dm
@@ -28,11 +28,16 @@
 	if(SEND_SIGNAL(src, COMSIG_HUMAN_EARLY_UNARMED_ATTACK, A, proximity) & COMPONENT_NO_ATTACK_HAND)
 		return
 	SEND_SIGNAL(src, COMSIG_HUMAN_MELEE_UNARMED_ATTACK, A, proximity)
+	var/rmb_stam_penalty = 0
+	if(istype(rmb_intent, /datum/rmb_intent/strong) || istype(rmb_intent, /datum/rmb_intent/swift))
+		rmb_stam_penalty = 4 // as opposed to 10 for a weapon; these are your hands, it's easier to move them
 	if(isliving(A))
 		var/mob/living/L = A
 		if(!used_intent.noaa)
 			playsound(get_turf(src), pick(GLOB.unarmed_swingmiss), 100, FALSE)
 //			src.emote("attackgrunt")
+		if(used_intent.releasedrain)
+			stamina_add(used_intent.releasedrain + rmb_stam_penalty)
 		if(L.checkmiss(src))
 			return
 		if(!L.checkdefense(used_intent, src))
@@ -61,6 +66,8 @@
 						visible_message(span_warning("[src] pushes [AM]."))
 					changeNext_move(CLICK_CD_MELEE)
 					return
+		if(used_intent.releasedrain)
+			stamina_add(used_intent.releasedrain + rmb_stam_penalty)
 		A.attack_hand(src, params)
 
 /mob/living/rmb_on(atom/A, params)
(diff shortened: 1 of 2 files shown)


Gyrandola:
Really nice

Gyrandola:
Astar aside felt a bit out of my depth reviewing the pathfinding code but I trust you. TMing.

Genevesh:
Is it possible to add the AI disabling when controlled by a player? Admin events and probably future content could really benefit from this given we already have quite a few playable simplemobs.

dwasint:
I'd recommend doing some tests with ai_controller mobs in particular, since you have raw aStar for all movement on mobs. I would recommend they use a hybrid where they either swap between astar and no astar when a player isn't nearby or not have mobs process when far away, because astar gets really expensive mainly from the proc overhead

code/game/turfs/open/openspace.dm
```diff
@@ -95,6 +95,15 @@ GLOBAL_DATUM_INIT(openspace_backdrop_one_for_all, /atom/movable/openspace_backdr
...
 
+/turf/open/transparent/openspace/can_traverse_safely(atom/movable/traveler)
+	var/turf/destination = GET_TURF_BELOW(src)
+	if(!destination)
+		return TRUE // this shouldn't happen
+	if(!traveler.can_zTravel(destination, DOWN, src)) // something is blocking their fall!
+		return FALSE
+	if(!traveler.can_zFall(src, DOWN, destination)) // they can't fall!
+		return FALSE
+	return TRUE
```
dwasint, line 106: doesn't this just mean they can always traverse openspace? which means mobs will like leap off ledges to get across gaps as its a cheaper path? for on z-level movements?

Windryn:
I will promptly use this to be a coward, and pull all the mobs chasing me into town so the guards can deal with them. I'm going to turn ratwood into Kenshi.

RadiantFlash7:
TM removed on request.

PR #2801, opened 2025-05-06
summon adjusting loot and required items

ARefrigerator:
## About The Pull Request
Does the following:

- makes all summons consistent requiring 2 of the BASIC materials needed for summoning.
- increases health and damage of behemoth, dryad, and void dragon
- fixes broken drops on some summons: behemoth, watcher, fiend

## Why It's Good For The Game
Most of this was oversights being fixed.

https://github.com/user-attachments/assets/85a79a09-4333-437d-8c8f-1fc4d599cba3

### DIFF
--- a/code/modules/mob/living/simple_animal/rogue/creacher/summoned/elemental.dm
+++ b/code/modules/mob/living/simple_animal/rogue/creacher/summoned/elemental.dm
@@ -274,11 +274,11 @@
 	mob_biotypes = MOB_ORGANIC|MOB_BEAST
 	ranged = TRUE
 	projectiletype = /obj/projectile/earthenfist
-	health = 800
-	maxHealth = 800
+	health = 1000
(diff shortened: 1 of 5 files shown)


RadiantFlash7 (approved):
Changes approved.

Genevesh:
> Good changes for the most part- Though I think the biggest issue right now for more dangerious mobs is that parrying is always 90%. In particular, we need a way to make upper level mobs have a higher chance to hit. Unfortunately, I don't think we can provide mobs combat skill since they lack a mind for it to take to. You got any suggestions?
> 
> I was working on increasing the void dragons health, probably trippling it, and giving it dragon flavored loot but I have been too busy with irl to do so. The current jump in it's health is decent enough at the moment.

Is it possible to make them actually damage the equipment they're hitting? I know they hurt armor REALLY well already

PR #2812, opened 2025-05-07
Fix elf sabre runtime when attacking NPCs

noelle-lavenza:
## About The Pull Request
Separated from #2794.

## Why It's Good For The Game
Bugfix. Not caused by me, this entire section of code might need to be ripped out entirely in favour of the magic item effect.

This was tested in the NPCs PR and worked fine, I'm just separating it out.


RadiantFlash7 (changes requested):
You can just tear out al of the funny effect blocks for the elf saber. 
We already added a silver effect to elf saber on initialize on line 793

RadiantFlash7 (approved):
Extraneous code removal as it has the magic silver trait.
Thank you, Approved.

PR #2813, opened 2025-05-07
Make embeds more punishing to resist out of

noelle-lavenza:
## About The Pull Request
Resisting an embed-grab will return the embedded object to its holder on success and twist the item in your wound on failure, preventing players and mobs from running off with a weapon in them and also making having something embed in you be just a lot more dangerous overall.
Twisting weapons now gives a guaranteed crit roll as well, because just 5-10 damage is kind of wimpy. That said, it's a 5-10 damage crit roll, so still not super powerful.

## Why It's Good For The Game
Weapon embeds should punish the victim more than the attacker.

This was in the NPC PR and worked fine, I'm just splitting it out.


RadiantFlash7 (approved):
No issues reporting during test merge, the code looks good.

PR #2814, opened 2025-05-07
Add extra woundpain to special wounds and increase chest and head max pain

noelle-lavenza:
## About The Pull Request
The head and chest can now take more pain than 100, which allows for making NPCs flee in pain while targeting only one zone. Before, some of them had such ludicrously high endurance that you couldn't hit the threshold by only attacking one bodypart.

## Why It's Good For The Game
Fixes issues with NPCs not fleeing because they have insane endurance right now.
Special injuries now have increased woundpain because they're special.

This was part of #2794 and worked fine.

### DIFF
--- a/code/datums/wounds/types/punctures.dm
+++ b/code/datums/wounds/types/punctures.dm
@@ -8,8 +8,6 @@
 	sewn_clotting_rate = 0.01
 	clotting_threshold = 0.2
 	sewn_clotting_threshold = 0.1
-	woundpain = 0
-	sewn_woundpain = 0
 	sew_threshold = 75
 	mob_overlay = "cut"
 	can_sew = TRUE
--- a/code/datums/wounds/types/slashes.dm
+++ b/code/datums/wounds/types/slashes.dm
@@ -8,8 +8,6 @@
 	sewn_clotting_rate = 0.02
 	clotting_threshold = 0.2
 	sewn_clotting_threshold = 0.1
-	woundpain = 0
-	sewn_woundpain = 0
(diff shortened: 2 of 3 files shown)


code/datums/wounds/types/punctures.dm
```diff
@@ -8,8 +8,6 @@
 	sewn_clotting_rate = 0.01
 	clotting_threshold = 0.2
 	sewn_clotting_threshold = 0.1
-	woundpain = 0
```
RadiantFlash7, line 11: I'm not sure we want to remove wound pain listings under each wound type. If you go back into datum/wound they're inherited the exact same; But for readability purposes, having it explicitly read as 0 in each file in subtypes of wound is probably better.

code/datums/wounds/types/slashes.dm
```diff
@@ -8,8 +8,6 @@
 	sewn_clotting_rate = 0.02
 	clotting_threshold = 0.2
 	sewn_clotting_threshold = 0.1
-	woundpain = 0
```
RadiantFlash7, line 11: Same concern as above.

RadiantFlash7 (commented):
Bug fixes and tweaks to wound pain should be fine- Though I question removing wound = 0 from the various wound types.
Yes, it inherits it from the parent, but the parents in another file. It's a bit more organized to have it restated as =0 in each individual file.

RadiantFlash7:
...missclicked the labeled. Eitherway, Get back to me on the var removal before this is approved or denied.

PR #2815, opened 2025-05-07
Removes Furry options from Seelie

Dilligentdog:
## About The Pull Request

Removes all wild-kin customization from Seelies. These customizations were added for no good reason against the creator's will

## Why It's Good For The Game

Seelie should never have been given furry customization without approval of the person who made it IMO.

image


noelle-lavenza (approved):
This can probably be added to Unseelie if we ever get them, but for now it definitely needs to be removed.

Gyrandola:
> Seelie should never have been given furry customization without approval of the person who made it IMO.
Weak argument. This is a open source project.

noelle-lavenza:
> Weak argument. This is a open source project.

Sure, but I do think leaving them for Unseelie only will not only help carve out an identity for Seelies themselves, rather than just "annoying and tiny half-kin," while also making it clear Unseelie are different.

Gyrandola:
> > Weak argument. This is a open source project.
> 
> Sure, but I do think leaving them for Unseelie only will not only help carve out an identity for Seelies themselves, rather than just "annoying and tiny half-kin," while also making it clear Unseelie are different.

I dont mind that but Unseelies have been supposedly WIP since before I left. What are the chances this is merged and they never actually come out? 

I think bundling this change with the actual Unseelie PR would be ideal.

Drakeven:
I agree with Gyrandola, there's nothing more permanent than a temporary solution- and if this is to make way for 'Unseelie' filling the gap than it should wait until they're actually in. Seelie are so limited in what they can do anyway that I don't feel limiting them even further in customization will do them any favors, besides, they're supposed to be creatures born of chaos.

PR #2819, opened 2025-05-07
Make pinning require grabs in both hands

noelle-lavenza:
## About The Pull Request
This was at one point included in #2794 but is being split out.
Readds the check to be grabbing the target with both hands in order to tackle/pin them.

## Why It's Good For The Game
This was, to my knowledge accidentally, undone in the grab rework. If it wasn't accidental, it was at the very least undocumented.

I'm also considering making it allow an empty hand as well, as long as you don't have a weapon in it, but that's up to others to decide. I'm just the coder not the balancer-er.

It *was* in the NPC PR but I don't know if it ever got deployed with this change in.

### DIFF
--- a/code/modules/mob/living/grabbing.dm
+++ b/code/modules/mob/living/grabbing.dm
@@ -180,6 +180,9 @@
 				if(user.loc != M.loc)
 					to_chat(user, span_warning("I must be on top of them."))
 					return
+				if(user.l_grab?.grabbed != M || user.r_grab?.grabbed != M)
+					to_chat(user, span_warning("I must grab them with both hands."))
+					return
 				user.stamina_add(rand(10,15))
 				M.visible_message(span_danger("[user] pins [M] to the ground!"), \
 								span_userdanger("[user] pins me to the ground!"), span_hear("I hear a sickening sound of pugilism!"), COMBAT_MESSAGE_RANGE)


RadiantFlash7:
I'm not sure about this part of it. I was hoping at some point to make tackling/pinning someone require the target to have been disarmed. Making Tackling someone take two hands against miight make it even more difficult to chain folks though. Hm.

Gyrandola:
It makes sense but IIRC this was done to incentivize non-lethal take-downs. I'm curious to see how combat would change with this revert.

Lord-Preston:
I believe Gyran is correct. When the old maintainer reworked grabbing he made it like it is to incentivize non-lethal takedowns. At the time a lot of people, especially on the keep side of things, were just killing people or when they did take them alive they skull cracked them (ruins chances of escaping, can be hard to get treated, limits RP/emotes since you’re paralyzed etc.). It sucks to be left in a cell skull cracked, which still happens but thankfully not as much anymore iirc. 

I agree that it makes more sense, but I can see it being very difficult to chain someone with both of your hands tied up unless you have help.

In my opinion, I am fond of keeping the status quo on grabs since it does incentivize non-lethal takedowns when a lot of players still seem to prefer summary executions and taking no prisoners. I think how things are now grab-wise helps a lot. However, I can see the other side of things too.

roarbark:
this should be TM'd and the IC consequences looked at. there is little other way to really tell how the comminity will react to this...

noelle-lavenza:
> I'm also considering making it allow an empty hand as well, as long as you don't have a weapon in it, but that's up to others to decide. I'm just the coder not the balancer-er.

i might go with this if we TM it tbh

PR #2820, opened 2025-05-07
Make grabs use wrestling skill for parrying/defense

noelle-lavenza:
## About The Pull Request
Originally in #2794, now separate.
Grabs now use the wrestling skill as their associated skill, enabling skill gain with grabs using the same logic as normal weapon training.
Unarmed intents can now specify a specific skill to use rather than just unarmed, so now grabbing uses the wrestling skill and not the unarmed skill. This may warrant some role skill adjustments.

Also cleans up some parrying code, it was a little too verbose for me.

## Why It's Good For The Game
Grabs should use wrestling skill rather than unarmed/no skill. This adds a way to train wrestling on par with normal fighting skills, plus I was told that this was the reason so many roles have high unarmed, meaning if that's correct we can reduce unarmed skills on most roles aside from classes that use it as their primary combat skill.

This was in #2794 before that was split out, and it worked fine.

### DIFF
--- a/code/game/objects/items/rogueweapons/intents.dm
+++ b/code/game/objects/items/rogueweapons/intents.dm
@@ -12,6 +12,7 @@
 	var/obj/item/masteritem
 	var/mob/living/mastermob
 	var/unarmed = FALSE
+	var/datum/skill/unarmed_skill = /datum/skill/combat/unarmed
 	var/intent_type
 	var/animname = "strike"
 	var/blade_class = BCLASS_BLUNT
@@ -418,6 +419,7 @@
 	candodge = TRUE
 	canparry = TRUE
 	item_d_type = "blunt"
+	unarmed_skill = /datum/skill/combat/wrestling
 
 /datum/intent/unarmed/grab/rmb_ranged(atom/target, mob/user)
 	if(ismob(target))
--- a/code/modules/mob/living/carbon/human/species.dm
+++ b/code/modules/mob/living/carbon/human/species.dm
@@ -1197,7 +1197,7 @@ GLOBAL_LIST_EMPTY(roundstart_races)
 			log_combat(user, target, "attempted to punch")
 			return FALSE
 */
-		var/selzone = accuracy_check(user.zone_selected, user, target, /datum/skill/combat/unarmed, user.used_intent)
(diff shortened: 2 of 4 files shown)


noelle-lavenza:
> Are you sure this won't lead to a bunch of people owning everyone because they can powergame wrestling every round

As opposed to people owning everyone because they have wrestling roundstart and no one can counter them through skillgain from getting their ass beat in a grab-fight? You can only train it by wrestling someone with 2 levels of wrestling or more above you, the same as other combat skills.

Purpleturtle3:
> > Are you sure this won't lead to a bunch of people owning everyone because they can powergame wrestling every round
> 
> As opposed to people owning everyone because they have wrestling roundstart and no one can counter them through skillgain from getting their ass beat in a grab-fight? You can only train it by wrestling someone with 2 levels of wrestling or more above you, the same as other combat skills.

Actually, it's just one level and any weapon, so the people of the garrison can all get wrestling five if they go to the veteran since they have swords 6 for example.

noelle-lavenza:
> Actually, it's just one level and any weapon, so the people of the garrison can all get wrestling five if they go to the veteran since they have swords 6 for example.

Oh that mega sucks LMAO. Fuck. Skill gain from parrying is just evil actually I really hate it. Ughhhhh. I should make it so wrestling can't be trained, but it should still check wrestling skill rather than using unarmed.

john-pawson:
The balance thing is probably going to be a big issue if you make grabs trainable. Since part of the balance already *is* that it's untrainable. 
It's what makes kidnapping a soilson or about any towner role alive considerably easier than, well, about every single combat class there is. Even if it is two levels without needing to find a trainer, when those two levels make you just 1 level below you average combat class instead of the usual 3, it starts to matter. That and it makes a random towner grab-spamming someone in the middle of the battle considerably more effective, since it means them landing an aggro grab will be considerably more likely.

That and.. i don't see a situation where you'd be grabbing someone *and* parrying something at the same time with that grab? It just sounds like you shouldn't be able to parry on the first place if you are grabbing someone or something other than for specifically training, unless there's some situation i'm just missing.

RadiantFlash7 (changes requested):
If you can make it so wrestling people isn't trainable, this can be testmerged. Wrestling impacts balance a little too much for me to want it to be trainable midround honestly. We want bandits to be able to wrestle down townsfolk and such.

PR #2821, opened 2025-05-07
Make unarmed attacks (including grabs) use stamina based on intent

noelle-lavenza:
## About The Pull Request
Actually implements the `releasedrain` var on intents, which was never used. Stamina is now drained regardless of hit or miss when you use an unarmed intent on a mob or a `WEIGHT_CLASS_GIGANTIC` or larger object, according to the `releasedrain` value on the intent.

The existing `releasedrain` value on punches has been lowered from 5 to 1 because I thought it was a little too high to keep punching viable. Similarly, using swift or strong intent only increases the stamina cost by 4 rather than 10 with armed attacks, because your hands aren't as heavy or unwieldy as weapons are. This means that creating a grab on someone costs 8 stamina, and 12 on strong or swift intent.

I also made item attacks check it, but it's set to `1` on everything but grabs, which have a cost of `2`. This does mean normal attacks with any intent will have a base stamina cost, but that's trivial to adjust by setting the default `releasedrain` to `0`. (I actually kind of like the idea that any action, if repeated quickly enough without resting, can exhaust you. It means battles of attrition are actually a thing.)

This new stamina cost ***DOES STACK*** with the existing stamina loss from missing.

## Why It's Good For The Game
Grab spamming now has a stamina cost, and punches now actually cost stamina when you miss. This was always intended, there were values set on the intents, they were just never used. This makes swift-spam punches less OP, and it means there is a cost to failing a grab or to losing it and having to regain it. You can actually wear someone out via grab-spam instead of them just having no consequence for it.

This was tested in #2794 before splitting it out, and the only issue was that clicking items on grab intent caused excessive stamina usage. This was an indentation error and has been fixed in this version of the PR.

### DIFF
--- a/code/_onclick/item_attack.dm
+++ b/code/_onclick/item_attack.dm
@@ -126,10 +126,12 @@
 			if(!user.used_intent.swingdelay)
 				user.do_attack_animation(M, visual_effect_icon = user.used_intent.animname)
 			return
-	if(istype(user.rmb_intent, /datum/rmb_intent/strong))
-		user.stamina_add(10)
-	if(istype(user.rmb_intent, /datum/rmb_intent/swift))
-		user.stamina_add(10)
+	var/rmb_stam_penalty = 0
+	if(istype(user.rmb_intent, /datum/rmb_intent/strong) || istype(user.rmb_intent, /datum/rmb_intent/swift))
+		rmb_stam_penalty = 10
+	// As of writing, releasedrain is 1 on everything but unarmed and grab intents,
+	// so it's primarily an unarmed thing.
+	user.stamina_add(user.used_intent.releasedrain + rmb_stam_penalty)
 	if(M.checkdefense(user.used_intent, user))
 		if(M.d_intent == INTENT_PARRY)
 			if(!M.get_active_held_item() && !M.get_inactive_held_item()) //we parried with a bracer, redirect damage


Skyglazer:
Having played with this for a little while, I think punches could do to cost a little less stamina on swift- I know it's already reduced, but I think having it be a little further reduced would make it so that swift is a viable intent to use when unarmed, which is when I think it's the most appropriate. Grabs can stay the same cost though-- them being stamina intensive is good. Just my opinion though.

noelle-lavenza:
> Having played with this for a little while, I think punches could do to cost a little less stamina on swift- I know it's already reduced, but I think having it be a little further reduced would make it so that swift is a viable intent to use when unarmed, which is when I think it's the most appropriate. Grabs can stay the same cost though-- them being stamina intensive is good. Just my opinion though.

wonder how I could make that happen. 1.5x stam cost instead of +4? since currently for punches it's a 5x multiplier and for grabs it's only 1.5x. 1.5x flat would make it exactly as punishing stam-wise for punches as it is for grabs

Skyglazer:
Having played with this further, I feel MUCH more negative about this PR. Grabs are SO unsatisfying to use now, and roles that are supposed to rely on them (acolyte, dualist, mortician) feel absolutely gutted. With the way "missed" grabs work, it's very possible to lose more than half your stamina to basically nothing-- combine that with the high stamina costs on activities like tightening a grab, resisting, tackling, let alone defensive actions like parrying/dodging- makes it nigh-impossible to manage stamina effectively in a fight as a grappler. Punching and grappling absolutely sucks now, even with high stats and skill; it shouldn't.

Skyglazer:
>Skyglazer's posted a video showing him simply grabbing repeatedly at empty ground and fully depleting his stamina in less than three seconds. 

Yep, here it is. 10 stats across the board, Swift intent. 9 empty clicks is all it takes to be COMPLETELY out of stamina with this PR.

https://github.com/user-attachments/assets/ff78aa30-ba9d-4b57-b09f-b5a1e422e56d

Foxentail:
This is gonna make it harder to non-lethally subdue people which in turn will make people just murder more. I hate how annoying grabs can be but I'm not sure if I'm too fond of this.

RadiantFlash7:
Would this feel better if the stamina cost from missing was removed, so you didn't lose stamina when you miss a grab? (not talking parries, I mean full on clicking on empty tiles in frantic clicking.)

GravyMaker:
I think grabs which miss (in the sense of failing to click a valid target) and grabs which can't happen in the first place due to being on cooldown should certainly not drain anything, personally. As for any other considerations, I think it would be wise to ask @Skyglazer since he's much more learned in the martial arts of this game than I am.

Skyglazer:
> Would this feel better if the stamina cost from missing was removed, so you didn't lose stamina when you miss a grab? (not talking parries, I mean full on clicking on empty tiles in frantic clicking.) @GravyMaker @Skyglazer

A lot better, but I still think the stamina consumption for punches (and swift punches) should be a bit lower. One of the whole upsides of unarmed combat was that it's very stamina-light, now you have all the downsides of having no weapon, with basically none of the upside.

noelle-lavenza:
> I'm going to be untestmerging this for now, pending a review of the other maintainers. I don't hate this PR, but it does have notable issues that have not been fixed for a few weeks.

azure has a fixed version that solves that problem, i'll take a look at theirs

PR #2826, opened 2025-05-08
Rewrite roguetownprocs.dm to fix issues and improve code quality.

noelle-lavenza:
## About The Pull Request
`roguetownprocs.dm` is by far the most infamous file in all Roguetown forks, given how a lot of the code is poorly-written, redundant, in violation of basic object-oriented programming principles, poorly-organised, redundant, buggy, outright broken, and redundant. This rewrites nearly all of it, fixing a few bugs in the process.

- `checkdefense()` is now split into three procs, `checkdefense()`, `checkparry()`, and `checkdodge()`. In terms of functionality they should be the same, the code is just way easier to understand (IMO) now.
- You can no longer dodge into turfs with dense objects.
- Simplemob `defprob` now actually applies; as-written it would never ever work because lol bad code.

## Why It's Good For The Game
Code quality good, bugs bad.

I have not tested it yet.

### DIFF
--- a/code/game/objects/lighting/roguelighting.dm
+++ b/code/game/objects/lighting/roguelighting.dm
@@ -309,24 +309,20 @@
 
 	if(on)
 		var/mob/living/carbon/human/H = user
-
 		if(istype(H))
-			H.visible_message(span_info("[H] warms \his hand over the fire."))
-
-			if(do_after(H, 15, target = src))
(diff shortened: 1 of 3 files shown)


noelle-lavenza:
Tested. I can dodge and parry just fine. You won't dodge into solid objects though, which means close-quarters fights are a little more punishing for dodge users, but since they can dodge attacks from behind that's probably fine.

Gyrandola (approved):

PR #2833, opened 2025-05-09
Runtimes generation fix + maybe server clangs but I dunno

DelineFortune:
## About The Pull Request

fixes pathfinder a bit, if mob cannot find path to escape more that 3 times = it deletes them

Naturally it happens RARE RARE when rous tries to reach the body from this direction but something blocks them

image

## Why It's Good For The Game

its low low chance but your round may start like this (5% chance) and fucking clang your game my gamer 

image

image

image

image

image


DelineFortune:
1 runtime per 30 minutes, I guess, its better than 10000. But would like to ask skilled maints to look into

PR #2834, opened 2025-05-10
Adds Realm-Colored Banner Flags

Alanii:
## About The Pull Request

All this does is add "realm color" banner flags. They pull lordcolor's and match with the duchy colors.

## Why It's Good For The Game

More flavor

image

Soon™️


RadiantFlash7:
if I recall correctly, the regular decorative flags are actually craftable. If you want to also make these craftable, you'll need a recipe made for em.

Alanii:
> if I recall correctly, the regular decorative flags are actually craftable. If you want to also make these craftable, you'll need a recipe made for em.

These are not intended to be craftable. This is so I can litter the new town with these and it makes sense.

RadiantFlash7 (approved):
Fine enough addition. Might want to make em craftable later, but for now purposes it's good.

PR #2837, opened 2025-05-11
Fix CI not downloading BYOND

noelle-lavenza:
Fixes it by adding a user agent, also should make the cache work.


RadiantFlash7 (approved):
Requested Speed merge to fix CI.

PR #2839, opened 2025-05-11
Buff NPC pathfinding and AI

noelle-lavenza:
## About The Pull Request
Buffs NPC pathfinding and AI.
- NPCs can now path up and down stairs.
- NPCs should no longer move diagonally (except when fleeing from being in too much pain, because that still uses BYOND's built-in pathfinding).
- NPCs will now take into account slowdown from turf type when pathing. This even includes using flowing water to travel upstream quicker, and avoiding entering it entirely if it's going the wrong way (unless there's no better option).
- NPCs fleeing in pain should work more reliably, and has a lower threshold.
- NPCs should now resist/stand up/etc. properly.
- Attacking an NPC interrupts their attempts to do timed actions like climbing (except for resisting or standing up).
- Humanmob NPCs can now path through unlocked doors.
- Mobs with the door-bashing trait can now path through doors and will bash them down, even if locked. Scary!
- **MOBS NOW ATTEMPT TO GRAB THEIR ENEMIES, UPGRADE GRABS, AND TWIST/SMASH/CHOKE/ETC.**
- Converted deadites will not have skills.
- Deadite AI has been completely rewritten because it was garbage before.
- In preparation for switching to the AI controller system for humanmobs, a lot of the AI logic has been pulled out into separate overrideable procs.
- Goblins now lose stamina.

All of the misc. changes have been PR'd separately.

## Why It's Good For The Game
No more stair cheese.

EiJXWdU27i-ezgif com-optimize

btLoOo3ZbY

### DIFF
--- a/code/modules/mob/living/carbon/human/npc/_npc.dm
+++ b/code/modules/mob/living/carbon/human/npc/_npc.dm
@@ -7,11 +7,12 @@
 	var/list/friends = list()
 	var/mob/living/target
 	var/obj/item/pickupTarget
+	var/atom/pathfinding_target
 	var/mode = NPC_AI_OFF
 	var/list/myPath = list()
 	var/is_currently_pathing = FALSE
 	var/list/blacklistItems = list()
-	var/maxStepsTick = 6
+	var/maxStepsTick = 7 // can move a screen's width in one turn
 	var/resisting = FALSE
 	var/pickpocketing = FALSE
 	var/del_on_deaggro = null
@@ -74,6 +75,10 @@
 		npc_stand()
 		resisting = FALSE
 		return // Standing passes your turn, you can't attack.
+	// If we're pathfinding somewhere, move along that path
+	if(length(myPath))
+		move_along_path()
+		// We could return here if we wanted to make moving use your turn.
 	if(!handle_combat())
 		if(mode == NPC_AI_IDLE && !pickupTarget)
 			npc_idle()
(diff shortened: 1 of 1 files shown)


Fn-OS:
When on a tree/upper Z level with no access, skeletons walk under the tile you are on.

noelle-lavenza:
> When on a tree/upper Z level with no access, skeletons walk under the tile you are on.

this was supposed to happen with the old code but was broken. this is an intentional feature. @Fn-OS is this bugged somehow or are people just mad they can't use tree cheese anymore

RadiantFlash7:
Resolve conflicts and we'll see about having this full merged soonish.

Gyrandola (approved):

Windryn:
_Elf coding_

code/modules/mob/living/living.dm
```diff
@@ -1487,6 +1487,10 @@
 		var/datum/species/seelie/seelie = H.dna.species
 		if(!seelie.has_wings(src))
 			canstand_involuntary = FALSE
+	if(canstand_involuntary)
+		mobility_flags |= MOBILITY_CANSTAND
+	else
+		mobility_flags &= MOBILITY_CANSTAND
```
GravyMaker, line 1493: Does this line do what it's meant to do? It looks like this is disabling _every_ movement flag except for MOBILITY_CANSTAND, and coincides neatly with a bug we've had reported where broken and missing legs lead to almost-complete inability to act.

noelle-lavenza: FUCK. Thank you

PR #2841, opened 2025-05-11
Fixed (most cases of) unintended stat changes on resurrection

GravyMaker:
## About The Pull Request

This PR fixes a highly common and visible bug, which I think has had a strong adverse effect on a lot of people's rounds for a while. In brief, previously, at the moment the zombie antag datum was applied to a character (typically through death, but occasionally via infection), their stats - or, those which would be modified by becoming a deadite - were cached. Those stats were then restored when the datum was removed (typically on either being derotted or revived). The issue with this is that if any status effects that modified stats were applied between these two operations, their application or removal would be "forgotten" when the previously-cached stats were restored, causing the stat adjustments to remain permanently.

This PR fixes this issue by removing the caching logic, instead replacing it with what is essentially a generated status effect which adjusts the mob's stats to be what they should be when they become a deadite. This means that the same logic as usual is applied for every other status effect (with the exception of certain temporary immunities from being a deadite), preventing situations where status effects are not correctly applied to or removed from stats once they're "restored".

Additionally, the PR slightly modifies the change_stat proc, moving an early return in the case of an indexed change with a value of 0. This is to prevent the addition of meaningless empty stat adjustments in the case that we attempt to remove an indexed change when no such change exists.

## Why It's Good For The Game

This issue is highly visible and could dramatically affect the round of a player who died. It's particularly noteworthy for combat roles and other physically-intensive roles due to the hefty debuff for exsanguination (-3 strength, -4 speed) which most people who die will suffer. The issue is so widespread, in fact, that it seems like a lot of players believe this is the correct behaviour, and it's quite common for players - again, especially in combat or otherwise highly-physical roles - to simply leave the game after dying due to these lasting status changes.

With the fix, death and revival will properly update the stats they affect, which should hopefully lead to reduced instances of players leaving after dying, while also allowing any changes to death debuff balancing to be made with the confidence of knowing they'll be correctly applied.

Me, the character I'll be killing and reviving, and his stats as seen in the View Variable interface.
image

The character, now dead, showing his stats - including the results of his debuffs (extreme blood loss).
image
d

The character, still dead but healed, showing his stats.
image

The character, now revived, showing the stats - including the debuff from Lux revival (which wasn't being correctly applied previously, also due to the bug fixed in this PR). There's also an additional -1 to endurance from thirst.
image

And for good measure, the character having been given water to remove the -1 endurance from thirst:
image

## Addendum
This has been one of those bugs that was simple to understand once I figured out why it happened, but tricky, initially, to devise a fix for. My thanks to @noelle-lavenza for proposing the solution that (I believe) covers the edge cases that my initial attempt at a fix didn't properly account for.

I have a few more bugs relating to death and revival that I expect to submit PRs for either later tonight or tomorrow - but this is probably the single most important of them.

### DIFF
--- a/code/modules/antagonists/roguetown/villain/zomble.dm
+++ b/code/modules/antagonists/roguetown/villain/zomble.dm
@@ -57,12 +57,6 @@
 	cmode_music = zombie.cmode_music
 	patron = zombie.patron
 
-	STASTR = zombie.STASTR
-	STASPD = zombie.STASPD
-	STAINT = zombie.STAINT
-	STACON = zombie.STACON
-	STAEND = zombie.STAEND
-
 
 	return ..()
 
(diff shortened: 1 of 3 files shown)


RadiantFlash7 (approved):
No issues reported on testmerge.

PR #2843, opened 2025-05-12
Returns old court magos +1 to arcane magic

ARefrigerator:
## About The Pull Request
brings it back. Seemed to get deleted when someone did not understand what the proc did?

## Why It's Good For The Game
He's supposed to have it. At least that's what a PR from before indicates and it never got fixed.

before: 
image
with PR:
image


RadiantFlash7 (approved):
Honestly, I'm not sure why that was ever removed. Old folks /should/ really get universal skill increases across all jobs.

PR #2844, opened 2025-05-12
Great mana potion buff does not depend on sleepiness

Heheniy:
## About The Pull Request
Description

## Why It's Good For The Game
It is weird that to get buff from great mana potion you need to be tired


RadiantFlash7 (approved):
Minor balance change which- frankly makes plenty of sense. Approved.

PR #2845, opened 2025-05-12
Fix for deadite infection causing uncontrollable biting

GravyMaker:
## About The Pull Request
This PR fixes an issue which causes characters infected with the deadite infection to bite those around them uncontrollably.

This issue is caused by the zombie antag datum's on_life proc not considering whether or not the character had actually turned into a deadite yet - hence, anybody infected would immediately begin biting anybody nearby, rather than only doing so after actually becoming a deadite.

The fix is a simple check against the has_turned variable before proceeding to attempt to bite anything.

## Why It's Good For The Game
This corrects the infection to its intended behaviour -  the victim is a danger because they'll soon turn into a deadite, but still retains control of themselves for the time being. The uncontrollable biting was unintended behaviour which didn't fit well with the player otherwise retaining full control of themselves, and additionally made further interaction with them difficult.

Myself, infected and not biting the character next to me. (force_apply_deadite_bite is a debug proc written specifically to test this and similar issues, and isn't included in the PR.)
image

And indeed after turning, I still assault the character as expected.
image

### DIFF
--- a/code/modules/antagonists/roguetown/villain/zomble.dm
+++ b/code/modules/antagonists/roguetown/villain/zomble.dm
@@ -172,9 +172,10 @@
 	return ..()
 
 /datum/antagonist/zombie/on_life(mob/user)
-	var/mob/living/carbon/human/deadite = owner?.current
-	deadite.try_do_deadite_bite()
-	deadite.try_do_deadite_idle()
+	if (has_turned)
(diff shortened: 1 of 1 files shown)


RadiantFlash7 (approved):
Good bug fix,. Thank you as always Gravy

PR #2849, opened 2025-05-12
Recipe name standardization

Heheniy:
## About The Pull Request
Standardizes ALL recipe name to be "[created item] - ([req items]; [req structure (if needed)] ; [req tool (if needed)]; [req skill]"
Removes few excessive subclasses of some items, that were created only to be used as craft result for some reason (which touches maps)

## Why It's Good For The Game
Standartization


Gyrandola:
Not a fan of hardcoding recipe requirements into the name itself. Theres definitely a dynamic way to do such a thing that will cause less headaches down the line and is more maintainable.

Heheniy:
not how crafting is coded- no
there are different things with same names, same things with different recepies, some results are just an obj and some are a list of obj and so on

Heheniy:
shut up!

RadiantFlash7 (approved):
Changes approved, no tm issues.

PR #2850, opened 2025-05-12
Fixes issues with Cure Rot

GravyMaker:
## About The Pull Request
This PR fixes two issues with Cure Rot:
- Cure Rot killing targets who have the deadite infection, but who have not yet become deadites.
- Cure Rot applying the Rot-Touched condition to targets who aren't deadites and haven't rotted (including uninfected targets!)

## Why It's Good For The Game
In gameplay, Cure Rot is intended to be the most potent and reliable way of dealing with the rot and associated deadite infection. However, at present, Cure Rot will reliably kill a target who is merely infected - as well as applying the Rot-Touched debuff, preventing the target from ever being cured of the rot again. Additionally, if Cure Rot is ever incorrectly targeted - which isn't difficult to do when an Unknown is being restrained by another Unknown - it will apply the Rot-Touched debuff to the incorrect target, despite there being no rot to remove.

This fix improves the situation by adding an incentive to treat the infection: treating it in time will save not only the target's life, but spare them the rot-touched condition. Additionally, Cure Rot now can't be used to maliciously give somebody the rot-touched debuff, and won't apply it in cases of accidental application; putting the spell on cooldown for a while is already punishment enough for targeting the wrong person, in my opinion.

A character is infected with the deadite infection, given an Eoran bud, and allowed to turn.
image

The deadite rises.
image

I use Cure Rot, correctly killing the deadite, and leaving it rot-touched.
image

I apply the infection to a new character.
image

I cure the infection, leaving the target alive and free of the rot-touched condition as a reward for my swift action.
image

### DIFF
--- a/code/modules/spells/roguetown/acolyte/pestra.dm
+++ b/code/modules/spells/roguetown/acolyte/pestra.dm
@@ -218,6 +218,9 @@
 	if(isliving(targets[1]))
 		testing("curerot1")
 		var/mob/living/target = targets[1]
+		if(target == user)
+			revert_cast()
+			return FALSE
 		if(HAS_TRAIT(target, TRAIT_ROTTOUCHED))
 			to_chat(user, span_warning("The Rot has spread too deeply for even Pestra to clean."))
 			revert_cast()


RadiantFlash7 (approved):
Thank you for the bug fixes, Gravy!

PR #2851, opened 2025-05-13
Scholar Patron Edit

Bapldap:
Allows Scholar to be either Noc OR Zizo devoted

## About The Pull Request

Allows Scholars/Archivists to worship either Noc OR Zizo. Literally a 1 line change. (mind blowing, incredible)

## Why It's Good For The Game

It'd be a pretty good role for zizoids, and it makes sense. The Magos can be Noc or Zizo and this falls in line with that, a scholar would be a perfect cultist - after all, they've access to tons of knowledge, and thus there is plenty of chance for them to become corrupted by it.

proofoftesting1


RadiantFlash7 (approved):
Makes perfect sense. seeking power through knowledge is half zizo's shtick, given they were a wizard.

PR #2852, opened 2025-05-13
random character updates voice type

ARefrigerator:
## About The Pull Request
When the proc to randomize your character is called it adjusts the voice type to default

## Why It's Good For The Game
It wont solve previously existing characters (you will have to set them!!!) but for new characters this means by default you wont have to in case you forget.

https://github.com/user-attachments/assets/265e207d-7825-40bc-87b2-4de3e0fffbf2


RadiantFlash7 (approved):
Quick bugfix, Code seems fine. Approved.

PR #2853, opened 2025-05-13
Fix for bandit liches

GravyMaker:
## About The Pull Request
This PR adds bandit to the list of roles that cannot roll lich.

## Why It's Good For The Game
As far as I know, bandits aren't supposed to be able to be liches - they can't be any other type of antag, so the omission of bandits from the exclusion list seems to just be an accident, probably dating back to when bandits were more formal antagonists rather than a selectable role. Additionally, this is an issue that seems to crop up alarmingly frequently in rounds - I've seen a number of ahelps by users who've managed to roll a bandit lich who want to be made into just a bandit, creating round-start work and depriving the round of a proper antagonist in cases where no alternative can be arranged. This fix would resolve these issues going forward.

image


RadiantFlash7 (approved):
Simple bugfix, long over due.

PR #2854, opened 2025-05-13
[BUG FIX] checks if there's any reagents in the pot, not if reagents holder exists

ARefrigerator:
## About The Pull Request
title
Also fixes reagent checks

## Why It's Good For The Game
was a little oversight where you could cook even without reagents

pot with 3 oz of oatmeal after only putting in 3oz of water
image


ARefrigerator:
Added some more functionality to the code and cleaned it up somewhat.
If an item now has reagents in it the pot will try to render it down to its reagents after making a timer based on the total reagents of that item.

**essence of clarity into pot:**
image

**adds to timer:**
image

**get results (if you meet the chem reaction requirements):**
image

**also put an apple in to check:**
image

PR #2855, opened 2025-05-13
Fix for left arm spawn on reapplied game preferences

GravyMaker:
## About The Pull Request
This PR fixes a longstanding issue in which reapplying a player's game preferences would cause a phantom left arm to spawn, due to a flaw in the logic which replaces their arms to account for the missing arm character flaws. The fix is simply calling qdel() on the created arm - this was already being done for the right arm, but seems to have been omitted for the left due to an oversight.

## Why It's Good For The Game
This fix saves admins time when performing a common operation (often applied multiple times per round); in particular it will, I think, no longer be necessary to aghost as part of the process to delete the phantom limb spawn. It also prevents players from seeing left arms popping into and out of existence and potentially impacting immersion.

I won't bore you with a screenshot of me with no arm on the ground, and then another screenshot of me with no arm on the ground.
image

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -1999,6 +1999,7 @@ Slots: [job.spawn_positions]</span>
 		O = character.get_bodypart(BODY_ZONE_L_ARM)
 		if(O)
 			O.drop_limb()
+			qdel(O)
 		character.regenerate_limb(BODY_ZONE_R_ARM)
 		character.regenerate_limb(BODY_ZONE_L_ARM)
 


Red-Honk:
At last, some man genius figured out the issue!

Now, I know it isn't all that required, but, could you kindly figure out a way to make the game re-apply organs when the gender gets changed?; For example, if a male character turns female or the other way around, it keeps the orientation (especially if its the same type of organ) to the same location, and only adjusts it, if another re-apply preferences is added to center them right.

GravyMaker:
I could take a look, though I'm not entirely clear from your description exactly what the problem is. It sounds like you're talking about sex organs not being swapped when reapplying preferences if it would change the character's sex? What do you mean by the second part - the sprites are in the wrong place?

Red-Honk:
> I could take a look, though I'm not entirely clear from your description exactly what the problem is. It sounds like you're talking about sex organs not being swapped when reapplying preferences if it would change the character's sex? What do you mean by the second part - the sprites are in the wrong place?
> 
> Edited to add - and despite the conversation happening here, if I take a look at it, it will be in a different PR. I prefer keeping things as atomic as possible, at least when it comes to bug fixes.

No, no, the issue in question arises due to the male and female sprite having a height difference pixel wise, with the gals being one pixel shorter, making things appear higher than it should and for guys at times covering their eyes with horns or markings that fail to cover the chest, as such preferences have to be re-applied when the gender is already set to the desired ones, since it'll aligned it correctly this time around because they are already x or y gender. 

(Due to both genders having their own organs, those get overwritten proper, so, that isn't a issue at all, pretty sure it even aligns them correctly because they are new ones and not just same thing but re-adjusted height wise for the other gender )

Red-Honk:
> I could take a look, though I'm not entirely clear from your description exactly what the problem is. It sounds like you're talking about sex organs not being swapped when reapplying preferences if it would change the character's sex? What do you mean by the second part - the sprites are in the wrong place?
> 
> Edited to add - and despite the conversation happening here, if I take a look at it, it will be in a different PR. I prefer keeping things as atomic as possible, at least when it comes to bug fixes.

No, no, the issue in question arises due to the male and female sprite having a height difference pixel wise, with the gals being one pixel shorter, making things appear higher than it should and for guys at times covering their eyes with horns or markings that fail to cover the chest, as such preferences have to be re-applied when the gender is already set to the desired ones, since it'll aligned it correctly this time around because they are already x or y gender.

(Due to both genders having their own organs, those get overwritten proper, so, that isn't a issue at all, pretty sure it even aligns them correctly because the game considers them as new entities and sprites, instead of being the literal same ones but height re-adjusted for either one or the other gender)

Gyrandola (approved):

PR #2856, opened 2025-05-13
Don't attempt to show hover over text for qdel'd turf/mob/atoms

tmyqlfpir:
## About The Pull Request

This PR fixes an issue that mostly affects high ping users when hovering over items that no longer exist on the server, most commonly collected items.

There is a rare chance that it'll attempt to render the offset position for a qdel'd item, which has been moved so far off screen that'll cause the client viewport to turn black, and require a reconnect to fix.

hmm

## Why It's Good For The Game

This fixes a techincal bug by adding additional checks to the hover over logic to avoid rendering items for qdel'd objects.

I've recorded a video of this bug in action, and the fix being shown working by logging in chat.
```dm
/atom/proc/handle_mouseover(location, control, params)
	var/mob/p = usr
	if(QDELETED(src))
		to_chat(p, span_notice("attempted to render deleted object [name]"))
		return FALSE
	...
```

https://github.com/user-attachments/assets/a0fbce00-38ae-48e9-a3ea-c2e6f81e17e5

By using the same checks when collecting items as for hovering, it'll avoid rendering hover over items that'll be invalid as the object no longer exists on the server, but still exists on the client due to ping latency.

### DIFF
--- a/code/_onclick/hud/mouseover.dm
+++ b/code/_onclick/hud/mouseover.dm
@@ -32,6 +32,8 @@
 
 /atom/proc/handle_mouseover(location, control, params)
 	var/mob/p = usr
+	if(QDELETED(src))
+		return FALSE
 	if(p.client)
 		if(!p.client.mouseovertext)
 			p.client.genmouseobj()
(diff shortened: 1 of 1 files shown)


Heheniy:
somewhat similar bug occurs when you hover over something at the edge of screen (usually at its top) at its name should apper over/behind screen it fucks up with your screen

RadiantFlash7 (approved):
I love bug fix PRs. Thank you for the effort on this. Code seems fine and video evidence proof is enough for a speed merge.

PR #2858, opened 2025-05-14
Prevents titles in name in character creator

Heheniy:
## About The Pull Request
Title
Works ONLY for character creator, any other way of changing character name works as before

## Why It's Good For The Game
Stupid idiots putting titles in their names

image
image

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -1378,7 +1378,7 @@ Slots: [job.spawn_positions]</span>
 							to_chat(user, "<font color='red'>Invalid name. Your name should be at least 2 and at most [MAX_NAME_LEN] characters long. It may only contain the characters A-Z, a-z, -, ' and .</font>")
 						else
 							for(T in title)
-								if(findtext(new_name, T))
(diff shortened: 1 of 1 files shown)


Skyglazer:
I think this is an issue that's better left to ahelp if you see bad examples of it rather than blanket disallow. Leave wiggle-room for people to add titles for roles that would make sense to have them, like refugee nobles or ecclesiasts, putting Lord/Lady in your name if you roll the noble blood special, so on. Players can be trusted with this bit of leeway- and if they misuse it, speaking from experience, it's trivially easy for admins to change a name.

Alanii:
Leave this to admin descretion. 

There may be instances where an admin may permit someone to use a "title" such as "Noble blooded" special, or "noble refugee" as currently noble refugees does not provide you with a title.

If you wish to do anything regarding titles, perhaps getting the system to work appropriately for those who receive "noble blood" trait.

RadiantFlash7:
This is definitely better left on admin discretion for now. I appreciate the effort you put in  for this. Normally I 100% prefer coding things so admins don't have to lift a finger, but in this case theres enough leeway to justify admin monitoring of it over hard code prevention.

PR #2860, opened 2025-05-14
brings up fishing cage catch time a bit

ARefrigerator:
## About The Pull Request
makes it take a little longer to pull fish out of cages

## Why It's Good For The Game
I think I did in fact make it a bit too low...

number changes.


Elli-Skala:
I really think it needs abit of rng

RadiantFlash7 (approved):
The never ending cycle of fine-tuning balance.
Hopefully this one feels a bit better, not too slow like prior, not to fast like now.

PR #2866, opened 2025-05-15
Adds additional peddler input validation

GravyMaker:
## About The Pull Request
This PR fixes a bug with peddler input validation, allowing certain unintended values to be entered. The fix itself is pretty simple, and copied from AP's fix for the same issue, so my thanks to their team for getting on top of it.

## Why It's Good For The Game
This issue is one I've noticed come up a few times in recent days and which has caused some disruption - following the fix, this disruption should no longer be possible.

image


RadiantFlash7 (approved):
Simple bug fix, high periority one.

PR #2867, opened 2025-05-15
[Experimental] Initial support for non-despawning summonable mobs

GravyMaker:
## About The Pull Request
This PR adds (very basic) support for non-despawning summonable mobs for the BYOS event, potential future events, and perhaps even dungeons and other PVE content down the line.

All that has changed is summonable mobs now have a "despawn_on_idle" variable, defaulting to true, which determines whether the mob should despawn when its AI enters the idle state. No further changes have been made.

## Why It's Good For The Game
The change potentially allows for future mapping opportunities - in the form of being able to place summonable mobs that don't despawn - without needing to create new versions of them, through setting the despawn_on_idle variable. It will be of particular utility in the upcoming BYOS event, which this was implemented for, but could prove useful later.

That said, it may need some further refinement (eg. the ability to turn off drops for these mobs). This hasn't been implemented yet as it wasn't requested for the upcoming event.

Not much to show off here, but the changes build fine.
image

### DIFF
--- a/code/modules/mob/living/simple_animal/rogue/creacher/summoned/elemental.dm
+++ b/code/modules/mob/living/simple_animal/rogue/creacher/summoned/elemental.dm
@@ -6,6 +6,7 @@
 		return FALSE
 /mob/living/simple_animal/hostile/retaliate/rogue/elemental
 	lighting_alpha = LIGHTING_PLANE_ALPHA_MOSTLY_VISIBLE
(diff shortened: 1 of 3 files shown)


GravyMaker:
Just an update to say that, in addition to my own quick testing, the mapper for whom this PR was submitted has confirmed it works as expected.

Useroth (approved):
lgtm

PR #2870, opened 2025-05-17
Fixed move speed not correctly updating

GravyMaker:
## About The Pull Request
This PR fixes an issue which causes movement speed to not correctly update when the speed stat is modified.

## Why It's Good For The Game
In many cases, speed buffs and debuffs are applied simply through modifying the STASPD value. However, these changes aren't immediately reflected in the player's movement speed, meaning that these buffs and debuffs don't function as intended. This PR allows these modifications to function correctly without any further work, either from developers or players.

A video of me running left and right with various speed values using the effects of Moondust.

https://www.youtube.com/watch?v=-tyQP4vnBts


RadiantFlash7 (approved):
Simple bugfix! Approved

PR #2872, opened 2025-05-17
Skeletons can tell if somebody is a zizoid

ddsaw:
## About The Pull Request

Skeletons will now be able to tell who is and who is not a zizoid.

## Why It's Good For The Game

Because it leads to nonstop "friendly fire" incidents. Because setting wise, this seems appropriate.  At the very least, if not this, then they should get an examine text for people in the undead faction such as apostates.

eIFzyQj


ARefrigerator:
I like it. We do need more means of telling friend from foe.

Gyrandola (approved):

PR #2873, opened 2025-05-17
Allows humans to vomit via emotes

ARefrigerator:
## About The Pull Request
adds vomit emote

## Why It's Good For The Game
Funny. Also there could be times where someone wants to RP vomitting and now they could add this into their actions. It still goes through the vomit emote so you still get all the negative effects that come with this (It is not an aesthetic action!!) so if you spam this or abuse it it **will** leave you completely open and defenseless. So don't go abusing it to puke on someone's floor or they'll get several seconds to beat your ass for free!

https://github.com/user-attachments/assets/398d1924-cbfc-4488-99da-878dc5d8618e

### DIFF
--- a/code/modules/mob/living/emote.dm
+++ b/code/modules/mob/living/emote.dm
@@ -1291,6 +1291,22 @@
 
 	emote("shake", intentional = TRUE)
 
+/* Vomit emote */
+/mob/living/carbon/human/verb/emote_vomit()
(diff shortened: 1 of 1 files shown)


LuckyLily4:
yes

Elli-Skala:
we dont need to buff tabaxi more

Gyrandola (approved):

PR #2874, opened 2025-05-17
Map change fix (QOL)

MarioTheShiba:
## About The Pull Request

Fixes map change not asking for a json file, resulting in maps not rendering z levels properly.

## Why It's Good For The Game

Allows for seamless map changing without needing to make bothersome PRs where you have to actually overwrite the rockhill.dm map file with the new map. This is part of the BYOS event QOL fixes.

This is a general band-aid fix to this issue and it should work as intended.

image_2025-05-17_201229348


RadiantFlash7 (approved):
No issues with the TM reported- Approved.

PR #2877, opened 2025-05-18
Lowers mercenary slots down to 2.

ARefrigerator:
## About The Pull Request
see title.

## Why It's Good For The Game
10 is way too high. These guys are more decked out than the MAA and so everyone ends up picking them.

https://github.com/user-attachments/assets/615f3f2c-1790-4cb1-8cbb-e76f1e8712da

### DIFF
--- a/code/modules/jobs/job_types/roguetown/mercenaries/mercenary.dm
+++ b/code/modules/jobs/job_types/roguetown/mercenaries/mercenary.dm
@@ -10,8 +10,8 @@
 	display_order = JDO_MERCENARY
 	selection_color = JCOLOR_MERCENARY
 	faction = "Station"
-	total_positions = 10
-	spawn_positions = 10 //subject to change, was originally 6
+	total_positions = 2
+	spawn_positions = 2 //subject to change, was originally 6
 	min_pq = 0
 	max_pq = null
 	subclass_cat_rolls = list(CTAG_MERCENARY = 20)


DelineFortune:
I think it will make another players to just play other servers
Raw content removal for some players without anything back
Being a guard is simply not interesting (low number of antags, weak bandits, only adventurers that do cause troubles)
-content for shunned races because we have no combat roles at all, no major buffs, no access to strong combat roles (templar has a lot of rp responsabilities, adventurers were added back a week ago)

Make their life more dangerous, put their spawn back into bog.
This server cannot offer a lot of content as Azure or Vanderlin has and there is another content removal

Foxentail:
Two slots is actually insane. This is either a salt PR or you're sort of dumb.

ARefrigerator:
It's not salt I just don't know what to do with them and the general consensus I get when I ask about mercs is 'lower their slots'

ARefrigerator:
> I think it will make another players to just play other servers Raw content removal for some players without anything back Being a guard is simply not interesting (low number of antags, weak bandits, only adventurers that do cause troubles) -content for shunned races because we have no combat roles at all, no major buffs, no access to strong combat roles (templar has a lot of rp responsabilities, adventurers were added back a week ago)
> 
> Make their life more dangerous, put their spawn back into bog. This server cannot offer a lot of content as Azure or Vanderlin has and there is another content removal Its not mercs problem that they are popular. Its other roles problems that they are not. Can we not dump/nerf things, but give them things to do/content to other roles instead?

moving their spawns back into the bog could work.

PR #2879, opened 2025-05-20
Drakian Statblock Buff

Skyglazer:
## About The Pull Request

The Drakian statblock is lame and underwhelming, and feels bad to play with. They have lore about being mighty and bloody conquerors, and yet the only physical claim to might they have is a +1 to constitution, at the cost of a -1 to speed, which is a pretty significant hit. I want to increase their might by giving them a +1 to strength, at the cost of a -1 to endurance-- they have to lug around those big-ass wings with them everywhere they go, after all. Probably gets tiring.

## Why It's Good For The Game

Big and strong dragons should be strong. Makes Drakian more of a competitive race pick for more roles, rather than an active handicap in most cases. I'd say it does well to even the scales as far as power goes. Previously, the statblock was just objectively bad. With the new proposed one, their statblock will be better- and the -1 to endurance will serve to make sure that it's not TOO much better, in a way that makes sense, and still embodies their identity as a race.

Just a numbers change, doesn't really need testing.

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/furry/dracon.dm
+++ b/code/modules/mob/living/carbon/human/species_types/furry/dracon.dm
@@ -57,6 +57,10 @@
 		"speed" = -1,
 		"fortune" = 0
 		)
+	specstats_f = list(
+		"constitution" = -1,
+		"endurance" = 1,
+	)
 	enflamed_icon = "widefire"
 	attack_verb = "slash"
 	attack_sound = 'sound/blank.ogg'


Meltharas:
I agree, I think that the Drakians need a slight buff. Their wings are for show so they likely gain more muscle from the extra weight, also thanks to the powerful dragon blood in them! I could see them as mighty warriors, even if they are slower and get an endurance hit. Makes them more desirable for combat roles in general.

YaBoiWes:
Given the fact that they are purposefully towering dragon people who are supposed to be more akin to half-orcs in size, I see no reason it would not make sense for them to have a more combat aligned statistical buff. This would see an increase of drakians overall, but given their nature as warriors in their lore? Not surprising and somewhat welcome.

BattleTitties:
Hard hitting and decently tough, but not as fast or enduring in the stamina department would be interesting and different. Historically we're speaking about a race of proud, ageless beings who's claim to fame is their great conquests and 'ruling' class. It does check out from the lore we do have on them.

Red-Honk:
I like it, tested out, works just fine, lore wise wise it makes sense, now for the male character they are a tad mightier, and for the female ones, it just makes it seem that there's no down side, besides a endurance minus, which isn't much if roleplay is the priority.

RadiantFlash7:
I don't consider the endurance penalty over crippling, as most roles that have need of endurance often have end bonus included in their role. So Yeah, it hurts, but it's atleast interesting over having the same stat block as other races with an increase in strength.

RadiantFlash7 (approved):
Changes seemed to work out fine, issues with potentially too much endurance loss resolved.

PR #2882, opened 2025-05-21
Sawbones Rework

envoyin:
## About The Pull Request

Changes the Sawbones into an ex-clergy type called the Defrocked.

They have a maximum of 350 devotion, churchling miracles, and have had some skills reduced or removed

Treatment reduced by 1
Farming removed
Athletics has not been touched.

But they have level 3 miracles, and gain +1 to their holy skill if they are OLD.

I believe they should keep their ability to purchase surgical tools and first aid kit from the horde master, as I think they would end up being the camp's doctor lorewise.

If this needs changes for balance, I am happy to do what I need to do.

_**I'm not exactly a coder for BYOND, so this may be sloppy and improvised.**_
## Why It's Good For The Game

Sawbones is a poorly designed class, a large part due to the physician nerfs. Treatment healing is good, sure, but it requires being stationary a lot, and is much more of a non-combat healer.

Bandits are an extremely mobile group of antags, and most of their healing is needed MID combat, as they do not have the luxury of derot or Astrata's revival miracle.

This, combined with bottles of red being a great alternative to a sawbones, makes the sawbone's one job to replace eyes and do surgery when they don't want to waste red (though many will ignore the sawbones and chug red anyway) and have time. 

image
image
image

### DIFF
--- a/code/modules/jobs/job_types/roguetown/bandits/types/sawbones.dm
+++ b/code/modules/jobs/job_types/roguetown/bandits/types/sawbones.dm
@@ -23,13 +23,8 @@
 	backpack_contents = list(/obj/item/rogueweapon/surgery/hemostat/improv = 1, /obj/item/rogueweapon/surgery/retractor/improv = 1, /obj/item/rogueweapon/surgery/saw/improv = 1, /obj/item/reagent_containers/glass/alembic = 1,)
 	H.mind.adjust_skillrank(/datum/skill/combat/knives, 3, TRUE)
 	H.mind.adjust_skillrank(/datum/skill/combat/wrestling, 3, TRUE)
-<<<<<<< Updated upstream
 	H.mind.adjust_skillrank(/datum/skill/craft/crafting, 2, TRUE)
 	H.mind.adjust_skillrank(/datum/skill/craft/carpentry, 2, TRUE)
-=======
-	H.mind.adjust_skillrank(/datum/skill/craft/crafting, 1, TRUE)
-	H.mind.adjust_skillrank(/datum/skill/craft/carpentry, 1, TRUE)
->>>>>>> Stashed changes
 	H.mind.adjust_skillrank(/datum/skill/labor/lumberjacking, 1, TRUE)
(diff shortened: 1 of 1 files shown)


DelineFortune:
Make them real inhumen clerics
https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/blob/d2e9305adaec65965b74efb8394a9c7c28643c18/modular_azurepeak/code/modules/spells/pantheon/inhumen/matthios.dm#L4
Why do they need skills removed even
They are only crafters/doctors that bandits have

Skyglazer:
Very against this; this is a MASSIVE buff to bandit in general, miracle healing is extremely potent and would turn the tide of battle more in their favor than their easy access to Red already does. Sawbones already has their niche in doing the eye surgery, being a knife-fighter and a bit of a wrestler that can even revive bandits with Lux. That, and this is amount of devotion is really high- so high that if you go into a fight with full devotion (as you usually will) you will likely never run out. Contrast with Templars, whose low devotion pool means they can't just spam miracles forever. There are ways to make Sawbones a better role without making it extremely strong like this.

GravyMaker:
Bandits have ready access to health potion, and though some acolytes have a habit of hanging around Lowtown, they're not really such a reliable presence to the vanguard or retinue that having a dedicated combat healer who could be expected to be glued to other bandits would be suitable to balance. I think there are considerably better ways to go about improving the bandit experience than buffing Sawbones to have miracles when the flavour is that they were a secular physician in the first place - indeed, I don't think bandits themselves need to be buffed closer to the state they were previously in at all; I think there should be a bit more for the town combat roles to do instead.

Gyrandola:
Not a fan of turnint sawbones into a miracle based healer.

RadiantFlash7:
> Would it be possible to discuss this once more, now that the hordemaster PR is fully merged?

After discussion with maintainer team and others. I'd be up for giving this a shot. My only request on changes would be to shift the outfit you gave them. Councillor nobles have the hierophant stuff right now, I don't think it's really 'monk' clothing. Toss 'em something else and Resolve the conflicts, and we'll see about a test merge.

PR #2885, opened 2025-05-22
Allows a few more races to choose the New Moon Spellblade mercenary class.

Skyglazer:
## About The Pull Request

This PR just allows a few more races to choose the New Moon Spellblade subclass.

## Why It's Good For The Game

More player freedom! It's a commonly requested change that people allow more races (especially Lupian!) to be New Moon Spellblades, and the lore on the role doesn't seem to imply the pretty strict race-restrictions the role has. They hail from Zybantia, and other Zybantian roles like the slavers and desert riders don't have as much restriction. It makes sense that the more intellectually-inclined races like Lupians, Fluvians, Drakians, etc. could find themselves as part of this order. I don't really see it as a balance concern either-- Tabaxi is already allowed (and the race you'll see most commonly in this role) and they make great use of dodge expert with their speed bonuses. It'll allow more players to be mercenary spellblades without having to resort to adventurer or refugee battle magos (though Battle Magos will probably be a bit better because of the armor proficiency).

Compiles and runs, spawned in as a Lupian New Moon Spellblade on the test map. Here is a video demonstrating spawning in as a Lupian mercenary and choosing the subclass:

[https://www.youtube.com/watch?v=JvCS7tulqrA](url)

### DIFF
--- a/code/modules/jobs/job_types/roguetown/mercenaries/types/newmoon.dm
+++ b/code/modules/jobs/job_types/roguetown/mercenaries/types/newmoon.dm
@@ -6,7 +6,11 @@
 		/datum/species/tabaxi,
 		/datum/species/elf/wood,
 		/datum/species/demihuman,
-		/datum/species/tieberian
+		/datum/species/tieberian,
+		/datum/species/dracon,
+		/datum/species/anthromorph,
+		/datum/species/moth,
+		/datum/species/lupian,
 	)
 	outfit = /datum/outfit/job/roguetown/mercenary/newmoon
 	maximum_possible_slots = 10


DelineFortune:
remove shunned race
add all furry races 
ok

Elli-Skala:
This is a fully unrelated rant but they are noc worshipers, why are they called new moon when a new moon is when the moon isn't in the sky? I feel like it should be fullmoon mercenaries. Is there some lore reasons?

Heheniy:
demis, elfs and tieflings should be removed

notafurryy:
> This is a fully unrelated rant but they are noc worshipers, why are they called new moon when a new moon is when the moon isn't in the sky? I feel like it should be fullmoon mercenaries. Is there some lore reasons?  

They shine Noc's light when the world is lacking it, shrimple

PR #2886, opened 2025-05-22
New dungeon, robust mobs that can use spells, random mobs generation

DelineFortune:
## About The Pull Request

Adds a dungeon beneath the city with random loot. NPCs that can use abilities. Bosses that can use abilities. Random NPC generation in rooms. 6 types of dungeon NPCs and 3 types of bosses. 1 type of NPC adventurer and 1 type of NPC bandit.

## Why It's Good For The Game

PvE content no mobs that cannot do anything against your 90% blocks


DelineFortune:
reworked in better version

PR #2891, opened 2025-05-24
Disables "Sleep" verb for standard player characters

GravyMaker:
## About The Pull Request
This PR disables the ability to use the "sleep" verb to fall asleep more quickly than is intended to be possible. It is still available to spirits and monkeys.

## Why It's Good For The Game
So, I recognise that this will be a controversial PR for some, so I'll try to be comprehensive in my justification of it.

Sleep has a few functions in Ratwood. It clears the tired status condition, allows players to learn and increase skills, and, critically, gives health regeneration for as long as it lasts, and restores your stamina and fatigue. For this reason, sleep is intended to be something that takes a bit of time to do - you remove uncomfortable armour, close your eyes, wait a while, and then eventually fall asleep - and once you're asleep, it takes a while to wake up once you "open" your eyes.

Why does sleep function in this way? The main use-case for instant sleeping would be during protracted fights - retreating from the frontline, using the sleep emote to sleep for a short period of time, bypassing the need to remove and re-equip armour, and replenish health, stamina, and fatigue, before waking up quicker than they would otherwise. and getting straight back into the fight.

Now, as far as I can tell, this is not intended - the _intended_ method of sleeping is laying down, closing your eyes, and waiting, and for good reason - sleeping in a dangerous environment is supposed to be dangerous. Sleeping close to a battlefield should be an extremely dangerous thing - almost never worth it. Sleeping using the emote is a holdover from base SS13, and one which allows those more inclined to powergame to bypass these design elements, allowing them to sleep far more quickly than they're intended, drastically lessening the danger of doing it close to combat, and similarly lessening the amount of time they might have to abandon a post. By having both methods, we allow certain players who know about it to bypass the intended design, giving themselves an advantage that shouldn't exist - being much more effective over long, drawn-out combats or watches than they should be, rather than exhausted; and we put players who don't know about it, or otherwise are reluctant to use it out of a sense of propriety, at a disadvantage. For that reason, I think this emote should be disabled on human players, requiring that everybody use the same, intended method of sleeping.

Whether the existing sleep and wake-up timers are a sensible length or too long is, I believe, a separate discussion; but ultimately, everybody should be on a level playing field, and there shouldn't be a means to use emotes to bypass design elements.

The emote now does nothing:
image

Sleeping itself still functions correctly, however:
image

### DIFF
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -53,7 +53,6 @@
 #ifdef MATURESERVER
 	sexcon = new /datum/sex_controller(src)
 #endif
-	verbs += /mob/living/proc/mob_sleep
 	verbs += /mob/living/proc/lay_down
 
 	icon_state = ""		//Remove the inherent human icon that is visible on the map editor. We're rendering ourselves limb by limb, having it still be there results in a bug where the basic human icon appears below as south in all directions and generally looks nasty.


ophaq:
I didn't even know you could do this. Wow. I wish we had a faster wakeup time too.

RadiantFlash7:
I was not aware this was a thing either. This is probably needed...

GravyMaker:
Since it's been brought to my attention in Discord - I should add that skeletons are apparently also able to use the verb to sleep, so this PR should fix that issue as well.

RadiantFlash7 (approved):
Removal of the sleep verb will help prevent abuse- so Approved.

PR #2894, opened 2025-05-25
Tabaxi now have purr efect when saying anything with R. 

Erol509:
## About The Pull Request

Tabaxi now have their own tongue, which affects their speach slightly. Somewhat simillar to the old drakians, and perhaps better example would be lizard people from TG. 

If its a problem, I can also just do it via selectable accent. But that list is only growing bigger and bigger, so I decided to stick with this.

## Why It's Good For The Game

Makes the way that tabaxi is speaking a bit more unique. Its similar to what cat tongue had in the past on Skyrat, its just multiplying R or r within the sentance, for slight purring effects. Thats the easiest way to achive it, instead of replacing whole words. 

This might not be liked by many people, but Im sure it adds a bit more flavour overall to the race, and making them slightly more unique compared to other races. 

Will add it soon, it works. But might refine the regex values a bit, not sure of it yet.

### DIFF
--- a/code/modules/surgery/organs/tongue.dm
+++ b/code/modules/surgery/organs/tongue.dm
@@ -76,7 +76,8 @@
 	var/static/regex/tabaxi_purr = new("r+", "g")
 	var/static/regex/tabaxi_Purr = new("R+", "g")
 	var/message = speech_args[SPEECH_MESSAGE]
-	if(message[1] != "*")
+	var/language = speech_args[5]
+	if(message[1] != "*" && (language != /datum/language/felid))
 		message = tabaxi_purr.Replace(message, "rrr")
 		message = tabaxi_Purr.Replace(message, "RRR")
 	speech_args[SPEECH_MESSAGE] = message


RadiantFlash7:
Iii am not a fan adding seperate organs to force speech patterns. Theres a reason we have it so you can pick your accent at present in game. While normally I would like races to have unique traits/effects- I don't think something that affects speech is a good idea.

Erol509:
> Iii am not a fan adding seperate organs to force speech patterns. Theres a reason we have it so you can pick your accent at present in game. While normally I would like races to have unique traits/effects- I don't think something that affects speech is a good idea.

I can make in an accent, if you prefer it. But adding yet another accent to the already long and still growing list, sooner or later might become a problem.

Erol509:
Besides the regex values are very little affecting, its the same as hissing accent  with and for only one letter.

Erol509:
> > Besides the regex values are very little affecting, its the same as hissing accent with and for only one letter.
> > Edit: I can also make it based on prob, so only lets say 50% will be affected
> > And also there is no other way for me to change the whisper, yell verbs without adding a new tongue. Which was the initial purpose of this PR.
> 
> I mean, if you check up on the VV there is a place where you can change, whisper, say, exclaim, ect, you could force it to, if Tabaxi to x y or z modify.

That would require a new proc which runs on roundstart with checks for race. While organs initialise on spawning, and tongue affects your speech, which is usefull for speech mod. Which would have to be reused again in said proc. 

Thats 2x the work for more complicated code.

ophaq:
Finally, I've been doing this manually.

Gyrandola:
> i personally don't like how cutesy this is. especially the new say words make little sense literally and thematically "x mrrps y" just seems way too terminally furry to really fit the setting...

I mean, good point. I think the "r" -> "rrr"s are good to have though

Erol509:
> > i personally don't like how cutesy this is. especially the new say words make little sense literally and thematically "x mrrps y" just seems way too terminally furry to really fit the setting...
> 
> I mean, good point. I think the "r" -> "rrr"s are good to have though

Might just end up with simple "meows" for say_mod

RadiantFlash7 (changes requested):
Riiiight, Coming back to this now.
@Erol509 A few things I'd like done. Keep their 'say' verb as the same as everyone else. Don't make it furrified with purrs or mrowls/meows.
Secondly, If we want this to go through, I'd like you to ensure that they only roll their R's in imperial. It wouldn't make sense to enforce R rolling in the tabaxi language, after all.

Erol509:
> Riiiight, Coming back to this now. @Erol509 A few things I'd like done. Keep their 'say' verb as the same as everyone else. Don't make it furrified with purrs or mrowls/meows. Secondly, If we want this to go through, I'd like you to ensure that they only roll their R's in imperial. It wouldn't make sense to enforce R rolling in the tabaxi language, after all.

Not sure if there is an easy way to do it, unless I try to nullify regex when someone uses ,f

Erol509:
image
Okay got it

RyszardParys:
Good lord, let me toggle it off at least, this is ass...

Erol509:
> Good lord, let me toggle it off at least, this is ass...

only if maintainers want me to do so

ophaq:
Yeah this is nice but it should be optional like the lupian accent.

Skyglazer:
Why have it be forced? Why not have it be optional?

Peter-Dolby:
As others have said, would rather this be optional.

Gyrandola:
not everything has to be imo. This is a racial trait.

Peter-Dolby:
If anything I'd expect this to be tied to lupians.

Peter-Dolby:
> We wanted to make races more distinct and unique and this ought to help towards that goal. No issues during tm, Approved.

I dunno, I think the fact that lots of people don't like it would count as an Issue during tm.

PR #2897, opened 2025-05-27
Fix for sewn armour not graphically updating when repaired & examine text not correctly identifying broken items

GravyMaker:
## About The Pull Request
This PR fixes two issues:
1) Sewn items, when repaired, did not correctly update their sprites from the broken state.
2) When examined, broken items' examine text would not necessarily say that the item was broken (ie. unable to provide protection) if its break integrity threshold was greater than 1% of the maximum integrity.

This is achieved by replacing direct mutations of the obj_integrity value with a call to a new mend_damage proc, which handles modifying the integrity, enforcing the maximum integrity, and updating the object's graphics in the event that it's mended from a broken state. This proc takes a second argument which defines whether it will be allowed to mend a broken item or not.

## Why It's Good For The Game
The "broken" sprites for equipment are - deliberately - unsatisfying and unappealing to look at, and remaining in that state makes it difficult to visually identify whether an item has been broken and subsequently mended, or is still broken. This fix makes it easier to tell at a glance whether the armour will still provide protection or not.

Additionally, examine text saying that an item is severely damaged when it's actually broken and can provide no protection is misleading; this fix makes the text more reflective of the item's actual ability to protect, reducing possible player confusion.

Myself wearing chain chausses and a padded gambeson, which I'll be damaging with the Crimson Fang.
image

The padded gambeson and chain chausses on a table, in the broken state and showing the appropriate graphics.
image

Metallic items still can't be repaired with blacksmithing < 4.
image

Having raised by blacksmithing to 4, the chausses have been repaired to a functioning (but still low-integrity) state, showing blacksmithing still works as expected.
image

Giving myself sewing 2, I've now repaired the padded gambeson to a functioning state. It now has the unbroken sprite again.
image

And finally, the object will now display as broken when the broken flag is set, even if its integrity isn't below 1%:
image


RadiantFlash7 (approved):
Classic bug fixes. Thank you.

PR #2903, opened 2025-05-28
Give Ecclesiast Monk Churchling Miracles

Skyglazer:
## About The Pull Request

This PR gives churchling miracles (Miracle and Diagnose) to the Refugee Ecclesiast subclass, Monk, while reducing its strength bonus by 1. 

## Why It's Good For The Game

This subclass has been pretty much entirely left in the dirt because of Refugee Apostate Dualist having more stat increases and traits based on the patron, and I think that giving them some access to very basic divine healing would be a good way to catch them back up. They already have 1 holy skill already, so might as well give them a way to use it! It will get more people playing the role and more able healers is always good, as people tend to avoid playing them. 

This change will make them decent, but not so great that they overshadow any town roles, or any refugee roles where miracles are their specific niche-- returning crusader and preacher are still better-equipped for miracles, as are Refugee Apostate Fanatics. Meanwhile, the reduced strength will make it so that the church's Acolyte and the Apostate Dualist will remain the best wrestlers, so that Ecclesiast Monk has a niche to call its own.

https://www.youtube.com/watch?v=EjzBmQpJvFs

### DIFF
--- a/code/modules/jobs/job_types/roguetown/refugees/types/ecclesiast.dm
+++ b/code/modules/jobs/job_types/roguetown/refugees/types/ecclesiast.dm
@@ -223,7 +223,7 @@
 			beltl = /obj/item/storage/belt/rogue/pouch/coins/poor
 			backl = /obj/item/storage/backpack/rogue/backpack
 			r_hand = /obj/item/rogueweapon/woodstaff
-			H.change_stat("strength", 3)
+			H.change_stat("strength", 2)
 			H.change_stat("intelligence", 2)
 			H.change_stat("endurance", 1)
 			H.change_stat("constitution", 1)


ddsaw:
> ## About The Pull Request
> This PR gives churchling miracles (Miracle and Diagnose) to the Refugee Ecclesiast subclass, Monk, while reducing its strength bonus by 1.
> 
> This subclass has been pretty much entirely left in the dirt because of Refugee Apostate Dualist having more stat increases and traits based on the patron, and I think that giving them some access to very basic divine healing would be a good way to catch them back up. They already have 1 holy skill already, so might as well give them a way to use it! It will get more people playing the role and more able healers is always good, as people tend to avoid playing them.
> 
> This change will make them decent, but not so great that they overshadow any town roles, or any refugee roles where miracles are their specific niche-- returning crusader and preacher are still better-equipped for miracles, as are Refugee Apostate Fanatics. Meanwhile, the reduced strength will make it so that the church's Acolyte and the Apostate Dualist will remain the best wrestlers, so that Ecclesiast Monk has a niche to call its own.
> 
> ## Proof of Testing (Required)
> https://www.youtube.com/watch?v=EjzBmQpJvFs

regarding apostate vs monk: simply incorrect. Monk has 9(8 after the str nerf) total statpoints, Apostate has at BEST(baotha) 8(with a graggarite apostate having 3(but to be fair- losing a lot of "bad" stats for good stats).
Apostates DO get god traits, most of which like good lover/bedbreaker, are combat useless and matter only for flavor. The best possible trait is undeadfaction, allowing for some cheesy looting.
Monks also get sewing 2, unlike Dualists, which is actually quite a big deal.

Skyglazer:
Well, I'll accept that then- you put up a pretty convincing argument. I suppose my perspective was mostly just informed by how Apostate Dualist is so popular while Ecclesiast Monk is basically never played, I figured it HAD to be better somehow. Would be worth looking into giving them churchling miracles too, if this gets accepted.

RadiantFlash7:
I'm skeptical about adding healing to the monk refugees. I'll see what the other maints think.

PR #2904, opened 2025-05-28
Stops fish multi initialize

ARefrigerator:
## About The Pull Request
Stops the runetime crash of fish calling their initialize multiple times. If you know why they still would try to call init twice let me know and I Can try fixing that too.

## Why It's Good For The Game
runtime.

runtime gone, fish still inits twice??
EDIT: never mind I did this. It works fine. Runtime is fixed and it only inits once.
image

### DIFF
--- a/code/modules/roguetown/roguejobs/fisher/rod.dm
+++ b/code/modules/roguetown/roguejobs/fisher/rod.dm
@@ -103,9 +103,7 @@
 		return
 
 	var/caught_thing = pickweight(baited.fishloot)
-	var/obj/item/I = new caught_thing()
-
-	I.forceMove(current_fisherman.loc)
+	var/obj/item/I = new caught_thing(current_fisherman.loc)
 
 	if(istype(I, /obj/item/reagent_containers/food/snacks/fish))
 		var/obj/item/reagent_containers/food/snacks/fish/F = I


DelineFortune:
Thank a lot
I have tried to increase chances of rarity from hot spring water but 
It: creates the fish first in your hands
Then: sets the rarity

code/modules/roguetown/roguejobs/fisher/rod.dm
```diff
@@ -103,14 +103,12 @@
...
-	var/obj/item/I = new caught_thing()
-
-	I.forceMove(current_fisherman.loc)
+	var/obj/item/I = new caught_thing(current_fisherman.loc)
 
 	if(istype(I, /obj/item/reagent_containers/food/snacks/fish))
 		var/obj/item/reagent_containers/food/snacks/fish/F = I
 		F.fished_from = target
-		F.Initialize()
+		F.set_rarity()
```
noelle-lavenza, line 111: this isn't necessary, Initialize() will be called when it's created with `new`, which will call `set_rarity()`. Frankly, I'd suggest adding `fished_from` as a parameter to Initialize() later, but it's not necessary for this fix
```suggestion
```

noelle-lavenza (changes requested):

code/modules/roguetown/roguejobs/fisher/rod.dm
ARefrigerator: but will that recognize the variable I'm assigning under it?

noelle-lavenza: Oh, in that case it won't. That also makes passing it as a parameter really annoying. This works fine.

PR #2905, opened 2025-05-28
Memory & Nerve Master job anonymisation improvements

GravyMaker:
## About The Pull Request
This PR aims to prevent metagaming by anonymising jobs of "foreigner" roles in the nerve master and memory interfaces, alongside preventing werewolves from appearing under their werewolf names in the nerve master bank interface.

In cases where a character's examine text would show them as a wanderer or foreigner (with the exception of mercenaries), these characters' jobs will now display as "Foreigner" in the memory and nervemaster interfaces, rather than revealing information that should be hidden (for example, revealing somebody to be an Apostate). In the case of a transformed werewolf, the nerve master will now show the werewolf's regular name, if they have an account, rather than the werewolf name assigned when they transform.

## Why It's Good For The Game
These changes restrict two avenues of metagaming that I've become aware that some players exploit - namely, that one can identify a refugee or wanderer's role by examining them and checking their memories, or by checking their account in the nervemaster, should they create one at a nervelock. I'm not sure of the precise extent to which these issues were exploited, but a friend has told me of one occasion on which they, as Duke, were told by the Steward to declare several people Outlaws without much explanation as to why, only later learning that the Steward had been exploiting this oversight.

With these oversights fixed, it will no longer be possible to gain knowledge of player roles or werewolf identities in this way, reducing both the temptation and ability to metagame.

Nerve master interface showing my character's job as "FOREIGNER" after joining as an Apostate Duallist:
image

The same, with compact mode enabled:
image

Having switched to another character and examined the original character, my memories, listing the original character's job as "Foreigner":
image

Image showing my character's regular name, even after being transformed into a werewolf:
image

And confirmation that the "unknown" fallback still functions correctly, when no job is available:
image

As does the fallback in the nerve master, for those with an account but no job:
image

### DIFF
--- a/code/datums/mind.dm
+++ b/code/datums/mind.dm
@@ -183,7 +183,12 @@
 						used_title = J.f_title
 				if(!used_title)
 					used_title = "unknown"
+
+				var/datum/job/roguetown/job = SSjob.GetJob(H.job)
+				if (job && job.should_anonymise_job())
+					used_title = "Foreigner"
 				M.known_people[H.real_name]["FJOB"] = used_title
+
 				M.known_people[H.real_name]["FGENDER"] = H.gender
 				M.known_people[H.real_name]["FSPECIES"] = H.dna.species.name
 				M.known_people[H.real_name]["FAGE"] = H.age


RadiantFlash7 (approved):
No isuses during test merge, Approved!

PR #2906, opened 2025-05-28
Undead swift intent fix

GravyMaker:
## About The Pull Request
This PR fixes undead antags being able to use the Swift combat intent.

Swift is a very powerful combat intent, allowing a character to attack (or grab) much more rapidly, at the expense of consuming considerably more stamina and thus tiring themselves out much more quickly - something which is likely to be deadly in combat. Skeletons and vampyres, however, do not consume stamina. For this reason, the Swift intent is actually removed from those creatures' combat intent options; however, due to an oversight, if a character has Swift intent selected when they're turned, they retain it even after it's no longer a valid intent for them to select.

This PR addresses this problem by checking, on conversion, whether the character has the Swift intent selected and, if so, forcing their combat intent back to Feint. Additionally, Swift is removed from the lich's available combat intents for the same reason; I suspect that the reason it wasn't already removed was, again, an oversight.

The PR does not modify deadites; this is because deadites don't actually appear to have the NOSTAMINA trait, and aren't player-controlled anyway.

## Why It's Good For The Game
Ultimately, creatures with TRAIT_NOSTAMINA make Swift intent overpowered - the intent is mainly held back by its hefty stamina usage, and when stamina is no longer a concern, it becomes extremely powerful. Indeed, this was clearly recognised at some point, hence why the intent was mostly made unavailable to undead - it's only due to the oversight of forgetting to modify the active intent on conversion that it's possible to use as an undead at all. This PR simply corrects that mistake - undead still have the benefit of not requiring stamina, but no longer benefit from the combat intent which is balanced only by its heightened stamina consumption.

Myself, with swift intent selected.
image

Myself, having been turned into a vampyre via the Traitor Panel. Note that my combat intent has been forced to Feint.
image

Additionally, the Lich can no longer select Swift intent:
image

I've also tested with vampyre spawn and animated skeletons, which also work correctly, and with swift deselection when becoming  alich, which also works correctly; but I'll spare you all the screenshots.

### DIFF
--- a/code/modules/antagonists/roguetown/villain/lich/lich.dm
+++ b/code/modules/antagonists/roguetown/villain/lich/lich.dm
@@ -53,17 +53,6 @@
 	greet()
 	save_stats()
 
-	// Undead have infinite stamina; they should not be using swift intent under any circumstances.
-	if (istype(owner.current.rmb_intent, /datum/rmb_intent/swift))
-		owner.current.swap_rmb_intent(null, 1)
-
-	// No swift intent for undead!!
-	owner.current.possible_rmb_intents = list(/datum/rmb_intent/feint,\
-		/datum/rmb_intent/aimed,\
-		/datum/rmb_intent/strong,\
-		/datum/rmb_intent/riposte,\
-		/datum/rmb_intent/weak)
-
 	return ..()
 
 /datum/antagonist/lich/greet()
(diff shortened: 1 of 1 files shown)


RadiantFlash7 (approved):
Fixes an old exploit. Thank you!

PR #2912, opened 2025-05-29
Fix for necromancer-summoned skeletons preserving Swift intent

GravyMaker:
## About The Pull Request
This PR extends #2906 by removing the ability to preserve Swift intent from skeletons raised with Raise Lesser Undead, which was missed from the previous PR due to a silly oversight on my part.

## Why It's Good For The Game
See #2906 

image


RadiantFlash7 (approved):
We doing what we call a speedmerge for something I missed on the last PR. My bad.

PR #2913, opened 2025-05-29
removes lich from antag roles

ARefrigerator:
## About The Pull Request
Removes lich from antag selection. I think this is all we need to do since the function to select antags fires through this list so removing anything should fix it I think. The only issue I could see arising is an issue where you have it enabled and saved and then it can't load the selection? But all my testing showed nothing.

If anything else has to be removed let me know but the glance I had at the usage of this list suggests this should be all one needs to do...

## Why It's Good For The Game
Every lich round turns into fragging hell. I suggest we disable the antag role lich and give the necromancer class of apostate more love to be fun and expandable. This would allow it to be a more direct opt in choice and starts them in the bog.

Removed from selection:
image

### DIFF
--- a/code/__DEFINES/role_preferences.dm
+++ b/code/__DEFINES/role_preferences.dm
@@ -65,8 +65,7 @@ GLOBAL_LIST_INIT(special_roles_rogue, list(
 	ROLE_ASPIRANT = /datum/antagonist/aspirant,
 	ROLE_WEREWOLF = /datum/antagonist/werewolf,
 	ROLE_ZIZOIDCULTIST = /datum/antagonist/zizocultist,
-	ROLE_NBEAST = /datum/antagonist/vampirelord,
-	ROLE_LICH = /datum/antagonist/lich
+	ROLE_NBEAST = /datum/antagonist/vampirelord
 ))
 
 //Job defines for what happens when you fail to qualify for any job during job selection


DelineFortune:
Ratwood is only one server that cannot handle with antags 
We had turned off for half of year:
Vampires (returned back with minor changes)
Goblins (returned back with minor changes)

Now we have:
Turned off maniac 
Turned off cult 

And there is goes lich

Cool

ARefrigerator:
> Cool
Liches are clearly the only antags right now that have no pretext of trying to RP. There's nothing wholly interesting to them either. I guess I could add to them to make it more interesting if people want but now we're in a situation where admins call round ends at 1:50 into the round because literally everyone is dead.


> I'm not necessarily opposed to disabling lich for a period, but I'd personally want there to be at least a plan on how to rework it, or some other antag to take its place, first.

My plan was to buff necromancer to take its place as an opt in antagonist. But if people have ideas for how to change lich or more antags to add I could try those as well here. I have had ideas of some other kinds but I do also think some being more role play orientated could be nice as well. I'll think about it but I'll leave this here for now.

DelineFortune:
> > Cool
> 
> Liches are clearly the only antags right now that have no pretext of trying to RP. There's nothing wholly interesting to them either. I guess I could add to them to make it more interesting if people want but now we're in a situation where admins call round ends at 1:50 into the round because literally everyone is dead.

Its server with fast revives with a lot of combat roles 
Join as a towner
Re take the city back
City is not supposed to survive or always win

ARefrigerator:
> Its server with fast revives with a lot of combat roles Join as a towner Re take the city back City is not supposed to survive or always win

That's not what I care about what I care about is there's basically no build up at all. Once people know a lich is about everyone has to turn into a soldier to kill it. All plans for the round vanish. I think if we bring back cult it would improve things as it would do the following
1. The cultists are human so killing them could be final
2. Cultists would have more strict requirements for winning they would have to follow
3. Upon succeeding those harder requirements they could perhaps end the round as well

You can't really have any deceit, build up, or discussion with a skeleton. now maybe if more antags get in there would be less liches just by chance but we'd have to get to that point first.

ARefrigerator:
I'm gonna close this for now while we get more antags ready. I guess that might need to be my main priority because it seems like common lich rounds are a pain.

Dioclex:
> My plan was to buff necromancer to take its place as an opt in antagonist. But if people have ideas for how to change lich or more antags to add I could try those as well here. I have had ideas of some other kinds but I do also think some being more role play orientated could be nice as well. I'll think about it but I'll leave this here for now.

Yeah I'd say if you intend to have another go at this PR, include the "planned" antag rework with it -- since it's been demonstrated time and time again that people are content to disable antags and provide no follow through by fixing the issues that warranted them being disabled to begin with.

PR #2917, opened 2025-05-29
Removes refugee noble gender-based class selection.

Skyglazer:
## About The Pull Request

This PR makes it so the Refugee Noble's subclass selection is the same regardless of character gender, renaming the existing subclasses into gender-neutral ones and making each of them available for both genders. The "widowed consort" role even gives male characters a shirt and pants rather than a dress.

## Why It's Good For The Game

The subclasses being gender-locked strikes me as more of a holdover from older versions of RT lore where women weren't allowed in combat roles. In our server, women can be knights, mercenaries, templars, all sorts of warriors-- so it makes sense to allow them to pick the archetypes available to male characters in the Noble refugee role.

Forgot to use the test map for this one, but here is a video of me spawning in as each role with a character of both genders. https://www.youtube.com/watch?v=EnSS5QFic6g

### DIFF
--- a/code/modules/jobs/job_types/roguetown/refugees/types/noble.dm
+++ b/code/modules/jobs/job_types/roguetown/refugees/types/noble.dm
@@ -20,18 +20,13 @@
 	..()
 	H.adjust_blindness(-3)
 
-	var/classes
-
-	if(H.gender == MALE && !isdarkelf(H )|| H.gender == FEMALE && isdarkelf(H))
-		classes = list("Disgraced Lord", "Failed Condottiero")
-	else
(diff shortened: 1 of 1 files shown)


GravyMaker:
Looks good to me. I can understand the argument for some roles or sub-roles being sex-restricted, but given the town and keep roles which aren't, all of these seem sensible to me to be unrestricted in that way.

AukariBunn:
The code seems solid to me.

RadiantFlash7 (approved):
This PR's been TM'd with no issues. Approved.

PR #2918, opened 2025-05-30
Tabaxi Statblock Nerf

Skyglazer:
## About The Pull Request

This PR changes the Tabaxi statblock a bit, giving them a -1 STR penalty in exchange for their -1 CON penalty being removed, bring it more in-line with other +2 SPD races.

## Why It's Good For The Game

This brings Tabaxi more in-line with other races that get SPD boosts- Kobold, Verminfolk, Fluvian-- all have a +2 bonus to SPD just like Tabaxi, except each of these races have a whopping -2 to STR to balance it out. Considering most races that have a STR bonus get a SPD penalty, I think it makes sense for tabaxi to take a (smaller than the others) penalty to STR in exchange for their high speed. It makes sense flavorwise as well- cats are known for elegance and their flavortext describes them as slender.

I would still wish that statpacks be implemented so people could play a tabaxi as like a big strong lion or tiger without as much speed if they so wished, but this is the system we live with for now, and this I think would help balance it a bit more. In my opinion, the current Tabaxi statblock is a bit overtuned.

Just a number change.

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/furry/tabaxi.dm
+++ b/code/modules/mob/living/carbon/human/species_types/furry/tabaxi.dm
@@ -37,10 +37,10 @@
 		OFFSET_SHIRT_F = list(0,0), OFFSET_ARMOR_F = list(0,0), OFFSET_UNDIES_F = list(0,0)
 		)
 	specstats = list(
-		"strength" = 0,
+		"strength" = -1,
 		"perception" = 1,
 		"intelligence" = 0,
-		"constitution" = -1,
+		"constitution" = 0,
 		"endurance" = -1,
 		"speed" = 2,
 		"fortune" = 0


ddsaw:
I think we should just entirely rework or remove dodgeexpert, the only thing that makes high spd a real problem, and suddenly speedsters are not as much of an issue.

Skyglazer:
> I think we should just entirely rework or remove dodgeexpert, the only thing that makes high spd a real problem, and suddenly speedsters are not as much of an issue.

Disagree. Movement speed is a HUGE component of what makes high SPD so good. Ever tried chasing a high SPD role as a low SPD role? They can just walk away from you. Kobold/Tabaxi knaves can pull around a cart full of bandits at breakneck speeds and unleash them on the masses, and then make a quick getaway before any of the low-SPD knights can catch them. High SPD roles get the liberty of choosing whether a fight in open areas like the bog continues or not-- if they don't like their odds, they can easily escape. If their lower-SPD opponent is trying to run, they can easily catch up. But that's all a different conversation. I still think them getting +2 SPD with no STR drawback is a bit much.

Skyglazer:
> Kobold are verminfolk are all 'small'. Fluvians don't have the same penalties.
Indeed, that's why the strength penalty for Tabaxi would be -1 instead of -2. I'm not sure what you mean about Fluvians-- they currently ARE -2 STR, +2 SPD. All the STR-increasing races I'm aware get a SPD penalty- why would we have the reverse be true for every +2 SPD race except Tabaxi?

DelineFortune:
> I'm not a fan of this change for stat blocks. Kobold are verminfolk are all 'small'. Fluvians don't have the same penalties.

Kobolds and verninfolks are also shunned and they cant get roles with high stats
Fluvians can
Instead of adventurers or mercs

ddsaw:
> it makes them resprited elves with fur also dont count shunned races and not shunned races as same shunned races have massive job lock in exchange of ~a bit better stats and people hate tabaxi ingame for their rogue-like stats why do we need to remove it

shunned races do not have any more stats than non shunned races, this was only ever a balance consideration for horcs, but you have dwarves which are respected and have almost horc stats..

Erol509:
-1 and sometimes even more endurance on minus is already punishing enough for tabaxi. Const for strenght isnt a good tradeoff, especially when rolling STR on - as tabaxi is quite often.

PR #2920, opened 2025-05-30
Adds slipping procs, uses athletic and perception as checks

ARefrigerator:
## About The Pull Request
- Uses both the athletics as a main check for slipping and a backup check for perception in case it fails as a fallback attempt to save you from falling.
- Adds this new proc to soap as a proof of concept.

## Why It's Good For The Game
Allows us to expand slipping with more precision and get more use out of some skills.

https://github.com/user-attachments/assets/397426f4-78bc-431f-b7ef-021c00185369

### DIFF
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -862,6 +862,11 @@ Try slip
 	if(STAPER >= chance)
 		failed = FALSE
 	
+	// we could also do a check here but I think it's fair you
+	// could sneak over almost anything without punishment.
+	if(m_intent == MOVE_INTENT_SNEAK)
+		failed = FALSE
+
 	if(failed)
 		slip(I)
 /*==


ARefrigerator:
> perhaps it should be something that isn't as lethal as falling over?
This isn't a complete stun it is just being made to lie down. I don't see how it would be any worse than getting mantrapped randomly.

ARefrigerator:
> not opposed to this at all, gives perception more use cases and doesn't heavily affect combat classes, would you still drop your weapon if slipped like this with 10 per 3 athletics (the average vanguard)?
> 
> also, its could be cool to implement this system for existing sources of trips and falls, like sprinting over a large rock or a bush.

I would like t expand it to other things and probably use the difficulty mod for them. I'd also need to change how it works for knocking someone over to allow them to keep their item in their hand if that is a function someone would want however.

roarbark:
> > perhaps it should be something that isn't as lethal as falling over?
> 
> This isn't a complete stun it is just being made to lie down. I don't see how it would be any worse than getting mantrapped randomly.

well, mainly that you cant throw mantraps properly... getting floored in combat against someone who knows what they're doing is extremely lethal, because they only need to quickly put a passive grab on you to stop you from standing up

ARefrigerator:
> well, mainly that you cant throw mantraps properly... getting floored in combat against someone who knows what they're doing is extremely lethal, because they only need to quickly put a passive grab on you to stop you from standing up. i think you're underestimating the balance implications of this.

Then perhaps I could start functionality with something more like large rocks and bodies on the floor? Things less likely to be moved about. Then perhaps lowering the difficulty modifer for soap so it's less guaranteed to work unless someone is very unlucky with their rolls.

ARefrigerator:
Maybe I'll change this to be more consistent and only if you're sprinting then.

Red-Honk:
> I would like t expand it to other things and probably use the difficulty mod for them. I'd also need to change how it works for knocking someone over to allow them to keep their item in their hand if that is a function someone would want however.

Maybe just make it so, it uses the same command as "laying down", you can't go wrong with it, or accidently making it trigger to make them stand up with it.

I personally think that if one hovers over the item with the mouse cursor or at least examine, the game should not that as "No way you'd actually fall for it, since you saw it coming." type of deal.

PR #2925, opened 2025-05-31
Allows for auto mining rock walls.

ARefrigerator:
## About The Pull Request
Uses do_after to help a miner auto dig a tile based on toolspeed (I think this is the same as the cooldown for its attack speed?)
I'm not entirely sure how to get the do_after to disable once you are stunned from being out of stamina but it only tries to fire but fails. It causes a double panting emote which isn't the worst thing in the world but if anyone has an idea on how to fix that I could do it.

You still manually hit rocks, if you think I should add it to them I could.

## Why It's Good For The Game
my fingers are sore...

https://github.com/user-attachments/assets/6427216b-816a-403a-a5cb-df8e113e7530

### DIFF
--- a/code/game/turfs/closed/minerals.dm
+++ b/code/game/turfs/closed/minerals.dm
@@ -66,17 +66,16 @@
 	..()
 	// Auto mining logic
 	if(istype(I, /obj/item/rogueweapon/pick))
+		if(!isliving(user))
+			return
+
+		var/mob/living/L = user
 		var/my_icon = icon_state
 		user.doing = FALSE
+
 		spawn(1)
 		while(get_turf(src))
-		
-			if(ishuman(user))
-				var/mob/living/carbon/human/H = user
-				if(H.IsStun() || H.IsParalyzed())
-					break
-
-			if(do_after(user, CLICK_CD_MELEE, TRUE, src))
+			if((L.energy > 0) && (do_after(user, CLICK_CD_MELEE, TRUE, src)))
 				..()
 				//does this even work?
 				var/olddam = turf_integrity
@@ -90,7 +89,6 @@
 				//Turfs never die, they simply change
 				if(icon_state != my_icon)
 					break
-
 			else
 				break
 


ARefrigerator:
Changing the do_after delay to `CLICK_CD_MELEE` because it seems more in line with what it should actually be. I think toolspeed is something else.

Red-Honk:
Maybe add another intent on it, so, people can mine the old fashion styled if they so, choose to do so, you can just keep it named pick I suppose, with the description of, "For those that want to do it the old fashioned way." or the like.

SMoto95:
Might wanna add this feature for the architect drill as well, would be nice considering the drill is currently just pickaxe but worse vs rocks...

ARefrigerator:
> Might wanna add this feature for the architect drill as well, would be nice considering the drill is currently just pickaxe but worse vs rocks...

I tried with the drill. It does work. I'm not sure if the charge the drill has even works at all honestly. Nor do I think its force helps.

code/game/turfs/closed/minerals.dm
```diff
@@ -64,13 +64,35 @@
...
+		
+			if(ishuman(user))
+				var/mob/living/carbon/human/H = user
+				if(H.IsStun() || H.IsParalyzed())
+					break
+
+			if(do_after(user, CLICK_CD_MELEE, TRUE, src))
+				..()
+				//does this even work?
+				var/olddam = turf_integrity
```
Gyrandola, line 82:
```suggestion
			if((user.stamina)&&(do_after(user, CLICK_CD_MELEE, TRUE, src)))
```
Try something like this (I forgot if we need to check for `energy `or `stamina`) (`user` needs to be made `mob/living`)

Gyrandola (approved):

code/game/turfs/closed/minerals.dm
```diff
@@ -64,13 +64,35 @@
...
+		user.doing = FALSE
+		spawn(1)
+		while(get_turf(src))
+		
+			if(ishuman(user))
+				var/mob/living/carbon/human/H = user
+				if(H.IsStun() || H.IsParalyzed())
+					break
+
+			if(do_after(user, CLICK_CD_MELEE, TRUE, src))
```
Gyrandola, line 79:
```suggestion
			if((user.stamina>0)&&(do_after(user, CLICK_CD_MELEE, TRUE, src)))
```

> I'm not entirely sure how to get the do_after to disable once you are stunned from being out of stamina but it only tries to fire but fails. It causes a double panting emote which isn't the worst thing in the world but if anyone has an idea on how to fix that I could do it.

Try something like this (I forgot if we need to check for `energy `or `stamina`) (`user` needs to be made `mob/living`)

ARefrigerator: It was energy. I added it in and it seems to be working.

Gyrandola (approved):

PR #2929, opened 2025-06-01
Fixed characters shoving targets they're carrying to move themselves

GravyMaker:
## About The Pull Request
This PR fixes an issue allowing players to shove while carrying a mob to propel themselves forwards.

The fix is fairly simple - Move_Pulled(), the proc which moves a character when they shove, now checks whether the shove target is a mob buckled to the shoving character, returning false if so. It otherwise behaves as before.

## Why It's Good For The Game
This is a bug I've seen abused previously on a couple of occasions to move at speeds which should be impossible, allowing those exploiting it to potentially outrun pursuers (something I've observed once with a necromancer). It's also achieved in a particularly unimmersive way - having to carry around a living (in the code sense; the mob can be dead) mob in order to propel oneself rapidly around the map.

GIF of me moving at normal speed while attempting to shove repeatedly, while still being able to shove a chest, showing that shoving is not completely disabled while carrying a person.
dreamseeker_S87BysuGJ5

GIF of me standing still while attempting to shove repeadedly, showing more clearly that I no longer move when doing so.
dreamseeker_VbvKsEZfRK

GIF of me shoving a chest normally, while carrying nothing.
dreamseeker_ddIuStSIzi

### DIFF
--- a/code/game/atoms_movable.dm
+++ b/code/game/atoms_movable.dm
@@ -196,6 +196,9 @@
 		if(L.buckled && L.buckled.buckle_prevents_pull) //if they're buckled to something that disallows pulling, prevent it
 			stop_pulling()
 			return FALSE
+		else if (L.buckled == src) // Prevent shoving ourselves if we're carrying somebody
+			// Don't stop pulling - that will make us drop the other character - just negate the action.
+			return FALSE
 	if(A == loc && pulling.density)
 		return FALSE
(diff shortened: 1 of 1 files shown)


ophaq:
There has been a carry bug I noticed too where if you pick someone up, you don't move at full walking speed until you go up and down stairs which fixes it.

Gyrandola (approved):

PR #2934, opened 2025-06-01
Gives Vanguards and Wardens a unique combat mode track

roarbark:
## About The Pull Request

I really dislike the combat_bog track, being just a ten second loop of what sounds like pots and pans clanging together... So, this PR adds combat_vanguard.ogg and gives it to Vanguards and Wardens. Also adds it to credits.txt....

## Why It's Good For The Game

https://www.youtube.com/watch?v=ZH5eHnLGRP8

The track is Felvidék from a game of the same name. It's meant to be heroic at its core but is somehow offputting and vaguely depressing, to mirror the main character, Pavol, who is a slovakian veteran suffering from alcoholism after giving in to the trauma of almost losing his life in Kosovo. Despite this, he always carries a shit-eating grin and jokes around, knowing that the show must go on. Personally I found it fitting! All of it gives it a laugh in the face of death feeling.

While the BPM on it is on the lower end, I thought that it would still work fine, especially with how there's a few other tracks that are a bit slower and heavier, like werewolf or brigand. Talking to a few other vanguard players on discord, i got positive responses on the choice.

https://github.com/user-attachments/assets/b10413d8-2763-4ffa-a46a-4bec9b6f0d03

### DIFF
--- a/code/modules/jobs/job_types/roguetown/garrison/bogmaster.dm
+++ b/code/modules/jobs/job_types/roguetown/garrison/bogmaster.dm
@@ -23,7 +23,7 @@
 	min_pq = 8
 	max_pq = null
 	can_leave_round = FALSE
-	cmode_music = 'sound/music/combat_bog.ogg'
+	cmode_music = 'sound/music/combat_vanguard.ogg'
 
 /datum/job/roguetown/bogmaster/after_spawn(mob/living/L, mob/M, latejoin = TRUE)
 	. = ..()


noelle-lavenza (approved):
relatively low size for a music addition, nice

Peter-Dolby:
I approve of a change in music though this track in particular is a bit... I dunno, tooty. High-key. You also don't want too much of an Intro given how it's something you're flicking on and off frequently.

roarbark:
> I approve of a change in music though this track in particular is a bit... I dunno, tooty. High-key. You also don't want too much of an Intro given how it's something you're flicking on and off frequently.

- the intro is about 4 seconds... this isn't much more than the combat tracks we have right now- in my opinion a short intro actually goes better for the few situations where you do flick it on and off quickly because you're not exploded by the music dropping every time
- i do agree that it is kind of high-key, though not much more than some other tracks we have, like vaquero, guard2 and clergy

i found this track fitting because it takes a classic fantasy tune with brass, flute etc and puts a unique spin on it, making it darker, even if subtly so, which just kind of spoke RT aesthetic to me...

Peter-Dolby:
I do think the vanguard theme is one of the ones that could do with a replacement, just not sure about this replacement in particular.

I also feel like the MAA/Knight combat theme could do with a replacement (sounds very out-of-place) and the dwarven quarter/guild smith soundtrack is in dire need of replacement, that one is really bad.

PR #2941, opened 2025-06-01
allows pickpocketing from sleeping and closed eye humans

ARefrigerator:
## About The Pull Request
adds two or statements to pickpocket checks to see if the other person has their eyes closed or is asleep. I don't think this caused any issues with seeing a message with the eyes closed because if it's a visible message neither of these situations should tip you off as far as I know.

## Why It's Good For The Game
Makes sense. They check if they can see you. They would not be able to see you in either of these situations.

https://github.com/user-attachments/assets/4ede9e6b-8f9f-431d-8059-1a072c2302e5


ARefrigerator:
okay I made it so without a client you get no experience. Tested it locally and it seemed to work.

RadiantFlash7 (approved):
Approved

PR #2944, opened 2025-06-03
Serial Killer: Maniac reimagined

Gyrandola:
## About The Pull Request

This PR aims to soft rework the disabled maniac into a non round-ending, non fourth-wall-breaking and truly minor Town of Salem inspired antag.

The Serial Killer, randomly selected among the towners, will feel the urge to murder at least one player per night, or risk succumbing to his own visions and commiting suicide.

## Gameplay

- Serial Killer rolls with a 15% probability as a Minor Antag.
- SK can craft totems (they are easier to craft, there's only one type but with varying icons) requirements: 3 bodyparts, 1 head.
- Totems are now destructible, their effect on bystanders is temporary (5 minutes).
- SK can still retrieve a knife and a surgical kit from a clock or a tree. The kit is no longer strictly necessary but I left it for flavor, in case the SK wishes to roleplay a certain murder pattern or whatever.
- SK starts with a +2 skill in knives (ended up giving this minor buff so any towner can at least properly wield a knife).
- SK is given a 5 minutes prep time on round start, after then a 25 minutes timer (roughly 1 dae-night cycle) starts ticking. The SK is forced to commit at least one murder before the timer ends, or receive an absolutely lethal heart attack.
- SK triumphs if he manages to survive the week (implying he got away with many murders).

I am willing to modify the killing timer if it ends up being too strict, but the idea is that it forces the SK to leave a trail of bodies wherever they go.

~~PS: I gave up trying to make it so the old maniac hallucinations temporarily stop after each murder, to give a feeling of relief of sorts. I believe there's something fundamentally wrong with the hallucination code that keeps them from being removable from memory: once they start, you'll never get rid of them.~~

### DIFF
--- a/code/modules/antagonists/roguetown/villain/serial killer/objectives.dm
+++ b/code/modules/antagonists/roguetown/villain/serial killer/objectives.dm
@@ -1,6 +1,6 @@
 /datum/objective/serial_killer
 	name = "KEEP KILLING"
-	explanation_text = "The visions haunt me... I know that if I do not kill at least one person every twenty-four hours, I will be the one to die."
+	explanation_text = "The visions haunt me... I know that if I do not kill at least one person every 2 days, I will be the one to die."
 	flavor = "The overwhelming need to kill."
 
 /datum/objective/serial_killer/check_completion()


Gyrandola:
> With the visions being temporary and round-ending, waking up shtick removed, I could see this being brought back.
> 
> One thing I /would/ suggest, is giving the serial kill knife skill with the 'up_to' proc of something like 4 or so. Maybe force-set their patron to graggar, so they can do graggar rituals? Would be a neat inclusion that grants them small stat buffs for killing targets, rather then being an unstoppable monster from night 1.

Could do that, but even if I didnt' they wouldnt be an unstoppable monster, the antag datum wont give any extra stats or skills.

Bapldap:
Opinions on totems giving the maniac a buff while on screen? Might add to encouragement for luring people into fucked up murder rooms full of viscera statues.

RadiantFlash7:
> > With the visions being temporary and round-ending, waking up shtick removed, I could see this being brought back.
> > One thing I /would/ suggest, is giving the serial kill knife skill with the 'up_to' proc of something like 4 or so. Maybe force-set their patron to graggar, so they can do graggar rituals? Would be a neat inclusion that grants them small stat buffs for killing targets, rather then being an unstoppable monster from night 1.
> 
> Could do that, but even if I didnt' they wouldnt be an unstoppable monster, the antag datum wont give any extra stats or skills. It will probably be a pretty hard antag where stealth and gameknowledge matter.

Agreed.

Kregle:
Love this. What if we let the Inquisitor examine the totems for clues regarding the SK's identity?

Red-Honk:
I do hope you add a way for the code to state, yeah, they killed them, instead of, they died due to oxygen loss and you didn't deal the final blow, would be a tad boring, maybe check the amount of damage they dealt?; Or give it a, they must witness them murder happen type of deal.

I think the timer should be 30 to 35 minutes, to give them at least some time to emote things out, instead of just being forced to kill something for the sake of killing it, because there just simply isn't any leeway when it comes to time.

Also, ensure that the totems don't give the Zizo effect, it simply sucks, it's bad for the health of the player behind it, filter only or not.

Gyrandola:
> I do hope you add a way for the code to state, yeah, they killed them, instead of, they died due to oxygen loss and you didn't deal the final blow, would be a tad boring, maybe check the amount of damage they dealt?; Or give it a, they must witness them murder happen type of deal.

I think thats tricky. lastattacker is the closest thing we have to determine the killer, and in a way it also encourages the SK not team up with others, and as we know most serial killers are lone wolves.

> I think the timer should be 30 to 35 minutes, to give them at least some time to emote things out, instead of just being forced to kill something for the sake of killing it, because there just simply isn't any leeway when it comes to time.

I'd like to test it with 25 minutes first but yeah I'm open to increasing it

> Also, ensure that the totems don't give the Zizo effect, it simply sucks, it's bad for the health of the player behind it, filter only or not.

So right now totems only give visions and they last for 5 minutes, they are also destructible. The visions help the SK blend in, because if a bunch of people are being affected by them, theres no way to tell who is the actual murderer.
If, as suggested above, the totems are changed to serve some other purpose, I can see this mechanic being removed or modified.

code/game/gamemodes/roguetown/roguetown.dm
```diff
@@ -369,9 +369,35 @@ var/global/list/roguegamemodes = list("Rebellion", "Vampires and Werewolves", "N
...
 	restricted_jobs = list()
 
-/datum/game_mode/chaosmode/proc/pick_maniac()
-	restricted_jobs = list("Duke", "Duke Consort")
-	antag_candidates = get_players_for_role(ROLE_MANIAC)
+/datum/game_mode/chaosmode/proc/pick_serial_killer()
+	restricted_jobs = list(
+	"Duke",
+	"Duke Consort",
+	"Dungeoneer",
```
RadiantFlash7, line 376: Dungeoneer being a serial killer might be fine.

Gyrandola: yeah

RadiantFlash7:
Feedback has shown that this still isn't quite right. The timer kind of hurts roleplay/scenes players want to create, and the +2 knife skill isn't enough to make a decent impacts.

Perhaps remove the timer before suicide, and ensuring the serial killer gets a level_up_to expert in knives, and a speed bonus?

Gyrandola:
I'll make the timer non-lethal, as in the only thing it will do is make your visions temporarily go away if you have murdered someone . And I'll give those buffs. Sounds good?

PR #2945, opened 2025-06-03
Verminfolk Underdweller & Shaman Subclass Nerf

Vizoplez:
## About The Pull Request

# TO MAKE IT CLEARER, THIS HAS EXPANDED BEYOND ITS ORIGINAL TESTMERGE. STILL NEEDS MORE TESTING AND FEEDBACK.

Nerfs Shaman AND underdweller

Slots -> 10 -> 4 YOU ARE NOT AN ANTAG FACTION OR A MAJOR ANTAG
Shaman -> seperated by a subpick now between MIRACLES and MAGIC, still keeps other skills
Shaman -> -1 STR changed to -2

## Why It's Good For The Game
These guys have proven problematic time and time again. Time to cull them down.

A non-major faction does not need 10 slots, especially when they get roles with very decent overall skills and utility (including wild eater which lets them basically survive off of anything forever), atop of yknow, the choice to have a nonaligned churchling (miracle spam go brrr) with no obligation to the church, and a potentially very solid mage start (apprentice book + decent int/end) with a scavenged book? Yeah.

On average it used to take these guys less than an hour 20 to get meteor storm. They were not fine as is, and were split into seperate roles and nerfed.
also reading EXP is way too easy to get so 'shit reading shit int' isn't a good excuse for how they were

### DIFF
--- a/code/modules/jobs/job_types/roguetown/refugees/types/underdwellers.dm
+++ b/code/modules/jobs/job_types/roguetown/refugees/types/underdwellers.dm
@@ -118,7 +118,6 @@
 			head = /obj/item/clothing/head/roguetown/helmet/leather/volfhelm
 			shoes = /obj/item/clothing/shoes/roguetown/shortboots
 			cloak = /obj/item/clothing/cloak/tribal
-			backpack_contents = list(/obj/item/rogueweapon/surgery/hemostat/improv = 2, /obj/item/rogueweapon/surgery/retractor/improv = 1, /obj/item/rogueweapon/surgery/saw/improv = 1, /obj/item/book/granter/spellbook/apprentice = 1)
 			H.change_stat("strength", -2)
 			H.change_stat("intelligence", 2)
 			H.change_stat("endurance", 1) //So they can use their miracles and spells more often in exchange for their lost speed.
@@ -130,11 +129,13 @@
 					var/datum/devotion/C = new /datum/devotion(H, H.patron)
 					C.grant_spells_churchling(H)
 					H.verbs += list(/mob/living/carbon/human/proc/devotionreport, /mob/living/carbon/human/proc/clericpray)
+					backpack_contents = list(/obj/item/rogueweapon/surgery/hemostat/improv = 2, /obj/item/rogueweapon/surgery/retractor/improv = 1, /obj/item/rogueweapon/surgery/saw/improv = 1)
 				if("Magos")
 					H.mind.adjust_skillrank(/datum/skill/magic/arcane, 1, TRUE)
 					H.mind.AddSpell(new SPELL_PRESTIDIGITATION)
 					H.mind.adjust_spellpoints(2)
 					H.mind.AddSpell(new SPELL_LEARNSPELL_WEAK)
 					H.mind.adjust_skillrank(/datum/skill/misc/reading, 1, TRUE)
+					backpack_contents = list(/obj/item/rogueweapon/surgery/hemostat/improv = 2, /obj/item/rogueweapon/surgery/retractor/improv = 1, /obj/item/rogueweapon/surgery/saw/improv = 1, /obj/item/book/granter/spellbook/apprentice = 1)
 			ADD_TRAIT(H, TRAIT_WILD_EATER, TRAIT_GENERIC)
 	H.set_blindness(0)


Rayforx:
Nah, giving a role 1 spell point at round start, and then removing the speed.
Dunno who you are making this PR for. But saying that because a role "can get better gear" means that we should nerf them, yeah, I don't know about that. Having a role that can heal and use spells is strong, but this is a very strange way to balance shit.

They used to only be able to get 1 spell point per night, with their current book. And what comes down to this PR being "they can be strong, if they are played under certain criteria," which can be said about every mage.

code/modules/jobs/job_types/roguetown/refugees/types/underdwellers.dm
```diff
@@ -126,12 +126,11 @@
 			cloak = /obj/item/clothing/cloak/tribal
 			backpack_contents = list(/obj/item/rogueweapon/surgery/hemostat/improv = 2, /obj/item/rogueweapon/surgery/retractor/improv = 1, /obj/item/rogueweapon/surgery/saw/improv = 1, /obj/item/book/granter/spellbook/apprentice = 1)
 			H.change_stat("strength", -1)
-			H.change_stat("speed", 1)
 			H.change_stat("intelligence", 2)
 			ADD_TRAIT(H, TRAIT_WILD_EATER, TRAIT_GENERIC)
 			H.mind.AddSpell(new SPELL_PRESTIDIGITATION)
-			H.mind.adjust_spellpoints(1)
-			H.mind.AddSpell(new SPELL_LEARNSPELL)
+			H.mind.adjust_spellpoints(-1) //You start with a malus.
```
RadiantFlash7, line 132: last tweak is to remove this adjust spellpoint part. Not really needed per say.

Skyglazer:
I've spoken about this role in the past-- I still think it's absolutely silly that this race that has this extremely unique thing of both spellcasting and miracles in one role is just tied to some random-ass small-people-only refugee role hidden in a subclass selection. We should be giving cool roles with powers like that to the town instead...

mkr3492:
This REEKS of spite. The way that this PR is written should have resulted in it thrown out immediately. Can you at least PRETEND to be an impartial voice that just wants to balance things instead of someone that has been clearly personally beaten by someone who was playing as this role? It's fucking pathetic.

Alanii:
Why is there a role that has both magic and divine spells?

This shouldn't be a thing.

RadiantFlash7:
After re-thinking this a fair bit. It's come to my attention theres some uinherent problems with the role.
For one, it's unlimited, where as refugee mages is very much a limited number, you can have as many cave dwelling shaman mage hybrids as you want.
For another, It's not immensely difficult for a shamman to get ahold of a strong book. That, combined with healing magic is fairly rough.

I personally have never been a major fan of hybrid casters of both divine and arcane froma  lore perspective- Though it might be a unique special at some point- I don't really know if we should keep that as a thing here.

AukariBunn:
> I've spoken about this role in the past-- I still think it's absolutely silly that this race that has this extremely unique thing of both spellcasting and miracles in one role is just tied to some random-ass small-people-only refugee role hidden in a subclass selection. We should be giving cool roles with powers like that to the town instead...

How dare you give a cool role to the races with the least content in the game and not to my self-insert hooman character
image

ddsaw:
> After re-thinking this a fair bit. It's come to my attention theres some uinherent problems with the role. For one, it's unlimited, where as refugee mages is very much a limited number, you can have as many cave dwelling shaman mage hybrids as you want. For another, It's not immensely difficult for a shamman to get ahold of a strong book. That, combined with healing magic is fairly rough.
> 
> I personally have never been a major fan of hybrid casters of both divine and arcane froma lore perspective- Though it might be a unique special at some point- I don't really know if we should keep that as a thing here.

I've said it before and I will say it again and again: the inherent problem is how much of a mages power is tied to the quality of their book vs their reading/arcane/int. You can't really have a shitty fucking caster either because the difference is so small between the different levels of arcane.

divifest:
New moon was tabaxi only, and then it got opened to like everyone

Forlorn hope was the same way, but for lupuan.

Having a super awesome powerful role that is made only available to the coder's slop-tier OC is dumb.

divifest:
> > New moon was tabaxi only, and then it got opened to like everyone
> > Forlorn hope was the same way, but for lupian.
> > Having a super awesome powerful role that is made only available to the coder's slop-tier OC is dumb. And it's a refugee subclass. Lol
> 
> Ah yes, famous races with almost no content in the game and with almost all roles restricted LUPIAN AND TABAXI

Yeah but we're talking about verminfolk here. That was an example of precedent in regards to the whole 'race specific roles'.

Subclasses aren't content, never have been. It's just super special snowflake bullshit that isn't required.

divifest:
> > Yeah but we're talking about verminfolk here. That was an example of precedent in regards to the whole 'race specific roles'.
> > Subclasses aren't content, never have been. It's just super special snowflake bullshit that isn't required.
> 
> Alright bro When are you allowing nobles and the priest for verminfolk and kobolds?

Imagine not being able to tell the difference between making a subclass exclusive for one race, and disallowing shunned races from positions of importance in a prejudiced setting. 

Holy

Rayforx:
With the current changes in:

The closest comparison would be the battle magos.
The battle magos get 2 endurance, 1 cont, 1 speed (with no negative stats)
The under dweller gets -2 str and 1 endurance (that is a 5 stats difference)

They get the same book and the same arcane.
The underdweller gets one less weapon stat.

In terms of items, it takes 2 minutes to make them, as they are makeshift ones... It's fiber and a stick it's barely anything worth anything.

The argument that they can get a better book is simply dumb, that is exactly the same for all magic roles, lol. We need to nerf every magic role, if that is the argument we are using.

The underdweller also gets fewer max spells, and no "medium armor training." 

The sharman also gets NO pouch, armor, or weapon. This means that they loose about 60 mammon worth of gear from the "closest" comparable role. 

> They should be slightly better at magic but get no bonus SPD as this made them really annoying in practice to catch. Also nerfs them down to 3 spells as if these little shits get ahold of an expert book somehow it could cause them to be able to do a little trolling

You changed the **arcane back to 1, so saying that they are "better" at magic is false. ** Fix this statement, instead of using it as a reason to nerf shit.
I am not even going to pretend that this makes sense. But to put it down, I will sum up all of this statement in terms of general balance.
The underdweller gets worse str, endurance, and speed now. Could you explain why you're removing the speed?
The whole 3 spells make no sense, as every magic role can get hold of an expert book. There is nothing special about the sharman that increases their chance of getting it.

> also reading EXP is way too easy to get so 'shit reading shit int' isn't a good excuse for how they were
Again, this is the same with every mage role, do not use this as a reason to nerf.

Peter-Dolby:
The base water/food gameplay cycle is not that hard.
I don't really see the shaman being that much more or less likely to join forces with bandits or more or less dangerous if they do join forces with bandits than any other refugee class.

PR #2947, opened 2025-06-03
Players can no longer tackle/wrestle armed players, requiring disarming.

RadiantFlash7:
## About The Pull Request
Edit 6/13: It is now impossible to tackle someone if they have both combat mode and either carry a shield or a normal/large/great sized weapon in hand, the only exception to this rule, is if it's done from behind them.

## Why It's Good For The Game
One of the major gripes people have with wrestling is that wrestling is massively binary. You either have enough to make it trivial to break out of, or not enough to get out of. And once someone grabs you, It's effectively ended the fight. 
As part of my quest to make it so fights don't end so abruptly (Vampire crit weakness removed, mobs taking 2 or so hits before you can crit someone), This should help matters considerably. It's still feasible to tackle players- But now you need to disarm them first. A feature that existed, but was rarely used because why bother when you can just tackle someone instead?

image
image
image

### DIFF
--- a/code/modules/mob/living/grabbing.dm
+++ b/code/modules/mob/living/grabbing.dm
@@ -209,7 +209,7 @@
 				else
 					I = M.get_inactive_held_item()
 			user.stamina_add(rand(3,8))
-			var/probby = clamp((((18 + (((user.STASTR - M.STASTR)/4) + skill_diff)) * 10) * combat_modifier), 5, 95)
+			var/probby = clamp((((4.5 + (((user.STASTR - M.STASTR)/4) + skill_diff)) * 10) * combat_modifier), 5, 95)
 			if(I)
 				if(M.mind)
 					if(I.associated_skill)


anconfuzedrock:
I will note an answer to the question of "why disarm?" if you grab someone's weapon hand, they can't stab you. disarming isn't useless as it is now- as it stops them attacking while you're performing the disarm.

SMoto95:
So my view on this;

Instead of giving armed people total immunity just because they hold something in their hands is kind of funny and downright unrealistic, especially if they are immune from being pushed down/tackled from the side where nothing is being held /or/ from behind.

The type of weapon used should also be taken into account, a cleaver should not magically give you footing because you hold a weapon, parrying grabs exist if you're skilled to use them for a reason and you can just slice someone who failed the attempt in the hand(s) with the cleaver to make them hurt enough from trying again, it's better if the weapons used are realistically long enough to help you stand like polearms or something that can block access to your chest like a shield or longsword/polearms/etc being held with two hands instead of one.

The idea itself is not bad, but it feels too powerful at the moment and needs some extra balancing.

Gyrandola:
> So my view on this;
> 
> Instead of giving armed people total immunity just because they hold something in their hands is kind of funny and downright unrealistic, especially if they are immune from being pushed down/tackled from the side where nothing is being held /or/ from behind.
> 
> The type of weapon used should also be taken into account, a cleaver should not magically give you footing just because you hold it, parrying grabs exist if you're skilled to use them for a reason and you can just slice someone who failed the attempt in the hand(s) with the cleaver to make them hurt enough from trying again, it's better if the weapons used are realistically long enough to help you stand like polearms or something that can block access to your chest like a shield or longsword/polearms/etc being held with two hands instead of one.
> 
> The idea itself is not bad, but it feels too powerful at the moment and needs some extra balancing.

I think we can incorporate both your ideas honestly.

Maybe:
1) Tackling someone from the sides or behind bypasses the need to disarm them
2) Tackling frontally requires you to disarm the opponent, but there needs to be some level of differentiation between an opponent wielding something like a knife and one with a spear or a halberd.

RadiantFlash7:
> > So my view on this;
> > Instead of giving armed people total immunity just because they hold something in their hands is kind of funny and downright unrealistic, especially if they are immune from being pushed down/tackled from the side where nothing is being held /or/ from behind.
> > The type of weapon used should also be taken into account, a cleaver should not magically give you footing just because you hold it, parrying grabs exist if you're skilled to use them for a reason and you can just slice someone who failed the attempt in the hand(s) with the cleaver to make them hurt enough from trying again, it's better if the weapons used are realistically long enough to help you stand like polearms or something that can block access to your chest like a shield or longsword/polearms/etc being held with two hands instead of one.
> > The idea itself is not bad, but it feels too powerful at the moment and needs some extra balancing.
> 
> I think we can incorporate both your ideas honestly.
> 
> Maybe:
> 
> 1. Tackling someone from the sides or behind bypasses the need to disarm them
> 2. Tackling frontally requires you to disarm the opponent, but there needs to be some level of differentiation between an opponent wielding something like a knife and one with a spear or a halberd.

I can see about adding checks for this. Though what weapons should block tackling and what shouldn't will be a bit of a topic, along with coding it cleanly.

john-pawson:
I honestly think that making tacking not automatically drop your weapon and pinning less easy would be a better change than something as specific as this, thing is that tackle does all,

Need an enemy disarmed? Why disarm them, just tackle them, and grab the weapon, now they're on the floor too!
Can't get through an enemy's defences? Why grab their arm and attack with the other or feint, tackle them, and they won't even be able to have a chance to block, beause the thing they were using to block is on the floor!
Need an enemy pinned? Just.. tackle as normal, there really isn't that much thought to it, you won the tackle, you won the pin.

It gives you an advantage for combat AND disarms the enemy AND neutralizes them (since following up with a pin isn't exactly hard due to it being harder to resist due to being prone *and* not needing to make a new grab) this solution just seems a bit.. overcomplicated. And weird.

Red-Honk:
> I honestly think that making tacking not automatically drop your weapon and pinning less easy would be a better change than something as specific as this, thing is that tackle does all,
> 
> Need an enemy disarmed? Why disarm them, just tackle them, and grab the weapon, now they're on the floor too! Can't get through an enemy's defences? Why grab their arm and attack with the other or feint, tackle them, and they won't even be able to have a chance to block, beause the thing they were using to block is on the floor! Need an enemy pinned? Just.. tackle as normal, there really isn't that much thought to it, you won the tackle, you won the pin.
> 
> It gives you an advantage for combat AND disarms the enemy AND neutralizes them (since following up with a pin isn't exactly hard due to it being harder to resist due to being prone _and_ not needing to make a new grab) this solution just seems a bit.. overcomplicated. And weird.

Maybe we should make it so, it no longer drops the weapon from their hands, but adds a tiny stun, forcing a pin to happen quite swiftly if not a counter-attack will come into play?

I think we might want to make tackles possible but, maybe add like auto-retaliate to play, you want to do it, without protection?; Sure, but prepare to be cut or the like if the weapon makes sense.

Perhaps the pinup should have a windup, so, it isn't just, easy to spam while someone else does something else.

Personally at least shields that aren't buckler or maybe even the wooden shield (maybe give it less chance) should be untackable from the front and sides, like the towering/kite shields.

Also, on a random note, I think the touch intent shake should have a windup and not a instant get up, it's a tad annoying, because if it's done to allies they might accidently go prone, if it's done to the enemy it's a tad annoying that oops, I meant to grab or increase my grip, but we managed to accidently get them up due to grabs swapping to touch intent when said grab it's broken out of.

Gyrandola (approved):

RadiantFlash7:
> This ain't it, chief. Many other people have said their piece on it, but it's just not it.

What alternative can you propose?
One of the worst feeling ways about the game is it doesn't matter what shield you have, what weapon or skill. If someone grabs at you twice on swift intent they bypass your parry. If they have either journeyman or expert wrestling and any strength skill- high chance you're just finished. Fights instantly over in an unsatisfying manner.

Something needed to be done, and theres still ways to handle people easier that isn't an immediate end.
Con checking knocks people to the floor, kicking people into structures knocks them over. This PR still allows for grabbing people from behind and tackling without any other checks, and reauires the target to be in combat mode.

code/modules/mob/living/grabbing.dm
```diff
@@ -209,7 +209,7 @@
 				else
 					I = M.get_inactive_held_item()
 			user.stamina_add(rand(3,8))
-			var/probby = clamp((((3 + (((user.STASTR - M.STASTR)/4) + skill_diff)) * 10) * combat_modifier), 5, 95)
+			var/probby = clamp((((18 + (((user.STASTR - M.STASTR)/4) + skill_diff)) * 10) * combat_modifier), 5, 95)
```
john-pawson, line 212: The 3 gets multiplied by 10 at the end, so this is an... 150% increase. Should be changed to 4.5

PR #2949, opened 2025-06-03
Removal of Grenzel Kommandant

RadiantFlash7:
## About The Pull Request
This PR Flat out removes Grenzel Kommandant

## Why It's Good For The Game
Grenzel's are a professional mercenary group yes. But at present, Kommandant is a flavored role with stupid stats and stupid gear. Further more, it's existance actively harms players that don't want to participate in clique-y behavior, and at worst is used by players as a gamer role without giving a damn about their supposed leadership needs.

All of that aside- Individual mercs should never have been given a 'leadership' role, beyond one for a general guild leadership. We do not want to encourage large amounts of mercs to take the same job if possible.

N/A


MickDick1:
I'm ok with this too.

Remove it, or I'll nerf it even harder in my existing nerf PR.

PR #2952, opened 2025-06-04
Refactors Gem information, creates arcane power global list, cleans remarks list.

ARefrigerator:
## About The Pull Request
Refactor to organize gems a bit. They were in different files and had variables assigned in them as well. Also turns arcane power into a list in case it would ever need to be expanded. Changes the logic for reading the list when using it in the spellbook code.

The remarks list chose to fight me so it is no longer a large single line but turned into something easier to read.

## Why It's Good For The Game
Should change nothing on the user end, mostly to clean a few things up.

test logs: 
image

proof gem changes the stored_gem value:
image


RadiantFlash7 (approved):
No issues during test merge, Approved

PR #2953, opened 2025-06-04
No more Pysdonite Duke

MickDick1:
## About The Pull Request

Duke is now restricted from rolling Psydon as a faith.

## Why It's Good For The Game

Psydonite Duke is silly and often sloppy beyond belief. That is the general sentiment, and this was greenlit by Useroth himself.

Worx locally yadda yadda. That said. IDK why but I had to list the patrons piecemeal because combining 2 defines (all divine patron and all inhumen) into a list doesn't work at all. You tell me.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/nobility/ruler.dm
+++ b/code/modules/jobs/job_types/roguetown/nobility/ruler.dm
@@ -13,6 +13,22 @@ GLOBAL_LIST_EMPTY(lord_titles)
 	allowed_races = RACES_TOLERATED_UP
 	allowed_sexes = list(MALE, FEMALE)
 	can_leave_round = FALSE
+	allowed_patrons = list(\
+	/datum/patron/divine/astrata,\
+	/datum/patron/divine/noc,\
+	/datum/patron/divine/dendor,\
+	/datum/patron/divine/abyssor,\
+	/datum/patron/divine/ravox,\
+	/datum/patron/divine/necra,\
+	/datum/patron/divine/xylix,\
+	/datum/patron/divine/pestra,\
+	/datum/patron/divine/malum,\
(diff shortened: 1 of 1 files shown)


RadiantFlash7:
Contrary to Gyran's point- This is actually a lore issue. It may be a salt PR as well, but it's one that I've mentioned a few times in the staff channels-
Pysdonions are tolerated heretics. Tolerated, But still heretics. The church is not going to crown a heretics, even if they're tolerated. 
Now; The reason we ALLOW inhuman heretic dukes is because- they're meant to keep that shit /quiet/ Or else the church would over throw them. They know that they're not meant to be public about it, and conform to that.

However- Theres... no reason for pysdonians to do that. They're tolerated, why bother hiding that you're a psydonian? Yet it doesn't make sense for the church to crown a psydonian- It's a conundrum. This reason alone is why I think it's better to prevent duke's from being psydonian.

Alanii:
If you're going to remove Psydon Duke, then remove the inhumen duke.

This is just weird.

Peter-Dolby:
If you're removing Psydon Duke you might as well remove Psydonism

MickDick1:
> If you're going to remove Psydon Duke, then remove the inhumen duke.
> 
> This is just weird.

I asked Useroth if I could also remove inhumen from duke but he said we're doing Psydon first for now.

Kregle:
The Duke is an antagonist. They should have the opportunity to be chaotic and rival the Church if that's what they want to do within the bounds of the rules. If you have issues with how Psydonism itself reacts with the Church- I'm working with the maintainers to change Psydonism to be more syncretic, a schism rather than a heresy; in that they believe Psydon has left us until we kill Zizo, and He created the Ten to guide us because He's nice, essentially.

PR #2956, opened 2025-06-05
Adds Goblins and Seelies to Zybantine Slave Roster

MickDick1:
## About The Pull Request

Adds Goblins and Seelies to zybantine slave role racial pool.

## Why It's Good For The Game

I think having a seelie slave would actually be pretty fun. The master even starts with a lamptern to stuff them into.
Goblin slaves are also funny. Consider it hard mode, potentially. 

Yes.


TheFenix323:
Not sure if intended but the Seelie wouldn't start with spells

PR #2964, opened 2025-06-06
Adds post_craft proc, offsets railing on craft

ARefrigerator:
## About The Pull Request
I originally wanted to make railings offset on craft to feel more like they're in the proper place. It seems like the easiest way was to just set a post_craft when they're made and call that. If there's a better way let me know but this could be useful for other objects that need to do something when they are crafted versus just initialized.

I might also consider making the south view be a rotated version of the north sprite but left it unchanged for now for compatibility with currently existing fences.

## Why It's Good For The Game
Almost every railing on the map is designed with a pixel offset in mind. When you craft them in game however you do not have this offset. I think it looks more aesthetically pleasing to have the offset even in crafted versions.

3 railings on 1 tile before:
image
3 railings on 1 tile after:
image

### DIFF
--- a/code/__HELPERS/crafting.dm
+++ b/code/__HELPERS/crafting.dm
@@ -206,15 +206,18 @@ proc/construct_item(mob/user, datum/crafting_recipe/R)
 							var/atom/movable/I = new IT(T)
 							I.CheckParts(parts, R)
 							I.OnCrafted(user.dir, user)
+							I.post_craft()
 					else
 						if(ispath(R.result, /turf))
 							var/turf/X = T.PlaceOnTop(R.result)
 							if(X)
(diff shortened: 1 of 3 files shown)


Red-Honk:
Problem with this is, would this from the tile away being made on the same tile or on the standing tile being made outwards?

Because this feels like you are just simply reverting how they are made when such an effect could be done differently, only this time this makes you wonder where they were build to begin with.

Red-Honk:
Tried it, don't like the look of them, don't like their positions and don't like the fact that it makes it look like your character is standing on them if one is north of them.

ARefrigerator:
Ah, probably best to remove it then. I know z-ordering in this game is a bit finicky.

PR #2968, opened 2025-06-08
Removes Inhumen/Psydonite/Zizoid Novitiate

Skyglazer:
## About The Pull Request

This PR is a very simple change: It prevents Novitiates from choosing any patron other than the Divine Pantheon ones. 

## Why It's Good For The Game

As it stands right now, Novitiate is a role that's a bit of an awkward leftover from old Roguetown's young roles, pretty much obsoleted by Acolyte. Still, it has a place as a learning role and something of a church-assistant, although the church is never so crowded that you wouldn't just be able to play Acolyte instead.

Still, the way it exists currently has a single, crucial flaw: there is no patron restriction on Novitiate. Whether unintentional or intended, I believe this to be a flaw in the role. It leads to the role being used not as a learning or helper role, but as a saboteur role. People only play as novitiate with the intention of being an inhumen or Zizo-worshipper that can sabotage the church from the inside. It's a completely baffling thing to have exist, and it's the only church role that isn't restricted in terms of patron. 

This simple change would make it so that these saboteurs actually have to work to get their way into a trusted position from which they can sabotage the church-- rather than being able to do so from roundstart. It makes sense that the church would be a bit strict on who exactly gets to wear the cloth, especially considering they all command miracles.

It also keeps a bit more consistency in the lore. Right now, Psydonites can be novitaites. Even though lorewise, Psydon-worshippers are not meant to have any sort of divine power, due to the way the role is coded, Psydonite novitiates still get a functional miracle. With this change, Psydonite novitiates will no longer be possible.

https://www.youtube.com/watch?v=ICol28zaNv4

### DIFF
--- a/code/modules/jobs/job_types/roguetown/youngfolk/novitiate.dm
+++ b/code/modules/jobs/job_types/roguetown/youngfolk/novitiate.dm
@@ -7,6 +7,7 @@
 	spawn_positions = 2
 
 	allowed_races = RACES_SHUNNED_UP
+	allowed_patrons = ALL_ACOLYTE_PATRONS
 	allowed_sexes = list(MALE, FEMALE)
 	allowed_ages = list(AGE_ADULT)
 


GravyMaker:
I'm interested to learn whether allowing heretic Novitiates was a deliberate decision or an oversight. Given that Psydonite novitiates are possible - when Psydonites typically shouldn't have miracles due to having no living god to channel power from - I suspect it may have been more of the latter than the former.

As it stands, I think I support this, if only to keep Novitiates in-line with the rest of the church roles.

mkr3492:
Heretical novitiates were always an intentional option. How do I know this? BECAUSE THEY HAVE UNIQUE MIRACLE / HEALING LINES CODED FOR THEM!!! Stop removing content for no good reason! Stop it! Stop meddling, we don't have a PR quota to meet!!

PR #2969, opened 2025-06-08
Fix spider AI issues (and maybe summons)

noelle-lavenza:
reverts the spider speed to 2.5 tiles per second instead of 5 tiles per second
fixes spiders having a vision range of 9 instead of 5
attempts to fix summon AI failing to target properly

spiders work fine now

https://github.com/user-attachments/assets/80ed8c1c-3ec8-4b23-8609-12fa013e05b9


Gyrandola:
Makes sense. Un-TMING

PR #2974, opened 2025-06-08
Alternative Shaman Nerf

envoyin:
## About The Pull Request

When the Caveling Shaman was buffed, the idea was produced that the Shaman stay buffed but lose access to most combat spells. I have also opened them up to all shunned races, minus Half-Orcs.

They have lost:

- Greater Fireball
- Meteor Storm
- Sundering Lightning
- Fireball
- Lightning Bolt
- Spitifre
- Lightning Lure
- Arcane Bolt
- Frost Bolt

They have kept Find Familiar and Ensnare as neither of them are nearly as fucked up as the other, and they would make sense for a Shaman to know (especially Find Familar).

Someone tell me if they should lose the Zizoid spells.

## Why It's Good For The Game

I am against nerfing unique classes in a terrible way that provide some interesting refugee flavor.

image
image

### DIFF
--- a/code/modules/spells/roguetown/wizard.dm
+++ b/code/modules/spells/roguetown/wizard.dm
@@ -139,8 +139,6 @@ Unless of course, they went heavy into the gameplay loop, and got a better book.
 	var/list/choices = list()//Current thought: standard combat spells 3 spell points. utility/buff spells 2 points, minor spells 1 point
 
 	var/list/spell_choices = list(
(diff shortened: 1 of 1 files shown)


ddsaw:
Frostbolt is very good, and arcane bolt is outright broken. Arcane bolt will break almost all armor in a single hit, and from then on every single hit is like being shot with a crossbow.

PR #2975, opened 2025-06-08
Remove skeleton genitalia

Gyrandola:
## About The Pull Request

Removes all genital organs from NPC & Lich-raised skeletons.
Also removes some unused, unticked duplicate deadite code I found that was ported by accident months ago.

## Why It's Good For The Game

They are skeletons, they shouldn't have those organs. 
In fact they should not have any organs at all, but that would make them immortal due to how stuff is coded.

Tested on NPC skeletons, not on lich minions.


Alanii:
Find out if Goblin NPCs have genitalia, and then this PR should be okay. 

I think only issue people have is they want skeleton genitals for genital muta-swap surgery.

EmilyFontaine:
Not the PR Ratwood needs, but the PR Ratwood deserves

divifest:
> Not sure why this PR is getting so many downvotes. It's a basic bug fix

If it was a basic bug fix, you'd find a way to remove all the organs without making skeletons invincible

PR #2976, opened 2025-06-08
Standardizes reagent container logic (adds features too plz read gyran!!)

ARefrigerator:
## About The Pull Request
works reagent containers to have common use interactions, sets fill define, fixes pot insertion so there's no more special cases. or hopes to...

I might still need to do more work on this but it's here as a proof of concept until I think it's done.

edit:

- Standardizes all reagent container intent order to be consistent.
- adds `INTENT_FILL` define
- Removes the "this person cannot be milked" message, unneeded when someone is standing there clothed.
- removes barrel update_icon (depreciated)
- prevents torches and their children types from being put into pots.
- allows pouring of reagent container's contents onto food (right click) (video below)
- allows precise pouring of 1 action via right click

Creates two main procs for reagent containers to prevent repeating code: `can_reagent_interact` and `do_reagent_interact`

`can_reagent_interact` is all the checks to see if it's possible to transfer while  `do_reagent_interact` is the functional aspect of the transfers.

## Why It's Good For The Game
some interactions around reagent containers can be iffy and inconsistent. This will try to unify it in as many was as possible.

https://github.com/user-attachments/assets/1cfd65bd-6ea2-424f-b407-02688d7c2012

### DIFF
--- a/code/modules/reagents/reagent_containers/glass.dm
+++ b/code/modules/reagents/reagent_containers/glass.dm
@@ -111,6 +111,7 @@
 
 	if(!spillable)
 		return
+
 	if(can_reagent_interact(target, user, user.used_intent.type))
 		do_reagent_interact(target, user, user.used_intent.type)
 		return
@@ -369,8 +370,6 @@
 	icon = 'icons/roguetown/items/surgery.dmi'
 	icon_state = "mortar_empty"
 	volume = 100
-	reagent_flags = OPENCONTAINER
-	spillable = FALSE
 	var/obj/item/grinded
 	var/grinding_started = FALSE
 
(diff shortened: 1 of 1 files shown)


ARefrigerator:
Fixed mortars to work the way you imagine they should. Can now pour and fill witha cup into out of them and vice versa.
image

code/modules/reagents/reagent_containers.dm
```diff
@@ -134,6 +134,120 @@
...
+			failed_msg = "[I] is full."
+			state = FALSE
+
+		//These checks come later to make failed message order work properly
+		if(!I.is_refillable())
+			failed_msg = "I can't fill \the [I]."
+			state = FALSE
+		if(!is_open_container())
+			failed_msg = "\The [src] isn't open."
+			state = FALSE
```
Gyrandola, line 189: Why don't you invert the order of the checks and add early returns?

ARefrigerator: I was getting conflicting logic with how things should work and the message it returned when the states were hard returned on the lines. It was also breaking pouring liquids onto foods.

PR #2772, opened 2025-04-26
Buffed Acolyte, Hospitaller, Druid, & Priest starting stats and skills

Skyglazer:
## About The Pull Request

This pull request modifies the starting stats and skills for Acolyte, Mortician, Priest, Hospitaller, and Mortician. Each role has been buffed, with Priest being more buffed than the rest. Acolytes were given crafting 1 to allow them to make crosses as soon as they spawn, and generally stats have been adjusted to be more suited to the church's role in the setting- as a militant organization that has been making a stand against Zizo's rot, and basically the last thing holding society together.

## Why It's Good For The Game

The Church, in-setting, is meant to be extremely important. The Ten are the state religion- and it is the foundation on which society is built. A Duke is coronated by Astratan priests, and the general followings of the Ten are what lawful people are expected to profess. It is an establishment not like the ones we know in our society today, of smiling faces that haven't really known much hardship. In Ratwood, the church is waging a constant war of attrition against the forces of Zizo and the Inhumen- and its attendants reflect this in their training and abilities- not just being healers, but being practiced martial artists. They hone their bodies to be able to protect and serve their church and its flock, and they hone their minds to absorb religious scripture.

Mechanically, it is my opinion that in the current state of the game, it is poorly represented. Church roles all get a VERY modest spread of stats, with their only real 'superpower' being their quite good unarmed/wrestling capabilities, and the healing and other utility they get from miracles. Of course, Templars are there to serve as the true powerhouses of the church, but it so often goes that ONLY the Templars will be able to engage in combat, because the others simply cannot compare. As well, their lacking stat spreads leads to a somewhat muddled identity. 

I would point to roles like Magician's Apprentice/Court Magos, Dungeoneer, and sort of thief/rogue role (refugee or bandit) and Knight for examples of jobs that have a good sense of stat identity. Many magos classes often have the physical stats neglected in favor of highly increased intelligence. This makes sense, for bookworms that spend their time entirely engaged in study. Meanwhile, dungeoneer is stupid-- yet bulky, and strong, reflected in their reduced intelligence and higher STR/CON/END. Knights are tanks-- high strength, and can take a lot of punishment, but their armor slows them down, reducing their speed. Rogues are speedy yet squishy, and focused on striking weakpoints- represented by their lower strength/constitution, but higher speed/perception, so on.

[...]

PR #2979, opened 2025-06-09
Acolyte nerf

divifest:
## About The Pull Request

VERY simple.

Pulls wrestling and unarmed down to 3 and 2 respectively. This leaves them stronger than your average towner, but doesn't keep them as the wrestling gods they are.

Templars do nothing most of the round anyways save for a lich or vampire attack, they should also be protecting the acolytes.

Stop wandering off into the bog alone as acolytes. The Church doesn't train acolytes to be combat experts.

## Why It's Good For The Game

Balance

Stat change lol

### DIFF
--- a/code/modules/jobs/job_types/roguetown/church/monk.dm
+++ b/code/modules/jobs/job_types/roguetown/church/monk.dm
@@ -92,8 +92,8 @@
 			shoes = /obj/item/clothing/shoes/roguetown/sandals
 			armor = /obj/item/clothing/suit/roguetown/shirt/robe/astrata
 	if(H.mind)
-		H.mind.adjust_skillrank(/datum/skill/combat/wrestling, 5, TRUE)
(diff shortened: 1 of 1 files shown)


divifest:
> current acolyte combat strength is very well explained in #2772
> 
> from personal experience, acolytes fold the moment they are outnumbered, which is perfect. they can make people regret blaspheming infront of them and are able to eject them from the church, which is something sorely needed in the setting, but fold to actual antags, that always work in groups, or outwrestle them anyways
> 
> outside of that, wrestling is being worked on to be nerfed right about now on the github, this seems like the wrong time to adjust somethinf like this.

8 paragraphs of cope and slop.

Acolytes don't need the highest wrestling stat bar hedge knights. They have access to free miracles that do crazy things.

RadiantFlash7:
I'm going to deny this for now. Acolytes are meant to be wrestling fiends, and we're presently working on nerfing/tweaking wrestling to be a bit more in depth

PR #2980, opened 2025-06-09
4 athletics to werewolves

DelineFortune:
## About The Pull Request

Too fast stamina drain, you cannot run away because it nukes your stamina too fast
You cannot swim


Red-Honk:
This is simply just 20 extra stamina points, 4 parries extra.

RadiantFlash7 (approved):
Test merge complete without issue, Approved.

PR #2987, opened 2025-06-11
Lets Mercenaries be Werewolves

Dedwolf:
## About The Pull Request

1 line code change to allow mercenaries to be werewolves.

## Why It's Good For The Game

Mercenaries are foreigners who've travelled to rockhill, you shouldn't trust them, they could be a were! 

Yeah works

### DIFF
--- a/code/game/gamemodes/roguetown/roguetown.dm
+++ b/code/game/gamemodes/roguetown/roguetown.dm
@@ -555,7 +555,6 @@ var/global/list/roguegamemodes = list("Rebellion", "Vampires and Werewolves", "N
 	"Warden",
 	"Knight",
 	"Mortician",
-	"Mercenary",
 	"Bandit",
 	"Goblin Chief",
 	"Goblin Cook",


Gyrandola:
Balance aside, I think this change is a bit hard to justify. Mercenaries like the grenzels are likely to travel, live and work as part of organized companies, how could they possibly conceal their lycanthropy?

I don't have a hard opinion on this PR, but it doesn't feel completely right to me.

roarbark:
instead of this, a better way to do this would make antags roll before your job and then put you in a lower priority job that can roll antags

werewolves are designed to be weak and vulnerable at day, which mercenaries simply are not.

roarbark:
i don't think latejoin adventures can roll WW with how its set up, as well as it being a random migration roll anyways.

it's extremely simple to have a 20 minute timer on your phone and just return to your equipment just before you're about to untf as ww...

i don't know, it's just another thing on the pile of reasons to play merc over something like towner. you'd the very best role in the game to antag as

RadiantFlash7:
I'm going to bite the bullet and say it's fine. Keep roles can be the werewolf, and those are far more likely to be suspect then the mercs.

PR #2988, opened 2025-06-12
Changes "jumpscare (1).ogg" to Fear & Hunger battle start sfx

ARefrigerator:
## About The Pull Request
The current jumpscare (1) sounds rough and stands out as the worst one. Much more painful to listen to than the others. I changed it to be the fear and hunger battle start sound instead. These are also only ever used in grass ambushes so it's kind of like an RPG encounter.

sounds:

https://github.com/user-attachments/assets/b22feffe-24d1-42d9-9648-3dfc2456c99e

## Why It's Good For The Game
It's less painful on the ears perhaps. 

sound file replacement...


ARefrigerator:
wtf... hang on

PR #2989, opened 2025-06-12
Changes "jumpscare (1).ogg" sound to F&H battle start sfx

ARefrigerator:
## About The Pull Request
The current jumpscare (1) sounds rough and stands out as the worst one. Much more painful to listen to than the others. I changed it to be the fear and hunger battle start sound instead. These are also only ever used in grass ambushes so it's kind of like an RPG encounter.

sounds:

https://github.com/user-attachments/assets/c74f490c-5ccb-456d-bf04-106dcd63d0e3

## Why It's Good For The Game
It's less painful on the ears perhaps.

sound file replacement...


Gyrandola (approved):

Peter-Dolby:
I prefer the old one - it was appropriately startling to match a startling event happening that you have a very short amount of time to immediately react to. This new sound effect worked better in the turn-based game it originated from.

PR #2990, opened 2025-06-12
Warhammer Nerf

Dedwolf:
## About The Pull Request

Changes the warhammer's damage type from pick to stab to prevent it from critting through all armor types, now armor that prevents stabbing crits will also prevent the warhammer critting through it. However, the warhammer will still deal damage through the armor, it just won't crit until the armor is broken. 

BCLASS_PICK ----> BCLASS_STAB

The PICK damage type was originally found on daggers before it was removed. The warhammer was the only other weapon using pick damage, now, there are no weapons that use pick damage. 

Slightly buffs the generic stab intent to 25 pen and 0.9 damfactor to bring it in line with the thrust with the goeden/great mace.

## Why It's Good For The Game

The can opener no more, hopefully this should require less braindead gameplay wth the warhammer, requiring players to aim for areas uncovered or vulnerable to stabbing. It'll still deal damage through armor, though. 

Wasn't going to nerf it but I realized its a wlength = normal weapon and is capable of hitting people's hands, also merely dealing damage through armor is enough to slow somebody down and make them bleed, you also still have smash intent which was being underutilized in favor of impale, hopefully this fixes that. 

Yep works.

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/blunt.dm
+++ b/code/game/objects/items/rogueweapons/melee/blunt.dm
@@ -220,21 +220,21 @@
 	damfactor = 0.9
 	item_d_type = "stab"
 
-/datum/intent/mace/warhammer/stab
+/datum/intent/mace/warhammer/stab // With the spike on the top of the hammer, quick thrust
 	name = "thrust"
 	icon_state = "instab"
 	blade_class = BCLASS_STAB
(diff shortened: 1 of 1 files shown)


anconfuzedrock:
Thank you, being 1 shot heart crit by a goblin with this thing through 2 layers of armor and 15 CON was funny, but like, not actually funny at all this should never happen

PR #2992, opened 2025-06-12
Slaver's License for Zybantine Party

MickDick1:
## About The Pull Request

Adds a slaver's license to the Zybantine wave.
Adds a copy of the tome of justice to the Zybantine wave.

## Why It's Good For The Game

Being an active slaver requires a license by the queen to be legitimate.
Zybantines were always supposed to have one, but we just never got around to it until now.
Zybantines must still abide by the tome of justice to be within the limits outlined in the tome.

image

### DIFF
--- a/code/datums/migrants/migrant_roles/slaver roles.dm
+++ b/code/datums/migrants/migrant_roles/slaver roles.dm
@@ -22,7 +22,7 @@
 	backl = /obj/item/storage/backpack/rogue/backpack
 	beltl = /obj/item/flashlight/flare/torch/lantern
 	beltr = /obj/item/rogueweapon/whip/antique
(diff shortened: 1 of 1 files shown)


RadiantFlash7:
I'm not sure on this one. Yeah, the tome of justice is the defacto laws of the duchy, but its changed fairly routinely to be other laws.

Gyrandola (changes requested):
The license is a neat addition,  but I think the Tome of Justice is a bit out of place.
I doubt the average zybantine slaver either knows how to read, or is interested in carrying the entire penal code in his backpack.

I understand the intention was to give the slavers the tools to argue their way out of annoying anti-slavery arguments, but the license should suffice. Please remove it.

MickDick1:
The master is educated with reading 3. In any case, it's removed.

PR #2993, opened 2025-06-12
Vanderlin-Azure alchemy port [2] Meet new butching

DelineFortune:
## About The Pull Request

Build butcherhook and pull the body to it to get +1 skill
Result depends of butching and random now
Changed drop from mobs so butchers and hunters are more valuable now
Will be more drop related to alchemy in future

### DIFF
--- a/code/modules/roguetown/roguecrafting/structure.dm
+++ b/code/modules/roguetown/roguecrafting/structure.dm
@@ -22,7 +22,7 @@
 
 /datum/crafting_recipe/roguetown/structure/meathock
 	name = "meathook"
-	/obj/structure/meathook
+	result = /obj/structure/meathook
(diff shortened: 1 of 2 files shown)


Skyglazer:
Looks good! I've had a bit of a look over it and my only concerns are there being no spidermeat from spiders (is that intentional?), a botched minotaur butchery not giving ANY hides, and the fur from goats being locked to only perfect butchery. Other than that, the changes all seem pretty sensible, it's good for Hunting to have more use as a skill.

DelineFortune:
image

image

RadiantFlash7 (approved):
This really should have a few things spanified, but per useroth we can touch that up later  guess. EIther way, I want the PR full merged, it's been test merged for nearly a month.

PR #2997, opened 2025-06-12
Buff Sword Weapon Defence

Dedwolf:
## About The Pull Request

Buffs every single sword barring the gladius to provide approximately double the original wdefence, from 4 to 8 for the basic steel and iron swords with the longsword getting a wdefence of 10.

## Why It's Good For The Game

Swords are fucking fantastic IRL for parrying, sword fights should be drawn out with frequent clashes and parrying. There are spark effects for parrying another sword opponent, but how often do you see that? I've only seen it once. Additionally, shields straight up provide 15 wdefence and take the damage instead of your weapon, meaning if you have a warhammer/axe with a shield you can damage he opponents weapon without risking your own, which is fine but its not exactly fair when juuust about all swords prefer to be two-handed, even the basic sword. 

It might make the sabre a bit overpowered because thats one sword with a great wdefence score AND can be one handed, meaning you can use a shield with it too, same for rapiers etc. 

Also maces and hammers have a wbalance of -1 meaning they deal additional stamina loss for parrying them depending on strength dif between the user and the opponent. With this in mind, I think that swords having really good wdefence scores is perfectly fine, you can use feinting if defeating a parry is difficult.

Yes.

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/swords.dm
+++ b/code/game/objects/items/rogueweapons/melee/swords.dm
@@ -28,7 +28,7 @@
 	smeltresult = /obj/item/ingot/steel
 	minstr = 7
 	sellprice = 30
-	wdefense = 4
+	wdefense = 8
 	var/accuracy_bonus = 20
 
 /obj/item/rogueweapon/sword/attack_right(mob/user)
@@ -153,6 +153,7 @@
 	inhand_y_dimension = 64
 	associated_skill = /datum/skill/combat/swords
 	throwforce = 15
+	wdefense = 10
 	thrown_bclass = BCLASS_CUT
 	smeltresult = /obj/item/ingot/steel
 	accuracy_bonus = 5	//large swords are unwieldy
@@ -375,7 +376,7 @@
 	thrown_bclass = BCLASS_CUT
 	minstr = 6
 	sellprice = 42
-	wdefense = 5
+	wdefense = 9
 	smeltresult = /obj/item/ingot/steel
 
 /obj/item/rogueweapon/sword/long/marlin/getonmobprop(tag)
@@ -489,7 +490,7 @@
 	possible_item_intents = list(/datum/intent/sword/cut/short, /datum/intent/sword/thrust/short)
 	gripped_intents = null
 	minstr = 4
-	wdefense = 3
+	wdefense = 6
 	wlength = WLENGTH_SHORT	//It's a shortsword for crying outloud. Give it it's accuracy bonus for being short.
 
 /obj/item/rogueweapon/sword/iron/short/chipped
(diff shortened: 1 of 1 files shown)


roarbark:
putting my thoughts from earlier here, instead of touching their wdef, maybe add a line in the proc for parrying that adds 5 to 10 ish % of flat parry chance per level of swords you have, as long as you're parrying with one

this means skilled swordsmen will find it easier to parry, while unskilled ones wont profit from very high wdef

Red-Honk:
This PR is basically, tell me you don't understand balance without telling me you don't understanding balance.

First of all, each of those points of defense is a 10% of parry chance, each skill level above your opponent adds a extra 20%.

The benefit of using a sword normally is the following, a hand is usually free for other things to be done, swords in most cases don't have any delayed or charge up attacks that either takes time to land, prevents stamina from recovering while charging up an attack, or have an attack that drains stamina while it's in motion.

Swords have the benefit, of, being easier to carry, usually faster to attack with, being more precise than most other weapons, having a free hand mixed with the ability to wrestle while doing it with greater ease.

This PR would make it, so, anyone that picks up a sword, will be competent at parrying anyone that does not, giving a more difficult time even to the skilled fighters, this thing is purely over tuned for what it tries to achieve.

This PR is an balance violation.

TL:DR, too unbalanced, here how work, 1 weapon def = 10% parry, 1 skill weapon above opponent = 20% parry chance, 10% + 20% = 30% parry chance.

roarbark:
> Look, the defense values were tweaked by me, giving every single sword its own purpose, what's being done to them now is basically laughing at the balance that was fine tuned to carry it out, hence why I am displeased.
> 
> The sabre, rapier and cutless were all given a purpose, they all have their own attacks, sabre was given less defense in exchange for being able to parry more strikes, the rapier was given the opposite treatment since it has a thinner blade mixed with the fact it was designed to parry, the cutless was given the in-between role due to lacking neither of their benefits, instead of just being this but worse.

a blanket buff changes nothing about this. none of the swords are being changed from the position they are in like this. as a matter of fact i dont really think anybody really uses these swords for the things you mention, outside of rapier for swift balance and the other two swords for classes that spawn with them...

> 
> Now than, for the entertaining it part, give them all a +0.5 to defense, instead of going overboard if you want to see more parrying being done, give the parry chance per skill level a 25% instead of just 20%, that should stack up nicely and really tell them the difference in skill.
> 
> wdefense from 3 to 3.5, as an example should do the trick.

a wdef buff is already ruled out from the comments, i also suggest against tying it to skill difference because we are trying to make swords useable in comparison to other weapon classes, so we want to avoid things like lvl 3 swords vs lvl 3 maces having no parry buff to the sword side, essentially still having it where the sword user would have been better off using a different weapon type. even matchups like 4 swords vs 3 polearms/maces would result in a meager 5% or 10% extra parry compared to the same situation but the swordsman has chosen a different role with 4 maces or polearms

PR #2999, opened 2025-06-13
(depreciated) Fixes hand intents when putting things into various objects

ARefrigerator:
## About The Pull Request
**NOTE: currently using https://github.com/Rotwood-Vale/Ratwood-Keep/pull/3010 to replace this.**

Adds the drop from equipment the actions need to fix the intents of the user to the following things:

- Oven insertion
- Hearth insertion
- Hearth insertion onto pan
- Quiver insertion

## Why It's Good For The Game
Previously it was just a forceMove which does not update intents of the hand. You currently have to swap hands to fix it. This solves that issue.

https://github.com/user-attachments/assets/e729df00-a7e4-4754-8bc3-17cf4823cc3e

### DIFF
--- a/code/modules/roguetown/roguejobs/blacksmith/smelter.dm
+++ b/code/modules/roguetown/roguejobs/blacksmith/smelter.dm
@@ -51,6 +51,7 @@
 			return
 
 	if((ore.len < maxore) && W.smeltresult)
+		user.dropItemToGround(W, force = TRUE)
 		W.forceMove(src)
 		ore += W
 		user.visible_message("<span class='warning'>[user] puts something in the smelter.</span>")


Actuallygoinginsanern:
does this also fix the glitch for when you put an arrow in a quiver and it has the same issue?

Skyglazer:
> > does this also fix the glitch for when you put an arrow in a quiver and it has the same issue?
> 
> It does now. I added it to the PR.

Another similar case is when you put items into a smelter, I think. Would be worth having a look at too if it's similar to this case.

ARefrigerator:
> Another similar case is when you put items into a smelter, I think. Would be worth having a look at too if it's similar to this case.

Added

GravyMaker:
Good to see a fix submitted for this irritating problem. I'd actually looked into this myself, but hadn't submitted a PR due to the amount of testing I wanted to do.

You might want to look into the forceMove proc that crops up in most of the snippets that you've changed - when I looked into this, I found that caching the human mob that had previously been holding the item (if such a mob existed) and then updating its intents after the move happened fixed the behaviour as well - if you can do the testing I never got round to, you might find, if we're lucky, that this approach catches every case, rather than requiring you to manually add logic to every interaction that causes the behaviour.

ARefrigerator:
I'll close this one when I get confirmation it's no longer in and or when the other one gets merged.

Useroth:
Merged  #3010 instead

PR #3000, opened 2025-06-13
Add export only value for stockpile datums

ARefrigerator:
## About The Pull Request
Allows some things to be export only in the stockpile menu. Adds it to gold and silver

## Why It's Good For The Game
Radiant request. Said it was to make getting the rarer ores riskier.

in game menu:
image
image

### DIFF
--- a/code/modules/roguetown/roguestock/_roguestock.dm
+++ b/code/modules/roguetown/roguestock/_roguestock.dm
@@ -15,7 +15,7 @@
 	//how many of the items are consumed/spawned when exporting/importing
 	var/importexport_amt = 10
 	var/import_only = FALSE //for importing crackers, etc
-	var/export_only = FALSE // for preventing exports
+	var/export_only = FALSE // for preventing imports of certain goods
 	var/stable_price = FALSE
 	var/percent_bounty = FALSE
 	/// How much to generate in the remote section each firing of the treasury system.


RadiantFlash7:
> This guts the Steward's gameplay loop, for what little there is of it. Merchant would be the only reliable source of treasure. Doubly so at low pop, if one even exists. High pop you've a chance I guess, with merchant and miners both present. At least with the import, you could shore up the vault roundstart and immediately have things put to work. Now, lowpop you're just screwed. Deadpop you've no chance whatsoever. May as well grab a pick yourself.
> 
> This is a pointless change that harms Steward, who'd be the only one benefiting from the import anyway. Silver weaponry aside. What's the actual consideration here? To make ore more rare? Who said it wasn't rare? Who's the one importing it, period? The Steward, that's who. Singular guy who's touched by this PR, realistically.
> 
> Maybe it's doomsaying but this is considerably worse than an alternative of just jacking up the price. Which would also be bad, to be clear. But not nearly as poor as this.

It's worth noting that the bog lava-ravine now has a much higher chance of providing silver and gold ore.
By default, Importing silver was always a... poor mechanic, considering it hard counters 2 of our antags from a single touch.

I think it's a more interesting system to have the steward hire a miner with guards to go to the ravine and mine the silver and gold, rather then just importing it and being done with it.

ARefrigerator:
> I get why the PR is being made, but I'm providing my input as a player who interacts with the vault. Importing exists and it's a horrible crutch, but it's one that's needed, given you cannot rely on any other method of vault income generation. Miners are fickle and may run off with coin, which fucks the vault and everyone else as a result in matters of income. IC issue I guess, but that's likely to become the normal. Which is not great for a mechanic so integral. Merchants may not exist. At least with importing you can smith the goods yourself, after a good while of farming skill level.

Is there a way to make money as a steward without the gold import?

roarbark:
i like this change, a lot of healthy things for the game imo
- silver is not as easy to mass produce
- towners are made more important
- steward gameplay is less braindead
- keep has less money and might actually have to ration. it's ok if you only make 100 mammon every 7 minutes instead of 800, if there's no miners, improvise. pay a beggar, pay a fisher, pay a thief, pay anyone who can get you items of value at all

ARefrigerator:
I feel like it'd make more sense if the treasury started with more in it. There's a laughable about of wealth in the keep.

ARefrigerator:
I could make it count mammons unless there's a reason why it doesn't.

john-pawson:
> Steward gameplay isn't particularly engaging now, and removing further methods of actively building mammon is just going to make it even more boring. You'd think that if the alternative was better, we'd be doing it, but no, the import is the best you have as the Steward right now. Unless you imagine Stewards particularly love to sit around waiting on mammon to build up, and give an "IOU <3" to everyone who comes about, because they don't have the mammon up front to pay them?

Have you ever heard of the whole "people will optimize the fun out of a game"? Well, i think this might be one of those cases.

Why go ahead and ever interact with any of the miners or the people that bring you mammon when you've got more than enough by just importing ASAP a bunch of gold and having the smith turn it into bars and valuables?
No smith? Not like that's an issue, considering how you already have the keys to the place you can easily just churn out three bars of gold, silver, steel, and blacksteel without much of an effort. (Which already gives you >150 mammon from roundstart, without counting anything else)
There's not even a need for gold and silver for that amount specifically, you can get close to the same amount by just putting three of every other importable mineral on the stockpile. (Again, not counting all the rings, statues, etc. that could be made if there's a smith.)

Maybe the steward *should* go ahead and need to put some effort and elbow grease to have a good amount of mammon, if they haven't been able to reach the amount they needed. Make them DO something to earn it at least.
Maybe they should start changing the prices of withdrawing and feeding because every mammon counts. 
Or send the MAA on an expedition to mine on the bog at the cost of leaving the keep undefended for a bit.
There's no MAA? Then that's a lot less spending! You've barely even need to do paychecks, the main way you spend your mammon, if there's nobody around. (Not like that's ever going to happen) that and, there's always towners or, well, litterally anyone else.
There's no merchants to buy valuables from? You don't even need one! Just get yourself a bodyguard and go on an expedition to the bog merchant, it's more expensive but it's not exactly impossible to afford considering how you already start with more than enough to.
That's not even mentioning all the other ways to get valuables... fishers (Rings, legendary fishes, pearls), farmers (You can probably farm so many crops in so little time that it's another passive income, if you give the farmers a reason to do it), hunters (hides, furs, saddles, bedrolls, quivers, armors, pouches, cloaks, essence of wilderness), seamsters (wizard robes, half-cloaks, silk clothing, padded gambesons), physicians (uncut moondust, spice, even LUX!), artificers (prosthetics), even cheesemakers (aged cheese) can all be used to boost up your mammon income. But you barely ever see a steward try ANY of these options. Because they don't need to. Why would they.

There's tons of interesting scenarios and options that simply don't happen because the steward never runs out of mammon. Because they won't. They got their mammon farm roundstart and are basically set for the rest of the round. Which ends up with them not having to do anything else unless they want to have more than they'll spend.

It's not removing methods of building mammon, it's making THE most efficient and proftiable method on the entire game need more than a loan from the duke and a click of a button to be set for the rest of the round.

PR #3001, opened 2025-06-13
Mercenaries Are Greedy

Dedwolf:
## About The Pull Request

Forces mercenaries to have the greedy flaw.
Mercenaries who pick the hunted flaw will not have the greedy flaw.
Mercenaries who pick missing eyes or wooden limbs will still have a missing eye or wooden limb.

## Why It's Good For The Game

Mercenaries should want MAMMONS and fucking lots of it, they should take heads or capture people for bounties or they should find work where they can. It's not a class where you can just- help the local guard out, you need mammon to eat and live, you don't live in Rockhill. 

~~I have attempted to write the code so that you could still take other flaws with it, primarily ones that modify your body like wooden limbs or missing eyes but it seems you can only have one flaw at a time.~~

image
image

### DIFF
--- a/code/modules/jobs/job_types/roguetown/mercenaries/mercenary.dm
+++ b/code/modules/jobs/job_types/roguetown/mercenaries/mercenary.dm
@@ -31,6 +31,17 @@
 		H.advsetup = 1
 		H.invisibility = INVISIBILITY_MAXIMUM
 		H.become_blind("advsetup")
-	if(H.mind)
-		if(H.charflaw && !H.has_flaw(/datum/charflaw/greedy))
-			H.charflaw = /datum/charflaw/greedy
+	if(H.charflaw)
+		if(H.charflaw.type == /datum/charflaw/noeyer)
+			var/obj/item/eyepatch = new /obj/item/clothing/mask/rogue/eyepatch(get_turf(H))
+			H.put_in_hands(eyepatch, forced = TRUE)
+			H.equip_to_slot_if_possible(eyepatch, SLOT_WEAR_MASK, FALSE, TRUE, FALSE, TRUE, TRUE)
+		if(H.charflaw.type == /datum/charflaw/noeyel)
+			var/obj/item/eyepatch_l = new /obj/item/clothing/mask/rogue/eyepatch/left(get_turf(H))
+			H.put_in_hands(eyepatch_l, forced = TRUE)
+			H.equip_to_slot_if_possible(eyepatch_l, SLOT_WEAR_MASK, FALSE, TRUE, FALSE, TRUE, TRUE)
+		if(H.charflaw.type != /datum/charflaw/greedy)
+			if(H.charflaw.type == /datum/charflaw/dead_or_alive)
+				return
+			else
+				H.charflaw = new /datum/charflaw/greedy


Peter-Dolby:
I don't like this particularly since I just don't like how the Greedy flaw works, with the way it only counts physical coins held on your physical person and not, like, any of the things you might use the money for.

I'd rather be making money so that I can spend it on drugs and whores rather than making money so I can just hold it.

roarbark:
> Probably gonna tick people off. I understand the change but an issue with Rockhill is that the threats against it are often existential; they're so harsh that mercs would be stupid not to help fight against them.
> 
> It's hard to earn any money if everyone's dead.

a character who thinks like this would not have become a mercenary, but rather a guardsman. mercenaries are disgusting profiteers who dont care if another settlement falls as long as their pouches are heavier.

Skyglazer:
I'm very skeptical of the design intention to force mercenaries to have one of the most annoying flaws in the game to "encourage" them to roleplay in a certain way. This isn't going to encourage merc players to be more cutthroat or look for jobs-- these are roleplay problems. What it IS going to encourage is mercnearies to grind hunting/etc. for mammons to satisfy their mechanical vice (which a lot of them already DO grind mammons roundstart to get better gear). Let alone the fact that mercenaries are seldom actually paid enough mammons for a contract to satisfy the vice.

This PR won't make mercenaries play better or look for more contracts; the town already barely hires them for anything. All this will do is force people playing the role to deal with an annoying flaw rather than pick one that suits their character. 

If we think that players aren't playing the role properly to the point of it needing correction, this should be dealt with by admin action, not spoiling the fun for everyone by annoying them with mechanics. People who play mercenary like guard+ can be met with admin directions and rolebans if necessary.

MrBity:
Yeah, this seems to be made to address a problem that should really be handled by QC. Mechanical problems like this are going to push people towards mechanical solutions. We need to address player quality with action against players.

GravyMaker:
I do have a few misgivings about this myself. Mercenaries _should_ certainly be willing to be, well, mercenary - profiteering, exploiting crises, selling their services to the highest bidder _within reason_. However, doing this via forcing them to have the Greedy flaw has a number of disadvantages, in my opinion:

- It disincentivises them from actually spending a large portion of their money on anything when, as a combat role, it has a huge gameplay incentive to do so. The conflict here is likely to push players to find ways of gaming the flaw.
- It makes the flaws system largely irrelevant to mercenaries; their only valid flaws are now Hunted (which a lot of mercenaries will ignore on their fellow mercenaries), a missing eye, or Greedy - absolutely nothing else.
- The implementation of Greedy itself is pretty lacklustre for various reasons that may be better discussed in the contributor channel, but some of these issues are likely to affect the manner in which at least some mercenaries play in undesirable ways. This could be something addressed through admin intervention; but then, so is the initial issue that this PR aims to address.
- As has been pointed out, the game economy and mechanics really in a state where the best way for mercenaries to satisfy this flaw is through mercenary work - there are other, typically more lucrative activities they can engage in which will likely bear more profit, and players may find themselves pushed towards these activities rather than selling their services as they ought to.

I also have a couple of comments with respect to the implementation of this PR:
- On a technical level, due to the way the code is structured, only Hunted will actually prevent you from getting Greedy due to the early return in its branch. If your flaw is a missing left or right eye, you'll still be missing the eye _and_ be Greedy. It's difficult to tell whether this was a deliberate decision or just forgetting to include early returns, as in the case of Hunted.
- Presumably missing eyes are supported in the first place due to being plausible battle wounds, which makes me also wonder why wooden arms aren't acceptable flaws (even if I doubt many mercenaries would take it).

Actuallygoinginsanern:
this PR is not good. And is a lazy way to get people to RP in a way you want

Red-Honk:
Played it, hated it, imagine being peckish and just get hit with tired and get the message, **I AM GETTING ON MY LIMITS!**, in a round that has, literally no bounties, people that want to hire a merc, a place hat has food to begin with, this encouraged my character to act like a bandit+ instead of a merc, because, what you want them to do?; Do things that break role expectations, by doing labor?; That feels off. (You can't even hunt bandits or skeletons anymore for bounties, so, the options are slim)

PR #3003, opened 2025-06-13
Nerfs wooden staff defense and integrity

john-pawson:
## About The Pull Request

For some inexplicable reason, the wooden staff has had more wdefense than any other weapon on the entire game excluding shields (It even has higher defense than the buckler! That's a shield!! Specifically made for blocking!!) which lead to certain people getting some incredible parry chances by just carrying a long stick around. 
Not only that, but since it had no integrity assigned, it had the default 200 integrity, which is normally used for steel weapons. Just to compare, the stone spear, which is a wooden staff with a stone tip, has 4 wdefense and 50 integrity.

## Why It's Good For The Game

Balance! If you want a durable and useful way to block attacks, you should get a shield or a proper weapon at least.

image

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/polearms.dm
+++ b/code/game/objects/items/rogueweapons/melee/polearms.dm
@@ -50,7 +50,7 @@
 	inhand_x_dimension = 64
 	inhand_y_dimension = 64
 	wdefense = 4
-	max_integrity = 50
+	max_integrity = 75
 	bigboy = TRUE
 	gripsprite = TRUE
 	associated_skill = /datum/skill/combat/polearms
@@ -282,7 +282,7 @@
 	blade_dulling = DULLING_BASHCHOP
 	walking_stick = TRUE
 	wdefense = 4
-	max_integrity = 50
+	max_integrity = 75
 	throwforce = 20
 
 /obj/item/rogueweapon/halberd


RadiantFlash7 (changes requested):
I don't think we need to nerf the staff's integrity.

What I /do/ think we need to do, is change how the calculation for wielded items works. Instead of adding a flat 3, create a new var on base items with a value of three- like wdefense- and then change that var for staff/polearms so it has a much higher defense value when wielded, but it's default wdefense value for polearms being terrible.

MickDick1:
The wooden staff having high wdefense is absolutely intentional. It's an entirely defensive focused tool, meant for pilgrims and the like when travelling long journeys to keep volves and the like at bay. That's the idea in general. Balance/design wise it was also the brigand hallmark, given you need a staff to travel the bog efficiently. The freeman fighting with staff and weapon is quintessential to their themes. A tool of freedom at all times. They would blend in very well with all the other refugees as a result too.

But in any case, I will say. If the staff is nerfed, I will get around to giving bandits bog trekking. They are **the** survivalists of the terrorbog. They know the place better than anyone anyways. And I agree with Radiant. Polearms should probably have pretty bad 1h defense while having excellent wielded. As long as it's brought back up to 13 wdefense when wielded it's a good change.

roarbark:
i'd say keep it at 10 (or maybe lower it to 7 ish so you dont get 90%s with it one handed) but lower the durability to 100 to 150 ish. a quarterstaff is simply easy to.parry with and in game its a really good tool for noncombat roles to not get their heads instantly ripped off

RadiantFlash7:
We'll instead tweak it so we can set weapons wielded def value

PR #3004, opened 2025-06-13
Gives vanguard axe skill

anconfuzedrock:
## About The Pull Request
Just gives vanguard footman 3 in axes, to go along with their general weapon skill, and vanguard archers 1 in axes, to go along their 1 mace skill.

## Why It's Good For The Game
"Lore"wise: Axes are something many are already familiar with even without martial training, it makes sense the the more inexperienced goons would be good with axes, more sense than the swords really. also parallels the MAA footman, which has axe skill. They even have axes in their armory!! Stone axes, but even if they're only meant to be used for cutting wood, woodcutters get axe skill still!
Balance wise:
Axes are seen as scary for balance to some, but MAAs and the warden have axes 4 without causing any real uproar. With Forlorn Hope mercs now having the option to start with axes 4 and a battleaxe, axes are more prevalent. But even if axes were particularly nasty, vanguard aren't going to be too terrifying with axes, because they still only get 3 weapon skill and stone ones! They'll have to put in effort to even get an iron axe, let alone a battle axe. This won't break anything
Fun wise: 
Axes are fun and convenient, good way to deal with the endless skeleton ambushes or trees, both things the vanguard might encounter. Let them enjoy weapon variety!

It's a 2 line change have mercy please


MickDick1:
They should train with the warden if they want to become specialized in fighting with axes.

ddsaw:
> Im always against stat and skill buffs unless its absolutely necessary as it generates slow and steady powercreep. I don't believe this change to be _necessary_. Leaving the last verdict to Radiant.

I'd prefer it if they had swords removed and replaced with axes entirely. Swords are cool, noble and expensive weapons and thus do not fit them at all. Hatchets could replace the swords in the armory.

RadiantFlash7:
Going to deny this for now. Vanguard don't get axes, theres no reason for them to have axe skill.

Peter-Dolby:
I think it's nice to have skill in something you don't have access to at round start for that occasional once-in-a-while opportunity where you come into possession of one

Y'know, you might get into a fight with a bandit or goblin, they drop a cool axe, and you think "oh boy, a cool axe!" but are disappointed when it turns out you can't use it.

PR #3009, opened 2025-06-13
Hoardmaster changes

Rayforx:
## About The Pull Request
Gives two health potions worth of reagents to the kidnapped victim when they are kidnapped and put into the hoardmaster.
Admins see that this is causing problems with people dying from being skull-cracked and left for dead in a spawn area.

## Why It's Good For The Game

Admin requested changes.
Why it's being requested by the admins: Having people with skull cracks in the dock causes problems, and makes them into a deadite that spawn camps people.

image
Nothing breaks, and 2 bottles worth of health potion are given.

### DIFF
--- a/code/modules/roguetown/roguemachine/hoardmaster.dm
+++ b/code/modules/roguetown/roguemachine/hoardmaster.dm
@@ -268,6 +268,7 @@
 
 	visible_message(span_warning("Runes of binding ignite on [H]'s temples, each sigil a searing brand of agony!"))
 	H.emote("scream")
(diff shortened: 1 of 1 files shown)


Red-Honk:
The way this is coded, it's a bad idea, make it so, it either modifies the reagent to be on 90, or give it 1 healthpot and tell it to modify it to 90, last thing we need is them dying from healing potion overdose.

Red-Honk:
> Only greater potion ODs, normal and lesser doesn't.

MY LIFE WAS A LIE!

Loldaddy5676:
Maybe have it check if the target is healthy? Would make sawbones a much better role to play as.

PR #3010, opened 2025-06-13
Alternative hand intent fix on forceMove

ARefrigerator:
## About The Pull Request
An alternative approach to the issue https://github.com/Rotwood-Vale/Ratwood-Keep/pull/2999 tries to fix.

This one puts a simple check in the forceMove code itself to double check if the location of the object is a human and if so to update the intents.

## Why It's Good For The Game
Would be less tedious and you wouldn't have to think about keeping two procs together should you use forceMove in these situations.

https://github.com/user-attachments/assets/870347ec-a35c-4cb4-955c-f5e9b5e1cc19

### DIFF
--- a/code/game/atoms_movable.dm
+++ b/code/game/atoms_movable.dm
@@ -479,11 +479,19 @@
 	A.Bumped(src)
 
 /atom/movable/proc/forceMove(atom/destination)
+	//Fix for human intents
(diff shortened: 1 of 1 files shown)


Red-Honk:
Than that settles it, pestering time.

PR #3012, opened 2025-06-14
Adds scrying orb failure effects, and reduces likelihood of the orb breaking

GravyMaker:
## About The Pull Request
This PR adds a variety of failure effects that can occur when using the scrying orb, based on the user's arcane skill and intelligence, and whether or not the orb is on cooldown. These failure effects are intended to be somewhat-punishing, but - mostly - more flavourful than mechanically frustrating, with those with the harshest mechanical penalties being the shortest-lived. I've also taken the opportunity to refactor the orb code a little and fix some bugs (eg. the examine text not working, and division by zero when a person with no arcane skill attempts to use it).

Failure effects include:
- Scrying a random person other than the person you'd intended to scry.
- The hand holding the orb being burned by magical backlash.
- Being confused for a few minutes.
- Being put to sleep for a little over one minute.
- Being rendered drunk for a few minutes.
- Being rendered high for a few minutes.
- Having your intelligence and speed reduced, and having a "simple speech" modifier applied.
- Being paralysed in a random selection of limbs for a couple of minutes.
- Losing knowledge of Imperial for a few minutes.
- Becoming obsessed with the target of the scrying attempt, as if they'd used a love potion on you.
- Having your intelligence drastically reduced and losing access to all languages for 10 minutes (the worst effect - and very unlikely to be seen in gameplay).

On failure to use the scrying orb, one of these failure effects will be applied to the user depending on a few factors and a little bit of randomisation. When a failure occurs, if the user has 2 or higher arcane skill, the orb enters an extended ten-minute cooldown during which success rate when using it is halved; if the user has 1 or 0, they still receive failure effects as described above, but the orb breaks as it did previously. Success rates while the orb is on cooldown have also been reduced - in most cases, by about one skill level's worth, relative to success rates when the orb is not on cooldown.

Finally - the "Simple speech" trait that was added as part of this is a new speech modifier. Specifically, this speech modifier means a character can only say the most common 1,000 English-language words unhindered. Words not on this list have a chance of being replaced or prefaced with "uh..." or "um...", and will be scrambled, leaving the first and last letters in place to aid legibility while still communicating the idea that the speaker is struggling with them.

Obviously, the specific "severities", durations, and so on of each effect can be tweaked fairly easily if necessary.

## Why It's Good For The Game
The scrying orb is an interesting item with a lot of potential - but as it stands, it's very risky to use if you aren't one of the roles that spawn with it, due to the penalty for failing the skill check. With this PR, failures would still be somewhat punishing - but wouldn't deprive the user of the orb for the rest of the round, or until they're able to find the funds to purchase a new one, thus making it a more worthwhile investment for roles that have some magical capability.

The failures themselves, I think, are also much more interesting than simply breaking the orb - often being effects that one can roleplay around for, hopefully, a more generally interesting experience than simply having to find money for a new orb.

I'll not show a fully comprehensive list of screenshots, but a few to demonstrate certain elements working correctly. I promise I _have_ tested everything I can think of, though.

Drunkenness failure condition:
image

Dim-witted failure condition:
image

High failure condition:
image

### DIFF
--- a/code/datums/status_effects/debuffs.dm
+++ b/code/datums/status_effects/debuffs.dm
@@ -677,25 +677,30 @@
 	var/list/traits_added = list()
 	var/list/bodyparts_disabled = list()
 	
+#define PARALYZE_L_ARM 0x1
+#define PARALYZE_R_ARM 0x2
+#define PARALYZE_L_LEG 0x4
+#define PARALYZE_R_LEG 0x8
+#define PARALYZE_SUM (PARALYZE_L_ARM | PARALYZE_R_ARM | PARALYZE_L_LEG | PARALYZE_R_LEG)
 /datum/status_effect/debuff/mishap_arcane_paralysis/on_apply()
 	. = ..()
-	var/limbs = rand(1, 15) // To be used as bits, NOT a meaningful integer value
+	var/limbs = rand(1, PARALYZE_SUM) // To be used as bits, NOT a meaningful integer value
 	// This method ensures our limbs are chosen randomly,
 	// that all limbs have a 50% chance of being paralyzed,
 	// and at least one limb is guaranteed to be paralyzed
-	if (limbs & 0x1)
+	if (limbs & PARALYZE_L_ARM)
 		ADD_TRAIT(owner, TRAIT_PARALYSIS_L_ARM, id)
 		traits_added.Add(TRAIT_PARALYSIS_L_ARM)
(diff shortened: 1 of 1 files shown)


Useroth (changes requested):

RadiantFlash7 (approved):
Test merge completed with little issues. A great change for something honestly.

PR #3014, opened 2025-06-14
Improves Tabaxi Speech and makes it Toggle'able

Erol509:
## About The Pull Request

Not much to say, this improves the speech, going byond the limitations of regex regules, by spliting the sentence in parts and treating each one seperately. This makes so there can be only one r multiplied in the word by the code. It usualy is the only first one at the front, if it starts by the R. It also now lowercase when it should.

So, it splits the sentence into words, and packs it into the list. Then the loop cycles through each letter and searches for the R in it, when it finds it, it gets modified, it runs through each word you say. It shouldnt be performance eating, since character modifications are usualy run very fast, and it doesnt do many calculations at the same time. 

## Why It's Good For The Game

Easier to read, with words that contain multiple syllables. 

image

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -61,6 +61,7 @@ GLOBAL_LIST_INIT(name_adjustments, list())
 	var/pda_style = MONO
 	var/pda_color = "#808000"
 	var/prefer_old_chat = FALSE
+	var/alternative_speech = TRUE
 
 	var/uses_glasses_colour = 0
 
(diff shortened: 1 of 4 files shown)


Erol509:
Gonna post Screenshot when I test it fully. So far it works just fine

Erol509:
image

Erol509:
image

Gyrandola:
A bunch of requests

-  Reduce the amount of  r's from 3 to 2
- Make the replacement probabilistic, not every word in every phrase has to have it.
- Make the replacement *not* happen to a r if its the first letter of the a word. (e.g. Retarrd instead of Rretard)

Erol509:
- 50% chance to affect a word it cycles.
- It only affects syllables inside the words
- It now only allows 2 R's at most. 

Regarding possible issues with computation. It shouldnt be a problem even with a long sentences, it runs only once per message. And there wont be 100 people outputing 4k word messages each millisecond, to affect the performance at all. 

Im sure it wont be a problem at all, with how the chat is being used.

RyszardParys:
Is is optional tho?

Issue #1536, opened 2024-11-11
[REQUEST] Text explaining Towner homes to Towners


Frienderanian:
## Is your feature request related to a problem? Please describe. 

I think a lot of players who pick Towner don't get that you LIVE in Rockhill and have a HOUSE in Rockhill! As Barkeep I have had players ask for a room as Woodcutter/Carpenter/Miner/Gambler/etc players. I've had to show them the towner homes and tell them to use the door!

## Describe the solution you'd like

Perhaps when you spawn in as towner, it can say 'As a towner, you live in Rockhill. Try clicking doors around town to find an unoccupied home to claim,' or maybe even spawn them outside of homes or inside of them with the key?

## Additional context

This is completely unrelated: I am not sure how the Towner home situation is going to be handled when families are implemented. I feel like the entire design of the town would need to be flipped around to accommodate this feature otherwise players in the same family accidentally claim multiple houses and have no idea where each person lives. 

Also, completely unrelated: Perhaps look into some ways of reworking the Tavern so the process of renting rooms and returning keys is a little more automated? Coin operated rooms perhaps? Might make a more in depth suggestion later.

PR #3015, opened 2025-06-14
Explains Towner houses to Towners

Soproxi:
## About The Pull Request

Adds an extra line to the Towner greet message, saying that you can claim a house as your own by clicking on it.

## Why It's Good For The Game

Addresses Issue #1536 .
This should really be written out somewhere, because it isn't in any documentation, or anything to my knowledge. This is a simple little fix until a future rework of Towners/families/houses etc.

image

### DIFF
--- a/code/modules/jobs/job_types/roguetown/towners/towner.dm
+++ b/code/modules/jobs/job_types/roguetown/towners/towner.dm
@@ -6,7 +6,7 @@
 	total_positions = 75
 	spawn_positions = 75
 	allowed_races = RACES_VERY_SHUNNED_UP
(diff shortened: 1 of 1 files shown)


Red-Honk:
--somedae, but for now, you'll just have to live here, probably somewhere your house lies, no better way to find out which than to check in your key under the sleeve works on one of them.

Soproxi:
Changed to
"You own one of the decrepit houses around here... you think you remember stashing a spare key just under the door."

Seems more immersive, I suppose!

And updated the old use of day to 'dae'!

RadiantFlash7:
Seems fine enough.

Actuallygoinginsanern:
I dont think something like this exactly requires immersive wording, if its for newer players to realize they can get a house, being clear about it is best.

PR #3017, opened 2025-06-14
Remove remaining bikini armors

Gyrandola:
## About The Pull Request

Title. Most of these removals are already unused items without recipes, except for the chain one.

## Why It's Good For The Game

Bikini armors are obviously out of place, and almost all of them are ports from Azure Peak, which says a lot.
I believe the decision to get rid of these was taken long ago, we just never got around to do it.


Red-Honk:
The last time I saw it, it was on a Dame.

PR #3019, opened 2025-06-15
Mage gameplay loop tweaks

RadiantFlash7:
## About The Pull Request
this PR does a few things:
It adds crafting for runed artifacts, 
Provides random chance to find runed artifacts when mining based on luck (like essence of nature)
Adjusts the ritual requirements to use the planar convergence rune, (from T2 meld to T1)
Fixes a bug related to planar convergence matrix and other one ritual spells.
Revamps how spellbooks are made. 
From apprentice upwards, You must now upgrade your existing spellbook instead of crafting an entirely new one. In this vein, Upgrading a spellbook required a runed artifact to imbue it, before you can apply the next level meld to your book.

## Why It's Good For The Game
Runed artifacts being a bit more reliable to get ahold of, now having a decent source of it.
Makes the Planar conversion ritual a bit more common, hopefully, compared to it's never ever used status.
Should hopefully make mages stop skipping lower level spellbooks to go straight to high level ones, since there was no reason to ever make a lower level one.

image
image

### DIFF
--- a/code/modules/spells/roguetown/_roguetown.dm
+++ b/code/modules/spells/roguetown/_roguetown.dm
@@ -51,7 +51,7 @@
 /obj/effect/proc_holder/spell/invoked/proc/on_deactivation(mob/user)
 	return
 
-/obj/effect/proc_holder/spell/invoked/InterceptClickOn(mob/living/caller, params, atom/target) 
+/obj/effect/proc_holder/spell/invoked/InterceptClickOn(mob/living/caller, params, atom/target)
 	. = ..()
 	if(.)
 		return FALSE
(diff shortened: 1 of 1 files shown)


anconfuzedrock:
by apprentice upwards, does this imply you need an apprentice book to craft an adept one? because apprentice books are unusually awkward to get, requiring a +10 enchanted rock. should probably be adept and up instead.

Genevesh:
I would still consider making wall runes explosion-proof as well as giving the walls themselves more durability to see them get some use.

Wiping the rune with a wet cloth would still be extremely easy, and even logically fireballs should not be able to destroy runes across a barrier.. that's just now how explosions work, but reworking them entirely will not be something we'll see in a while, so this is probably the only option

anconfuzedrock:
I think there might be a bug where you can't upgrade an adept spellbook to an expert one with a saffira because it gets used temporarily like an amythortz instead? not sure yet.

PR #3020, opened 2025-06-17
Start of loadout framework

Dilligentdog:
## About The Pull Request

Adds the loadout items framework from Azurepeak to Ratwood. Adds many clothing items and stuff like cards or dice to the loadouts. Easily modifiable so if we decide to add more clothing to it or remove items from it

## Why It's Good For The Game

Players have been asking for this for ages and i was finally approved to work on these changes. Adds customization to character in with extra clothing they may want for their role

image
image

### DIFF
--- a/code/datums/loadout.dm
+++ b/code/datums/loadout.dm
@@ -44,10 +44,6 @@ GLOBAL_LIST_INIT(loadout_items, init_loadout())
 	name = "Straw Hat"
 	path = /obj/item/clothing/head/roguetown/strawhat
 
-/datum/loadout_item/witchhat
-	name = "Witch Hat"
-	path = /obj/item/clothing/head/roguetown/witchhat
-
 /datum/loadout_item/bardhat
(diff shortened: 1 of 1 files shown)


Skyglazer:
Hey. As per Useroth's request (after my own suspicions in-game shortly after seeing the available options), I've gone through and tested the options selectable in the loadouts.

The following selectable items don't have correct worn-sprites or are otherwise non-functional:

Hijab
Witch Hat
Volf Mantle
Cloth Anklets
Collar (has a sprite, but is a debug item)

Lord-Preston:
I noticed that the loadouts are saved across all characters rather than per character like on AP's code base, would it be possible for that to be changed? It is annoying to need to change the loadout each time you switch characters.

Useroth:
Let's work iteratively since I need to touch other code in that place and I don't want it to conflict

But yes, we'll make them per character

PR #2227, opened 2025-01-15
Changes how players are counted for roundstart antags, adds vampire and werewolves (seperately) back to lowpop.

Spaghetti-bit:
## About The Pull Request

Number of players is now determined by the amount of players connected, not just the ones readied up.
Adds vampire back to lowpop. 50/50 between peasant rebellion and vampires.

## Why It's Good For The Game

An unhealthy amount of people don't ready up and play migrants which affects the roundstart antags. Some purposefully late join to do this as well.

Adding vampire back to lowpop as an experiment to see how they are balanced, a rework is planned after bandits are changed.

PR #3021, opened 2025-06-18
Fix antag rolling not checking if players were readied

noelle-lavenza:
`num_players()` checks number of players connected
`num_players(TRUE)` checks number of players readied
untested because how the fuck am I going to do that on local


john-pawson:
This was an intended change made in https://github.com/Rotwood-Vale/Ratwood-Keep/pull/2227 since people would just not ready up and then latejoin to prevent any antags from spawning. Ever.

noelle-lavenza:
> This was an intended change made in #2227 since people would just not ready up and then latejoin to prevent any antags from spawning. Ever.

Yeah, I'm thinking of making players in lobby count as 25% or 50% instead of not counting at all?

Gyrandola:
> This was an intended change made in #2227 since people would just not ready up and then latejoin to prevent any antags from spawning. Ever.

Ditto. If your intention is to lower antag spawn rates just modify those directly.

noelle-lavenza:
> > This was an intended change made in #2227 since people would just not ready up and then latejoin to prevent any antags from spawning. Ever.
> 
> Ditto. If your intention is to lower antag spawn rates just modify those directly.

no, my intention is to fix the "bug" people have been complaining about for months where antag spawns are based off of players connected rather than players readied

PR #3022, opened 2025-06-18
Ports actors menu (lobby manifest) from Azure Peak

noelle-lavenza:
lazy-ass port of https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/245. somewhat tested in local testing but no promises

also fixes the title for popups not actually using the specified font

### DIFF
--- a/code/datums/browser.dm
+++ b/code/datums/browser.dm
@@ -107,7 +107,7 @@
 	</head>
 	<body scroll=auto>
 		<div class='uiWrapper'>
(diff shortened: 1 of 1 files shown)


noelle-lavenza:
oh wait I didn't realize the test merge label was used to actually TM things automatically. anyway, this should be TM'd.

RadiantFlash7:
> oh wait I didn't realize the test merge label was used to actually TM things automatically. anyway, this should be TM'd.

It doesn't automatically test merge things. It does however, label stuff in TGS for us, bringing TM labeled PR's to the top of the list of PRs when Tming things.

PR #3023, opened 2025-06-18
Better Bandit Posters

Soproxi:
## About The Pull Request

Adds additional functionality to Bandit Posters and the Scholar role:

- Bandit Notices can be 'torn down' by bandits, taking a little time, making a loud noise and an alert to anyone who sees!
- Wardens, Knight Captains, Sheriff, Mayor, or any Nobility can create a bandit notice by placing a paper on a wall, and scrawling the bandit's names, with a sufficient Reading check.
- Scholar can post Noble bulletins throughout the town in a similar fashion, for educating the peasantry.

## Why It's Good For The Game

Adds more of a notoriety mechanic to bandits. Allowing more back-and-forth gameplay. Bandit posters shouldn't be completely indestructible as they are now, and there should also be a way to place them back, if bandits are allowed to tear them down.

The Scholar change just gives the scholar more to do, as I felt like it was a RP-focused role, it would be nice to have a bulletin-style RP-focused ability.

image
image

(still in draft! new contributor too, please be kind!)

### DIFF
--- a/code/game/turfs/closed/walls.dm
+++ b/code/game/turfs/closed/walls.dm
@@ -198,7 +198,6 @@
 			if("icon")
 				SStitle.icon = icon
 
-
 // Override the try_wallmount proc instead of attackby
 /turf/closed/wall/try_wallmount(obj/item/W, mob/user, turf/T)
 	// Check if user has Scholar role and has paper/parchment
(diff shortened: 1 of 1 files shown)


Soproxi:
I will try to make bandit posters able to be placed on palisades-- meaning the bog bridge could have a bandit poster placed upon it, with some planning. This would allow vanguard easier bandit-identifying, and generally makes more sense.

PR #3028, opened 2025-06-20
Faster round announce

Useroth:


Useroth:
Nevermind, not worth fucking with for now

PR #3039, opened 2025-06-26
Craftable nooses.

Useroth:


anconfuzedrock:
you can already craft nooses. this adds noose crafting a second time to the same structure crafting file it's already in. am I missing something?

PR #3042, opened 2025-06-27
Tetris inventory: Useroth porting it edition

Useroth:
Ported from Vanderlin

Also made bags eject contents on tile in front of you instead of under you


ophaq:
- Sacks currently don't hold their full amount. Felt like they held more on other codebases.
- There might be a bug where sometimes a sack will only grab one item (such as fruit) and you can't fill it any more.

RadiantFlash7:
From my understanding, storage enchantments are still broken with the new system. Shouldn't be too hard to fix though. Would just need to make the enchantment tweak the new vars.

Useroth:
Merging this for now to prevent conflicts with some other things.
Will fix storage enchantment in another PR.

PR #3044, opened 2025-06-29
Tracking & Thieves' Cant, also singing because why not

Useroth:
Code is there but skills and language are not added to anyone yet, it's TODO.

To sing, just put % in front of your message.

Also some misc azure ports along the way like dream lines for skills.


Useroth:
Let's add said skills to people in next PR

PR #3054, opened 2025-07-04
Restores Adventurer

Xaytan:
## About The Pull Request
Re-adds adventurer.

They use the same spawn points as refugees, nobody ever bothered to remove that.

## Why It's Good For The Game

image

Teethling Barbarian Gaming

image

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/adventurer.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/adventurer.dm
@@ -16,8 +16,8 @@
 
 	outfit = null
 	outfit_female = null
-	total_positions = 50
-	spawn_positions = 50
+	total_positions = 20
+	spawn_positions = 20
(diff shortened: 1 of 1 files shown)


code/modules/jobs/job_types/roguetown/adventurer/adventurer.dm
```diff
@@ -16,14 +16,16 @@
 
 	outfit = null
 	outfit_female = null
-
+	total_positions = 50
+	spawn_positions = 50
```
Useroth, line 20:
```suggestion
	total_positions = 20
	spawn_positions = 20
```

Useroth:
Also make sure that the spawn points exist on the map @Xaytan

Peter-Dolby:
Feels like there's going to be a lot of fiddling to do with how much of it is also covered by refugee.

PR #3056, opened 2025-07-04
Rot Zombification timer increase, Town deaths protected from zombification (still rot, but after a longer delay)

RadiantFlash7:
## About The Pull Request
Players won't turn intto deadites until after 7 minutes after death
Players that die in town won't turn to deadites at all, and rot after 10 minutes.

## Why It's Good For The Game
One of the most frequent requests players have been asking for, a long while. This should hopefully make rotting less punishing, allowing players doing surgeries more time to treat the dead, as well as providing more safety in town. 

N/A

### DIFF
--- a/code/datums/components/rotting.dm
+++ b/code/datums/components/rotting.dm
@@ -64,7 +64,7 @@
 		qdel(src)
 		return
 
-	if(amount > 4 MINUTES)
+	if(amount > 7 MINUTES)
 		if(is_zombie)
 			var/datum/antagonist/zombie/Z = C.mind.has_antag_datum(/datum/antagonist/zombie)
 			if(Z && !Z.has_turned && !Z.revived && C.stat == DEAD)
--- a/code/modules/mob/living/carbon/human/death.dm
+++ b/code/modules/mob/living/carbon/human/death.dm
@@ -152,4 +152,4 @@
 	return ..()
 
 /proc/can_death_zombify(mob/living/carbon/human)
-	return TRUE
+	return hasomen(OMEN_NOPRIEST) || !is_in_roguetown(human)


Peter-Dolby:
I like deadites, deadites are fun :(
The chaos it can cause is hugely entertaining to me. I'd rather there be particular actions you can take to extend the timer (winding sheets, more access to embalm fluid, maybe laying on tables) for dilligent players trying on purpose to slow it down rather than just making it take long enough for it to rarely happen. Also just outright disabling it from happening *in town* at all is tremendously sad, that's where everything is and all the fun parts happen!

Peter-Dolby:
> This also returns the old system of requiring the Priest to be dead. Excellent stuff.

What part requires the priest to be dead?

Genevesh:
> > > If they're in the town and the Priest is alive, they can't turn after rotting. Once the Priest is dead, the rot takes the town entirely. Reason to kill him / save him.
> > 
> > 
> > And what happens if there isn't a priest in the round?
> 
> No omen potential, from their death, which means there can never be a deadite turning in town. Only outside of it.

I really fw this

Peter-Dolby:
> I can see why you'd not want it, though, absolutely. As it stands, death is a very real issue. This doesn't really change that, though. It just makes it so the dead don't get up and maul you to death inside the town. You also can't really do a lot with bodies RP wise in the town as is, whether that be sacrificing them, propping them up to be found, etc, without them just up and walking away.

It's why I think the solution is to add more ways for players to deliberately extend/halt the timer beyond the relatively rare reagent that is embalming fluid with things like winding sheets or just laying them out on a table or with an acolyte's magical blessing or through removing the brain, so that a body that people are really intentionally trying to save is less likely to go (part of this is a bit of a skill issue from people who have embalming fluid never remembering to use their embalming fluid) but a body that gets kinda forgotten about still comes back to cause some chaos in an amount of time that isn't a total bore for a watching ghost.

I feel like Priests could do with some special feature, like an Anti-Rot Aura that activates when they're specifically around, like within ~x amount of tiles, or within the church when the priest is in the church. Make it feel like something that the priest is actively bringing to the table rather than a baseline default that a joining priest can only serve spoil by dying (in a way that I feel will usually result in a lot of OOC salt)

PR #3059, opened 2025-07-05
AP Port: Spell & Orison

ProbablyCarl:
## About The Pull Request
Jesus Christ someone atomize the wizard file. I beg you.
I ain't doing it, but SOMEONE has to. Probably something I'll do in another PR if no one else gets to it, but good lord it's AWFUL.
-  - -
New:
 - Blink, 4 cost, allows jumping 7 tiles after charging for five seconds. Fourteen second cooldown. Has a bunch of limiting factors for what you can jump do. For example, it cannot be through windows that aren't climbable, can't go through bars / gates / doors, etc. Primarily going to be used for crossing very large gaps. Weaker here than it is on AP, due to the increased charge / cooldown, given I didn't want people using it to blink about in battle. I'd buffed it significantly as a tradeoff, in that you can now freely blink down Zs, and, as long as there's an overhang, you CAN blink upwards. You cannot blink onto structures directly, as the 'walls' will stop you. It was oversight, but I'd left it as an intended mechanic. Should stop some shenanigans and prevents jumping atop most structures.

 - Guided Bolt, 3 cost, far slower arcane bolt. 5 less damage. Arced, allowing it to hit targets through foes / down Z levels.

 - Acid Splash, 3 cost, AOE burn spell. ~80 damage on an indirect hit over time, 5 damage per tick. 10 projectile damage, so not great without that buildup. Burn also just fucking sucks, so you'll need to do upwards of 600+(~700 something?) damage to deathcrit someone, and it doesn't destroy gear unlike fireball. The primary use is for crowd dispersal / mass paincrit and crippling. We're going to need to tweak this I figure. Quite heavily.

 - Orison. T0 miracle for ALL clerics / clergy. Exactly as it is on AP. Allows you to amplify your voice, which also has a chance, 3% per miracle level, to broadcast what you say to a scom on the map. Every scom is checked, so it could be a good number, or it could be one or two at best. Secondly, it allows you to fill glasses with cursed or blessed water, depending on your patron. Very small amounts of heals for those who're corresponding to blessed/cursed/whatever, and the dead can be harmed with blessed water. Whether that be by tossing it on them, or drinking it. Lastly, you can imbue yourself, or a companion, with stackable light.

The above also includes icons (and the blink sound) from AP(I think?), though recycles their shadowstep icon for blink, instead of the icon they had, because the shadowstep one was neat by comparison.

## Why It's Good For The Game
I 'unno. I felt it'd be neat to have some more spells. Break up the monotony. Add variety. I'm too lazy to port the rest, though there's plenty to choose from.
It works. I just don't have the ability to record a video as of the moment. Or crop / edit one, for that matter.
image
image


RadiantFlash7:
> Out if these 4 you're porting, I think guiding bolt may be something we can do without. It does less damage for seemingly little gain, and I don't really see occasions people will use it over other spells.
> 
> Acid splash is in a similar issue in this PR, albiet to a lesser degree. Arcane storm as it functions serves this role fairly well; though acid splash providing DoT effect for entering it is interesting, and unique.

I miss understood how acid splash worked. It's definitely fine. Guiding bolt might still be a little underwhelming, but- we'll have to see I suppose.

PR #3074, opened 2025-07-11
AP Port: Slings

ProbablyCarl:
## About The Pull Request
Ports slings from AP. Skills, crafting, etc. Doesn't include the alternative ammunition types, and instead just focuses on stone. Had to port the weapon sprite, but otherwise we already had the others. Sounds, of course, had to be ported, too.

This PR is almost entirely just a copy paste of AP's code, with modifications to allow compiling here. This isn't my work, beyond the aforementioned changes.

Original PR, on AP, can be viewed [here](https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/1550).
## Why It's Good For The Game
An alternative ranged weapon that makes use of a plentiful resource.
Variety. Also more thematic for tribals and goblins
image


wow50000:
why are you adding shadowlings?

PR #3077, opened 2025-07-19
Updates Dirt Crafting to Fix Possible Exploit

Trexdude:
## About The Pull Request
This PR aims to fix a possible exploit involved with crafting dirt turfs using the crafting menu. Prior to this PR, dirt turfs could be crafted on any `/turf/open/floor/rogue` turf: this led to a possible exploit, wherein someone trying to break into a building that's located a z-level below where they're standing could, hypothetically, craft a dirt tile on the floor, which would enable shoveling straight through the floor into the building below. To fix that, this PR simply adds a check for crafting dirt, that checks if the turf you're crafting the dirt on is itself dirt. If not, you can't craft it there.

## Why It's Good For The Game
In all honesty, this exploit was probably never used, but I figure it could help to prevent sneaky farmers (as low as beginner skill) from breaking into places they wouldn't normally be able to. It also hopefully adds some more realism/balancing with farm soil creation, in that dirt crafting can't be used to make a possible farm plot wherever you want now, if you wouldn't normally be able to farm there beforehand. This includes crafting dirt on grass; you have to remove the grass first before you can access the dirt underneath, and therefore craft new dirt on it - as well as cases such as farming on cobblestone streets, rooftops, etc.
However, this PR might affect people who use dirt crafting as a way of in-game beautification, or as a way of adding dirt to rocky floors, such as in caves (for the few plants that grow underground). It's also possible there'll be cases of people trying to farm on rocky outcrops exposed to the surface, who will be unable to craft dirt on the rock floors to farm. I'll try to keep an eye out for any feedback from players on this.

CompileProof
GrassNoCraftDirt
RoadDirtCraftDirt
WoodNoCraftDirt

### DIFF
--- a/code/modules/roguetown/roguecrafting/turfs.dm
+++ b/code/modules/roguetown/roguecrafting/turfs.dm
@@ -240,7 +240,7 @@
 
 /// DURT
 /datum/crafting_recipe/roguetown/turfs/nrich
-	name = "soil plot - (ash, fiber, dirt clod; COMPETENT)"
+	name = "soil plot - (ash, fiber, 2 dirt clods; MASTER)"
 	result = /turf/open/floor/rogue/dirt/nrich
 	reqs = list(/obj/item/ash = 1,
 				/obj/item/natural/fibers = 1,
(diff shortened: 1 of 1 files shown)


Trexdude:
Found a possible bug with this - crafting cobblerock roads seems impossible on grass now. I'll check it out.

Trexdude:
Ah, I think I found it. The cobblerock crafting recipe also checks to see if you're crafting it on dirt or not, and if you are it disallows crafting. Not sure why.. Maybe that can be another PR?

PR #3079, opened 2025-07-21
Leatherworking Patchwork

ProbablyCarl:
## About The Pull Request
Checking used intents is dumb and it should be bladeclass.
But this SHOULD make it capable for all knives to once again work with hide on a rack. The same can't be said about axes or swords, which have a handful of cuts, so only one or two will still work for it, and they need to be wielded. Someone can fix this to check bladeclass at some point. But this patchwork fix should be fine.
## Why It's Good For The Game
Technically not a bug, but it's oversight and is proper.
It should work. I've not tested it.

### DIFF
--- a/code/modules/roguetown/roguejobs/tailor/tanning.dm
+++ b/code/modules/roguetown/roguejobs/tailor/tanning.dm
@@ -36,7 +36,7 @@
 			return
 		else
 			to_chat(user, span_warning("The rack is already occupied!"))
-			return
+			return//Todo, CHANGE THE USED INTENT CHECK TO BLADECLASS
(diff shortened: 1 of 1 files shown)


Grigorix:
Its because the code is tied to knives/ daggers having the huntingknife parent which currently the iron and steel dagger lack. Its just a master of fixing that as mechanics like dipping are all tied to hunting knives

ProbablyCarl:
[This PR](https://github.com/Rotwood-Vale/Ratwood-Keep/pull/3083) is proper. The current one will be closed shortly and um-TM'd.

PR #3083, opened 2025-07-22
Tanning.dm simplification.

SmiLeYre:
## About The Pull Request

I've simplified the tanning to make it easier for players to make cured hide.

## Why It's Good For The Game

In the current conditions, cured hide can only be obtained with a stone, a two-handed axe and scissors, and etc.


ProbablyCarl:
This supersedes [this fix](https://github.com/Rotwood-Vale/Ratwood-Keep/pull/3079) and should be used instead. I'll close the other PR shortly.

PR #785, opened 2024-08-12
Gives Squires the ability to repair. Adds polishing metal tools

GuppyLaxx:
This PR adds the ability to polish anything metallic, giving it a nice shine and some extra durability and the likes. It's some small boons for now, though I will work on the exact perks it provides later on.
To polish something you need cream, and a brush. They can be purchased as a kit from the shop. Squires and smiths get some at the start.
You then put cream, brush it harshly, brush it finely, and then clean it. The effect lasts until the item is damaged.

dreamseeker_niqlbRBPY3

This PR also gives squires a unique trait that lets them do certain smithing related things, primarily, the new polishing and repairing stuff. They can ONLY repair things at an anvil, at a fixed rate.
This trait removes their ability to smith anything, or to properly do anything smithing related, UNLESS they manage to become skilled enough at smithing. In which case their trait becomes moot and they are now a proper smith!

The Squires now get a little room outside the manor, next to the tower and the gatehouse.
image

The whole purpose of this PR is to go with my 2 other Armor PRs nerfing sleeping in armor, and nerfing being able to put on armor on your own. Along with other intended effects that accompany the broader changes like the little extra room and polishing cream.

PR #3087, opened 2025-07-27
Cream & Squire Changes

ProbablyCarl:
## About The Pull Request
It's just [THIS](https://github.com/Rotwood-Vale/Ratwood-Keep/pull/785), revived.
New:
 - Polishing cream and brush. Added to the luxury tab for merchant purchasing, and otherwise placed into Squire inventories roundstart. These allow you to apply the cream and then scrub the item. If done appropriately, the item is given a +2 bonus to force and an integrity bonus. Alongside a fancy metal glint.

 - Squire trait added, allowing them to repair and sew without the skill, albeit at a worse rate than a smith for repairs. This trait also helps for applying cream and brushing equipment. Squires below the skill of master smithing cannot work with metal. Period. Fullstop. They can repair it, but they can't smith.
---
Other:
 - Squire given a number of description changes, and their backpack contents are separated per subclass, as opposed to being doubled on spawn. A page spawns with a rondel, while the other two still get a misericorde.
 - Squire subclasses have 3 slots now, given there's 3 slots for the class itself. Instead of 2 per. I should just remove the limit altogether, but ehhhhhhh. There was no reason for the restriction, really.
---
## Why It's Good For The Game
More content. Thematically fits. Just something really neat to promote roleplay or otherwise encourage engagement between a Squire and Knight.
It just works.
image
image
image
image


wow50000:
Can smiths get these as well or is it only going to be squires?

